PR #23, opened 2025-06-11
feat: allow models from build.nvidia.com to be used

LunaticMaestro:
NVIDIA build models use the format `nvdia/vila` or `abc/xyz`. 

Change: The pydantic regex was throwing 500 validation error as the regex pattern was not inclusive of `/`. So updated the pattern to include it.


zac-wang-nv:
it was for a previous release and the code path is different

PR #31, opened 2025-07-03
VSS 2.3.1

nv-unnikrishnankgs:


zzv2:
One thing, since you added the `parakeet` docker image, shouldn't the instructions in the jupyter notebook be updated to say:
> After running the following cell, you should be able to see _**four**_ containers, one for each model.

PR #90, opened 2026-04-17
feat: add fusion search skill

maxervo:
## Description
Add a video-search skill so agents know how to run natural-language video archive search on a deployed VSS stack, using fusion search (semantic embeddings + CV attributes).

Steer agent towards the main user flows, add troubleshooting feedback loop for common scenarios where data is missing, user query is ambiguous, optional verification step with frontier model etc.

For now it uses the VSS agent top level endpoint `/generate` (secondary agentic loop) to leverage more agentic searches and abstract away the lower level components (embed search, attribute search). In the future, we may expose the [lower level pieces](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/82) and enable the use of a single agentic loop.

## Checklist


zac-wang-nv:
LGTM, can you put 'discovery_modes.md' and 'troubleshooting.md' to references folder? thanks

soumilinandi:
@maxervo @zac-wang-nv does it make sense to add optional parameter info like `source_type` as rtsp as well? And also adding`curl -s -X POST http://${HOST_IP}:8000/api/v1/rtsp-streams/add` in troubleshooting for RTSP streams along with default video_file support? 

We can also include start time and end time in search payload

zac-wang-nv:
yes, I think it makes sense to support both input source types from the skill. By default, it uses 'video' as source types and switching to RTSP when user specifies.

maxervo:
> We can also include start time and end time in search payload

@soumilinandi To double check, the top level agent endpoint `/generate` does not have these explicit knobs but we instruct it to pass these params in natural language. Let us know if this approach (i.e. only exposing top agent for now) works for you
```
# Search by time context
curl -s -X POST http://${HOST_IP}:8000/generate \
  -H "Content-Type: application/json" \
  -d '{"input_message": "what happened at the entrance between 2pm and 3pm?"}' | jq .
```

The top agent then calls the underlying `search_agent` that maps those to the proper payload params for `/search`. In the future, I guess we will expose those underlying endpoints with deterministic knobs via skills and let user agents (claude code / openclaw / opencode etc.) orchestrate them directly?

PR #92, opened 2026-04-17
Feat/vss orchestrator mcp server

nvleeryan:
Summary
This MR updates the VSS orchestrator MCP surface to enable agents to deploy VSS with various profiles by making the tool call to the MCP server. VSS orchestrator MCP is NAT compliance.

MCP tool API
Exposed MCP tools:
profiles
prereqs
docker_generate
docker_read
docker_list
docker_logs
docker_up
docker_down
docker_status
Documentation

Added src/vss_agents/orchestrator/README.md with:
startup instructions
full tool list
workflow diagram
payload examples
response examples
environment-specific config notes (including mdx_data_dir and output_dir)

Test plan
Start server:
uv run nat mcp serve --config_file src/vss_agents/orchestrator/vss_orchestrator_mcp_config.yml --port 9902
Validate tool registration and run smoke calls:
docker_profiles (no payload)
docker_prereqs (no payload)
docker_generate with sample profile/env overrides
docker_read using returned docker_compose_id
docker_up + docker_status polling
docker_list / docker_logs
docker_down + docker_status polling
Confirm old tool names are no longer referenced in config/include paths.


agent/src/vss_agents/orchestrator/tools.py
```diff
@@ -0,0 +1,931 @@
...
+    # ---------------------------------------------------------------------------
+    # Tool: docker_logs
+    # ---------------------------------------------------------------------------
+
+    if "docker_logs" in _config.include:
+
+        async def _docker_logs(input: ContainerLogsInput) -> dict:
+            """Fetch docker logs by container name."""
+            result = subprocess.run(
+                ["docker", "logs", "--tail", str(input.tail), input.container_name],
```
zac-wang-nv: **Flag-smuggling risk.** `container_name` is inserted into the argv without a `--` separator, so an agent that chooses `-f` / `--follow` / `--since=...` / `--help` as the container name will have `docker logs` interpret it as an option rather than a positional. `-f` in particular will block the MCP event loop until the container dies or the server is killed.

Minimal fix:
```python
["docker", "logs", "--tail", str(input.tail), "--", input.container_name]
```
Consider also rejecting `container_name` values that begin with `-`. Same pattern would apply to any future tool that appends a free-form string after options in a `docker ...` argv.

agent/src/vss_agents/orchestrator/tools.py
```diff
@@ -0,0 +1,931 @@
...
+from .docker_compose_util import parse_env_overrides
+from .docker_compose_util import SUPPORTED_PROFILES
+from .prereqs_check import run_prereqs_checks
+from .storage import ArtifactKind
+from .storage import ModelArtifact
+from .storage import ensure_data_directories
+from .storage import ensure_model_artifacts
+
+_COMPOSE_OPS_LOCK = threading.Lock()
+_COMPOSE_OPERATIONS: dict[str, dict[str, object]] = {}
```
zac-wang-nv: **Unbounded memory growth.** `_COMPOSE_OPERATIONS` and `_COMPOSE_SPECS` are module-level dicts that only ever grow. Every `docker_generate` adds a spec; every `docker_up` / `docker_down` adds an op entry with a `deque(maxlen=4000)` log buffer. Nothing is evicted when an operation finishes or when `docker_down` tears a deployment down. For a long-running MCP server driven by an agent that retries or iterates, this is an unbounded leak.

Two options that fit the current shape:
1. LRU-cap each dict and evict completed ops older than some TTL.
2. Add a `docker_forget` tool that lets the agent drop a `docker_compose_id` / `docker_compose_ops_id` explicitly.

I'd lean toward (1) for predictability.

services/agent/src/vss_agents/orchestrator/tools.py
```diff
@@ -0,0 +1,931 @@
...
+
+    if "docker_list" in _config.include:
+
+        async def _docker_list(input: ComposeContainersInput) -> dict:
+            """List docker container names."""
+            args = ["docker", "ps", "--format", "{{.Names}}"]
+            if input.all_containers:
+                args.insert(2, "--all")
+
+            result = subprocess.run(
```
zac-wang-nv: **Blocking subprocess on the event loop.** `_docker_list`, `_docker_logs`, and `_docker_prereqs` are `async def` but call blocking `subprocess.run(...)` directly, which blocks the event loop for the duration of the subprocess. `docker logs` with a large tail, and especially the `docker run nvidia/cuda:...` toolkit check in `prereqs_check.py` (120s timeout, pulls ~1 GB image on first run), will freeze any concurrent MCP request.

Wrap with `await asyncio.to_thread(subprocess.run, ...)` (or switch to `asyncio.create_subprocess_exec`) so the loop stays responsive. The `_start_compose_op` path is already fine because it spawns a dedicated watcher thread.

services/agent/src/vss_agents/orchestrator/docker_compose_util.py
```diff
@@ -0,0 +1,441 @@
...
+def parse_env_overrides(entries: list[str]) -> Dict[str, str]:
+    overrides: Dict[str, str] = {}
+    for raw in entries:
+        if "=" not in raw:
+            raise ValidationError(f"Invalid --env entry '{raw}'. Expected KEY=VALUE.")
+        key, value = raw.split("=", 1)
+        key = key.strip()
+        if not VALID_ENV_KEY.match(key):
+            raise ValidationError(f"Invalid env key '{key}'. Must match {VALID_ENV_KEY.pattern}.")
+        overrides[key] = value
```
zac-wang-nv: **Missing value validation.** `parse_env_overrides` validates the key but not the value. A value containing `\n` (or `\r`) will append additional `KEY=VALUE` lines to the generated env file when `_set_env_line` writes it back out, silently injecting or hijacking other variables.

Reject control characters in values:
```python
if any(ch in value for ch in ("\n", "\r")):
    raise ValidationError(f"Invalid value for '{key}': control characters not allowed.")
```

zac-wang-nv (commented):
Thanks for getting this up. The design is clean and the 9-tool surface reads well. Left ~14 inline comments below; most are polish, a handful are worth addressing before merge — the ones I'd prioritize are the `docker_logs` flag-smuggling, blocking `subprocess.run` inside `async def`, and the unbounded in-memory ops / specs dicts.

Two open questions I'd like your take on before we land this:

**1. Tests.** The `orchestrator/` package adds ~2000 LOC of subprocess + threading + compose lifecycle with no unit tests. The preempt / cancel logic in `_start_compose_op` and the env-merge rules in `build_resolved_env` are exactly the kind of thing where a small suite pays for itself the first time behaviour regresses. At minimum, the pure functions are cheap wins: `parse_env_overrides`, `parse_env_file`, `first_non_placeholder`, `sanitize_resolved_compose`. What's the plan here — follow-up MR, or can we get a first pass in this one?

**2. Tool surface: 9 tools vs. 1 tool with actions.** We currently expose `docker_profiles`, `docker_prereqs`, `docker_generate`, `docker_read`, `docker_list`, `docker_logs`, `docker_up`, `docker_down`, `docker_status` as nine separate MCP tools. Have we considered collapsing these into a single `docker` tool with an `action` field (e.g. `{"action": "up", "docker_compose_id": "..."}`)? Pros: smaller tool surface for the agent to reason about, one entry point to document, easier to extend with a new verb later. Cons: per-action input schemas become a discriminated union and some MCP schema validators are finicky with those. Curious whether this was discussed already — happy to go either way, mostly want the choice to be intentional rather than default to one-tool-per-verb.

agent/src/vss_agents/orchestrator/tools.py
```diff
@@ -0,0 +1,931 @@
...
+                pass
+
+    def _start_compose_op(docker_compose_id: str, action: str, action_args: list[str]) -> dict:
+        if action not in _SUPPORTED_COMPOSE_ACTIONS:
+            return {
+                "status": ComposeStatus.ERROR.value,
+                "error": f"Unsupported action '{action}'. Supported actions: {sorted(_SUPPORTED_COMPOSE_ACTIONS)}.",
+            }
+
+        def _running_op_error(existing: dict[str, object], conflicts: list[dict[str, object]]) -> dict:
```
zac-wang-nv: `conflicts` is never used inside `_running_op_error`; please drop it from the signature (both call sites would only need the `existing` arg anyway). Minor trailing whitespace after the closing `}` here and on line 425 — ruff will complain under the repo's existing lint config.

agent/src/vss_agents/orchestrator/tools.py
```diff
@@ -0,0 +1,931 @@
...
+            return f"Resolved compose file not found: {compose_path}. Run the 'docker_generate' tool first."
+        return None
+
+    def _compose_env() -> dict[str, str]:
+        compose_env = os.environ.copy()
+        compose_env.setdefault("COMPOSE_PROGRESS", "plain")
+        compose_env.setdefault("COMPOSE_ANSI", "never")
+        return compose_env
+
+    def _terminate_running_op(op: dict[str, object]) -> None:
```
zac-wang-nv: **PID-reuse race.** Between the SIGTERM and the final SIGKILL in the `else` branch, the OS can recycle `pid` for an unrelated process, and the subsequent `os.kill` would hit that innocent PID.

Since the watcher thread already holds the `Popen` object, the cleaner pattern is to stash it in `_COMPOSE_OPERATIONS[...]["process"]` and call `process.terminate()` / `process.kill()` / `process.wait(timeout=...)` here, which is race-free. Low probability in practice, but it's a trivial structural fix.

services/agent/src/vss_agents/orchestrator/tools.py
```diff
@@ -0,0 +1,931 @@
...
+
+        group.add_function(name="docker_logs", fn=_docker_logs, description=_docker_logs.__doc__)
+
+    # ---------------------------------------------------------------------------
+    # Tool: docker_up
+    # ---------------------------------------------------------------------------
+
+    if "docker_up" in _config.include:
+
+        async def _docker_up(input: ComposeOperationInput) -> dict:
```
zac-wang-nv, line 996: The docstring says `docker compose up -d --force-recreate --build`, but the actual argv on line 864 also passes `--quiet-pull`. Please sync the docstring, the README, and the tool description so all three agree.

While you're there, worth a sentence in the README noting that `--force-recreate --build` means every `docker_up` rebuilds images and recreates containers. That's fine for dev workflows but surprising for an agent that polls `docker_up` repeatedly after a failed deploy.

services/agent/src/vss_agents/orchestrator/vss_orchestrator_mcp_config.yml
```diff
@@ -0,0 +1,155 @@
...
+#   vss_orchestrator__docker_generate - resolve env + compose YAML artifacts
+#   vss_orchestrator__docker_read - fetch generated env/yaml by docker_compose_id
+#   vss_orchestrator__docker_list - list docker container names
+#   vss_orchestrator__docker_logs - fetch docker logs by container name
+#   vss_orchestrator__docker_up - docker compose up -d --force-recreate --build
+#   vss_orchestrator__docker_down - docker compose down -v --remove-orphans
+#   vss_orchestrator__docker_status - poll docker operation status/logs
+#
+# Start with:
+#   nat mcp serve --config_file deployments/developer-workflow/vss-orchestrator-mcp/vss_orchestrator_mcp_config.yml
```
zac-wang-nv: This comment points at `deployments/developer-workflow/vss-orchestrator-mcp/...`, but the file actually lives at `agent/src/vss_agents/orchestrator/...`. The port also disagrees: this comment says `:9901`, the README says `:9902`.

Please pick one port, and reconcile all three places (this comment, the README, and any downstream operator docs) so they agree.

.gitignore
```diff
@@ -0,0 +1,3 @@
+*.pyc
```
zac-wang-nv: Per the existing convention in the internal deep-search repo, the `.gitignore` belongs under `agent/` rather than the repo root. Can we reuse the canonical `agent/.gitignore` from deep-search instead of introducing this 3-line list at the repo root?

agent/pyproject.toml
```diff
@@ -3,7 +3,7 @@ requires = ["hatchling", "hatch-vcs"]
 build-backend = "hatchling.build"
 
 [tool.hatch.version]
-source = "vcs"
+source = "vcs"
```
zac-wang-nv: Accidental trailing whitespace on a line that otherwise didn't need to change. Please revert so the diff stays scoped to the orchestrator addition.

services/agent/src/vss_agents/orchestrator/network_util.py
```diff
@@ -0,0 +1,122 @@
...
+
+def run_text_command(command: list[str], *, timeout_seconds: int = DEFAULT_COMMAND_TIMEOUT_S) -> str:
+    try:
+        result = subprocess.run(command, capture_output=True, text=True, timeout=timeout_seconds)
+    except (FileNotFoundError, subprocess.TimeoutExpired):
+        return ""
+    return result.stdout.strip() if result.returncode == 0 else ""
+
+
+def detect_internal_ip() -> str:
```
zac-wang-nv, line 56: `bash -lc` sources the user's login shell (`/etc/profile`, `~/.bash_profile`, `~/.bashrc` depending on flags) and resolves aliases, so a user with `alias ip=...` or a noisy `~/.bashrc` either contaminates stdout or adds hundreds of ms of startup per call. No shell metachar here actually requires bash.

Doing it in pure Python is tighter:
```python
result = run_text_command(["ip", "-o", "route", "get", "1.1.1.1"])
# then split on " src " in Python
```
That removes the bash dependency, makes the parse explicit, and isolates the call from shell-environment surprises.

services/agent/src/vss_agents/orchestrator/tools.py
```diff
@@ -0,0 +1,931 @@
...
+
+        group.add_function(name="docker_profiles", fn=_docker_profiles, description=_docker_profiles.__doc__)
+
+    # ---------------------------------------------------------------------------
+    # Tool: docker_prereqs
+    # ---------------------------------------------------------------------------
+
+    if "docker_prereqs" in _config.include:
+
+        async def _docker_prereqs(input: DockerPrereqsInput) -> dict:
```
zac-wang-nv: **Completeness gap: structured prereq output.** All the useful prereq detail (GPU count, driver version, Docker version, Compose version, disk free, NVIDIA Container Toolkit state) gets `print(...)`'d to the server's stdout and discarded — the MCP caller only sees `"Prerequisite checks passed."`.

For an orchestrator MCP whose whole point is letting an agent reason about whether the host can deploy a profile, this is the single most useful structured payload we could return. Refactoring `run_prereqs_checks()` (and the `_run_*` helpers) to return something like:
```python
{"gpus": [...], "gpu_count": N, "driver_version": "...",
 "docker_version": "...", "compose_version": "...",
 "container_toolkit_ok": True / False,
 "disk_free_gib": ..., "disk_total_gib": ...}
```
and exposing it as `result["details"]` here would unblock downstream agent decisions with minimal new code.

services/agent/src/vss_agents/orchestrator/tools.py
```diff
@@ -0,0 +1,931 @@
...
+    deployments_dir = Path(_config.deployments_dir).resolve()
+
+    # ---------------------------------------------------------------------------
+    # Shared helpers
+    # ---------------------------------------------------------------------------
+
+    configured_output_dir = Path(_config.output_dir).expanduser().resolve()
+    mdx_data_dir = Path(_config.mdx_data_dir).expanduser().resolve()
+    configured_mdx_data_directories = tuple(_config.mdx_data_directories)
+    configured_model_artifacts_by_profile: dict[str, tuple[ModelArtifact, ...]] = {
```
zac-wang-nv, line 452: **Silent profile-typo failure mode.** A profile-name typo in `vss_orchestrator_mcp_config.yml` (e.g. `serach:` instead of `search:`, `alert:` instead of `alerts:`) silently disables artifact downloading for the intended profile: `_start_compose_op` will look up the misspelled profile, get `None`, skip the download, and the subsequent `docker compose up` will fail deep inside a container with a confusing "model file not found" error.

One-line fail-fast at startup:
```python
unknown = set(_config.model_artifacts) - SUPPORTED_PROFILES
if unknown:
    raise ValueError(f"model_artifacts contains unsupported profile keys: {sorted(unknown)}")
```
Same shape as the `mdx_data_dir` bootstrap guard on line 342.

agent/src/vss_agents/orchestrator/tools.py
zac-wang-nv: **Byte-size cap alongside line cap.** The `tail` upper bound (`_CONTAINER_LOG_TAIL_BOUNDS.maximum = 10000`) caps line count but not total byte size. A container that emits 10k lines at ~8 KB each (JSON-per-line VLM logs, stack traces with embedded payloads) produces ~80 MB in a single MCP response, which most MCP clients won't tolerate gracefully and which will also block the event loop on serialization.

Add a byte-size cap (e.g. truncate at 1 MB and append `\n... [truncated, N bytes more]`) in addition to whatever `to_thread` fix comes from the other comment on this tool.

nvleeryan: good point.  Will update as suggested

nvleeryan: fixed

agent/src/vss_agents/orchestrator/tools.py
nvleeryan: Fixing with LRU.  This one was on my TODO list :)

agent/src/vss_agents/orchestrator/tools.py
nvleeryan: Missed that dangling unused param after refactoring...Removed

agent/pyproject.toml
nvleeryan: fixed

services/agent/src/vss_agents/orchestrator/docker_compose_util.py
nvleeryan: fixed
BTW, source .env file is resided in the targeted host machine, and we don't allow user to edit .env remotely.  So strictly the source .env must be coming from the VSS repo.  If user has access to the targeted VSS deployment machine to change the .env, she/he already has access to the system anyways and probably no need to hijack by modifying the .env

services/agent/src/vss_agents/orchestrator/network_util.py
nvleeryan: fixed

services/agent/src/vss_agents/orchestrator/tools.py
nvleeryan: fixed 

`{
  "status": "success",
  "message": "Prerequisite checks passed.",
  "details": {
    "gpus": [
      {
        "index": 0,
        "name": "NVIDIA RTX PRO 6000 Blackwell Server Edition",
        "driver_version": "580.126.09",
        "memory_total_mib": 97887,
        "memory_total": "97887 MiB"
      },
      {
        "index": 1,
        "name": "NVIDIA RTX PRO 6000 Blackwell Server Edition",
        "driver_version": "580.126.09",
        "memory_total_mib": 97887,
        "memory_total": "97887 MiB"
      }
    ],
    "gpu_count": 2,
    "driver_version": "580.126.09",
    "docker_version": "Docker version 29.4.1, build 055a478",
    "compose_version": "Docker Compose version v5.1.3",
    "container_toolkit_ok": true,
    "disk_free_gib": 295.3,
    "disk_total_gib": 484.4
  }
}`

services/agent/src/vss_agents/orchestrator/tools.py
nvleeryan: good point.  fixed

agent/src/vss_agents/orchestrator/tools.py
nvleeryan: fixed
Also does the same for docker_status

services/agent/src/vss_agents/orchestrator/tools.py
nvleeryan: Fixed
Wrapped _docker_list, _docker_logs, and _docker_prereqs are async def into await asyncio.to_thread(subprocess.run, ...) to unblock event loop

agent/src/vss_agents/orchestrator/tools.py
nvleeryan: fixed

services/agent/src/vss_agents/orchestrator/vss_orchestrator_mcp_config.yml
nvleeryan: fixed
updated outdated path and port

agent/tests/unit_test/embed/test_cosmos_embed.py
zac-wang-nv, line 1: why is this file removed?

tools/scripts/nemoclaw/init_ollama.sh
zac-wang-nv, line 1: can we remove ollama?

tools/scripts/nemoclaw/setup_ollama.sh
zac-wang-nv, line 1: can we remove ollama?

skills/deploy/references/alerts.md
zac-wang-nv, line 1: can we remove skills additions to avoid conflicts when merging from feat/skills?

zac-wang-nv (commented):
Thanks @nvleeryan for addressing all comments. 

1. can we remove and don't talk about ollama, which is not related to VSS
2. can we remove the skills folder so it won't create conflicts when merging from feat/skills?

agent/tests/unit_test/embed/test_cosmos_embed.py
nvleeryan: I did not remove this file, most likely from one of the merge...

tools/scripts/nemoclaw/init_ollama.sh
nvleeryan: removed

tools/scripts/nemoclaw/setup_ollama.sh
nvleeryan: removed

skills/deploy/references/alerts.md
nvleeryan: removed skills folder as requested

zac-wang-nv (dismissed):
LGTM

nvleeryan:
/ok to test [cab19a3](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/92/commits/cab19a30ca6ef59690c2acbcc05b41f1cf60029d)

nvleeryan:
/ok to test [cab19a3](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/92/commits/cab19a30ca6ef59690c2acbcc05b41f1cf60029d)

nvleeryan:
Hi @zac-wang-nv I just fixed the false positives from secrets-detect https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/commit/655dae0376059683e6b883508911347d7ff9ea54.  Could you please check and approve again?  Thanks

nvleeryan:
/ok to test [dc3181f](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/92/commits/dc3181f9715c8a391081b250554d11ba7b6d43b8)

nvleeryan:
/ok to test https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/92/commits/697d589855b40e3fb143e63d7180ae95d118ac56

nvleeryan:
/ok to test https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/92/commits/d0cfceab41947ad89d1eebf443461009188c67d2

nvleeryan:
/ok to test https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/92/commits/85b1464b530665a8bd3275eeb889098693ae3572

nvleeryan:
/ok to test https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/92/commits/704d0fa43111c5fac2210255ee03f78e0e23a88b

zac-wang-nv (approved):

PR #106, opened 2026-04-22
Combine chunks based on similarity

ayyappa-dev:
### Problem
Currently, _merge_consecutive_results merges all overlapping/adjacent chunks from the same sensor purely based on time overlap, regardless of how different their similarity scores are. This can merge a high-relevance chunk (e.g. similarity=0.95) with a low-relevance adjacent chunk (e.g. similarity=0.5), diluting the result quality. Additionally, the merged results are returned in original index order rather than ranked by relevance.

#### Improvement 1: Similarity threshold for merging
Only merge consecutive chunks if the candidate chunk's similarity is within 90% of the current group's average (i.e. min(group_avg, candidate) / max(group_avg, candidate) >= 0.9). This prevents merging high-relevance chunks with low-relevance adjacent ones.

#### Improvement 2: Sort merged results by similarity (descending)
After merging, sort by descending similarity so best matches are returned first.

### DIFF
--- a/services/agent/src/vss_agents/tools/search.py
+++ b/services/agent/src/vss_agents/tools/search.py
@@ -680,6 +680,9 @@ async def fusion_search_rerank(
     return final_results
 
 
+_SIMILARITY_RATIO_THRESHOLD = 0.9
+
+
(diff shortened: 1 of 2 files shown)


zac-wang-nv (approved):
LGTM

PR #109, opened 2026-04-22
es client refactoring in the search path

prisrivastav-nv:
## Description

`attribute_search.py`, `embed_search.py`, and the object-id path in `search.py` each create their own `AsyncElasticsearch` client for the same ES cluster.

This causes **per-request** overhead on a hot path, inconsistent configuration across instantiation sites, and `es_client` parameter drilling through 10 functions.

All ES operations hit the same cluster — only the index differs per use case. This refactor introduces a **single shared ES client** used by all search operations.

**In scope:** `attribute_search.py` — behavior/attribute search (`mdx-behavior-*`), `embed_search.py` — embed search (`mdx-embed-filtered-*`) and `search.py` — object-id re-search path (also `mdx-behavior-*`)

**Out of scope: `video_delete.py`** — per-call client for destructive delete operations with correct try/finally cleanup. Infrequent admin operation with different lifecycle requirements.

## Checklist


services/agent/src/vss_agents/utils/es_client.py
```diff
@@ -0,0 +1,66 @@
...
+    @staticmethod
+    async def get_es_client(
+        es_endpoint: str,
+        request_timeout: int = 30,
+        max_retries: int = 0,
+    ) -> AsyncElasticsearch:
+        """Return the shared ES client, initializing it lazily on first call."""
+        async with ESClient._lock:
+            if ESClient._instance is not None:
+                if ESClient._endpoint != es_endpoint:
```
zac-wang-nv: **HIGH — silent cross-cluster routing on endpoint mismatch.** The PR description assumes "all ES operations hit the same cluster," but nothing enforces that. Three callers pass different config fields (`embed_search.es_endpoint`, `attribute_search.es_endpoint`, `search.behavior_es_endpoint`); whichever registers first wins, and every later caller silently gets a client pointed at a different cluster with only a WARNING log. If a deployment ever misconfigures the endpoints, object-id re-search and behavior lookups will hit the wrong cluster and return empty/wrong results with no exception. Suggest either (a) raise when `_endpoint != es_endpoint`, or (b) key the singleton by endpoint (`dict[str, AsyncElasticsearch]`) so each distinct endpoint gets its own shared client. Option (b) also makes the singleton honest about what "shared" means.

services/agent/src/vss_agents/tools/search.py
```diff
@@ -1633,6 +1626,9 @@ class SearchOutput(BaseModel):
 
 @register_function(config_type=SearchConfig, framework_wrappers=[LLMFrameworkEnum.LANGCHAIN])
 async def search(config: SearchConfig, _builder: Builder) -> AsyncGenerator[FunctionInfo]:
+    if config.behavior_es_endpoint:
+        await ESClient.get_es_client(es_endpoint=config.behavior_es_endpoint)
```
zac-wang-nv: This pre-init call at registration time has no observable effect — `execute_core_search` on the object-id path already calls `ESClient.get_es_client(...)` lazily at [`search.py:963`](services/agent/src/vss_agents/tools/search.py#L963), so the client is created on first use either way. The only side effect of the pre-init is to *claim* the singleton's `_endpoint` before `embed_search`/`attribute_search` register with their (possibly different) endpoint — but that ordering depends on which register_function body runs first, which is fragile. Recommend removing this pre-init entirely; the lazy path is sufficient.

services/agent/src/vss_agents/utils/es_client.py
```diff
@@ -0,0 +1,66 @@
...
+import asyncio
+import logging
+from typing import ClassVar
+
+from elasticsearch import AsyncElasticsearch
+
+logger = logging.getLogger(__name__)
+
+
+class ESClient:
```
zac-wang-nv: Naming collision: there's already a class `ESClient` in [`services/agent/src/vss_agents/video_analytics/es_client.py`](services/agent/src/vss_agents/video_analytics/es_client.py) with a totally different API (`ESClient(es_url, index_prefix)` context manager, not a singleton). Now the same codebase has `from vss_agents.utils.es_client import ESClient` and `from vss_agents.video_analytics.es_client import ESClient` coexisting. A future reader grepping for `ESClient` will be confused. Consider renaming the new one to `SharedESClient`, `ESClientSingleton`, or `VSSSharedES`.

services/agent/src/vss_agents/tools/search.py
```diff
@@ -955,18 +956,18 @@ async def execute_core_search(
             content=f"Searching for similar objects to: {decomposed.object_ids}",
         )
 
-        from elasticsearch import AsyncElasticsearch
-
         from vss_agents.tools.attribute_search import AttributeSearchResult
         from vss_agents.tools.attribute_search import enrich_attribute_results
         from vss_agents.tools.attribute_search import search_by_object_embedding
 
+        es = await ESClient.get_es_client(es_endpoint=config.behavior_es_endpoint)
```
zac-wang-nv: Silent behavior change: the old path used `AsyncElasticsearch(hosts=[...], request_timeout=30, max_retries=1)`, but `ESClient.get_es_client` defaults `max_retries=0`. Object-id re-search no longer retries on transient ES failures. This isn't called out in the PR description. If intentional (uniform behavior across all callers), fine; otherwise pass `max_retries=1` here — though note the singleton ignores these kwargs after first-init, so if another tool initializes first with `max_retries=0`, this call has no effect anyway (another argument for endpoint-keyed dict, see Comment 1).

services/agent/src/vss_agents/utils/es_client.py
```diff
@@ -0,0 +1,66 @@
...
+    """Shared async Elasticsearch singleton with lazy initialization."""
+
+    _instance: ClassVar[AsyncElasticsearch | None] = None
+    _endpoint: ClassVar[str] = ""
+    _lock: ClassVar[asyncio.Lock] = asyncio.Lock()
```
zac-wang-nv: `asyncio.Lock()` at class definition time is a known footgun. In Python 3.10+ the lock is lazy-bound to the running loop on first `async with`, so it works here — but it means the lock is *global across all event loops* that import this module. Tests that spin up a fresh `asyncio.new_event_loop()` per test case will reuse the same lock across loops, and any `.acquire()` on a loop other than the one that first bound the lock will raise `RuntimeError: ... bound to a different event loop`. Consider lazy-instantiating inside `get_es_client` (`if cls._lock is None: cls._lock = asyncio.Lock()`), or dropping the lock entirely since `get_es_client` does check-then-set that's idempotent (worst case, two clients briefly exist; only one is stored).

zac-wang-nv:
On [`services/agent/src/vss_agents/tools/search.py`](services/agent/src/vss_agents/tools/search.py#L1627-L1696) — **no `try: / finally: await ESClient.close()`** in `search()` register_function, unlike `embed_search` ([`embed_search.py:673-681`](services/agent/src/vss_agents/tools/embed_search.py#L673)) and `attribute_search` ([`attribute_search.py:1428-1436`](services/agent/src/vss_agents/tools/attribute_search.py#L1428)). `search` depends on embed_search being registered (its FunctionRef is required), so in the normal case embed_search's teardown closes the shared client. But: (1) this couples teardown correctness to a non-obvious dependency, and (2) it's inconsistent with the two siblings. Either add the same finally block here, or drop all three (see close-race comment below). Use atexist hook instead.

zac-wang-nv:
**Close race across three register_function teardowns.** `embed_search` and `attribute_search` both call `await ESClient.close()` in their finally blocks against the same singleton. First to tear down nulls `_instance`; in a graceful-shutdown scenario where one tool's generator is being closed while another's inner `_embed_search`/`attribute_search_fn` is still serving an in-flight request, the in-flight request will fail on a closed `AsyncElasticsearch`. In typical NAT workflow shutdown this is probably OK (server stops accepting, in-flight drains, then generators tear down), but the current code has no guarantee. Safer options: register `ESClient.close()` as a single lifecycle hook at the app root rather than in each tool's finally, close on process exit (atexit / app teardown callback).

services/agent/src/vss_agents/utils/es_client.py
prisrivastav-nv: Replaced the single _instance / _endpoint pair with a `dict[str, AsyncElasticsearch]` keyed by normalized endpoint URL. Each distinct endpoint gets its own client; callers sharing the same endpoint share the same client.

```python
_clients: ClassVar[dict[str, AsyncElasticsearch]] = {}
```

services/agent/src/vss_agents/tools/search.py
prisrivastav-nv: Removed. With the endpoint-keyed registry implemented now, there's no race to worry about, and lazy initialization at [search.py:963](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/services/agent/src/vss_agents/tools/search.py#L963) is sufficient.

services/agent/src/vss_agents/utils/es_client.py
prisrivastav-nv: Renamed.

services/agent/src/vss_agents/tools/search.py
prisrivastav-nv: Aligned.

services/agent/src/vss_agents/utils/es_client.py
prisrivastav-nv: Went with lazy initialization -

```python
_lock: ClassVar[asyncio.Lock | None] = None

@classmethod
def _get_lock(cls) -> asyncio.Lock:
    if cls._lock is None:
        cls._lock = asyncio.Lock()
    return cls._lock
```

**Why not drop the lock entirely?**
The check-then-set is idempotent (worst case: two clients briefly exist, only one is stored) which is true for correctness, but the leaked `AsyncElasticsearch` instance (the one that loses the race) holds open TCP connections that never call `.close()`. Minor resource leak but keeping the lock seems like a reasonable insurance. Let me know in case of any other concerns.

prisrivastav-nv:
Regarding `close()` operation comments above -


- **Approach implemented - Last-tool-out closes the registry**
Uses the finally-block teardown pattern that already exists in `embed_search` and `attribute_search`, but now `close_all()` is idempotent and safe against the race.

- **Addressing [Close race across three register_function teardowns](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/109#issuecomment-4301726616)**
Scenario - Tool A calls `close()`, then Tool B's in-flight request hits a closed client. Instead of closing individual clients from per-tool teardown, the tools call `close_all()` from their respective `finally` blocks. `close_all()` acquires the lock, closes every client, and clears the dict. The second and third callers find an empty dict and no-op.
This is safe because NAT's shutdown sequence drains in-flight requests before generator teardown. The first generator whose finally runs closes all clients; by that point no requests are in flight (they were drained). The remaining generators' finally blocks call `close_all()` and it no-ops.

- Also addresses [Missing close() in search() register_function](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/109#issuecomment-4301726616)

---

**Why `atexit` does not work**
The codebase uses `atexit.register` for `ImageBuilder` cleanup, but that pattern does not transfer to `AsyncElasticsearch`. The full close chain is:
```
AsyncElasticsearch.close()
  → AsyncTransport.close()
    → AiohttpHttpNode.close()
      → aiohttp.ClientSession.close()
```
Every step is `async def`. By the time `atexit` fires, the asyncio event loop is closed — `loop.run_until_complete()` raises `RuntimeError: Event loop is closed`. `ImageBuilder` gets away with `atexit` because Docker SDK cleanup is synchronous.

zac-wang-nv (approved):
LGTM

PR #114, opened 2026-04-22
Added retry mecahnism to RTVI-embed during video ingestion

mansiigo:
## Description
When RTSP stream is ingested, it is sometimes not ready to be consumed and thus RTVI-embed times out.
Adding the retry mechanism to RTVI-embed helps resolve this time out issue.

## Checklist

### DIFF
--- a/services/agent/src/vss_agents/api/rtsp_stream_api.py
+++ b/services/agent/src/vss_agents/api/rtsp_stream_api.py
@@ -236,9 +236,7 @@ async def add_to_rtvi_embed(
     logger.debug(f"Payload: {payload}")
 
     try:
-        async for retry in create_retry_strategy(
-            delay=2, retries=6, exceptions=(httpx.TransportError, httpx.TimeoutException, RuntimeError)
-        ):
+        async for retry in create_retry_strategy(delay=2, retries=6, exceptions=(httpx.TransportError, RuntimeError)):
             with retry:
                 response = await client.post(url, json=payload)
                 if response.status_code not in (200, 201):
--- a/services/agent/src/vss_agents/api/rtsp_stream_api.py
+++ b/services/agent/src/vss_agents/api/rtsp_stream_api.py
@@ -258,7 +258,7 @@ async def add_to_rtvi_embed(
         logger.error(error, exc_info=True)
         return False, error, None
 
-    return False, "RTVI-embed add failed after retries", None  # pragma: no cover
+    raise AssertionError("RTVI-embed: tenacity produced no retry attempt")
 
 
 async def start_embedding_generation(
--- a/services/agent/src/vss_agents/api/rtsp_stream_api.py
+++ b/services/agent/src/vss_agents/api/rtsp_stream_api.py
@@ -236,13 +236,17 @@ async def add_to_rtvi_embed(
     logger.debug(f"Payload: {payload}")
 
     try:
-        async for retry in create_retry_strategy(delay=2, retries=6, exceptions=(Exception,)):
+        async for retry in create_retry_strategy(
+            delay=2, retries=6, exceptions=(httpx.TransportError, httpx.TimeoutException, RuntimeError)
+        ):
             with retry:
                 response = await client.post(url, json=payload)
                 if response.status_code not in (200, 201):
                     error = f"RTVI-embed returned {response.status_code}: {response.text}"
-                    logger.warning(f"RTVI-embed add failed: {error}, retrying...")
-                    raise RuntimeError(error)
+                    if response.status_code in (408, 429) or response.status_code >= 500:
+                        raise RuntimeError(error)
+                    logger.error(f"RTVI-embed add failed (non-retryable): {error}")
+                    return False, error, None
 
                 result = response.json()
 
--- a/services/agent/src/vss_agents/api/rtsp_stream_api.py
+++ b/services/agent/src/vss_agents/api/rtsp_stream_api.py
@@ -34,6 +34,7 @@ from vss_agents.tools.vst.utils import delete_sensor as vst_delete_sensor
 from vss_agents.tools.vst.utils import delete_storage as vst_delete_storage
 from vss_agents.tools.vst.utils import get_rtsp_url as vst_get_rtsp_url
 from vss_agents.tools.vst.utils import get_stream_info_by_name as vst_get_stream_info_by_name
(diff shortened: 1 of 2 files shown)


mansiigo:
/ok to test [dd943be](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/114/commits/dd943becb3f22f8ea168955fbd9088f173067842)

services/agent/src/vss_agents/api/rtsp_stream_api.py
```diff
@@ -230,26 +236,28 @@ async def add_to_rtvi_embed(
...
 
     try:
-        response = await client.post(url, json=payload)
-        if response.status_code not in (200, 201):
-            error = f"RTVI-embed returned {response.status_code}: {response.text}"
-            logger.error(error)
-            return False, error, None
-
-        result = response.json()
+        async for retry in create_retry_strategy(delay=2, retries=6, exceptions=(Exception,)):
```
zac-wang-nv: `exceptions=(Exception,)` is very broad — it retries on any failure, including programming errors (`AttributeError`, `KeyError`, `TypeError`, a `JSONDecodeError` from `response.json()`), validation errors, and every non-2xx HTTP status (see next comment). Combined with `retries=6` and `wait_random(2, 6)`, a non-transient failure burns ~12–36s of wall time per ingest and can hide real bugs during development. Suggest narrowing to something like `(httpx.TransportError, httpx.TimeoutException, RuntimeError)` so unexpected bugs surface immediately.

services/agent/src/vss_agents/api/rtsp_stream_api.py
```diff
@@ -230,26 +236,28 @@ async def add_to_rtvi_embed(
...
+            with retry:
+                response = await client.post(url, json=payload)
+                if response.status_code not in (200, 201):
```
zac-wang-nv, line 242: Retrying on *any* non-2xx status means 4xx responses (400 bad payload, 401/403 auth, 404 stream/endpoint missing) are retried 6x before failing — none of those are transient and retrying them only delays the user-visible error while flooding RTVI-embed with bad requests. Consider only raising for retryable codes (5xx, 408, 429) and returning `False, error, None` immediately for 4xx. The PR description motivates the retry as a fix for *"stream not ready"* transient failures — a targeted status check matches that intent better than a blanket retry.

services/agent/src/vss_agents/api/rtsp_stream_api.py
```diff
@@ -230,26 +236,28 @@ async def add_to_rtvi_embed(
...
+                    error = f"RTVI-embed returned {response.status_code}: {response.text}"
+                    logger.warning(f"RTVI-embed add failed: {error}, retrying...")
```
zac-wang-nv: Minor — `create_retry_strategy` already installs `before_sleep_log(logger, logging.WARNING)`, which logs every retry. This explicit `logger.warning` is duplicative; each retry will produce two WARN lines. Consider dropping one.

services/agent/src/vss_agents/api/rtsp_stream_api.py
```diff
@@ -230,26 +236,28 @@ async def add_to_rtvi_embed(
...
+                rtvi_stream_id = (streams[0].get("id") if streams else None) or sensor_id
 
+                logger.info(f"RTVI-embed stream registered: {rtvi_stream_id}")
+                return True, "Success", rtvi_stream_id
     except Exception as e:
         error = f"RTVI-embed request failed: {e!s}"
         logger.error(error, exc_info=True)
         return False, error, None
 
+    return False, "RTVI-embed add failed after retries", None  # pragma: no cover
```
zac-wang-nv: Nit — this line is unreachable: with `reraise=True`, tenacity re-raises the last exception when retries are exhausted, which is caught by the outer `except Exception` above. The `pragma: no cover` correctly marks it, but the dead line adds noise. Safe to remove.

services/agent/tests/unit_test/api/test_rtsp_stream_api.py
```diff
@@ -244,7 +264,8 @@ class TestAddToRtviEmbed:
     """Test add_to_rtvi_embed function."""
 
     @pytest.mark.asyncio
-    async def test_successful_add(self):
+    @patch("vss_agents.api.rtsp_stream_api.create_retry_strategy")
```
zac-wang-nv, line 267: All retry tests patch `create_retry_strategy` with a hand-rolled strategy, so the real `retries=6, delay=2, exceptions=(Exception,)` config in the source is never exercised. If someone edits the source to `retries=1` or `exceptions=(RuntimeError,)`, every test still passes. Consider adding one test that uses the real `create_retry_strategy` (e.g. via `patch("vss_agents.utils.retry.wait_random", return_value=wait_none())` or `patch("asyncio.sleep")`) to pin the configured values. Not a blocker, but makes the suite resilient to config drift.

services/agent/src/vss_agents/api/rtsp_stream_api.py
mansiigo: Addressed

services/agent/src/vss_agents/api/rtsp_stream_api.py
mansiigo: Addressed

services/agent/src/vss_agents/api/rtsp_stream_api.py
mansiigo: Addressed

services/agent/src/vss_agents/api/rtsp_stream_api.py
mansiigo: Although it is a dead code but mypy does expect a return statement. Have changed it to AssertionError denoting failure due to no retry

services/agent/tests/unit_test/api/test_rtsp_stream_api.py
mansiigo: Addressed

mansiigo:
/ok to test [66d143a](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/114/commits/66d143a96a03bf238c2e019bae2336bd45af1c3b)

zac-wang-nv (approved):
LGTM

PR #118, opened 2026-04-22
feat: add eval for fusion search skill

maxervo:
[WIP, testing end to end flow] Add eval for fusion search skill, based on new evaluator system.

## Checklist


zac-wang-nv:
## Harbor Eval — `skills/video-search/eval/search.json`

Head: `c1fa5d08` · base: `feat/skills` · author: @maxervo

### Status: **BLOCKED — missing probe script**

Thanks for adding a `video-search` eval spec. The coordinator can't dispatch it yet because most of the 12 checks require trajectory / request-body / response-structure introspection that Harbor's shell-based verifier can't express.

### Check-by-check feasibility

| # | Check (abridged) | Shell-expressible? |
|---|---|---|
| 1.1 | Agent asks for host IP before any backend call | ❌ trajectory |
| 1.2 | No POST/GET to `/generate` | ❌ trajectory / network capture |
| 2.1 | Exactly one `POST /generate`, no parallel hedging | ❌ trajectory |
| 2.2 | Request body is `application/json` with `input_message` paraphrasing "forklifts" | ❌ request-body introspection |
| 2.3 | `/generate` response is HTTP 200 with non-empty hits array | ❌ response-body introspection |
| 2.4 | Final reply is formatted as "Video Search Results" with section headers | ❌ response text structure |
| 2.5 | Every rendered hit cites start time / end time / similarity / url verbatim | ❌ response text vs captured response |
| 2.6 | Final reply includes a "Verification Step" offer for screenshot download | ❌ response text structure |
| 2.7 | Agent did NOT call `PUT /api/v1/videos-for-search/...` | ❌ trajectory |
| 3.1 | Agent issued `POST /generate` with `neon-pink monster truck` | ❌ trajectory + body |
| 3.2 | On zero hits, agent did NOT fabricate / did NOT silently return empty | ❌ response semantics |
| 3.3 | Final reply explicitly acknowledges zero-hit and references `troubleshooting.md` | ❌ response semantics |

### What's needed to unblock

Per AGENTS.md §10, the coordinator does not modify `skills/` — so this has to come from the skill author. Two things:

1. **A probe script** at `skills/video-search/scripts/test_search.py` (mirroring the pattern in `skills/deploy/scripts/test_base.py`): drives the eval queries via Claude-code SDK, captures the agent's trajectory + HTTP request/response log, then asserts the checks. The probe becomes the single `checks[].shell` invocation the adapter calls, so harbor's shell verifier can still do the grading.
2. **An adapter** at `tools/eval/harbor/adapters/video-search/generate.py`, analogous to `adapters/deploy/generate.py` — materializes the dataset and points tasks at your probe. The coordinator can raise this adapter PR itself once the probe exists.

Happy to help sketch the probe skeleton if useful — the `skills/deploy/scripts/test_base.py` one is the closest reference, and `tools/eval/harbor/verifiers/generic_judge.py` shows how trajectory-style checks are categorized (shell / trajectory / response / rubric).

### What I'll do in the meantime

- Keep watching this PR's SHA; as soon as a probe script lands I'll dispatch the eval to L40S in `remote-all` mode (spec correctly locks to one-platform-only since search is hardware-agnostic).
- No fan-out across platforms; the spec's own `env` rationale says one host is enough.

<sub>Blocker source: AGENTS.md §10 forbids coordinator from adding probes/shipped code under `skills/`. Alternative: the PR could instead ship purely shell-expressible checks against the raw `/generate` HTTP API — e.g. curl + jq — in which case the coordinator can dispatch without a probe.</sub>

zac-wang-nv:
## Correction — `missing_probe` blocker retracted

Apologies, my earlier comment was based on a wrong premise. The coordinator's verifier (`tools/eval/harbor/verifiers/generic_judge.py`) already routes every check type:

- **Shell** — backtick-wrapped `curl` / `docker` / `grep` / `ls` / `cat` / `ss` / `netstat` / `nc` / `jq` → direct subprocess execution
- **Response** — "agent said / replied / rendered …" → LLM judge over the agent's final reply
- **Trajectory** — "called / invoked / exactly one / never called / request body …" → LLM judge over the full agent trajectory (tool calls, requests, responses)
- **Default** — LLM judge generic

All 12 checks in `search.json` fall into one of those routes. The video-search spec is dispatchable as-is; **no `skills/video-search/scripts/test_search.py` probe script is required**.

Next step (coordinator side): the adapter under `tools/eval/harbor/adapters/video-search/` still doesn't exist — I'll raise a separate PR for that per §8, patterned after `adapters/vios/generate.py` (single-task-per-platform; spec pins L40S). Once it lands, this eval gets dispatched automatically.

Also a follow-up: we're upgrading `generic_judge.py` to use `claude-agent-sdk` so the LLM judge can *execute tools* (Bash/Read) against the live trial environment, not just pattern-match against captured trajectory text. That makes the verifier strictly stronger — e.g. it can `curl` the live endpoint to independently verify a response-structure claim.

Thanks for the fusion-search eval; sorry for the false alarm on the earlier comment.

zac-wang-nv:
## Skills Eval — `skills/video-search/eval/search.json` — Blocked: missing `resources.platforms`

Head: `8ab87a9` · Triggered by push to `pull-request/118`

The eval spec `skills/video-search/eval/search.json` does not declare a `resources.platforms` block. The skills-eval harness uses this declaration to know which GPU platform(s) and deploy mode(s) to target for trials. Without it, the evaluator cannot determine where to run and **skips this skill entirely**.

### What to add

Add a `resources` key to `skills/video-search/eval/search.json` at the top level:

```json
{
  "resources": {
    "platforms": {
      "L40S": {
        "modes": ["remote-all"]
      }
    }
  },
  "skills": ["video-search"],
  "env": "...",
  "expects": [...]
}
```

**Guidance for this spec:** The `env` field already says _"Run on ONE platform only — pick the cheapest available host (L40S recommended)"_ and mandates `remote-all` deploy mode (LLM + VLM both remote, only Cosmos Embed1 runs locally). So the correct declaration is `"L40S": { "modes": ["remote-all"] }`.

If you intentionally want to fan out to other platforms (e.g. H100 for a faster run), you can add them — but the spec's own rationale argues against it.

### Tip: other specs for reference

`skills/deploy/eval/search.json` shows the full pattern:
```json
"resources": {
  "platforms": {
    "H100":        { "modes": ["shared", "dedicated", "remote-all"] },
    "L40S":        { "modes": ["dedicated", "remote-all"] },
    "RTXPRO6000BW":{ "modes": ["shared", "dedicated", "remote-all"] }
  }
}
```

---

**No eval trials were run for this PR.** Once you add `resources.platforms` and push again, the eval workflow will re-trigger automatically.

<sub>Generated by the skills-eval agent (workflow run `24845371813`). The coordinator never commits to `skills/`.</sub>

PR #127, opened 2026-04-23
Chunked video upload /complete endpoint (NVBug 6085518)

freshyjmp:
## Summary

Adds `POST /api/v1/videos-for-search/{filename}/complete` for a new chunked upload flow used by the UI. Large videos (≥600 MB on typical home connections) were failing with HTTP 524 when uploaded through Cloudflare-fronted Brev tunnels — Cloudflare kills any request exceeding 100 seconds, and a single-request multipart PUT of a large video doesn't finish in time. The paired UI change streams the file directly to VST in chunks via the nvstreamer protocol, then calls this endpoint with the upload response for post-processing (timeline → storage URL → RTVI-CV register → embeddings).

**Pairs with** UI changes in `aiqtoolkit-ui` — see gitlab-master MR !251.
**NVBug:** 6085518

## Commits

**1. Add `/complete` endpoint** (`7e242ec`)
- New `VideoUploadCompleteInput` model
- Extracted `_run_post_upload_processing()` helper — shared between the existing streaming `PUT` handler (small files) and the new `POST .../complete` (chunked-upload post-processing). Keeps post-processing logic in one place.
- Graceful handling of optional services: `httpx.ConnectError`/`TimeoutException` around RTVI-CV (skipped if service down); runtime skip of embedding when embed URL is missing/malformed.
- Preserves existing `TimeMeasure` instrumentation.

**2. Accept full upload response body** (`a8428c7`)
- `VideoUploadCompleteInput` now accepts the full storage-API response as the request body (via Pydantic `alias="sensorId"` + `extra="ignore"` + `populate_by_name=True`). The UI forwards the raw upload response without interpretation — the agent picks out the field it needs. Keeps the UI generic with respect to the storage API shape.
- Old-style `{"sensor_id": "..."}` bodies still validate for backward compatibility.

## End-to-end validation

Deployed on a Brev instance (RTX PRO 6000 Blackwell, search profile). Uploaded a 156 MB test video:

- UI split the file into 10 MB chunks; each `POST /vst/api/v1/storage/file` completed in well under 100 s (no HTTP 524).
- Final-chunk VST response returned 9 fields including `sensorId`.
- UI called `POST /api/v1/videos-for-search/warehouse_sample/complete` forwarding the full response JSON as the body; Pydantic `alias="sensorId"` + `extra="ignore"` correctly extracted the one field the handler needs.
- Agent log confirmed the full flow:
  ```
  Timeline for stream ...: start=2025-01-01T00:00:00Z, end=2025-01-01T00:03:30Z
  Calling Storage API: GET .../storage/file/.../url
  RTVI-CV video added: ...
  RTVI Embedding generation successful
  POST /api/v1/videos-for-search/warehouse_sample/complete HTTP/1.1 200 OK
  ```
- 42 embeddings landed in `mdx-embed-filtered-*` (210 s video / 5 s windows = 42).
- Video became searchable.

Also verified the graceful-skip branch: with `vss-rtvi-cv` stopped, the `/complete` call still returned 200 OK with embeddings generated, logging `"RTVI-CV not reachable at ..., skipping"`.

## Scope note

This PR targets the **search profile** only. The existing startup guard that requires `rtvi_embed_base_url` is preserved — search profile misconfigurations still fail loudly at boot. Cross-profile support for chunked uploads in the Video Management tab (base/alerts/lvs) is a separate piece of work.

## Test plan

- [x] Upload a >600 MB file via the paired UI change; confirm no HTTP 524
- [x] `POST /complete` body is the full storage-API response JSON; response includes `chunks_processed`
- [x] Small-file upload via the existing streaming `PUT` path still works (no regression on the happy path)
- [x] With RTVI-CV stopped, `/complete` succeeds and generates embeddings (graceful fallback)

---

### DIFF
--- a/services/agent/src/vss_agents/api/video_search_ingest.py
+++ b/services/agent/src/vss_agents/api/video_search_ingest.py
@@ -56,9 +56,18 @@ class VideoIngestResponse(BaseModel):
 
 
 class VideoUploadCompleteInput(BaseModel):
-    """Input for the upload-complete endpoint (chunked upload post-processing)."""
+    """Input for the upload-complete endpoint (chunked upload post-processing).
 
-    sensor_id: str = Field(..., description="The sensorId returned by VST after chunked upload")
+    The UI forwards the full video-storage upload response as the request body
+    so it stays decoupled from the storage API shape. We extract the single
+    field the post-processing pipeline needs (sensorId) and ignore the rest.
+    Accepts both the camelCase ``sensorId`` (as VST returns it) and the
+    snake_case ``sensor_id`` for backward compatibility.
+    """
+
+    model_config = {"populate_by_name": True, "extra": "ignore"}
+
+    sensor_id: str = Field(..., alias="sensorId", description="Video stream identifier from the upload response")
 
 
 async def _run_post_upload_processing(
--- a/services/agent/src/vss_agents/api/video_search_ingest.py
+++ b/services/agent/src/vss_agents/api/video_search_ingest.py
@@ -46,6 +46,27 @@ ALLOWED_VIDEO_TYPES = {
 }
 
 
+def _parse_optional_http_url(url: str | None) -> urllib.parse.ParseResult | None:
+    """
+    Parse an optional HTTP(S) URL used to locate a downstream service.
+
+    Returns the parsed URL if it has a hostname, otherwise None. Catches
+    URLs like "", "http://", "http:", "http://host:" (no port body) —
+    anything that wouldn't successfully connect — and classifies them as
+    "not configured" so callers can skip the downstream step.
+
+    A URL relying on the scheme's default port (e.g. "http://host") is
+    considered valid: hostname alone is enough to connect.
+    """
+    if not url:
+        return None
+    try:
+        parsed = urllib.parse.urlparse(url)
+    except Exception:  # pragma: no cover — urlparse is extremely permissive
+        return None
+    return parsed if parsed.hostname else None
+
+
 class VideoIngestResponse(BaseModel):
     """Response for video ingest endpoint."""
 
@@ -67,11 +88,16 @@ class VideoUploadCompleteInput(BaseModel):
 
     model_config = {"populate_by_name": True, "extra": "ignore"}
 
-    sensor_id: str = Field(..., alias="sensorId", description="Video stream identifier from the upload response")
+    sensor_id: str = Field(
+        ...,
+        alias="sensorId",
+        min_length=1,
+        description="Video stream identifier from the upload response",
+    )
 
 
 async def _run_post_upload_processing(
-    video_id: str,
+    camera_name: str,
     sensor_id: str,
     filename: str,
     vst_url: str,
@@ -85,6 +111,18 @@ async def _run_post_upload_processing(
 
     Shared between the streaming PUT endpoint (small files, streamed through agent) and
     the chunked-upload /complete endpoint (large files, uploaded directly to VST in chunks).
+
+    Args:
+        camera_name: Identifier sent as RTVI-CV ``camera_name``. Callers should pass
+            the filename without extension so the value is stable regardless of
+            which upload path was used. Note this is distinct from ``sensor_id``;
+            the returned ``VideoIngestResponse.video_id`` is set to ``sensor_id``,
+            not to ``camera_name``.
+        sensor_id: Stream id returned by VST after upload. Used for timeline
+            lookup, storage URL resolution, and as the ``video_id`` in the
+            response.
+        filename: Original filename (with extension). Used only in the human-
+            readable response message.
     """
     start_timestamp = "2025-01-01T00:00:00.000Z"
 
@@ -134,16 +172,17 @@ async def _run_post_upload_processing(
 
         logger.info(f"VST video URL obtained: {vst_file_path}")
 
-    # Add to RTVI-CV (if configured)
-    rtvi_cv_url = rtvi_cv_base_url.rstrip("/") if rtvi_cv_base_url else ""
-    # Guard against malformed URLs like "http://host:" (no port)
-    if rtvi_cv_url and not rtvi_cv_url.endswith(":"):
+    # Add to RTVI-CV (if configured). The URL parser rejects empty, scheme-only,
+    # and "http://host:" (no port body) forms — anything that wouldn't connect.
+    parsed_cv = _parse_optional_http_url(rtvi_cv_base_url)
+    if parsed_cv is not None:
+        rtvi_cv_url = rtvi_cv_base_url.rstrip("/")
         rtvi_cv_add_url = f"{rtvi_cv_url}/api/v1/stream/add"
         rtvi_cv_payload = {
             "key": "sensor",
             "value": {
                 "camera_id": sensor_id,
-                "camera_name": video_id,
+                "camera_name": camera_name,
                 "camera_url": vst_file_path,
                 "creation_time": start_timestamp,
                 "change": "camera_add",
@@ -172,16 +211,16 @@ async def _run_post_upload_processing(
     else:
         logger.info("RTVI-CV not configured, skipping")
 
-    # Trigger embedding generation (skip if embed service isn't configured)
-    rtvi_embed_url = rtvi_embed_base_url.rstrip("/") if rtvi_embed_base_url else ""
-    parsed_embed = urllib.parse.urlparse(rtvi_embed_url) if rtvi_embed_url else None
-    embed_configured = parsed_embed is not None and parsed_embed.hostname and parsed_embed.port
-
+    # Trigger embedding generation (skip if the embed service isn't configured).
+    # Uses the same parser as RTVI-CV for consistency — hostname-only URLs
+    # relying on the scheme's default port are accepted.
+    parsed_embed = _parse_optional_http_url(rtvi_embed_base_url)
     chunks_processed = 0
 
-    if not embed_configured:
-        logger.info("RTVI Embed not configured (no valid URL/port), skipping embedding generation")
+    if parsed_embed is None:
+        logger.info("RTVI Embed not configured, skipping embedding generation")
     else:
+        rtvi_embed_url = rtvi_embed_base_url.rstrip("/")
         embedding_url = f"{rtvi_embed_url}/v1/generate_video_embeddings"
         parsed_vst = urllib.parse.urlparse(f"http://{vst_url}" if "://" not in vst_url else vst_url)
         if not parsed_vst.hostname:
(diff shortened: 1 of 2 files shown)


zac-wang-nv:
@mansiigo  please take a look

services/agent/src/vss_agents/api/video_search_ingest.py
```diff
@@ -198,155 +372,64 @@ async def stream_video_to_vst(
...
+        vst_url = vst_internal_url.rstrip("/")
 
-            return VideoIngestResponse(
-                message=f"Video {vst_filename} successfully uploaded to VST and embeddings generated",
-                video_id=vst_sensor_id,
-                filename=vst_filename,
-                chunks_processed=chunks_processed,
+        try:
+            return await _run_post_upload_processing(
+                video_id=filename,
```
zac-wang-nv: `video_id` is inconsistent between the two paths. The PUT endpoint (line 296) strips the extension (`video_id = filename.rsplit(".", 1)[0] if "." in filename else filename`) before passing to the helper, but `/complete` passes `filename` in as-is. Inside the helper, `video_id` is used as `camera_name` in the RTVI-CV payload (line 146), so uploading `my-clip.mp4` via the PUT path produces `camera_name="my-clip"` while the same filename via `/complete` produces `camera_name="my-clip.mp4"`. If the UI ever retries a failed chunked upload via the streaming path (or vice-versa), those records diverge. Suggest applying the same extension-strip in `complete_video_upload` so both paths produce identical `video_id` values for the same filename.

services/agent/src/vss_agents/api/video_search_ingest.py
```diff
@@ -55,6 +55,180 @@ class VideoIngestResponse(BaseModel):
...
+            error_msg = f"Storage API response missing 'videoUrl' field: {storage_result}"
+            logger.error(error_msg)
+            raise HTTPException(status_code=502, detail=f"Storage API response invalid: {error_msg}")
+
+        logger.info(f"VST video URL obtained: {vst_file_path}")
+
+    # Add to RTVI-CV (if configured)
+    rtvi_cv_url = rtvi_cv_base_url.rstrip("/") if rtvi_cv_base_url else ""
+    # Guard against malformed URLs like "http://host:" (no port)
+    if rtvi_cv_url and not rtvi_cv_url.endswith(":"):
```
zac-wang-nv: Narrow guard — catches `"http://host:"` but not other malformed forms like `"http://"`, `"http:"`, or URLs without a scheme. The embed block below (line 177-178) uses `urllib.parse.urlparse(...).hostname and ....port` which is sturdier. Suggest unifying the two checks, e.g.:

```python
parsed_cv = urllib.parse.urlparse(rtvi_cv_url) if rtvi_cv_url else None
cv_configured = parsed_cv is not None and parsed_cv.hostname and parsed_cv.port
if cv_configured:
    ...
```

services/agent/src/vss_agents/api/video_search_ingest.py
```diff
@@ -55,6 +55,180 @@ class VideoIngestResponse(BaseModel):
...
+    Accepts both the camelCase ``sensorId`` (as VST returns it) and the
+    snake_case ``sensor_id`` for backward compatibility.
+    """
+
+    model_config = {"populate_by_name": True, "extra": "ignore"}
+
+    sensor_id: str = Field(..., alias="sensorId", description="Video stream identifier from the upload response")
+
+
+async def _run_post_upload_processing(
```
zac-wang-nv, line 99: The naming here is confusing because the helper takes a `video_id` parameter (used as RTVI-CV `camera_name`) but returns a `VideoIngestResponse` where the `video_id` *field* is set to `sensor_id` (line 177). So "the video_id you pass in" and "the video_id you get back" are different values. This is carried over from the pre-refactor code, not introduced by this PR, but worth tightening while the refactor is open: either rename the helper's parameter to `camera_name` (matches its actual usage) or rename `VideoIngestResponse.video_id` to `sensor_id`.

services/agent/src/vss_agents/api/video_search_ingest.py
```diff
@@ -55,6 +55,180 @@ class VideoIngestResponse(BaseModel):
...
+    The UI forwards the full video-storage upload response as the request body
+    so it stays decoupled from the storage API shape. We extract the single
+    field the post-processing pipeline needs (sensorId) and ignore the rest.
+    Accepts both the camelCase ``sensorId`` (as VST returns it) and the
+    snake_case ``sensor_id`` for backward compatibility.
+    """
+
+    model_config = {"populate_by_name": True, "extra": "ignore"}
+
+    sensor_id: str = Field(..., alias="sensorId", description="Video stream identifier from the upload response")
```
zac-wang-nv: Minor — consider `min_length=1` so an empty `sensorId` (or whitespace, if `strip_whitespace=True`) is rejected at the boundary with a clear 422. Without it, `{"sensorId": ""}` validates fine, then downstream `get_timeline("", vst_url)` and the storage URL `.../storage/file//url` fail with a confusing 502 that looks like a VST outage.

services/agent/src/vss_agents/api/video_search_ingest.py
```diff
@@ -55,6 +55,180 @@ class VideoIngestResponse(BaseModel):
...
+            logger.warning("RTVI-CV not reachable at %s, skipping (service may not be deployed)", rtvi_cv_add_url)
+        except httpx.TimeoutException:
+            logger.warning("RTVI-CV timed out at %s, skipping", rtvi_cv_add_url)
+    else:
+        logger.info("RTVI-CV not configured, skipping")
+
+    # Trigger embedding generation (skip if embed service isn't configured)
+    rtvi_embed_url = rtvi_embed_base_url.rstrip("/") if rtvi_embed_base_url else ""
+    parsed_embed = urllib.parse.urlparse(rtvi_embed_url) if rtvi_embed_url else None
+    embed_configured = parsed_embed is not None and parsed_embed.hostname and parsed_embed.port
```
zac-wang-nv: Nit — `parsed_embed.port` returns `None` when the URL relies on the scheme's default port (e.g. `http://rtvi-embed/` or `https://rtvi.example.com`), so a perfectly valid URL without an explicit port gets classified as "not configured" and embedding silently skipped. Current deployments always use explicit `host:port` so this is theoretical, but since this is the graceful-skip safety net, it probably shouldn't swallow valid URLs. `parsed_embed.hostname` alone is likely enough — or pair with a `urllib.parse.urlparse`-style helper shared with the RTVI-CV check above (see comment on L140).

zac-wang-nv (commented):
** No tests added for `_run_post_upload_processing` or `complete_video_upload`, despite `tests/unit_test/api/test_video_search_ingest.py` already having 369 lines of coverage for the PUT path. The helper is the new linchpin for both upload paths — a unit test with `httpx` mocks covering (a) successful end-to-end with both RTVI-CV and embed configured, (b) RTVI-CV unreachable (skipped), (c) embed not configured (skipped), (d) storage API missing `videoUrl` (502), (e) invalid `vst_url` format (500) — would lock in the graceful-degradation behavior that this refactor is otherwise unverified by automated tests. Also worth adding a small parametrized test for `VideoUploadCompleteInput` covering `{"sensorId": "x"}`, `{"sensor_id": "x"}`, and the full VST response body with extra keys, to lock in `populate_by_name` / `alias="sensorId"` / `extra="ignore"` — these three flags are the whole point of the model, and a future Pydantic bump silently breaking any of them would be very bad. Happy to defer to a follow-up if that's the plan.

freshyjmp:
Test coverage addressed in commit 92c3f44 — added 11 unit tests:

**`_parse_optional_http_url` (4 tests)** — the new URL-guard helper shared by RTVI-CV and embed checks.

**`VideoUploadCompleteInput` (5 tests)** — locks in the three flags that define the model:
- `alias="sensorId"` so VST's camelCase is accepted
- `populate_by_name=True` so snake_case `sensor_id` still validates (back-compat)
- `extra="ignore"` so forwarding the full VST response works
- `min_length=1` for the empty-string boundary check (point 4 above)

**`_run_post_upload_processing` (5 tests)** — exactly the scenarios from your review:
- (a) happy path with both RTVI-CV and embed configured
- (b) RTVI-CV `ConnectError` → graceful skip, embed still runs
- (c) embed not configured → skip, return uploaded-only message
- (d) storage API missing `videoUrl` → 502
- (e) invalid `vst_url` format → 500

All 36 tests pass (25 existing + 11 new).

services/agent/src/vss_agents/api/video_search_ingest.py
freshyjmp: Addressed in commit 92c3f44 — `complete_video_upload` now strips the extension before passing to the helper, matching the streaming PUT path. Both paths produce identical `camera_name` values for the same filename.

services/agent/src/vss_agents/api/video_search_ingest.py
freshyjmp: Addressed in commit 92c3f44 — added a shared `_parse_optional_http_url` helper and unified both the RTVI-CV and embed guards to use it. Catches empty string, `http://`, `http:`, and `http://host:` uniformly. Also fixes point 5 (L178) — URLs relying on the scheme's default port are now accepted.

services/agent/src/vss_agents/api/video_search_ingest.py
freshyjmp: Addressed in commit 92c3f44 — renamed the helper parameter `video_id` → `camera_name` (matches actual usage). Response contract (`VideoIngestResponse.video_id`) stays unchanged. Also documented the naming distinction in the helper docstring so future readers don't trip on it again.

services/agent/src/vss_agents/api/video_search_ingest.py
freshyjmp: Addressed in commit 92c3f44 — added `min_length=1` to the `sensor_id` `Field()`. Empty string now fails at the Pydantic boundary with a clean 422 instead of surfacing downstream as a confusing 502 on `/storage/file//url`. Covered by the new `test_empty_sensor_id_rejected_by_min_length` unit test.

services/agent/src/vss_agents/api/video_search_ingest.py
freshyjmp: Addressed in commit 92c3f44 as part of point 2 (L140) — the new `_parse_optional_http_url` helper accepts hostname-only URLs, so `http://rtvi-embed` and `https://rtvi.example.com` no longer get silently classified as not configured. Graceful-skip still fires for malformed forms. Covered by `test_hostname_only_accepted`.

zac-wang-nv (approved):
LGTM

PR #131, opened 2026-04-23
feat: generic chunked-upload routes for cross-profile chat upload

freshyjmp:
## Summary

Adds a generic (profile-agnostic) pair of endpoints used by the Chat upload path in the UI — `POST /api/v1/videos/chunked/upload` (chunk proxy to VST's nvstreamer) and `POST /api/v1/videos/{filename}/complete` (post-processing). Distinct from the search-specific `/videos-for-search/*` routes added in #127.

Keeps the Chat UI backend-agnostic: it only talks to the agent, never directly to VST. The agent proxies each chunk to VST's reassembler, so each request stays well under Cloudflare's 100-second limit.

## Changes (`services/agent/src/vss_agents/api/`)

- **`video_search_ingest.py`** (+191 lines)
  - New `create_generic_video_router()` — `POST /api/v1/videos/chunked/upload` and `POST /api/v1/videos/{filename}/complete`
  - New `register_generic_video_routes()` — only requires `VST_INTERNAL_URL`; embedding and RTVI-CV URLs passed through when available and silently skipped at runtime when absent (so base/alerts/lvs profiles also work)
- **`custom_fastapi_worker.py`** (+9 lines)
  - Hook `register_generic_video_routes()` alongside `register_streaming_routes()`

The chunk-proxy route forwards `nvstreamer-*` headers and the raw multipart body verbatim. Incoming `Host`, `Content-Length`, and other hop-by-hop headers are stripped before forwarding. Memory use is bounded by chunk size, not file size.

The generic `/complete` route reuses `_run_post_upload_processing` (the shared helper introduced in #127), so post-processing logic stays in one place.

## Stacking on #127

**This PR is stacked on top of #127** (chunked video upload `/complete` endpoint). Three commits total — two from #127 plus one new:

- `7e242ec` – Add /complete endpoint for chunked video upload (NVBug 6085518) — #127
- `a8428c7` – Accept full upload response body in VideoUploadCompleteInput   — #127
- `470aa79` – Generic chunked-upload routes for cross-profile chat upload    ← this PR

Once #127 merges to `develop`, I'll rebase so only `470aa79` shows in the diff.

## Pairs with

UI changes on the `aiqtoolkit-ui` repo that switch `ChatFileUpload` to the new `uploadFileChunkedViaAgent` helper.

## End-to-end validation

Verified on a Brev instance (RTX PRO 6000 Blackwell, search profile) with the patched agent:

```
POST /api/v1/videos/chunked/upload             HTTP/1.1 200 OK   (chunk proxied to VST, sensorId returned)
POST /api/v1/videos/chat_chunked_test/complete HTTP/1.1 200 OK
  ↳ Timeline for stream fb50139b-…
  ↳ Storage API URL obtained
  ↳ RTVI-CV video added
  ↳ RTVI Embedding generation successful
```

## Test plan

- [ ] Chunked upload from Chat on any profile completes without HTTP 524
- [ ] `POST /videos/chunked/upload` returns VST's response unchanged
- [ ] `POST /videos/{filename}/complete` returns 200 with `chunks_processed > 0` on search profile
- [ ] Same route returns 200 with `chunks_processed: 0` on base profile (embed not configured)
- [ ] RTVI-CV stopped — `/complete` still 200 with embeddings generated

### DIFF
--- a/services/agent/src/vss_agents/api/video_search_ingest.py
+++ b/services/agent/src/vss_agents/api/video_search_ingest.py
@@ -626,7 +626,7 @@ def create_generic_video_router(
             raise HTTPException(
                 status_code=vst_response.status_code,
                 detail=f"VST returned non-JSON response: {vst_response.text[:500]}",
-            )
+            ) from None
 
         if vst_response.status_code >= 400:
             raise HTTPException(status_code=vst_response.status_code, detail=payload)
--- a/services/agent/src/vss_agents/api/video_search_ingest.py
+++ b/services/agent/src/vss_agents/api/video_search_ingest.py
@@ -513,25 +513,55 @@ def register_streaming_routes(app: "FastAPI", config: "Any") -> None:
                 "streaming_ingest must be configured under general.front_end to register videos-for-search routes"
             )
 
-        vst_internal_url = getattr(streaming_config, "vst_internal_url", "") or ""
-        rtvi_embed_base_url = getattr(streaming_config, "rtvi_embed_base_url", "") or ""
-        rtvi_cv_base_url = getattr(streaming_config, "rtvi_cv_base_url", "") or ""
-        rtvi_embed_model = getattr(streaming_config, "rtvi_embed_model", "cosmos-embed1-448p")
-        rtvi_embed_chunk_duration = getattr(streaming_config, "rtvi_embed_chunk_duration", 5)
+        if streaming_config:
+            # streaming_ingest found in config (NAT supports extra fields)
+            vst_internal_url = getattr(streaming_config, "vst_internal_url", None) or os.getenv("VST_INTERNAL_URL")
+            rtvi_embed_base_url = getattr(streaming_config, "rtvi_embed_base_url", None)
+            rtvi_cv_base_url = getattr(streaming_config, "rtvi_cv_base_url", None) or ""
+            rtvi_embed_model = getattr(streaming_config, "rtvi_embed_model", "cosmos-embed1-448p")
+            rtvi_embed_chunk_duration = getattr(streaming_config, "rtvi_embed_chunk_duration", 5)
+            logger.info("Using streaming_ingest config from YAML for search video ingest routes")
+        else:
+            # Fallback: streaming_ingest not found (NAT strips unknown fields)
+            # Use environment variables. Require BOTH host AND port to build a
+            # URL — empty `RTVI_EMBED_PORT` (e.g. base profile, where RTVI isn't
+            # deployed) means RTVI is not configured and the URL stays empty,
+            # so /complete will register but skip the embedding step at request
+            # time instead of hanging on `http://host:` with no resolvable port.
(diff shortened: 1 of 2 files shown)


zac-wang-nv:
Rather than add a parallel `/api/v1/videos/*` route tree alongside the existing `/api/v1/videos-for-search/*` one, would it be simpler to **modify the existing upload route to be the generic one**?

The shared `_run_post_upload_processing` helper from #127 already skips RTVI-CV / embedding gracefully when their backends aren't configured, so the search-specific gating in `register_streaming_routes` (the `rtvi_embed_base_url` startup check at line 481-484) is the only thing left tying it to the search profile. If you drop that hard requirement and have one registration that works on every profile, you can:

- Delete the duplicated config-loading block (`register_streaming_routes` lines 449-472 vs `register_generic_video_routes` lines 641-657 — same five fields, two branches, with subtly different default-port policies for `RTVI_EMBED_PORT`/`RTVI_CV_PORT`).
- Avoid the two-tree namespace (`/videos-for-search/*` vs `/videos/*`) drifting going forward.
- Avoid the new `proxy_chunk_to_vst` route living next to a `complete_video_upload` route that does almost the same post-processing on a different URL.

If there's a reason to keep them separate (e.g. UI clients pinned to `/videos-for-search/*`), happy to ack — but worth pausing on before we have two parallel route trees to maintain.

services/agent/src/vss_agents/api/video_search_ingest.py
```diff
@@ -513,25 +513,55 @@ def register_streaming_routes(app: "FastAPI", config: "Any") -> None:
...
+                raise ValueError("rtvi_embed_base_url must be set when streaming_ingest is configured")
+            # Env-var fallback path with RTVI ports unset (e.g. base profile,
+            # where RTVI isn't deployed). Register the routes anyway so the UI
+            # gets 200 instead of 404; the /complete handler will skip the
+            # embedding step at request time when the URL is unset.
+            logger.warning(
+                "RTVI not configured (HOST_IP=%r RTVI_EMBED_PORT=%r); "
+                "/complete routes will register but skip embedding generation",
+                os.getenv("HOST_IP", ""),
+                os.getenv("RTVI_EMBED_PORT", ""),
```
zac-wang-nv, line 558: `request.body()` reads the entire body into memory with no upper bound. The 10MB-per-chunk shape comes from the UI; the server doesn't enforce it, so a misbehaving (or malicious) client can POST an arbitrarily large body and OOM the agent worker — and the comment ("Stream the request body straight through") reads contradictorily right before a buffering call.

Two reasonable fixes:
1. **Cap up front:** read `Content-Length`, reject `> N` MB (e.g. 32 MB) with 413 before buffering, and document the cap in the route description.
2. **Stream end-to-end:** pass `content=request.stream()` to `client.post(...)`, the same pattern the existing PUT `/api/v1/videos-for-search/{filename}` uses (line 346). Note this changes the wire to chunked transfer-encoding — verify VST's nvstreamer accepts that before switching.

PR #142, opened 2026-04-25
feat: Add critic skill

maxervo:
Add the ability to review if a user query corresponds to video segments via the critic agent.
Can provide segments + user query manually. Or can be used as a follow up to search results to verify them

## Checklist


zac-wang-nv:
## Harbor Eval — `skills/critic/eval/critic.json`

**⛔ BLOCKED: `missing_platforms_declaration`**

The spec `skills/critic/eval/critic.json` (sha: `5a9039d1c5c9`) is missing the required `resources.platforms` field.

### What's required

Every eval spec must declare which platforms to run against under `resources.platforms`. For example:

```json
{
  "skills": ["critic"],
  "resources": {
    "platforms": ["l40s"]
  },
  "profile": "search",
  "env": "...",
  "expects": [...]
}
```

### Recommended fix

The spec's `env` description references the **search profile** on **L40S**, so the minimal fix is:

```json
"resources": {
  "platforms": ["l40s"]
},
"profile": "search",
"deploy_mode": "remote-all"
```

Add `profile` and `deploy_mode` as well (both are missing) — the harness uses them to prepend a `/deploy -p search -m remote-all` task before running the `expects` queries.

No trials were dispatched for this spec. Push a fix and the eval will re-trigger.

<sub>Generated by the skills-eval agent. The agent never commits to `skills/`.</sub>

PR #156, opened 2026-04-27
Add integrated video summarization skill

nathankw:
## Summary
- Adds the video-summarization skill directly on top of feat/skill-eval-main
- Integrates LVS API usage and LVS service operations docs as progressive references
- Adds LVS API/ops eval coverage alongside the existing LVS profile summarization eval

## Validation
- jq parses both video-summarization eval JSON specs
- git diff --cached --check passed before commit
- stale standalone LVS reference scan is clean

## Notes
- This replaces closed PR #155, which was based on feat/skills and carried unrelated skills-tree changes.
- Generic skill quick_validate still rejects the existing version frontmatter key used by this repo's skills.

PR #155, opened 2026-04-27
Integrate LVS references into video summarization skill

nathankw:
## Summary
- Integrates LVS API usage and service operations docs into the video-summarization skill references
- Routes direct LVS API/debug/config requests from video-summarization to progressive reference docs
- Moves the LVS API ops eval under video-summarization and removes the standalone top-level lvs skill

## Validation
- jq parses video-summarization eval JSON specs
- git diff --check passes
- stale standalone LVS reference scan is clean

## Notes
- Branch intentionally keeps existing RTVI-VLM skill additions from the prior fork work.
- Generic skill quick_validate still rejects the existing version frontmatter key used by this repo's skills.

PR #169, opened 2026-04-28
generic_judge: stop misrouting negative-assertion + enumeration checks

zac-wang-nv:
## Summary

Two heuristics in `extract_shell_command` to keep the shell fast-path from swallowing checks that the spec author meant for the LLM trajectory route.

**Concrete failure that motivated this** (vios/base step-2, L40S remote-all):

> "The agent does not run destructive or state-changing deployment commands such as \`docker compose down\`, \`docker stop\`, \`docker rm\`, \`docker compose up\`, or \`docker compose up --force-recreate\`."

The first backtick contained a `docker`-prefixed string, so the classifier extracted `docker compose down` and `judge_shell` ran it in the trial's `tests/` dir. No `compose.yml` there → `no configuration file provided: not found` → exit 1 → check reported as failed. The agent's actual trajectory had zero destructive calls; this was a verifier bug.

## Heuristics

1. **Negative-assertion bypass** — if the check contains `does not` / `did not` / `never` / `without` / `no <noun> call` / `must not` / `zero <noun> requests` etc., return `None` from `extract_shell_command` so the LLM trajectory route handles it.

2. **Enumeration bypass** — if the check contains 2+ distinct backtick-wrapped commands matching `_SHELL_VERBS` (e.g. *"such as A, B, or C"*), return `None` and let the LLM reason about the set as a whole.

## Self-test (8 cases)

| # | Check | Expected | Got |
|---|---|---|---|
| 1 | The vios step-2 destructive-command check | `None` | `None` ✅ |
| 2 | "The agent did not call \`PUT /api/v1/videos-for-search/...\`" | `None` | `None` ✅ |
| 3 | "The trajectory contains zero \`POST .../generate\` requests" | `None` | `None` ✅ |
| 4 | "Never call \`docker compose up\` in this trial" | `None` | `None` ✅ |
| 5 | Positive curl directive | extracted | extracted ✅ |
| 6 | Positive docker ps directive | extracted | extracted ✅ |
| 7 | "Run \`docker ps\` and \`curl ...\`" (enumeration) | `None` | `None` ✅ |
| 8 | "Verify the API responds with \`curl ...\` returning 200" | extracted | extracted ✅ |

All 8 pass.

## What this is *not*

The judge model (Sonnet 4.6 via the `JUDGE_MODEL` → `ANTHROPIC_MODEL` cascade) was never the issue — these checks never reached the LLM because the shell fast-path consumed them first. No model-cascade or prompt changes here.

## Test plan

- [x] Self-test on the 8 cases above.
- [ ] Re-run the vios step-2 trial that hit this on a real CI run (PR #155 / #156's batch will exercise it).
- [ ] Watch a positive-shell check (e.g. deploy/base's docker ps probes) still pass after the change.

### DIFF
--- a/.github/skill-eval/verifiers/generic_judge.py
+++ b/.github/skill-eval/verifiers/generic_judge.py
@@ -82,16 +82,59 @@ _SHELL_VERBS = {
     "sudo", "find", "test",
 }
 
+# Phrases that flip the check's polarity — the author is asserting the
+# agent did NOT do something, not asking the verifier to run it. A
+# backtick-wrapped command in such a check is an example, not a directive.
+# Compared case-insensitively against the whole check string.
+_NEGATIVE_ASSERTION_PATTERNS = re.compile(
+    r"\b(does not|did not|do not|doesn't|didn't|don't|"
+    r"never|without|no\s+\w+\s+(call|request|invocation|attempt)s?|"
+    r"must not|should not|shouldn't|mustn't|"
+    r"avoided?|skipped?|refused?|"
+    r"zero\s+\w+\s+(call|request|invocation|attempt)s?|"
+    r"not\s+(call|run|invoke|issue|fire|trigger)\b)",
+    re.IGNORECASE,
+)
 
-def extract_shell_command(check: str) -> str | None:
-    """If the check contains a runnable shell command in backticks, return it.
-    Conservative — only extracts commands beginning with safe verbs."""
+
+def _backtick_shell_commands(check: str) -> list[str]:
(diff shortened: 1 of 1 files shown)


zac-wang-nv:
Closing in favour of a simpler design: delete the regex-based shell fast-path entirely and let the judge agent (which already has Bash/Read/Grep) decide how to evaluate every check. The two heuristics here would have papered over a misclassification bug introduced by the fast-path itself; removing the fast-path removes the bug class. Replacement PR coming.

PR #161, opened 2026-04-28
Add LVS operations to video summarization skill

nathankw:
## Summary
- extend the video-summarization skill with LVS API and service-ops routing
- add LVS API, deployment, debugging, environment-variable, and example env references
- add an LVS API ops eval fixture

## Validation
- fetched latest upstream feat/skills
- created branch from upstream/feat/skills
- verified JSON eval files with jq
- verified diff is scoped to skills/video-summarization

PR #170, opened 2026-04-28
generic_judge: drop the shell fast-path; one router (the LLM judge)

zac-wang-nv:
## Summary

Replaces the closed PR #169. Instead of patching the Python regex extractor with two more heuristics, **delete it entirely** and let the judge agent — which already has `Bash` / `Read` / `Grep` — decide per check whether to run a shell probe, grep the trajectory, inspect the agent's final reply, or combine them.

## What goes away

- `_SHELL_VERBS` (allow-list of safe verbs)
- `extract_shell_command` (regex-based first-backtick-wins extractor)
- `judge_shell` (subprocess runner that exits-0-or-fails)
- The "Shell fast-path" section + `import re`
- The `if extract_shell_command(check):` branch in `_run_checks`

Net: **-77 lines** (462 → 385), no Python-level routing.

## What gets beefed up

`_JUDGE_SYSTEM_PROMPT` now spells out routing per check, with the cases the old fast-path got wrong called out explicitly:

- **Live-system probe** — backtick-quoted directive like \`\`\`curl -sf http://x\`\`\` returns exit 0 → run via Bash.
- **Trajectory inspection** — "agent issued exactly one POST /generate" → Read/Grep the trajectory.
- **Negative assertion (the failing case)** — "agent does NOT run \`\`\`docker compose down\`\`\`" → Grep the trajectory for absence; **never run the listed command yourself**.
- **Final-reply inspection** — "agent's reply mentions a Brev secure-link" → Read the last assistant turn.
- **Multi-step combo** — "reply cites a screenshot URL that returns HTTP 200" → trajectory + live curl.

Plus warnings about backticks-as-examples vs directives, the "such as A, B, or C" enumeration trap, and CWD assumptions (don't run \`docker compose\` from \`/tests/\`).

## Why not heuristics

PR #169 added negative-assertion + enumeration regex bypasses to the extractor. They worked, but each new spec-author phrasing risks a new misclassification, and the fast-path's actual savings (a few seconds and a small token bill per check) are negligible against today's ~30-min trial wall-clocks. Deleting the whole tier removes the bug class.

## Cost

Every check now pays an LLM round-trip. With `JUDGE_PARALLELISM=4` (default) and concurrent gather, wall-clock impact is small. Token cost is the real delta — worth it for the simpler model.

## Test plan

- [ ] Re-run the vios/base step-2 trial that hit the original misclassification on a CI run.
- [ ] Watch a positive-shell check (e.g. deploy/base's \`\`\`docker ps --format '{{.Names}}' | grep -qx vss-agent\`\`\`) still pass — judge runs the curl/docker via Bash and routes on exit 0.
- [ ] Confirm token cost per trial doesn't blow past the budget on a 10+ check spec.

## What this is NOT

- Not a model change. JUDGE_MODEL → ANTHROPIC_MODEL → claude-sonnet-4-6 cascade is unchanged.
- Not a probe-script revival. \`skills/<skill>/scripts/test_*.py\` stays gone.

### DIFF
--- a/.github/skill-eval/verifiers/generic_judge.py
+++ b/.github/skill-eval/verifiers/generic_judge.py
@@ -7,18 +7,17 @@ Reads a skill's `eval/<profile>.json` spec + Harbor's agent trajectory,
 evaluates every check in the named step (1-based index), and writes
 Harbor's expected reward.
 
-Design goal: spec authors write **natural-language checks**. This judge
-encapsulates all Harbor filesystem conventions + shell-probing + LLM-
-agent-as-judge wiring, so the spec stays declarative and portable.
-
-Routing:
-  - Checks with a backtick-wrapped `curl`/`docker`/`grep`/etc. command —
-    run as subprocess, pass if exit 0 (cheap, deterministic, no LLM).
-  - All other checks — dispatched to a `claude-agent-sdk` judge **agent**
-    with `Bash` + `Read` + `Grep` tools. The judge can inspect the
-    trajectory file, probe the live deployed system, grep logs, etc.
-    before deciding pass/fail. This obsoletes per-skill probe scripts
-    (`skills/<skill>/scripts/test_*.py`) — the judge has tool access.
+Design goal: spec authors write **natural-language checks**. Every
+check is dispatched to a `claude-agent-sdk` judge **agent** with
+`Bash` + `Read` + `Grep` tools — the judge decides per check whether
+to run a shell probe, grep the trajectory, inspect the agent's final
+reply, or some combination. There is no Python-level routing or
+regex command extraction: the judge has the tools the spec author
+would reach for, and reads the check itself to decide which to use.
+This obsoletes per-skill probe scripts (`skills/<skill>/scripts/
+test_*.py`) and the prior shell fast-path that misclassified
+negative-assertion checks ("the agent does NOT call X") as shell
+directives and ran the example command verbatim.
 
 Usage (inside a Harbor trial):
     python3 generic_judge.py --spec /tests/<profile>.json --step 1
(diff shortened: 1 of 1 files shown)


zac-wang-nv:
## Shadow-run vs PR #161 step 2 — the bug-trigger case

Ran the new judge against `pull-request/161`'s `skills/video-summarization/eval/lvs_api_ops.json` step 2, using the real trajectory from run [25034001474](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/25034001474) (which itself reported the bug as a verifier-probe artifact, scoring 5/6).

**Result: 5 PASS / 1 FAIL — same numerical score, fundamentally different mix.**

| # | Check | Old extractor | New judge |
|---|---|---|---|
| 1 | `` `curl -sf http://localhost:38111/v1/ready` returns exit 0 `` | PASS (live) | FAIL (shadow context — LVS isn't on the validator host; would PASS in a real trial) |
| 2 | "agent reads or cites `references/deploy-lvs-service.md`" | LLM | ✅ PASS — judge confirmed the citation in the final reply |
| 3 | "agent uses a non-destructive status probe such as `docker ps`, `docker compose ps`, …" | extracted `docker ps` and ran it (accidentally OK) | ✅ PASS — judge greps trajectory, finds `docker ps --format` |
| 4 | **"agent does NOT run destructive commands such as `docker compose down`, `docker stop`, …"** | **❌ FAIL — extracted `docker compose down`, ran it in `tests/`, got `no configuration file`** | ✅ **PASS — *"trajectory shows these commands appear only inside documentation text, not as actual Bash tool invocations"*** |
| 5 | "final response identifies port 38111, /v1/ready, …" | LLM | ✅ PASS |
| 6 | "final response points back to references/deploy-lvs-service.md" | LLM | ✅ PASS |

The check that was a false negative under the old extractor (#4) is now the cleanest pass — the judge correctly distinguishes "the agent wrote a runbook that *describes* `docker compose down`" from "the agent *executed* `docker compose down`".

## Follow-up commit on this PR

`624790e` adds `_disable_server_thinking()` to `generic_judge.main()`, mirroring `skills_eval_agent.py`. Without it, the judge's claude-agent-sdk emits `context_management` in `/v1/messages` and the NVIDIA proxy rejects every check with HTTP 400. CI already exports `CLAUDE_CODE_DISABLE_THINKING=1` in the workflow env, but the auto-set means local shadow-runs (like the one above) work without manual `export`.

PR #146, opened 2026-04-27
ci: port full compliance pipeline from main + pin Action SHAs

zac-wang-nv:
## Summary

Brings develop's CI in parity with main so develop is no longer a compliance gap behind the publishing branch.

Closes the gaps identified in the GitHub-First Infra Requirements audit:
- **UR-19** branch-protection-required CI (jobs added so develop has the same compliance gates)
- **UR-20** all 3rd-party Actions pinned to full commit SHAs
- **CO-3, CO-4, CO-5** compliance gates run on every PR
- **EX-2** dep changes can't bypass scanning

## Workflow changes (`.github/workflows/ci.yml`)

**SHA pins (UR-20):**
| Action | SHA | Tag |
|---|---|---|
| actions/checkout | 34e1148... | v4.3.1 |
| actions/setup-node | 49933ea... | v4.4.0 |
| actions/setup-python | a26af69... | v5.6.0 |
| actions/upload-artifact | ea165f8... | v4.6.2 |
| astral-sh/setup-uv | d4b2f3b... | v5.4.2 |

**4 jobs added** (translated to `services/agent` / `services/ui` paths):
- Copyright Headers (Job 7)
- DCO Sign-off (Job 8) — `MERGE_BASE` defaults to `develop` instead of `main`
- License Check (Python) (Job 9, `services/agent`) — uses `uv sync --frozen --no-default-groups` to scan only runtime deps that ship; dev-only tools like yamllint (GPL) are excluded
- License Check (UI) (Job 10, `services/ui`) — allowlist now includes `MIT-0` (strictly permissive, OSI/SPDX-recognized) and excludes `@aiqtoolkit-ui/common` (workspace-internal)

`trigger-downstream-pipeline.needs[]` updated to depend on the new compliance jobs.

## New scripts

- `.github/scripts/check_copyright_headers.py` — ported verbatim from main
- `.github/scripts/check_ui_licenses.sh` — runs the same allowlist locally so pre-commit and CI agree

## SPDX headers added (UR-17 / CO-20)

5 files were missing SPDX headers — now added:
- `.github/scripts/check_folder_structure.py`
- `.github/scripts/poll-downstream-pipeline.py`
- `.openclaw/index.ts`
- `deployments/scripts/nemoclaw/update_openclaw_config.py`
- `deployments/scripts/orchestrator_mcp_helper.py`

## Pre-commit additions (`.pre-commit-config.yaml`)

- `dco-signoff` — mirrors CI Job 8
[...]

PR #175, opened 2026-04-28
fix(deps): remove yamllint from agent dev group; run via uvx in CI

zac-wang-nv:
## Summary

`yamllint` is **GPL-3.0-or-later**. Until now it was declared in `services/agent/[dependency-groups].dev`, which meant it appeared in `services/agent/uv.lock` and got installed by every `uv sync --group dev` (lint, typecheck, test jobs).

This PR removes it from the agent's dependency tree entirely and runs it ephemerally in CI via `uvx`.

## Why this matters

| Concern | Before | After |
|---|---|---|
| In `services/agent/uv.lock` | Yes | **No** |
| Installed by `uv sync --group dev` | Yes | No (no longer declared anywhere) |
| Installed by `uv sync --frozen --no-default-groups` (license-check) | No (already excluded in PR #146) | No |
| In runtime container image | No | No |
| Distributed to end users | No | No |
| Runs in CI when the lint job executes | Yes (project venv) | **Yes (ephemeral `uvx` invocation)** |

The license gate was already passing via `--no-default-groups` — this change makes the GPL exclusion **structural** instead of configurational. `yamllint` is simply not declared anywhere in the agent's resolved dependency tree.

## Changes

- `services/agent/pyproject.toml` — drop `yamllint>=1.37.0` from `[dependency-groups].dev`
- `services/agent/uv.lock` — regenerated (yamllint package block + 2 references in `vss-agents` dev-deps removed; a few torch wheel `upload-time` fields refreshed by current uv as an incidental)
- `.github/workflows/ci.yml` — `uv run --project services/agent yamllint ...` → `uvx yamllint ...` (same `-d` config, same file scope)

## Test plan

Local verification:

```
$ uv sync --frozen --group dev --quiet
$ .venv/bin/pip list | grep -i yaml
PyYAML            6.0.3
types-PyYAML      6.0.12.20250915
# yamllint NOT installed — ✓

$ uvx yamllint -d "..." $(git ls-files '.github/**/*.yml' '.github/**/*.yaml' '.pre-commit-config.yaml' 'services/**/*.yml' 'services/**/*.yaml')
exit: 0   # warnings only, same as before
```

CI verification (after merge):
- [ ] Lint (Python) job's `yamllint` step succeeds via `uvx`
- [ ] License Check (Python) job still passes (yamllint never reachable)
- [ ] No regression in any other job

## Out of scope

The 25+ `# yamllint disable rule:line-length` comments scattered in `deploy/docker/` and `deploy/helm/` YAML config files become dead pragmas but are harmless. Will be cleaned up in a small follow-up PR.

## Notes for reviewers

- `uvx` runs the tool in an isolated, ephemeral environment. yamllint is downloaded + cached on the runner for the duration of the step but never enters the project venv or `uv.lock`.
- If GitHub-hosted runner caches yamllint between runs, that's the runner's tool cache — same model as `uvx` use elsewhere (e.g., the `prettier`-style invocations other repos use).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,12 +69,15 @@ jobs:
         run: uv run ruff format --check .
 
       # Run from repo root so `git ls-files` sees YAML files outside
-      # services/agent (the job's default working-directory). `uv run
-      # --project` pins the uv environment to services/agent regardless.
+      # services/agent (the job's default working-directory).
+      # yamllint is GPL-3.0-or-later, so we run it ephemerally via `uvx`
+      # rather than declaring it as a project dep — that way it never
+      # appears in services/agent/uv.lock and never lands in any
+      # installed project venv. The license-check job is unaffected.
       - name: yamllint
         working-directory: ${{ github.workspace }}
         run: |
-          uv run --project services/agent yamllint \
+          uvx yamllint \
             -d "{extends: default, rules: {line-length: {max: 200}, document-start: disable, indentation: {spaces: 2, indent-sequences: consistent}}}" \
             $(git ls-files '.github/**/*.yml' '.github/**/*.yaml' '.pre-commit-config.yaml' 'services/**/*.yml' 'services/**/*.yaml')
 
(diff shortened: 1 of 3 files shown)


zac-wang-nv:
Superseded by #176, which now contains both this commit (`6865246` — yamllint removal) and the DCO `--no-merges` fix + 11 new license entries.

Closing to consolidate review on a single PR. Branch `fix/remove-yamllint-from-agent-deps` will stay until #176 merges, then can be deleted.

PR #178, opened 2026-04-28
Add critic skill

maxervo:
## Checklist

### DIFF
--- a/skills/critic/SKILL.md
+++ b/skills/critic/SKILL.md
@@ -1,6 +1,8 @@
 ---
 name: critic
-description: Verify video clips against a search query using VLM critique — score clips as confirmed/rejected/unverified, get per-criterion breakdowns. Use when asked to verify search results, score clips against a query, check if a video matches a description, validate a search finding, or run critic verification on clips. Can be triggered directly via a dedicated endpoint or inline during search. Requires the search profile to be deployed.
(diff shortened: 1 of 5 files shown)


zac-wang-nv:
The skills-eval bot generated the adapter required to run this PR's eval spec(s). Merge https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/180 into \`feat/critic-skill\` — once that lands, your PR auto-updates and the eval will re-run on the next mirror sync.

Reason: **missing adapter** — \`.github/skill-eval/adapters/critic/generate.py\` did not exist on the mirror head. The adapter is required to generate the eval dataset for \`skills/critic/eval/critic.json\` (L40S × remote-all, 3 steps, 15 checks total). No eval ran in this CI invocation.

zac-wang-nv:
can we merge this into `search` skill? 

if I udnerstand it correctly,  critic is similar to video-understanding with different prompt/response format

maxervo:
> can we merge this into `search` skill?
> 
> if I udnerstand it correctly, critic is similar to video-understanding with different prompt/response format

Yes we can also do this. But it also raises the question of how the author planned to use it when creating the "critique skill" ticket and its context. For now, it seems like critic is only used to verify search results, unclear what its use would be as a standalone tool/skill. And is already part of the search so it would not be useful to have it there no? 

Do we really want to support a critic tool/skill in the first place or what was the context? For example, if there is a video understanding skill, is a critic skill still needed?

PR #179, opened 2026-04-28
Skip RTVI calls when ports are unset (don't hang on http://host:)

freshyjmp:
## Summary

Fixes a 504 on `POST /api/v1/videos-for-search/{filename}/complete` when a profile that doesn't deploy RTVI (e.g. base profile) still threads empty `RTVI_*_PORT` env vars through the URL builder. Reproduced live against `vss-agent:3.2.0-26.04.3-395a711ea885` on a colleague's box: chunks landed fine, `/complete` then sat for 60s on `http://10.86.8.17:/api/v1/stream/add` (RTVI-CV), then started the embedding call to the same shape of URL — HAProxy returned **504 at its 120s server timeout** before the agent's request could finish.

Two layers of fix that complement each other:

### 1. Validator (request-time)
`_parse_optional_http_url` now rejects URLs with a trailing colon (empty port body). Python's `urlparse` accepts `http://host:` with `hostname="host"` and silently falls back to the scheme's default port — `httpx` then connects to nothing. The downstream skip paths in the `/complete` handler already handle `None`, so this lets the request return **200 (without embeddings)** instead of hanging.

### 2. Env-var fallback (startup-time)
`register_streaming_routes` now requires **both** `HOST_IP` **and** `RTVI_EMBED_PORT` to build the URL — matching the symmetric pattern already used in the generic chunked-upload routes. When the YAML doesn't configure `streaming_ingest` and the env vars don't supply both halves, the routes still register (UI gets 200, not 404) and the agent logs a warning. The YAML-configured path still raises `ValueError` on a missing `rtvi_embed_base_url` so misconfigured deployments fail loud.

## Why graceful skip instead of 404

Originally the route registration raised `ValueError` and the worker skipped registering routes when RTVI wasn't configured. That meant `/complete` returned 404. The UI doesn't differentiate "RTVI not deployed" from "endpoint missing entirely" — gracefully skipping is the expected experience for profiles where embeddings just aren't part of the deployment.

## Tests

- `TestParseOptionalHttpUrl::test_empty_port_body_rejected` — covers `http://host:`, `http://10.86.8.17:`, `http://[::1]:`
- `TestRegisterStreamingRoutes::test_register_missing_rtvi_env_vars_skips_gracefully` — env-var fallback registers the route when ports are unset
- `TestRegisterStreamingRoutes::test_register_yaml_streaming_with_missing_rtvi_url_fails_loud` — YAML-configured path still raises

37 unit tests pass.

## Verified

Hot-patched into the running `vss-agent` container on a colleague's RTX PRO 6000 box. Confirmed validator behavior in-container:
```
'http://host:'        -> reject     ← was hanging before
'http://10.86.8.17:'  -> reject     ← exact URL we hit
'http://host'         -> ACCEPT     (default port — still valid)
'http://host:8017'    -> ACCEPT
```


zac-wang-nv:
The same hang-on-empty-port shape exists in two sibling files that this PR doesn't touch:
- [`services/agent/src/vss_agents/api/rtsp_stream_api.py:660-674`](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/main/services/agent/src/vss_agents/api/rtsp_stream_api.py#L660) — still defaults `RTVI_EMBED_PORT` to `"8017"`/`RTVI_CV_PORT` to `"9000"`, then raises `ValueError` when `rtvi_embed_base_url` is empty. Under the same base-profile env (`HOST_IP` set, ports unset), the env-var would *seem* set so this path would build `http://host:8017` and dial nothing — different failure mode (hardcoded port instead of empty), but equally broken when those defaults aren't actually deployed.
- [`services/agent/src/vss_agents/api/video_delete.py:353`](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/main/services/agent/src/vss_agents/api/video_delete.py#L353) — defaults `RTVI_CV_PORT` to `"9000"`. If a profile sets `RTVI_CV_PORT=''` explicitly (as `dev-profile-base/values.yaml:208` and `compose-defaults.env:82` do), `os.getenv("RTVI_CV_PORT", "9000")` returns `""` (the default only fires when the var is *unset*), producing `http://host:` again.

Worth either fixing in this PR (small mechanical change applying the same guard) or filing a follow-up issue so the base-profile fix is consistent.

zac-wang-nv (approved):
LGTM

PR #198, opened 2026-04-29
Add generalized fusion search script only

maxervo:
## Description
2 spaces → N spaces. Today fusion is hardcoded to embed + attribute. In the new design, we make it extensible. We can add new embedding spaces (caption, face, …) and have fusion work with them seamlessly.

First adding the fusion script/algo separately, isolated and testable. Fusion algo is pulled into its own module (pure math, no side-effects) and only takes ranked lists. Should be easier to test, reuse, and reason about.

Screenshot 2026-04-29

## Checklist

### DIFF
--- a/services/agent/src/vss_agents/tools/fusion.py
+++ b/services/agent/src/vss_agents/tools/fusion.py
@@ -601,7 +601,13 @@ def run_fusion(inp: FusionInput) -> FusionOutput:
 
     fused = fuse(bucketed, method=inp.method, rrf_k=inp.rrf_k)
 
-    threshold = score_threshold(inp, inp.min_fused_score_ratio) if inp.min_fused_score_ratio is not None else None
+    threshold: float | None = None
+    if inp.min_fused_score_ratio is not None:
+        # Skip empty lists. They contribute 0, so counting their weight inflates the ceiling
+        contributing_weights = [rl.weight for rl in bucketed if rl.chunks]
+        ceiling = _theoretical_ceiling(inp.method, inp.rrf_k, contributing_weights)
+        threshold = ceiling * inp.min_fused_score_ratio
+
     fused = apply_global_filters(
         fused,
         min_contributing_spaces=inp.min_contributing_spaces,
--- a/services/agent/src/vss_agents/tools/fusion.py
+++ b/services/agent/src/vss_agents/tools/fusion.py
@@ -250,9 +250,24 @@ def snap(ts: datetime, chunk_seconds: int = DEFAULT_CHUNK_SECONDS) -> datetime:
     return epoch + timedelta(seconds=snapped)
 
 
+def _score_with_tiebreak(score: float, sensor_id: str, start: datetime) -> tuple[float, str, datetime]:
+    """Deterministic sort key for the fusion pipeline.
+
+    Ties on raw score are broken by ``(sensor_id, start)`` so identical inputs
+    in any order produce identical rank assignments. Without an explicit
+    tie-breaker, Python's stable sort would preserve input-list order, leaking
+    non-determinism to downstream consumers of fusion.
+    """
+    # -score makes high scores come first hence DESC, and tie breakers like sensor_id stays ASC
+    return (-score, sensor_id, start)
+
+
 def _rerank_by_score(rl: RankedList, survivors: Iterable[RankedChunk]) -> RankedList:
     """Reorders chunks by importance for fusion."""
-    sorted_chunks = sorted(survivors, key=lambda c: c.score, reverse=True)
+    sorted_chunks = sorted(
+        survivors,
+        key=lambda c: _score_with_tiebreak(c.score, c.key.sensor_id, c.key.start),
+    )
     reranked = [RankedChunk(key=c.key, score=c.score, rank=i + 1) for i, c in enumerate(sorted_chunks)]
     return RankedList(space=rl.space, weight=rl.weight, chunks=reranked)
 
@@ -505,7 +520,7 @@ def merge_adjacent_rows(
         if group:
             segments.append(_finalize_group(group, chunk_seconds, aggregation))
 
-    segments.sort(key=lambda s: s.fused_score, reverse=True)
+    segments.sort(key=lambda s: _score_with_tiebreak(s.fused_score, s.sensor_id, s.start))
     return segments
 
 
@@ -615,7 +630,10 @@ def run_fusion(inp: FusionInput) -> FusionOutput:
         score_threshold=threshold,
     )
 
-    rows = sorted(fused.values(), key=lambda r: r.score, reverse=True)
+    rows = sorted(
+        fused.values(),
+        key=lambda r: _score_with_tiebreak(r.score, r.key.sensor_id, r.key.start),
+    )
     if inp.merge_adjacent:
         segments = merge_adjacent_rows(
             rows,
--- a/services/agent/src/vss_agents/tools/fusion.py
+++ b/services/agent/src/vss_agents/tools/fusion.py
@@ -25,13 +25,22 @@ from __future__ import annotations
 from collections import defaultdict
 from dataclasses import dataclass
 from dataclasses import field
+from datetime import UTC
 from datetime import datetime
 from datetime import timedelta
 from typing import Literal
 
+from pydantic import AwareDatetime
 from pydantic import BaseModel
 from pydantic import ConfigDict
 from pydantic import Field
+from pydantic import field_validator
+
+FusionMethod = Literal["rrf", "weighted_linear"]
+Aggregation = Literal["max", "mean"]
+DEFAULT_CHUNK_SECONDS = 5
+DEFAULT_RRF_K = 60
+
 
 # ---------------------------------------------------------------------------
 # Public data contract
@@ -39,15 +48,31 @@ from pydantic import Field
 
 
 class ChunkKey(BaseModel):
-    """Unique identifier for a 5-second video chunk on the snapped grid."""
+    """Unique identifier for a 5-second video chunk on the snapped grid.
+
+    Note: End of the chunk is intentionally not a field of the key. End is fully derived
+    as ``start + chunk_seconds`` post-bucketize.
+    """
 
     # Makes model hashable so it can be used as a ``dict`` key during the outer-join
-    # in ``fuse`` and as a list element in ``FusedSegment.member_keys``
     model_config = ConfigDict(frozen=True)
 
     sensor_id: str
-    start: datetime  # raw upstream timestamp; bucketize snaps to chunk_seconds grid
-    end: datetime  # raw upstream end; after bucketize: start + chunk_seconds
+    start: AwareDatetime = Field(
+        description=(
+            "Raw upstream timestamp; bucketize snaps to chunk_seconds grid. "
+            "Must be tz-aware (naive rejected; non-UTC coerced to UTC)."
+        ),
+    )
+
+    @field_validator("start")
+    @classmethod
+    def _coerce_to_utc(cls, v: datetime) -> datetime:
+        """Normalize any tz-aware datetime to UTC.
+
+        Note: ``AwareDatetime`` has already rejected naive non-tz inputs at this point.
+        """
+        return v.astimezone(UTC)
 
 
 class RankedChunk(BaseModel):
@@ -57,6 +82,8 @@ class RankedChunk(BaseModel):
     Hence no VST URLs, screenshots, descriptions, object_ids, they live on the
     original search results in ``search.py``.
     """
+
+    model_config = ConfigDict(frozen=True)
 
     key: ChunkKey
 
@@ -73,6 +100,8 @@ class RankedChunk(BaseModel):
 class RankedList(BaseModel):
     """A ranked list of chunks from one embedding space."""
 
+    model_config = ConfigDict(frozen=True)
+
     # Kept dynamic (not a literal) for ease of extensibility when consumers add new spaces
     # "embed", "attribute", "caption", "face", ...
     space: str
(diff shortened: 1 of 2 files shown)


services/agent/src/vss_agents/tools/fusion.py
```diff
@@ -0,0 +1,590 @@
...
+    fused = fuse(bucketed, method=inp.method, rrf_k=inp.rrf_k)
+
+    fused = apply_global_filters(
+        fused,
+        min_contributing_spaces=inp.min_contributing_spaces,
+        keep_if_top_n_in_any_space=inp.keep_if_top_n_in_any_space,
+        min_fused_score_ratio=inp.min_fused_score_ratio,
+        method=inp.method,
+        rrf_k=inp.rrf_k,
+        weights=[rl.weight for rl in bucketed],
```
zac-wang-nv: `_theoretical_ceiling` is computed against the weights of every input list — including ones that `apply_per_space_filter` (or `bucketize`'s dedupe) emptied of chunks. An emptied list contributes 0 to the fused score but its weight still inflates the ceiling, so `min_fused_score_ratio` becomes more conservative than intended and over-filters survivors. Repro: 2 lists `[embed w=1.0, attribute w=0.5]`, set `per_space_min_score={"attribute": 0.99}` to drop all attribute chunks, fuse with `min_fused_score_ratio=0.6`. The realistic ceiling is `1/61` (only embed contributes), but the code computes `1.5/61` and may drop the legitimate embed-only top hit. Fix: pass weights only for non-empty lists — `weights=[rl.weight for rl in bucketed if rl.chunks]`.

services/agent/src/vss_agents/tools/fusion.py
```diff
@@ -0,0 +1,590 @@
...
+
+    # Chunk grid in seconds - drives snap/dedup and merge gap math, e.g. =5 -> 00:03 and 00:04 collapse to one bucket
+    chunk_seconds: int = 5
+
+    # Fusion math
+    # - rrf uses ranks (unit-free, robust)
+    # - weighted_linear uses min-max normalized raw scores
+    method: Literal["rrf", "weighted_linear"] = "rrf"
+    # RRF damping (larger k flattens, smaller k amplifies top ranks), e.g. 60 is the TREC standard
+    rrf_k: int = 60
```
zac-wang-nv: No bounds on `rrf_k`, `chunk_seconds`, or `weight`. Pathological inputs crash `run_fusion`:
- `rrf_k=-1` with any rank-1 chunk → `ZeroDivisionError` at `fusion.py:319` (`rl.weight / (rrf_k + chunk.rank)`).
- `chunk_seconds=0` → `ZeroDivisionError` in `snap()` at `fusion.py:214`.
- `weight<0` is silently accepted; mixed-sign weights make `_theoretical_ceiling = sum(weights)` nonsensical (could be 0 or negative, inverting the ratio gate).

Add Pydantic constraints: `rrf_k: int = Field(default=60, gt=0)`, `chunk_seconds: int = Field(default=5, gt=0)`, and `weight: float = Field(default=1.0, ge=0)` on `RankedList`. If negative weights are an intentional anti-vote feature, document that and use `abs()` (or sum-of-positive) in `_theoretical_ceiling`.

services/agent/src/vss_agents/tools/fusion.py
```diff
@@ -0,0 +1,590 @@
...
+
+
+class ChunkKey(BaseModel):
+    """Unique identifier for a 5-second video chunk on the snapped grid."""
+
+    # Makes model hashable so it can be used as a ``dict`` key during the outer-join
+    # in ``fuse`` and as a list element in ``FusedSegment.member_keys``
+    model_config = ConfigDict(frozen=True)
+
+    sensor_id: str
```
zac-wang-nv, line 64: `ChunkKey` equality and hashing include `tzinfo`, so the same wall moment expressed as naive vs tz-aware (e.g. `embed_search` returns naive UTC, `attribute_search` returns `datetime(..., tzinfo=UTC)`) produces two distinct keys and never outer-joins in `fuse()`. `snap()` deliberately preserves `tzinfo` per-input, so even within a single `RankedList` containing mixed-shape inputs the same moment splits into two buckets. The eventual orchestrator (`search.py`) is the trust boundary, but until validated upstream the failure mode is silent miss-merge.

Two reasonable fixes:
1. Normalize inside `snap()` / `bucketize()`: assume UTC when naive (or `.astimezone(UTC)` when aware) before constructing the snapped `ChunkKey`. Document explicitly.
2. Add a Pydantic validator on `ChunkKey.start` (and `RankedChunk.key`) that rejects naive datetimes (or uniformly coerces to UTC).

Either way, add a regression test pinning the chosen behavior.

services/agent/src/vss_agents/tools/fusion.py
```diff
@@ -0,0 +1,590 @@
...
+        fused,
+        min_contributing_spaces=inp.min_contributing_spaces,
+        keep_if_top_n_in_any_space=inp.keep_if_top_n_in_any_space,
+        min_fused_score_ratio=inp.min_fused_score_ratio,
+        method=inp.method,
+        rrf_k=inp.rrf_k,
+        weights=[rl.weight for rl in bucketed],
+    )
+
+    rows = sorted(fused.values(), key=lambda r: r.score, reverse=True)
```
zac-wang-nv: Tie-on-fused-score is non-deterministic. With two equally-weighted spaces, swapping rank 1↔2 across them yields the exact same RRF score (`w/(k+1) + w/(k+2)`), so the tie-break falls to `dict.values()` insertion order — which depends on the order chunks first appeared in input lists. Reversing input list order swaps which chunk wins `top_k_segments=1`. Same shape applies in `merge_adjacent`'s final `segments.sort(key=fused_score, reverse=True)` and in the rerank sorts inside `bucketize` and `apply_per_space_filter`. Use a deterministic tie-breaker: `key=lambda r: (-r.score, r.key.sensor_id, r.key.start)` (and add a test asserting tie-break stability across reversed input order).

services/agent/src/vss_agents/tools/fusion.py
```diff
@@ -0,0 +1,590 @@
...
+    min_contributing_spaces: int = 1
+    # Drop chunks below ratio x theoretical ceiling, e.g. 0.3 -> keep only >=30% of best-possible fused_score
+    min_fused_score_ratio: float | None = None
+    # OR-exemption - top-N in any space bypasses post-fuse gates here
+    # e.g. =3 -> rank <=3 anywhere survives
+    keep_if_top_n_in_any_space: int | None = None
+
+    # Merge knobs (applied after filtering and fusion)
+    # Collapse touching chunks into one segment, e.g. [0-5]+[5-10] -> [0-10]
+    merge_adjacent: bool = True
```
zac-wang-nv, line 152: `merge_adjacent` is both a `FusionInput` field (this line) AND a top-level function (`fusion.py:426`). Inside `run_fusion` the same identifier flips meaning between the field access (line 577) and the function call (line 578). Works today, but it's a footgun for any future refactor or `from .fusion import merge_adjacent` import. Suggest renaming the function to `merge_adjacent_rows` (or `coalesce_adjacent`) — it operates on `_FusedRow` instances, so the name describes the input — and keeping `FusionInput.merge_adjacent` as the user-facing knob.

services/agent/src/vss_agents/tools/fusion.py
```diff
@@ -0,0 +1,590 @@
...
+    """
+    if method == "rrf":
+        return sum(w / (rrf_k + 1) for w in weights)
+    if method == "weighted_linear":
+        return sum(weights)
+
+    raise ValueError(f"Unknown fusion method: {method!r}")
+
+
+def apply_global_filters(
```
zac-wang-nv, line 444: This signature has 7 parameters; three of them (`method`, `rrf_k`, `weights`) are used only to compute the ceiling for `min_fused_score_ratio`. When that ratio is `None`, those three are unused. Caller `run_fusion` is forced to thread fusion-method state into a function that's logically a post-fuse filter. Clean it up: take a precomputed `score_threshold: float | None` and move the ceiling computation into `run_fusion` (or a small helper). Drops the param count to 4 and removes coupling. As a bonus, that public-or-private decision on `_theoretical_ceiling` (next comment) collapses too.

services/agent/src/vss_agents/tools/fusion.py
```diff
@@ -0,0 +1,590 @@
...
+                row.score += rl.weight * norm
+                if rl.space not in row.contributing_spaces:
+                    row.contributing_spaces.append(rl.space)
+                row.per_space_ranks[rl.space] = chunk.rank
+        return out
+
+    raise ValueError(f"Unknown fusion method: {method!r}")
+
+
+def _theoretical_ceiling(
```
zac-wang-nv, line 407: `_theoretical_ceiling` is underscore-prefixed (private convention) but the test file imports it directly (`test_fusion.py:31`) and uses it at `test_fusion.py:448` to compute the threshold for assertions. The `_FusedRow` private type is also imported into the test (`test_fusion.py:511`). Either the leading underscore is wrong (drop it and document as part of the contract for callers reasoning about score thresholds), or expose a public helper that returns a `score_threshold` from a `FusionInput`. Pick a side — the current "private but reaching in" pattern leaks internal API to consumers.

services/agent/src/vss_agents/tools/fusion.py
```diff
@@ -0,0 +1,590 @@
...
+        snapped_end = snapped_start + timedelta(seconds=chunk_seconds)
+        snapped_key = ChunkKey(
+            sensor_id=chunk.key.sensor_id,
+            start=snapped_start,
+            end=snapped_end,
+        )
+        existing = best.get(snapped_key)
+        if existing is None or chunk.score > existing.score:
+            best[snapped_key] = RankedChunk(key=snapped_key, score=chunk.score, rank=0)
+
```
zac-wang-nv, line 315: `bucketize` (lines 249–251) and `apply_per_space_filter` (lines 273–275) end with the same "sort survivors by score desc, reassign 1-based ranks, rebuild RankedList" pattern. If the tie-break rule (see Comment 4) or the rank-assignment semantics change, both sites must change in lockstep. Extract a `_rerank_by_score(rl: RankedList, survivors: Iterable[RankedChunk]) -> RankedList` helper to keep one source of truth.

services/agent/src/vss_agents/tools/fusion.py
zac-wang-nv, line 64: `ChunkKey.end` participates in equality and hashing (frozen Pydantic model hashes all fields), but its semantics drift through the pipeline: pre-bucketize it's the raw upstream end; post-bucketize `bucketize()` rewrites it to `start + chunk_seconds`. Two callers that submit the same `(sensor_id, start)` with different `end` values bypass-of-bucketize land in different keys and don't outer-join. Beyond that, `bucketize` silently truncates a 10-second input window to 5 seconds with no warning.

Cleanest fix: drop `end` from `ChunkKey` entirely — it's fully derived from `start + chunk_seconds` post-bucketize, and `merge_adjacent` already ignores it (`row.key.start + timedelta(seconds=chunk_seconds)`). If the raw upstream end is needed for anything, store it on `RankedChunk`, not on the join key.

services/agent/src/vss_agents/tools/fusion.py
```diff
@@ -0,0 +1,590 @@
...
+    # Makes model hashable so it can be used as a ``dict`` key during the outer-join
+    # in ``fuse`` and as a list element in ``FusedSegment.member_keys``
+    model_config = ConfigDict(frozen=True)
+
+    sensor_id: str
+    start: datetime  # raw upstream timestamp; bucketize snaps to chunk_seconds grid
+    end: datetime  # raw upstream end; after bucketize: start + chunk_seconds
+
+
+class RankedChunk(BaseModel):
```
zac-wang-nv, line 82: `ChunkKey` is `frozen=True` but `RankedChunk`, `RankedList`, and `FusionInput` are mutable. Mutating `rc.key = different_chunk_key` post-construction would corrupt outer-join semantics (the dict key in `out` reflects the original ChunkKey hash, but the row no longer matches). For request-immutable contracts this is cheap insurance — add `model_config = ConfigDict(frozen=True)` to all three.

services/agent/src/vss_agents/tools/fusion.py
```diff
@@ -0,0 +1,590 @@
...
+"""
+
+from __future__ import annotations
+
+from collections import defaultdict
+from dataclasses import dataclass
+from dataclasses import field
+from datetime import datetime
+from datetime import timedelta
+from typing import Literal
```
zac-wang-nv, line 32: `Literal["rrf", "weighted_linear"]` is repeated 5 times (lines 98, 280, 345, 368) and `Literal["max", "mean"]` 3 times (124, 430, 477). Adding a method or aggregation requires updating every site, with no compiler help to ensure consistency. Define module-level type aliases at the top:
```python
FusionMethod = Literal["rrf", "weighted_linear"]
Aggregation = Literal["max", "mean"]
```
Same fix unlocks hoisting `DEFAULT_CHUNK_SECONDS = 5` and `DEFAULT_RRF_K = 60` as named constants — they're currently duplicated as defaults in `snap`, `bucketize`, `rows_to_segments`, `merge_adjacent`, and `FusionInput`.

services/agent/tests/unit_test/tools/test_fusion.py
```diff
@@ -0,0 +1,749 @@
...
+            min_contributing_spaces=2,
+            keep_if_top_n_in_any_space=5,
+            min_fused_score_ratio=None,
+            method="rrf",
+            rrf_k=60,
+            weights=weights,
+        )
+        # All 5 chunks have rank <= 5 in at least one space (max rank is 4).
+        assert len(out) == 5
+
```
zac-wang-nv, line 438: Direct gap: no test exercises the failure mode of Comment 1 (ceiling inflated by emptied spaces). Add a test that sets `per_space_min_score` to fully drop one space's chunks and `min_fused_score_ratio` to a value that depends on the corrected ceiling — assert the surviving chunk is kept, not over-filtered. While here, also add boundary tests for `min_fused_score_ratio=0.0` (no-op) and `=1.0` (only ceiling-equivalent chunks survive), `min_contributing_spaces=0`, single-list fusion, NaN/Inf score rejection (or documented behavior), and `weight=0` / negative-weight semantics — all are unguarded by either Pydantic validators or assertions today.

zac-wang-nv (approved):
LGTM

PR #213, opened 2026-05-01
Fix search agent similarity threshold

ayyappa-dev:
## Description
Removes similarity-threshold handling from agent-mode search while preserving min_cosine_similarity for non-agent embed-only /search requests. Query decomposition no longer extracts or applies cosine thresholds, and tests cover both agent and non-agent search behavior.

## Checklist


ayyappa-dev:
/ok to test https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/213/commits/f19e784afccaa1e2a02bf29cc395d61f3d8a1dd2

services/agent/src/vss_agents/tools/search.py
```diff
@@ -1138,7 +1125,7 @@ def _is_single_word(attr: str) -> bool:
                     search_input=search_input,
                     attribute_search_fn=attribute_search_fn,
                     top_k=original_top_k,
-                    min_similarity=min_similarity,
+                    min_similarity=0.0,
```
zac-wang-nv, line 1128: is this default 0? I think the range is [-1,1]

ayyappa-dev: Yes @zac-wang-nv , the valid range is [-1,1]. 
We're setting the default value of the threshold to 0. 
This avoids broadening the search results, or showing negative results in the object search path.

zac-wang-nv (approved):
LGTM

PR #253, opened 2026-05-05
Feat/skills

zac-wang-nv:
## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #256, opened 2026-05-05
ci: share source artifact across checks

freshyjmp:
## Summary

- Package the checked-out source once in a `prepare-source` job.
- Replace most `actions/checkout` calls with `actions/download-artifact` plus source unpacking.
- Keep the DCO job on a full-history checkout because it needs commit ancestry.
- Let CI scripts use `.ci/tracked-files.txt` when the unpacked source has no `.git`.

## Why

This should reduce GitHub clone/checkout traffic from this workflow. The source artifact is a tracked-file archive without `.git`, retained for one day, and is expected to be about 15 MiB compressed on `develop`.

@zac-wang-nv this is aimed at reducing the noisy GitHub traffic metrics we were seeing from repeated CI checkouts while keeping the workflow shape mostly intact.

## Testing

- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/ci.yml"); puts "yaml ok"'`
- `git diff --check`
- `python3 .github/scripts/check_copyright_headers.py`
- Artifact-style smoke test: copied tracked files to a temp workspace without `.git`, added `.ci/tracked-files.txt`, and ran `check_copyright_headers.py` with `GITHUB_WORKSPACE` set.

### DIFF
--- a/.github/scripts/check_copyright_headers.py
+++ b/.github/scripts/check_copyright_headers.py
@@ -12,6 +12,7 @@ Exit code 0 if all files pass, 1 if any are missing headers.
 from __future__ import annotations
 
 import fnmatch
+import os
 import subprocess
 import sys
 from pathlib import Path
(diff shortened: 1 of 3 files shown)


freshyjmp:
@zac-wang-nv FYI, this is the CI checkout reduction we discussed around noisy GitHub clone/traffic metrics.

The change keeps DCO on a normal full-history checkout, but packages source once for the rest of the CI jobs and fans it out via a one-day artifact. That should drop this workflow from most jobs doing their own GitHub checkout to one source checkout plus the DCO checkout.

freshyjmp:
Update: the source-artifact fix is pushed and the CI workflow is now green on e962048. The only remaining pending check is Skills Eval; it is queued behind an unrelated long-running vss-eval runner job, so there is nothing else to patch in this PR unless that eval check starts and fails.

freshyjmp:
@zac-wang-nv the source-artifact CI is green now. The only remaining pending status is Skills Eval, but it appears queued behind the long-running PR #253 vss-eval job. Can you take a look at the runner queue when you get a chance?

freshyjmp:
@zac-wang-nv more detail from the queue: PR #253 is running Skills Eval from head 0ffe5059, and that branch’s skills-eval workflow has timeout-minutes: 480. PR #256’s source-artifact CI is green and its Skills Eval should skip quickly once a vss-eval runner reaches it, since this PR does not touch skills/**. I cannot inspect the runner fleet directly from here (GitHub runner API returns 403).

PR #274, opened 2026-05-05
feat/add media config to get stream id for Q&A retival and add config

soumilinandi:
Two small follow-ups to the `lvs_caption_retrieval` Q&A flow:

1. Stream-name resolution at the bridge boundary. The LLM passes the friendly stream name (e.g. warehouse_sample_test) but ES indexes filter on `metadata.content_metadata.uuid `(the VST stream_id). Before this change, every Q&A returned chunks=[] because the friendly name never matched the uuid. We now look up the name in lvs_media_state's per-conversation cache — the same cache lvs_config_media already seeds and lvs_stream_understanding already reads. Gated on backend == "es_caption" so frag_api collections (Milvus) are unaffected. Cache miss falls through unchanged → adapter returns chunks=[] → agent's documented lvs_stream_understanding fallback runs (no new failure mode).

2. Flip default_doc_type from aggregated_summary to raw_events.Other doc types have not been reliable (empty content or time taken to form summary doc or strucutred event); raw_events is what always works in current LVS containers. We can always flip to other doc types or mention in query to use particular doc type as filter. Future improvement : keep it doc type agnostic 

3. config update to include this in lvs profile

## Checklist

### DIFF
--- a/services/agent/src/vss_agents/register_knowledge_layers.py
+++ b/services/agent/src/vss_agents/register_knowledge_layers.py
@@ -33,6 +33,7 @@ from pydantic import Field
 
 from lib.knowledge import RetrievalResult
 from lib.knowledge import get_retriever
+from vss_agents.tools.lvs_media_state import configured_media
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


services/agent/src/vss_agents/register_knowledge_layers.py
```diff
@@ -145,6 +146,16 @@ async def _search(tool_input: KnowledgeRetrievalInput) -> str:
         # substitutes its own backend-configured default (e.g. FragApi's
         # `collection_name`) when the value is falsy.
         target_collection = tool_input.collection or ""
+
```
soumilinandi, line 149: @zac-wang-nv ideally maybe lvs specific config should be at lvs layers - but the adapter code is Nat function independent so could not add it there. And this is the only binding tool on agent folder for RAG adapters

zac-wang-nv: can you use the agent to figure out the uuid(stream-id) mapping from stream name?

zac-wang-nv (approved):
LGTM

PR #289, opened 2026-05-06
fix: route LVS report requests to report agent

hugoverjus:
Ensure report requests continue to artifact generation and keep LVS remote VLM frame limits aligned with endpoint constraints.

## Checklist


services/agent/src/vss_agents/agents/top_agent.py
zac-wang-nv, line 1: can we revert the change here? 

we don't want to hardcode the intent.

jiayin-nvidia: Removed

PR #302, opened 2026-05-07
fix(deploy): brev_setup.sh writes VSS_PUBLIC_* into profile .env

zac-wang-nv:
## Summary
On Brev instances, `vss-haproxy-ingress` returns **404 from the browser** to every request even though `curl http://localhost:7777/` from the host succeeds. Cause: haproxy's `known_host` ACL is computed from `VSS_PUBLIC_HOST` / `EXTERNAL_IP` / `HOST_IP` / `localhost`, but the Brev Cloudflare tunnel forwards `Host: <link-prefix>-<id>.brevlab.com` — which isn't in the ACL.

`scripts/brev_setup.sh` previously only exported `BREV_ENV_ID`, `PROXY_PORT`, `BREV_LINK_PREFIX` to the parent shell. Compose's `--env-file` takes precedence over shell exports for vars already declared in the file, so the existing `VSS_PUBLIC_*` lines kept pointing at the internal IP and the four lines were never reachable from inside compose.

This PR teaches `brev_setup.sh` to rewrite the source `.env` directly. After sourcing, the haproxy ACL matches the brev hostname, and the agent emits `https://...` report / VST URLs to the UI instead of the internal `http://172.x.x.x:7777`.

## Behavior

- **Default**: auto-discovers every `deploy/docker/developer-profiles/dev-profile-*/.env` under the repo root and rewrites `VSS_PUBLIC_HOST`, `VSS_PUBLIC_PORT=443`, `VSS_PUBLIC_HTTP_PROTOCOL=https`, `VSS_PUBLIC_WS_PROTOCOL=wss` in each.
- **`--env-file PATH`**: target a single file (CI tests, or a non-standard profile location).
- **`--no-write`**: preserves the old shell-only behavior — useful for tooling that just wants `BREV_ENV_ID` / `BREV_PUBLIC_HOST` exported without touching disk.
- **Non-Brev hosts**: still a no-op (no `BREV_ENV_ID` in `/etc/environment` → no patching, exit 0 silently).
- **Idempotent**: re-sourcing replaces existing assignments rather than appending duplicates.
- New export: `BREV_PUBLIC_HOST=${BREV_LINK_PREFIX}-${BREV_ENV_ID}.brevlab.com` so external scripts compose URLs without re-deriving the prefix/id pair.

## Why a script change rather than a doc-only fix
Tried a doc-only approach first (a `sed` snippet in `references/brev.md`) and bounced — the gap is mechanical (precedence rules around `--env-file`), not conceptual, so the right place to fix it is in the script that the skill already tells the agent to source. One more reliable, zero new steps for the agent to remember.

## Reproduction (RTXPRO6000BW Blackwell, Brev launchable, alerts real-time)
- Pre-fix: `curl https://77770-<id>.brevlab.com/` → Cloudflare Access login → after auth → 404 from haproxy. `curl -H "Host: 77770-<id>.brevlab.com" http://localhost:7777/` also 404 (proves haproxy is the gate, not Cloudflare).
- After `source skills/deploy/scripts/brev_setup.sh && docker compose up -d --force-recreate vss-haproxy-ingress vss-agent vss-ui`: same brev URL → 200 with the VSS UI.

## Tests
- [x] Auto-discover patches all four developer profiles (`base`, `lvs`, `search`, `alerts`)
- [x] `--env-file` patches only the named file
- [x] `--no-write` leaves files untouched (verified with `diff`)
- [x] `--quiet` produces no stdout
- [x] Re-sourcing is idempotent (no duplicate lines, values stable)
- [x] End-to-end on a live RTXPRO6000BW Brev box: brev URL serves the UI after `source` + `--force-recreate haproxy/agent/ui`

## Out of scope
- The general "GPU memory caps for shared devices" fix from the closed #301 is **not** in this PR. Will resubmit separately if needed.
- Doc-only changes (`references/brev.md`, `SKILL.md` Step 1c) intentionally not touched here. The script change makes them obsolete; can clean up the docs in a follow-up.


zac-wang-nv:
## Harbor Eval — BLOCKED: L40S CSP unavailable

**PR #302** · head `7c3fc99` · run `25505974054`

All three L40S instance providers failed to provision a healthy instance:

| Provider | Machine Type | Result |
|---|---|---|
| MassedCompute | `massedcompute_L40Sx2` | UNHEALTHY (build stuck) |
| Scaleway | `scaleway_L40Sx2` | UNHEALTHY (build stuck) |
| Launchpad | `gpu-l40s-d.2gpu-64vcpu-384gb` | FAILURE |

**Changed skill:** `skills/deploy` — `scripts/brev_setup.sh`  
**Required platform:** L40S × `remote-all` (5 specs: `base`, `alerts_cv`, `alerts_vlm`, `lvs`, `search`)

The harness adapter bug (undefined `underlying` variable in `generate.py` line 615) was fixed in the workflow artifact — no changes to `skills/` were made.

**Next steps:**
1. A human operator should verify L40S capacity with `brev search --type L40S`.
2. Re-trigger this eval by pushing a new commit or running `/ok to test` once capacity is confirmed.

<sub>Generated by the skills-eval agent (run 25505974054). The adapter fix lives in the workflow artifact `skills-eval-results-pr-302-25505974054.tar.gz`.</sub>

PR #327, opened 2026-05-07
Feat/fix openclaw plugin install with notebook

nvleeryan:
## Description

   - Switch sandbox install from raw `openshell sandbox upload skills/` to `npm pack` + `openclaw plugins install` so the plugin's `register()` runs.
   - Add `prepack`/`postpack` to `.openclaw/package.json` to vendor `../skills`; resolve skills from both `./skills` and `../skills`.
   - Add `OPENCLAW_PLUGIN_VARIANT` env var + `_${variant}/` md overlay system in `.openclaw/index.ts` (with a sticky marker so gateway reloads pick the same variant).
   - Ship `_nemoclaw/` overlay (`AGENTS.md`, `BOOTSTRAP.md`, `TOOLS.md`) with MCP-only deploy guidance and handshake recipes.
   - Refactor `init_nemoclaw.sh`: unify `OPENCLAW_PLUGIN_DIR`, drop per-skill upload helpers, install via tarball with the variant env prefix.
   - Register the VSS Orchestrator MCP server and set `agents.defaults.workspace` 
   - Drop the first-run "ask for a name" flow; identity is hardcoded to "VSS assistant".
   - Tighten `docker_generate` input schema with `Literal` types and `examples`; trust profile `.env` for LLM/VLM names + slugs (drop client-side `supported_models`).
   - Poll `docker_status` every 30s and emit a 1-line per-poll chat summary.
   - Notebook: add `plugins list` / `plugins doctor` / skills-list / workspace-md verification and refresh the GPU-layout doc.

## Checklist


nvleeryan:
/ok to test https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/327/commits/672120d233e9afe13ccc12c15fdae3d69744fda9

nvleeryan:
/ok to test https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/327/commits/7a25d4e142ff58becc032b10d4545a5359fe6bb1

nvleeryan:
/ok to test https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/327/commits/4014cc9addebad15ec609d8281765bca9018ac23

nvleeryan:
/ok to test https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/327/commits/85de02d1ea2d83639bceeb821b78d78e2bb56308

nvleeryan:
/ok to test https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/327/commits/53390eb1ca9ba4a9602fdef2268d84fae199efd4

nvleeryan:
/ok to test https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/327/commits/a460d0e32ad038c9a13d8247a4c2c0b58f21bdeb

nvleeryan:
/ok to test https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/327/commits/d9910f4f2dda2f3facef8d6d1b8ee5c18e026211

nvleeryan:
/ok to test https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/327/commits/906c7a43b28ef4ac9e2b6dcba5df9050c459a09a

zac-wang-nv (commented):
Notebook cell 7 keeps a ~50-line commented-out "host-alias staging" block (and a matching commented `import tempfile`, `env["VSS_PLUGIN_DIR"] = ...`, and print line) for possible re-enabling. A few asks here:

- The disabled block references `PLUGIN_DIR` (e.g. `plugin_src_dir = PLUGIN_DIR.resolve()`), but the live cell renamed that to `OPENCLAW_PLUGIN_DIR`. If anyone uncomments without spotting the rename, the cell will `NameError`.
- Long-lived commented blocks reliably drift out of sync with the surrounding code (this one already has).

Suggest either (a) deleting it now and recovering from git history if you need to re-enable, or (b) gating it behind a single boolean (`STAGE_PLUGIN_FOR_HOST_ALIAS = False`) so the commented identifiers actually live in code that the linter sees. At minimum, fix `PLUGIN_DIR` → `OPENCLAW_PLUGIN_DIR` in the disabled block so re-enabling isn't a foot-gun.

.openclaw/package.json
```diff
@@ -5,6 +5,10 @@
   "type": "module",
   "keywords": ["openclaw", "openclaw-plugin", "vss", "nvidia", "video-search"],
   "files": ["skills/", "workspace/", "openclaw.plugin.json", "README.md", "index.ts"],
+  "scripts": {
+    "prepack": "rm -rf skills && cp -r ../skills skills",
```
zac-wang-nv: Two related concerns with this line:

1. The `["./skills", "../skills"]` entry you added in `openclaw.plugin.json` declares `./skills` as a real source location. Anything that legitimately stages `.openclaw/skills/` (e.g. for a future bundled-skill use case, or a local debugging copy) gets silently wiped on every `npm pack`. The unconditional `rm -rf` should at least be guarded — e.g. only remove a sentinel-tagged directory, or fail if the dir is non-empty and not a copy of `../skills`.
2. `.openclaw/skills/` and `.openclaw/*.tgz` aren't in `.gitignore`. If `npm pack` is interrupted between `prepack` and `postpack`, or if `kubectl exec` later fails before `init_nemoclaw.sh:318` rms the tarball, those artifacts sit in a tracked directory and will show up in `git status`. Worth adding both patterns to the root `.gitignore` (or a new `.openclaw/.gitignore`).

deploy/docker/scripts/nemoclaw/init_nemoclaw.sh
```diff
@@ -293,46 +268,56 @@ apply_vss_policy() {
...
   fi
 
-  remote_skills_dir="$(resolve_vss_remote_skills_dir "${container_name}")"
-  log "Using OpenClaw skills directory ${remote_skills_dir}"
-
-  log "Preparing sandbox skills staging directory inside ${NEMOCLAW_SANDBOX_NAME}"
-  sudo docker exec "${container_name}" kubectl exec -n "${VSS_NAMESPACE}" "${NEMOCLAW_SANDBOX_NAME}" -- sh -lc \
-    "rm -rf '${remote_upload_dir}' && su - sandbox -c \"mkdir -p '${remote_upload_dir}'\""
+  log "Packing VSS OpenClaw plugin in ${plugin_dir}"
+  tgz_name="$(cd "${plugin_dir}" && npm pack 2>/dev/null | tail -n1)"
```
zac-wang-nv: Suppressing all of `npm pack`'s stderr makes failures hard to debug. If `prepack` fails (e.g. `${VSS_REPO_DIR}/skills` doesn't exist, or `cp -r` hits a permission error), `tgz_name` ends up empty and the script falls into the generic `ERROR: npm pack did not produce a tarball in ...` message with no underlying reason logged. Suggest dropping `2>/dev/null` (npm prints progress on stdout, not stderr; `tail -n1` still grabs the tarball name) or routing stderr through the `log()` helper. Bonus: add an early `if [ ! -d "${VSS_REPO_DIR}/skills" ]; then log "skills/ missing — cannot pack"; return 1; fi` before `npm pack` to surface the most common failure mode directly.

deploy/docker/scripts/nemoclaw/update_openclaw_config.py
```diff
@@ -251,6 +268,15 @@ def main() -> int:
         default=DEFAULT_CONFIG_PATH,
         help=f"Path to openclaw.json in the pod (default: {DEFAULT_CONFIG_PATH})",
     )
+    parser.add_argument(
+        "--openclaw-workspace-dir",
```
zac-wang-nv: The flag is added with a default of `/sandbox/.openclaw/workspace`, but no caller actually passes it. `init_nemoclaw.sh:237` invokes the script with only `--config-path "$VSS_REMOTE_CONFIG_PATH"`, and the notebook reads `OPENCLAW_PLUGIN_WORKSPACE_DIR` only for its own client-side verification (cell 9's `expected_workspace_md` check) — nothing pipes that value through to the sandbox.

In practice this is fine because the default matches the sandbox layout, but the flag is currently dead surface area: if anyone overrides the source-side workspace dir, they'll get a silent mismatch between what they verified locally and what was actually written to `agents.defaults.workspace` on the sandbox. Either drop the flag and hard-code the default at the single call site, or thread it through `init_nemoclaw.sh` (and the notebook env var) so it's exercised in the standard flow.

deploy/docker/scripts/nemoclaw/init_nemoclaw.sh
```diff
@@ -293,46 +268,56 @@ apply_vss_policy() {
...
+  openshell sandbox upload "${NEMOCLAW_SANDBOX_NAME}" "${tgz_path}" "/tmp/"
 
-  log "Copying staged skills into sandbox workspace inside ${NEMOCLAW_SANDBOX_NAME} via ${container_name}"
+  # --dangerously-force-unsafe-install: the plugin's index.ts uses child_process (npx skills add agent-browser,
+  # systemctl daemon-reload), which OpenClaw's install-time scanner flags. We trust this first-party plugin.
+  install_cmd="OPENCLAW_PLUGIN_VARIANT=${OPENCLAW_PLUGIN_VARIANT} openclaw plugins install ${remote_tgz} --force --dangerously-force-unsafe-install"
+  log "Installing plugin via: ${install_cmd}"
   sudo docker exec "${container_name}" kubectl exec -n "${VSS_NAMESPACE}" "${NEMOCLAW_SANDBOX_NAME}" -- sh -lc \
-    "rm -rf '${remote_skills_dir}' && su - sandbox -c \"mkdir -p '${remote_skills_dir}' && cp -r '${remote_upload_dir}/'* '${remote_skills_dir}/'\""
+    "su - sandbox -c '${install_cmd}' && rm -f '${remote_tgz}'"
```
zac-wang-nv: Medium confidence — defensive concern only. `${install_cmd}` (which embeds `${remote_tgz}`, derived from `npm pack` output) is interpolated into a single-quoted `su - sandbox -c '…'` string. npm package names today restrict to `[a-z0-9._-]`, so a quote can't actually appear in the tarball name with the current `package.json`. But if the package name/version ever changes to something more exotic (or someone copies this pattern elsewhere), an unescaped `'` would break out of the quoting and end up running as the `sandbox` user inside the pod via `kubectl exec` (which is itself running under `sudo docker exec`). Cheap fix: use `printf -v install_cmd '%q' …` to shell-escape, or pass the tarball path via argv to the inner shell instead of building a quoted string.

deploy/docker/scripts/nemoclaw/init_nemoclaw.sh
```diff
@@ -293,46 +268,56 @@ apply_vss_policy() {
...
 
-  log "VSS skills installed"
+  rm -f "${tgz_path}"
```
zac-wang-nv: Medium confidence — minor cleanup. The local tarball is rm'd only after the `kubectl exec` succeeds. If the install fails (network blip, kubectl error, plugin scanner rejects, etc.), `${plugin_dir}/${tgz_name}` is left behind in `.openclaw/`. Adding a `trap 'rm -f "${tgz_path}"' RETURN` (or an explicit unconditional cleanup before `return 1` in the error paths) would keep the working tree clean across retries.

.openclaw/package.json
nvleeryan: fixed 1 + 2 except not ignoring .openclaw/skills/ as it could be an actual user created folder later

deploy/docker/scripts/nemoclaw/init_nemoclaw.sh
nvleeryan: fixed

deploy/docker/scripts/nemoclaw/update_openclaw_config.py
nvleeryan: fixed.  note that it will require code change (1 line) if sandbox openclaw workspace path changes

deploy/docker/scripts/nemoclaw/init_nemoclaw.sh
nvleeryan: fix ==> surface npm pack errors, guard skills/, and rm tarball via RETURN trap on failure paths

deploy/docker/scripts/nemoclaw/init_nemoclaw.sh
nvleeryan: log intent only, no literal command flags

nvleeryan:
/ok to test https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/327/commits/a6e00bba91c6caa86777d9856633c975bedb4628

nvleeryan:
/ok to test https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/327/commits/a6e00bba91c6caa86777d9856633c975bedb4628

zac-wang-nv (approved):
LGTM

PR #331, opened 2026-05-08
fix: include runtime search params to top agent

ayyappa-dev:
## Description
To address the case of LLM directly returning the result from conversation history instead of calling search_agent, when the runtime params (use_critic, etc.,) are toggled without modifying the query.

## Checklist

### DIFF
--- a/deploy/helm/developer-profiles/dev-profile-search/configs/vss-agent/config.yml
+++ b/deploy/helm/developer-profiles/dev-profile-search/configs/vss-agent/config.yml
@@ -219,7 +219,7 @@ workflow:
       and stream names from the query text — do not filter these out.
     - agent_mode: always set to true (enables query decomposition)
     - use_attribute_search: always set to true (enables fusion reranking)
-    - use_critic: always set to true (verify results with VLM critic agent)
+    - use_critic: use the current request option value (verify results with VLM critic agent)
     Optional:
     - max_results: Number of final results to return to the user (default: 5)
     - top_k: Number of internal candidates to fetch before reranking (default: 10)
@@ -230,6 +230,14 @@ workflow:
     - end_time: Leave as None unless the user provides explicit timestamps or time references.
 
     Directly return the entire search result to user, no filtering or summarization.
+
+    Runtime request options:
+    - request_options is passed automatically to search_agent; do not fill it manually.
+    - On follow-up turns, the system prompt may include current_request_options and previous_request_options.
+    - For search requests and search follow-ups, treat search_source_type as the current media source selection
+      and use_critic as the current critic selection.
+    - If search_source_type or use_critic differs from previous_request_options, call search_agent again for
+      the relevant search query instead of answering from previous search results.
 
     Conversation history guidelines:
     - Always call search_agent for any search query, even if it looks similar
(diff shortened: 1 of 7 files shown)


zac-wang-nv (dismissed):
can you use system prompt for this? we shouldn't code the agent for a specific use case

services/agent/src/vss_agents/agents/top_agent.py
```diff
@@ -441,6 +445,38 @@ def _tool_accepts_param(self, tool_name: str, param_name: str) -> bool:
             return param_name in schema_fields
         return False
 
+    def _runtime_params_for_tool(self, options: AgentRequestOptions, tool_name: str) -> dict[str, Any]:
```
zac-wang-nv: we can parse input options to subagent and let subagent handles the logic

services/agent/src/vss_agents/agents/top_agent.py
```diff
@@ -441,6 +445,38 @@ def _tool_accepts_param(self, tool_name: str, param_name: str) -> bool:
...
+            return ""
+        params_context = json.dumps(
+            {
+                "current_search_params": current_params,
+                "previous_search_result_params": state.last_search_runtime_params,
+            },
+            sort_keys=True,
+            default=str,
+        )
+        return (
```
zac-wang-nv, line 470: this should go to the config

deploy/helm/developer-profiles/dev-profile-search/configs/vss-agent/config.yml
```diff
@@ -236,6 +236,13 @@ workflow:
 
     Directly return the entire search result to user, no filtering or summarization.
 
+    Runtime request options:
```
zac-wang-nv, line 234: We already have those options as an object, that can be passed to subagents as default. the idea is options provide an abstraction(interface) between the frontend and sub-agents. 

from top-agent, we have a check for if toolcalls are sent to subagents, if yes, then that option interface can be passed. 

wrap that from plan then tool call may have downside that the agent has fill tool call parameters correctly in to steps, while if we abstract and pass that, it will be used consistently

services/agent/src/vss_agents/agents/top_agent.py
ayyappa-dev: Addressed.

The request options are now modeled as a shared interface:
- top_agent now passes request_options generically to any tool/subagent that declares that field
- search_agent accepts these request_options and owns the parsing the needed runtime params

services/agent/src/vss_agents/agents/top_agent.py
ayyappa-dev: Addressed.

The search-specific instruction has been moved to the Search profile config (docker and helm). top_agent no longer has search-specific stale-param/rerun policy; it only exposes neutral current/previous request-options context when the prompt opts in.

deploy/helm/developer-profiles/dev-profile-search/configs/vss-agent/config.yml
ayyappa-dev: Addressed.

The latest update uses request_options as the abstraction instead of requiring the planner/LLM to manually fill search runtime parameters.
- top_agent auto-passes request_options only when the target schema accepts it
- search_agent derives effective source_type/use_critic from request_options

zac-wang-nv (approved):
LGTM

PR #375, opened 2026-05-12
feat(lvs): add Nemotron Nano Omni audio-aware VLM support

jiayin-nvidia:
## Description
- Adds Nemotron Nano V3 Omni opt-in to dev-profile-lvs.

## Checklist


jiayin-nvidia:
/ok to test [30bf317](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/375/commits/30bf31727335f19b9ecb70a826d5c71a1e9395a1)

jiayin-nvidia:
/ok to test [731f561](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/375/commits/731f5614d382ab0c24a756fecedbcf07a2c9a59c)

deploy/docker/scripts/dev-profile.sh
```diff
@@ -1194,10 +1194,17 @@ function state_up() {
...
     # RTVI_VLM_MODEL_TO_USE and RTVI_VLM_ENDPOINT come from the profile .env defaults for local
     # (see dev-profile-{alerts,lvs}/.env). Remote VLM overrides these in the block above via VLM_BASE_URL/vlm.
+    # Nemotron Nano Omni opt-in: when source .env has an uncommented
+    # RTVI_VLM_MODEL_TO_USE=vllm-compatible line (i.e. RT-VLM loads an HF git
+    # model in-pod), preserve the user's VLM_NAME / RTVI_VLM_MODEL_PATH instead
+    # of forcing the cosmos NGC defaults. VLM_BASE_URL stays http://<host>:8018
+    # either way.
     if [[ "${vlm_mode}" != "remote" ]]; then
-      set_env_var "VLM_NAME" "nim_nvidia_cosmos-reason2-8b_hf-1208"
+      if ! grep -q "^RTVI_VLM_MODEL_TO_USE=vllm-compatible" "${_source_env}" 2>/dev/null; then
```
zac-wang-nv: The grep+override block (lines 1203–1206) is redundant — `dev-profile-lvs/.env` already has `VLM_NAME=nim_nvidia_cosmos-reason2-8b_hf-1208` (line 75) and `RTVI_VLM_MODEL_PATH="ngc:nim/nvidia/cosmos-reason2-8b:hf-1208"` (line 212) as the active defaults, with the Nemotron alternatives commented out at lines 77 and 219. The standard opt-in flow (comment cosmos / uncomment Nemotron) flows through to `generated.env` without any help from `dev-profile.sh`, and `set_env_var` already handles both commented and uncommented variants for other vars. Suggest dropping the inner `if ! grep ... fi` and the two `set_env_var` calls entirely, keeping only the `VLM_BASE_URL` line (which does need runtime `host_ip` injection since `.env` has it empty). That removes the brittle grep (misses quoted values, leading whitespace, case variants) and the silent-revert footgun in one shot.

deploy/docker/developer-profiles/dev-profile-lvs/.env
```diff
@@ -209,6 +211,19 @@ RTVI_VLLM_GPU_MEMORY_UTILIZATION=''
...
+# Optional: Configuration for Nemotron Nano Omni
+# Uncomment below and set HF_TOKEN
+# -----------------------------------------------------------------------------
+#RTVI_VLM_MODEL_TO_USE=vllm-compatible
+#RTVI_VLM_MODEL_PATH='git:https://huggingface.co/nvidia/Nemotron-Nano-V3-Omni-GA0420-FP8'
+#VLM_MODEL_SUPPORTS_AUDIO=true
+#VLM_TRUST_REMOTE_CODE=true
+#INSTALL_PROPRIETARY_CODECS=true
+#HF_TOKEN=<your_hf_token>
+#ENABLE_AUDIO=true
```
zac-wang-nv, line 224: `ENABLE_AUDIO` and `VST_DISABLE_AUDIO` must be flipped together for audio to flow end-to-end. There's no validation, so the failure mode is asymmetric and silent:

- `ENABLE_AUDIO=true` alone → agent forwards `enable_audio=true` to LVS, but VST has already stripped audio during transcode → Nemotron sees empty audio track, no error.
- `VST_DISABLE_AUDIO=false` alone → audio kept in stored clip, but agent never forwards `enable_audio=true` → VLM ignores it, just wasted storage.

Two cleaner options: (a) collapse into one user-facing knob — derive `VST_DISABLE_AUDIO` from `ENABLE_AUDIO` in `dev-profile.sh state_up` so the user only flips one var; or (b) at minimum, add a comment in the .env block making the lockstep requirement explicit ("⚠ Both must be set together — setting only one silently disables audio.").

services/agent/tests/unit_test/tools/test_lvs_video_understanding.py
```diff
@@ -94,6 +94,16 @@ def test_missing_hitl_template_fails(self):
                 hitl_objects_template="Objects template",
             )
 
+    def test_enable_audio_defaults_false_and_accepts_true(self):
```
zac-wang-nv, line 103: This test only validates the Pydantic field's accept/default behavior — it doesn't exercise the actual code path at `lvs_video_understanding.py:691` that conditionally adds `enable_audio` to the LVS request body. The matching tool (`lvs_config_media`) does have a payload-level positive test (`test_config_media_payload_includes_enable_audio_when_set`) — mirror that here so a future refactor of the `if config.enable_audio:` guard can't silently drop audio forwarding without a test failure.

services/agent/src/vss_agents/tools/lvs_config_media.py
```diff
@@ -129,6 +129,10 @@ class LVSConfigMediaConfig(FunctionBaseConfig, name="lvs_config_media"):
         default=None,
         description="Optional VLM input frame height (pixels). When set, forwarded to LVS to bound the visual-token count.",
     )
+    enable_audio: bool = Field(
```
zac-wang-nv, line 132: The `disable_audio` field on `StreamingIngestConfig` got a detailed five-line description explaining the VST transcode behavior and when to flip it; the two matching `enable_audio` fields (here and at `lvs_video_understanding.py:171`) just say "Enable audio processing". A two-line description (e.g., "When True, forwards `enable_audio=true` to the LVS `/summarize` payload — required for audio-capable VLMs like Nemotron Omni. Pairs with `streaming_ingest.disable_audio=False`.") would help anyone reading the config without the PR context.

zac-wang-nv:
[`deploy/docker/developer-profiles/dev-profile-lvs/vss-agent/configs/config_rag.yml`] (general comment — file not modified in this PR, so no inline anchor)

`config_rag.yml` is the RAG-variant config for this same profile, but it's missing both new audio knobs: `streaming_ingest` has no `disable_audio: ${VST_DISABLE_AUDIO:-true}` and `lvs_video_understanding` (line 98) has no `enable_audio: ${ENABLE_AUDIO:-false}`. A user who selects the RAG variant and follows the .env opt-in (`ENABLE_AUDIO=true`, `VST_DISABLE_AUDIO=false`) will silently get audio-stripped clips at the VST boundary and no `enable_audio=true` in the LVS payload — Nemotron Omni effectively degrades to a vision-only model with no error. Either mirror the two lines added to `config.yml` (`disable_audio: ${VST_DISABLE_AUDIO:-true}` under `streaming_ingest`, `enable_audio: ${ENABLE_AUDIO:-false}` under `lvs_video_understanding`) into `config_rag.yml`, or explicitly document that audio support is `config.yml` only.

zac-wang-nv:
For VST disable audio option, can we just reuse the enable audio option? It is hard to track it's consistence

jiayin-nvidia:
/ok to test [b8f005c](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/375/commits/b8f005cf7d10541b53b988a48a921610ae59035c)

deploy/docker/developer-profiles/dev-profile-lvs/.env
jiayin-nvidia: Followed https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/375#issuecomment-4433609151 to remove `VST_DISABLE_AUDIO` and only keep `ENABLE_AUDIO`.

jiayin-nvidia:
/ok to test [fa2b22d](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/375/commits/fa2b22dd1c578b86ad1540dbfaeb39a3ddf4e628)

zac-wang-nv (dismissed):
LGTM

PR #416, opened 2026-05-13
fix(agent,ui): align media_type='rtsp'/'video' end-to-end; UI stops hardcoding 'video-stream' (NVBug 6171391)

zac-wang-nv:
## Summary

The UI's `+ Chat` button on `StreamCard` hardcoded the chat-context tag to `type: 'video-stream'` for **every** VST asset — uploaded MP4 or RTSP stream. On the base profile the LLM trusted that tag and called `report_agent(media_type='stream', sensor_id=streamId)`, which fails (no `lvs_stream_understanding_tool` configured). On LVS the same call ran but routed through the stream path even for uploaded MP4s — also wrong.

Minimal fix in two places:

1. **UI** — stop hardcoding. Use the existing `isRtspStream(stream)` helper. Emit `type: 'rtsp'` for live streams, `type: 'video'` for uploaded files. End-to-end the UI tag now matches the agent's `media_type` enum value, so the LLM mirrors it correctly.

2. **Agent value rename** `'stream'` → `'rtsp'` on the two pydantic models (`VideoReportGenInput`, `VideoReportAgentInput`) and on the `vst_video_list` output. Default stays `'video'`. Field name stays `media_type` (no `source_type` rename — that's a separate concept the search side uses).

## Why minimal is enough

- **`lvs_stream_understanding`** has no `media_type` / `source_type` input — tool-name-discriminated. No change.
- **`report_agent`'s validator** already rejects mixed-type / multi-stream input (`'rtsp' requires a single sensor_id`, `'rtsp' requires both start_time and end_time`). UI doesn't duplicate.
- **Agent prompts** should not reference UI payload internals. An earlier iteration of this PR had "Chat-context payload mapping" bullets in base + LVS prompts coupling the agent to the UI tag — wrong layer. Removed.
- **Typed queries (no `+ Chat`)** were never broken; this PR only fixes the `+ Chat` path.

## Changes (8 files)

**Agent (Python)**
- `services/agent/src/vss_agents/tools/video_report_gen.py`: `VideoReportGenInput.media_type: Literal["video", "rtsp"]` (was `"stream"`). Validator + error messages + docstrings updated.
- `services/agent/src/vss_agents/agents/report_agent.py`: same value change on `VideoReportAgentInput`.
- `services/agent/src/vss_agents/tools/vst/video_list.py`: emits `media_type: 'rtsp' | 'video'`.
- `services/agent/tests/unit_test/tools/vst/test_video_list_coverage.py`: asserts `'rtsp'`.

**Prompts (docker + helm, LVS only)**
- Routing examples: `report_agent(media_type='rtsp', …)`.

**UI**
- `services/ui/.../components/StreamCard.tsx`: `type` derived from `isRtspStream(stream)`.
- `services/ui/.../__tests__/components/StreamCard.test.tsx`: separate tests for uploaded (`'video'`) and RTSP (`'rtsp'`) chips.

## Test plan

- [x] Pydantic round-trip on `VideoReportGenInput`: defaults to `'video'`, accepts `'rtsp'`, rejects legacy `'stream'`, validator enforces start/end for `'rtsp'`.
- [x] YAML lint passes on all touched profile configs.
- [x] AST parse + ruff check + ruff format + mypy + DCO all pass.
- [x] StreamCard unit tests cover both uploaded-MP4 and RTSP payload shapes.
- [ ] Deploy base profile, upload an MP4, click `+ Chat`, send "Generate report for this sensor" → report generates (NVBug 6171391 repro).
- [ ] Deploy LVS profile, add an RTSP stream, click `+ Chat`, send "Generate report from 0 to 60s" → RTSP-stream report path runs.
- [ ] On both profiles, type "Generate report for warehouse_safety_0002" without `+ Chat` → still works (regression check).

## History

PR has two commits; the second (`2175d76`) scopes down the first (`2b6b16e`) in response to review. Net diff against `develop` is the slim version. Squash on merge will give a single clean commit.

Fixes [NVBug 6171391](https://nvbugspro.nvidia.com/bug/6171391).

### DIFF
--- a/services/ui/packages/nv-metropolis-bp-vss-ui/video-management/lib-src/components/StreamCard.tsx
+++ b/services/ui/packages/nv-metropolis-bp-vss-ui/video-management/lib-src/components/StreamCard.tsx
@@ -2,7 +2,7 @@
 import React, { useState, useEffect, useRef, useCallback } from 'react';
 import { Button } from '@nvidia/foundations-react-core';
 import type { StreamInfo, ChatSidebarQueryContext } from '../types';
-import { getFileExtension, isRtspStream, fetchPictureWithQueue } from '../utils';
+import { getFileExtension, isRtspStream, fetchPictureWithQueue, getStreamType } from '../utils';
 import { createApiEndpoints } from '../api';
 import { copyToClipboard } from '@nemo-agent-toolkit/ui';
 import { IconCheck } from '@tabler/icons-react';
@@ -119,14 +119,18 @@ export const StreamCard: React.FC<StreamCardProps> = ({
   };
 
   const handleCopyContext = useCallback(async () => {
-    const type = isRtspStream(stream) ? 'rtsp' : 'video';
-    const data = { sensorName: stream.name, streamId: stream.streamId };
+    const data = {
+      sensorName: stream.name,
+      streamId: stream.streamId,
+      mediaType: getStreamType(stream),
+    };
     const text = JSON.stringify(data, null, 2);
 
     onAddChatQueryContext?.({
       id: `video-mgmt-stream:${stream.streamId}`,
       label: stream.name,
-      type,
+      // contextType: UI-only (chip tooltip / future grouping); not sent to the backend — see Chat onSend.
+      contextType: 'media/video',
       data,
     });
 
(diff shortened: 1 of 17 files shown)


liamy-nv (dismissed):
Renaming looks good

services/ui/packages/nv-metropolis-bp-vss-ui/video-management/lib-src/components/StreamCard.tsx
```diff
@@ -119,13 +119,14 @@ export const StreamCard: React.FC<StreamCardProps> = ({
   };
 
   const handleCopyContext = useCallback(async () => {
+    const type = isRtspStream(stream) ? 'rtsp' : 'video';
```
akshayanv: We will use getStreamType func.

I shall push a change

akshayanv (changes requested):

akshayanv (approved):

PR #434, opened 2026-05-14
AGENT UI: Video Management Tab: Remove icon info when add rtsp

ghost:


akshayanv (commented):
@hntran-nvidia 
So removing the info icon was a good thing?

Instead should we have made the info icon functional?

akshayanv (commented):
@hntran-nvidia 

add NVBUG link to the commit message

PR #453, opened 2026-05-14
fix (agent): Omni/audio VLM wiring and VST clip URL handling

liamy-nv:
## Description

- disable_audio / VST disableAudio: Wire a configurable flag through vst.video_clip (get_video_url query param), video_report_gen (clip injection via _generate_single_report / _inject_video_clips), with small test fixture updates where get_video_url is mocked.
- Frame-mode clip downloads (video_understanding): Smarter handling when the agent downloads the clip—avoid host rewrites that break presigned URLs, prefer internal VST URLs for /vst/… paths, keep translate_url for modes where the VLM fetches the URL; add validation when the HTTP body is not real video (HTML/JSON/empty).
- vst.video_clip playback URLs: Add use_external_playback_url so server-side consumers can keep internal VST URLs when the API returns a proxy/external host; tests for that path.
- Dev-profile-base deploy: Helm config.yml updates (enable_audio, use_external_playback_url, disable_audio, RTVI/VLM naming + VLM_BASE_URL, url_validator.internal_ip → EXTERNAL_IP, report eval reference_base_dir default); Docker config.yml aligned with Helm; Docker .env documents ENABLE_AUDIO (default false).

## Checklist


liamy-nv:
/ok to test [c63a6dd](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/453/commits/c63a6dd5cf840ef4e2b4a50b089ad485c79d4f91)

liamy-nv:
/ok to test [afc69d9](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/453/commits/afc69d93156242bb513b344aab0f05ae1231ba0c)

deploy/docker/developer-profiles/dev-profile-base/vss-agent/configs/config.yml
```diff
@@ -73,6 +75,9 @@ functions:
     _type: vst.video_clip
     vst_internal_url: ${VST_INTERNAL_URL}
     vst_external_url: ${VST_EXTERNAL_URL}
+    use_external_playback_url: false
+    # VST disableAudio=false so audio-capable VLMs receive audio in video_understanding / UI clips.
+    disable_audio: true
```
zac-wang-nv: can we change this option to enable_audio and use the same env, ENABLE_AUDIO?

zac-wang-nv:
can you also check helm and sync helm and mypy tests?

liamy-nv:
/ok to test [b9868fb](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/453/commits/b9868fbcbbdfc77e49703dbeb6d6a9c41ed0f90e)

liamy-nv:
/ok to test [ef632e2](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/453/commits/ef632e2b62259e4f44e2631845b35e107f5bc4ca)

liamy-nv:
/ok to test [9d17592](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/453/commits/9d17592f064e29ab1a97de393bd5f0f566f7ca10)

liamy-nv:
/ok to test [fd60e15](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/453/commits/fd60e15b0db621224ea6aa7221da7ee27810fbb3)

deploy/docker/developer-profiles/dev-profile-base/vss-agent/configs/config.yml
zac-wang-nv: Done in commit `9d17592f` — every `.env` now uses `ENABLE_AUDIO`, every `config.yml`/`values.yaml` uses `enable_audio: ${ENABLE_AUDIO:-false}`, and the Pydantic configs (`VSTVideoClipConfig.enable_audio`, `VideoReportGenConfig.enable_audio`) match.

deploy/docker/developer-profiles/dev-profile-base/vss-agent/configs/config.yml
```diff
@@ -478,7 +484,7 @@ eval:
     report_evaluator:
       _type: report_evaluator
       eval_metrics_config_path: ${EVAL_DIR}/report_eval_metrics.yaml
-      reference_base_dir: ${REPORT_REFERENCE_BASE_DIR}
+      reference_base_dir: ${REPORT_REFERENCE_BASE_DIR:-${EVAL_DIR}}/gt
```
zac-wang-nv: As shipped, Docker users hit a doubled `/gt`. `dev-profile-base/.env:271` defines `REPORT_REFERENCE_BASE_DIR=/vss-agent/agent_eval/dataset/vss-devx-base/gt` (already ends in `/gt`), and the new template unconditionally appends `/gt`, so `reference_base_dir` resolves to `…/vss-devx-base/gt/gt`. Pick one of:
- (a) drop the trailing `/gt` from `.env` so the template owns the suffix,
- (b) drop the appended `/gt` from the template and keep the env var pointing at the final directory,
- (c) move the `/gt` to only the default branch: `${REPORT_REFERENCE_BASE_DIR:-${EVAL_DIR}/gt}` and require the env var to be the final dir.

Same fix needed in `deploy/helm/developer-profiles/dev-profile-base/configs/vss-agent/config.yml:487`, `deploy/helm/developer-profiles/dev-profile-base/configs/vss-agent/config_rag.yml:504`, and in `deploy/docker/developer-profiles/dev-profile-base/vss-agent/configs/config_rag.yml:504` (which still has the old `${REPORT_REFERENCE_BASE_DIR}` — see the other comment).

zac-wang-nv:
**[unchanged file: `deploy/docker/developer-profiles/dev-profile-base/vss-agent/configs/config_rag.yml:504`]**

This is the only one of four base-profile configs you did NOT update with the new `${REPORT_REFERENCE_BASE_DIR:-${EVAL_DIR}}/gt` template:
- `deploy/docker/.../dev-profile-base/.../config.yml:487` → new template
- `deploy/helm/.../dev-profile-base/.../config.yml:487` → new template
- `deploy/helm/.../dev-profile-base/.../config_rag.yml:504` → new template
- `deploy/docker/.../dev-profile-base/.../config_rag.yml:504` → **unchanged**

Decide on one shape (per the inline comment on `config.yml:487`) and apply it identically to all four files.

zac-wang-nv:
**Scope / design suggestion** — recommend dropping the VST clip URL handling additions and shipping this PR as just the `enable_audio` wiring + the `reference_base_dir` fix. Rationale, with file/line cites:

1. **`_rebuild_vst_clip_url_with_internal_base` (`services/agent/src/vss_agents/tools/video_understanding.py:139`)** duplicates `utils/url_translation.rewrite_to_internal_vst_url`, which is already called two lines above the new block at `video_understanding.py:614`. For any URL whose path starts with `/vst/`, both functions emit the same result. The new second-rewrite block at lines 611-622 (`if use_video_base64 and not signed_q: rebuilt = _rebuild_vst_clip_url_with_internal_base(...)`) is unreachable on the happy path because `rewrite_to_internal_vst_url` already prefixes the URL with `internal_clip_base`, so the `not video_url.startswith(internal_clip_base)` guard never fires.

2. **`_query_looks_like_signed_object_storage` (`video_understanding.py:118`)** + the skip-rewrite branch (`video_understanding.py:602-608`) is for presigned object-storage URLs. VSS doesn't return those from VST today, so the helper + the tests + the AWS-only / Azure-SAS / GCS-V4 footgun are all guarding a scenario that hasn't shipped.

3. **`use_external_playback_url=false` on `VSTVideoClipConfig`** (and the internal-rebuild branch in `_vst_video_clip` at `services/agent/src/vss_agents/tools/vst/video_clip.py:311-329`) tries to keep the URL internal so server-side downloads work behind a proxy. But every server-side consumer of `vst_video_clip`'s output already calls `rewrite_to_internal_vst_url` on it (`video_understanding.py:614`, `lvs_video_understanding.py:662`), so the URL is already internal by the time it matters. Setting the flag at the *producer* also causes the inline comment on `config.yml:75`: the same `vst_video_clip` is wired as `video_report_gen.video_url_tool`, so the report's "Video Playback" link becomes internal-only.

4. **`_response_body_looks_like_video_error` (`video_understanding.py:124`)** mostly catches HTML/JSON content that `aiohttp.read()` + OpenCV would already surface as a different error; the friendlier message it adds is built around "broken signing" framing that doesn't apply to VSS.

Shipping the PR with these additions kept will mean the inline comments on `config.yml:75` and `video_understanding.py:138` need follow-up fixes. Removing them is the smaller change set and leaves the working `enable_audio` fix intact.

deploy/docker/developer-profiles/dev-profile-base/vss-agent/configs/config.yml
```diff
@@ -73,6 +75,9 @@ functions:
     _type: vst.video_clip
```
zac-wang-nv, line 75: Only matters if you keep `use_external_playback_url` per the scope comment. `vst_video_clip` here has `use_external_playback_url: false`, and the same tool is wired as `video_report_gen.video_url_tool` (line 96). `_generate_single_report` calls it via `video_url_tool.ainvoke({"sensor_id": sensor_id})` (`video_report_gen.py:1888`). I verified both input schemas (`VSTVideoClipOffsetInput`, `VSTVideoClipISOInput`): `start_time`/`end_time` are `… | None = Field(None, ...)`, so the call succeeds and returns a URL. With the flag false, that URL is `VST_INTERNAL_URL`, and `_generate_single_report` appends it verbatim under `## Resources / Video Playback` (line 1897). Net: in dev-profile-base, the report's trailing playback link is internal-only and won't open in a browser. (The `[Watch Clip]` links inside the body are unaffected — `_inject_video_clips` always rewrites to `vst_external_url`.) Same pairing exists in `config_rag.yml:92`.

If you do keep the flag, either point `video_report_gen.video_url_tool` at a different `vst.video_clip` instance with the flag at default true, or have `_generate_single_report` rewrite to `vst_external_url` before appending.

services/agent/src/vss_agents/tools/video_understanding.py
```diff
@@ -113,6 +115,61 @@ def _is_remote_vlm(vlm_mode: str | None) -> bool:
...
+
+def _rebuild_vst_clip_url_with_internal_base(video_url: str, vst_internal_url: str | None) -> str | None:
+    """If *video_url* is a /vst/... playback URL, rebuild it using *vst_internal_url* (scheme+host+port)."""
+    if not vst_internal_url:
+        return None
+    parsed = urlparse(video_url)
+    path = parsed.path or ""
+    if "/vst/" not in path:
+        return None
+    base = urlparse(vst_internal_url.rstrip("/"))
```
zac-wang-nv: If you keep this helper (per the scope comment I'd suggest dropping it), three small fixes:

1. **Naming:** the function reads as a `bool` predicate (`_looks_like…`) but is a void-returning raiser. A future caller will write `if _response_body_looks_like_video_error(...)` and get a TypeError or pass a truthy `None` past their guard. Rename to `_assert_video_response_or_raise`.
2. **Dead branch:** `head.startswith(b"<!")` is unreachable — `head.startswith(b"<")` already catches `<!DOCTYPE`. Drop the `or head.startswith(b"<!")` clause. (Greptile's review thread on this is marked resolved upstream but the line in the diff is unchanged.)
3. **Log content:** the error messages embed up to 400 chars of upstream body. If the upstream is an auth gateway HTML page, the preview can include tenant IDs, internal hostnames, or CSRF tokens that end up in framework logs. Consider trimming to ~80 chars or gating the preview behind a debug-level log.

zac-wang-nv:
**[unchanged file: `deploy/docker/developer-profiles/dev-profile-base/vss-agent/configs/config_rag.yml:222`]**

Two related drifts in the same block, worth a quick decision:

1. In `dev-profile-base/config.yml:213-214`, `rtvi_vlm` now uses `${VLM_NAME}` + `${VLM_BASE_URL}/v1`. Here in `config_rag.yml` it still uses hardcoded `nvidia/cosmos-reason2-8b` + `${RTVI_VLM_BASE_URL}/v1`. If the two configs are *meant* to diverge (base shares the main VLM, RAG keeps a dedicated RTVI service), a one-line comment would save a future debug session. If unintentional, both should move in lockstep and `RTVI_VLM_BASE_URL` can be retired from `.env`.

2. The hardcoded `model_name: nvidia/cosmos-reason2-8b` will drift the day the cosmos-reason NIM ships a renamed model id; source it from `${RTVI_VLM_MODEL_TO_USE}` or `${VLM_NAME}` (both already in `.env`).

zac-wang-nv:
**Testing gap** — `_inject_video_clips(enable_audio=True)` has no unit-test coverage. The new `TestFrameModeClipUrlHelpers` and `test_get_video_url_disable_audio_false` cover the helpers and the VST URL builder, but the report-gen plumb-through (`video_report_gen.py:1875-1879` passing `config.enable_audio` into `_inject_video_clips` → `get_video_url(disable_audio=not enable_audio)`) is only exercised on the default `enable_audio=False` path. Add a sibling to `test_get_video_url_disable_audio_false`: drive `_inject_video_clips` with `enable_audio=True` and assert the generated clip URL contains `disableAudio=false`. Cheap and prevents the whole Omni audio path from regressing silently.

deploy/docker/developer-profiles/dev-profile-base/vss-agent/configs/config.yml
liamy-nv: changed to ${REPORT_REFERENCE_BASE_DIR}

liamy-nv:
/ok to test [4e32f31](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/453/commits/4e32f31a51aee6b0b226c2aff37e2973f73ab83a)

zac-wang-nv (dismissed):
lgtm

PR #478, opened 2026-05-15
docs: fix Brev launchable defaults

freshyjmp:
## Description

Fixes Brev Launchable notebook defaults from the doc audit:

- Adds optional `NGC_CLI_ORG` and leaves `org` unset by default so non-internal users use their NGC account default instead of the internal `nvstaging` org.
- Changes the GitHub clone default from `develop` to `release/3.2.0` so launchable deployments track the GA release branch.

## Validation

- `jq empty deploy/docker/scripts/deploy_vss_launchable.ipynb`
- `git diff --check`


zac-wang-nv:
@freshyjmp  can you fix DCO?

zac-wang-nv (approved):
LGTM

PR #482, opened 2026-05-15
fix(lvs): override per-chunk frame count for stream captioning

jiayin-nvidia:
## Description
- The LVS-side default of 1 frame per chunk made the VLM emit ~1 ms event durations on stream-mode reports.
- Pass `num_frames_per_second_or_fixed_frames_chunk` (default 10) in the `/v1/generate_captions` payload via a new `num_frames_per_chunk` config field on `lvs_config_media`. 
- Include `num_frames_per_chunk: 10` in the `dev-profile-lvs` agent config (docker + helm).

## Checklist

### DIFF
--- a/services/agent/src/vss_agents/tools/lvs_config_media.py
+++ b/services/agent/src/vss_agents/tools/lvs_config_media.py
@@ -124,6 +124,10 @@ class LVSConfigMediaConfig(FunctionBaseConfig, name="lvs_config_media"):
         default=10,
         description="Frames per chunk sent to the VLM. Forwarded as `num_frames_per_second_or_fixed_frames_chunk`.",
     )
+    use_fps_for_chunking: bool = Field(
+        default=False,
+        description="If True, interpret `num_frames_per_chunk` as FPS instead of fixed frames per chunk.",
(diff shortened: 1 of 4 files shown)


services/agent/src/vss_agents/tools/lvs_config_media.py
```diff
@@ -120,6 +120,10 @@ class LVSConfigMediaConfig(FunctionBaseConfig, name="lvs_config_media"):
     conn_timeout_ms: int = Field(default=5000, description="Connection timeout in milliseconds.")
     read_timeout_ms: int = Field(default=600000, description="Read timeout in milliseconds.")
     chunk_duration: int = Field(default=10, description="Duration of each stream chunk in seconds.")
+    num_frames_per_chunk: int = Field(
```
zac-wang-nv, line 123: can you also set use_fps_for_chunking=False?  in this case, even if downstream service changes the default, it will still work.

zac-wang-nv (approved):
LGTM

zac-wang-nv (approved):
LGTM

PR #559, opened 2026-05-19
support ENABLE_CRITIC mirroring dev-profile.sh

nvleeryan:
## Description
Support ENABLE_CRITIC for search profile mirroring dev-profile.sh

  1. services/agent/src/vss_agents/orchestrator/docker_compose_util.py — added enable_critic field to
  DryRunRecipe/create_dry_run_recipe; in build_resolved_env for search profile:
    - enable_critic.lower() == "false" → writes ENABLE_CRITIC=false AND forces VLM_NAME_SLUG=none (skip local VLM)
    - otherwise → writes ENABLE_CRITIC=true
    - non-search → no write (matches dev-profile.sh)
    - env_overrides still win as last write
  2. services/agent/src/vss_agents/orchestrator/tools.py — added enable_critic: str = Field(default="",
  validation_alias="ENABLE_CRITIC") to OrchestratorToolRuntimeSettings, added to the strip-validator, and passed through to
  create_dry_run_recipe.
  3. services/agent/tests/unit_test/orchestrator/test_docker_compose_util.py — added TestEnableCritic with 6 cases:
    - search + enable_critic="false" → ENABLE_CRITIC=false + VLM_NAME_SLUG=none
    - search + enable_critic="FALSE" → case-insensitive
    - search + enable_critic="true" → ENABLE_CRITIC=true, VLM_NAME_SLUG unchanged
    - search + enable_critic=None → defaults to ENABLE_CRITIC=true
    - non-search + enable_critic="false" → ENABLE_CRITIC not written
    - env_overrides override the logic

## Checklist


nvleeryan:
/ok to test https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/559/commits/68d141d01eff3c1e2bea7f38130dc803f991c365

zac-wang-nv:
can it work with the compose edit or env edit tools?  we don't want to build all those knobs into the mcp server, which will make it coupled with profile specific settings

nvleeryan:
> can it work with the compose edit or env edit tools? we don't want to build all those knobs into the mcp server, which will make it coupled with profile specific settings

@zac-wang-nv - sure, i think editing via the compose / edit tools could be a better approach.  We can remove the notebook and ENV knob.  However, we should still keep the ENABLE_CRITIC logic as dev-profile.sh was enforcing it internally in the code (see below rules that dev-profile.sh enforces)

- enable_critic == "false" →  forces VLM_NAME_SLUG=none (skip local VLM)
- otherwise → writes ENABLE_CRITIC=true
- Non Search profile → Ignore ENABLE_CRITIC

hugoverjus:
please don't forget to update agent tag before merging

PR #535, opened 2026-05-19
docs(vss-deploy-profile): flip Brev secure-link to current 7777-<id> form

zac-wang-nv:
## Summary

Brev no longer adds the trailing `0` to launchable-created secure links — the canonical form today is `7777-<BREV_ENV_ID>.brevlab.com` (matching the haproxy ingress port directly). The deploy skill's brev.md and SKILL.md still used the legacy `77770-` form throughout, so any new user following the docs hits a 404 when they paste the URL.

## Changes

- **`references/brev.md`**: architecture diagram, URL format header, sed override in Setup flow, verify-deploy echo, Agent-report-URL troubleshooting row — all flipped to `7777-<id>`.
- **`references/brev.md`**: explanation rewritten — `7777` is just the haproxy port, with an inline note that older launchables used to add a trailing `0`.
- **`references/brev.md`**: troubleshooting "link won't load" row inverted — current default is `7777-`; `77770-` is the legacy form to watch for if a user inherits an older instance.
- **`SKILL.md`** Step 1 EXTERNAL_IP example and Step 1d `sed` override both flipped to `7777-`.

Two intentional `77770-...` mentions remain in brev.md (in the legacy-callout text) — those flag the deprecated form for users who inherit older instances.

## Test plan

- [x] Render brev.md on the GitHub PR diff view — confirm the architecture diagram, URL formats, and the troubleshooting table render cleanly.
- [x] Verify SKILL.md Step 1d `sed` produces `EXTERNAL_IP=7777-<id>.brevlab.com` against a real Brev instance.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #567, opened 2026-05-20
eval(vss-summarize-video, vss-search-archive): pin to RTXPRO6000BW (LVS × 1, search × 2)

zac-wang-nv:
## Summary

Three eval specs declared `L40S` × `gpu_count: 1` (the prior "pick the cheapest available host" default). Operator allocation has shifted — L40S nodes are spoken for by other sweeps, and the `RTXPRO6000BW` × 2-GPU host has been carrying these workloads in practice. Pin the resource declaration so the dispatcher stops trying L40S (which queues indefinitely against an empty pool) and the env-preamble guidance no longer points at the wrong hardware.

## Changed

| Spec | Resource change | Env preamble change |
|---|---|---|
| `skills/vss-summarize-video/evals/lvs_api_ops.json` | `L40S × 1` → `RTXPRO6000BW × 2` | none — preamble already said "independent of GPU model once the lvs profile is deployed" |
| `skills/vss-summarize-video/evals/lvs_profile_summarize.json` | `L40S × 1` → `RTXPRO6000BW × 2` | "Pick the cheapest available host (L40S recommended)" → "Pinned to `RTXPRO6000BW` with `gpu_count: 2`" |
| `skills/vss-search-archive/eval/search.json` | `L40S × 1` → `RTXPRO6000BW × 2` | same one-line rephrase as the summarize profile |

## Out of scope

Two stale items I noticed but **didn't** roll into this PR to keep the diff narrow:

- The `BREV_LINK_PREFIX defaulting to 77770 per launchable convention` prose in both env preambles is wrong post-#535 (correct value is `7777`). Will sweep that across every spec that still references the legacy prefix in a follow-up.
- `vss-summarize-video/evals/` uses the plural form; repo convention is singular `eval/` (see #531). Renaming touches the dispatcher's path resolution — handled separately.

## Test plan

- [ ] Dispatch the skill-eval workflow for `vss-summarize-video` against `develop` after merge — confirm both specs land on the `RTXPRO6000BW` 2-GPU node and don't queue against L40S.
- [ ] Same for `vss-search-archive`.
- [ ] Confirm the existing reward values from prior `RTXPRO6000BW` runs hold (no platform-driven regression).


zac-wang-nv:
Thanks for the review. The P1 (\`gpu_count\` mismatch) is resolved as-is — Greptile reviewed an earlier snapshot when the PR title still said *"pin to RTXPRO6000BW × 2 GPUs"*. After the second commit (\`a76e34f\`) the PR title + description were updated to spell out the intentional per-skill asymmetry: **search × 2, LVS × 1**.

Reasoning per workload's actual local-GPU footprint:

| Spec | Local-GPU work | Pin |
|---|---|---|
| \`vss-search-archive/eval/search.json\` | Cosmos Embed1 runs on the GPU host (the search profile uses local Embed1 even in \`remote-all\`); benefits from extra VRAM for parallel embedding | **RTX × 2** |
| \`vss-summarize-video/evals/lvs_api_ops.json\` | API-surface probe only (\`/v1/ready\`, \`/models\`, \`/metrics\`); no GPU work on the host | **RTX × 1** |
| \`vss-summarize-video/evals/lvs_profile_summarize.json\` | LVS is throughput-bound on the **remote** VLM; GPU host only runs LVS / VST / postgres / phoenix containers | **RTX × 1** |

The Greptile flowchart's failure mode (*"only 2-GPU nodes in pool → queues indefinitely"*) doesn't apply — the operator's RTX pool vends 1-GPU and 2-GPU allocation units independently. Pinning LVS to × 1 saves capacity for other concurrent work on the same host.

Re-triggering Greptile now so the next pass sees the current title + body alongside the code.

PR #574, opened 2026-05-20
skill(vss-summarize-video): keep agent in flow after VIOS sub-skill

jiayin-nvidia:
## Summary

Fixes the `lvs_profile_summarize` eval failure (1/8 -> 8/8 expected) where the agent invoked the `vss-manage-video-io-storage` sub-skill, got the clip URL back, and then ended its turn instead of continuing to `POST /v1/summarize`.

## Root cause

Step 1 in `SKILL.md` was framed as `Resolve the video to a clip URL (delegate to vss-manage-video-io-storage)`. The word "delegate" read to the agent as a hand-off: it produced a `Result for vss-summarize-video: ...` block with the clip URL and exited. The actual deliverable (the LVS summary) in Step 2 was never reached.

HITL was not a factor - the eval query said "Run autonomously without prompting for confirmation" and the agent correctly skipped HITL.

Full trace analysis from CI run 26172961781:
- Agent invoked `Skill: vss-summarize-video`
- Per Step 1, invoked `Skill: vss-manage-video-io-storage`
- Sub-skill returned `streamId`, duration=120s, `videoUrl`
- Final message: *"Returning these details to the summarization skill. Result for vss-summarize-video: ..."* - session ended

## Fix

`skills/vss-summarize-video/SKILL.md`:

1. Renamed Step 1 from *"Resolve the video to a clip URL (delegate to ...)"* to *"Get the clip URL via vss-manage-video-io-storage (sub-task, NOT the final answer)"*.
2. Added an explicit "you are not done when it returns" paragraph at the top of Step 1 and a "proceed immediately to Step 2" instruction at the end.
3. Mirrored the rule in the Tips section so the agent has multiple anchor points to recall it from cache after the nested `Skill` invocation.

Used plain ASCII hyphens in new content; existing em-dashes left alone to keep the diff minimal.

## Test plan

Verified locally against a live VSS lvs deployment with the verbatim eval query:

|                          | Before (CI run 26172961781) | After       |
|--------------------------|------------------------------|-------------|
| `POST /v1/summarize` calls | 0                            | 1           |
| Final answer             | Clip URL, no summary         | LVS `video_summary` rendered verbatim |
| `num_turns`              | 10 (exited early)            | 13 (completed) |
| Result subtype           | n/a                          | `success`   |
| Duration                 | 37s (premature)              | 161s (real LVS work) |

Trajectory: agent traversed Step 1 (VIOS sub-skill returned clip URL) and continued into Step 2b, made exactly one POST to `/v1/summarize` with the three user-supplied events, and rendered the LVS response verbatim with the correct header format.

## Out of scope

`lvs_api_ops` failure (5/8) is a separate issue - the CI eval host runs a mock Python LVS stub that doesn't implement `/recommended_config` or `/metrics`. Will follow up with a separate PR after deciding between extending the stub vs relaxing the eval spec.

### DIFF
--- a/skills/vss-summarize-video/SKILL.md
+++ b/skills/vss-summarize-video/SKILL.md
@@ -165,27 +165,37 @@ done
 
 ---
 
-## Step 1 — Resolve the video to a clip URL (delegate to `vss-manage-video-io-storage`)
+## Step 1 - Get the clip URL via `vss-manage-video-io-storage` (sub-task, NOT the final answer)
 
-**Use the `vss-manage-video-io-storage` skill for all VIOS interactions** — it owns the
+**Use the `vss-manage-video-io-storage` skill for all VIOS interactions** - it owns the
 canonical curl recipes, parameter defaults, and delete/upload flows. Do not
 fabricate URLs or hand-roll VIOS calls here; they will drift.
 
-From `vss-manage-video-io-storage`, you need exactly three things for summarization:
+Calling `vss-manage-video-io-storage` is a sub-task. You (the summarization
+agent) are not done when it returns. The clip URL and duration are inputs
+to Step 2 below, which is where summarization actually happens. Do NOT
+end your turn after this step; do NOT return the clip URL as the final
+answer to the caller.
 
-1. **`streamId`** for the video (via `sensor/list` → `sensor/<id>/streams`,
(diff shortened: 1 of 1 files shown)


zac-wang-nv (approved):
LGTM

PR #579, opened 2026-05-20
fix(vss-search-archive): hard-stop before /generate if source unverified

zac-wang-nv:
## Summary

Tightens the `vss-search-archive` mandatory-workflow rule so the agent **cannot** fire `POST /generate` until it has confirmed the user's named source actually exists in VIOS.

Diagnosed via the eval trial [`step-1__Ej3rDoP`](https://harbor-13xh5gpe7.brevlab.com/jobs/26186987009__2026-05-20__20-11-25/tasks/rtxpro6000bw/claude-code/aws/anthropic%2Fbedrock-claude-sonnet-4-6/nvidia-vss%2Fvss-search-archive-search-rtxpro6000bw-step-1/trials/step-1__Ej3rDoP) on Skills Eval run [#26186987009](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/26186987009) (eval step 1: \"Find all instances of forklifts in the airport video\" — and there's no airport video registered).

## What the agent did

| Step | Action |
|---|---|
| 2-4 | Loaded \`vss-search-archive\` |
| 5-6 | Probed \`/docs\` + ES health |
| 7 | \"Both services are up. Now running the forklift search against the **airport video**.\" |
| **8** | ❌ \`POST /generate\` with \`input_message: \"find all instances of forklifts in the airport video\"\` — no \`/sensor/list\` had been called yet |
| 9 | Got back \"No sensors are ingested\" |
| 10+ | Recovered: read troubleshooting → used \`vss-manage-video-io-storage\` → \`/sensor/list\` → final answer to user |

Final reply was correct (acknowledged the airport video doesn't exist, asked user to clarify), but step 8 violated the eval's check #3 (`zero POST .../generate requests` for missing-source step) — reward 0.75 instead of 1.0.

## Root cause

The skill's mandatory workflow only had **one** hard stop (between \"resolve \$HOST_IP\" and \"fire /generate\"). The agent was free to fire optimistically and let `/generate` echo back \"no sensors\" as the discovery mechanism. Reinforced by a contradicting Gotcha that read *\"Assume video sources are already ingested. No need to search for them locally.\"*

## Fix

`skills/vss-search-archive/SKILL.md`:

- **New mandatory-workflow step 2 — HARD STOP**: before any `POST /generate`, list sensors:
  ```bash
  curl -s \"http://\${HOST_IP}:30888/vst/api/v1/sensor/list\" | jq '.[].name'
  ```
  - Named source IN the list → proceed.
  - Named source NOT in the list → STOP, do NOT fire \`/generate\` as a probe, ask the user.
  - Generic query naming no source → still require \`sensor/list\` non-empty.

  Why-it's-a-hard-rule paragraph explains: \`/generate\` will happily accept any natural-language \`input_message\` and reply \"no sensors / no results\" regardless. Listing sensors first is cheap and avoids the second-pivot recovery.

- **Renumbered** workflow steps 3-6 (was 2-5).

- **Rewrote** the contradicting Gotcha to align with the new precondition: the skill still assumes sources are already ingested in VIOS — but step 2 is the gate that proves it, and the agent must not search local disk or trigger ingest on the user's behalf.

## Test plan

- [x] Re-run the Skills Eval on \`vss-search-archive\` against this PR. Step 1 (\"airport video\" — no such source) should now hit reward 1.0 instead of 0.75.
- [x] Steps 2-4 (which name real sources) should be unaffected — the \`sensor/list\` check is cheap and just adds one curl call to the head of the trajectory.

### DIFF
--- a/skills/vss-search-archive/SKILL.md
+++ b/skills/vss-search-archive/SKILL.md
@@ -86,29 +86,17 @@ When using this skill, ALWAYS follow this high-level workflow:
    ```
 
    Then:
-   - **If the named source (or a clearly substring-matching name) IS in the list** → record the exact registered name (call it `<sensor_id>`) and proceed to step 3.
+   - **If the named source (or a clearly substring-matching name) IS in the list** → proceed to step 3. Forward the user's natural-language query verbatim — the agent's own search tool decomposer (`services/agent/src/vss_agents/tools/search.py`) extracts `video_sources` from the prose given the available sources, so the skill does NOT need to construct a structured `video sources` payload.
    - **If the named source is NOT in the list** → STOP. Do NOT fire `/generate` as a probe. Respond to the user with the registered source names and ask whether they meant one of those, want to ingest the missing video, or want to abandon the query. Wait for clarification.
    - **If the query names no specific source** ("find forklifts in the ingested videos", "search across all sources") → skip the substring check, but `sensor/list` must still return non-empty (otherwise no sources are ingested → HARD STOP).
-3. **Scope the `/generate` request to the resolved source.** When step 2 produced a specific `<sensor_id>`, the `/generate` payload MUST pass it via the `video sources` control knob — do NOT rely on the natural-language `input_message` alone to constrain the search:
-
-   ```bash
-   curl -s -X POST http://${HOST_IP}:8000/generate \
-     -H "Content-Type: application/json" \
-     -d '{
-       "input_message": "find all instances of forklifts",
-       "video sources": ["<sensor_id>"]
-     }' | jq .
-   ```
-
-   `/generate` does not reliably infer source scope from `input_message` prose ("find forklifts **in warehouse_cam_3**") — the `video sources` array is the authoritative filter. Omit it only for the no-specific-source branch from step 2.
-4. Run the search(es) via approach chosen
-5. Present the results to the user query. Format response as a professional inspection report but name it `Video Search Results`:
+3. Run the search(es) via approach chosen
+4. Present the results to the user query. Format response as a professional inspection report but name it `Video Search Results`:
    — Use clear section headers
    - Organize findings individually with supporting detail, and close with a summary
    - Use tables where comparisons help. Write like a technical report, not a chat message.
-6. CRITICAL: Verify the results and explain this to the user concisely.
+5. CRITICAL: Verify the results and explain this to the user concisely.
    If search fails, or returns unexpected results (i.e. videos that do not appear to match user query, zero matches, zero videos returned, error etc.), STOP. Do not proceed without reading [troubleshooting.md](references/troubleshooting.md) to iterate with feedback loops until proper results are found and presented like a professional inspection report.
-7. Final verifications:
+6. Final verifications:
    - ALWAYS inform user that final and further verifications can be run. Present this as a `Verification Step`
    - ONLY IF user agrees, download screenshots using the `screenshot_url` of the best candidates (highest similarity scores) from the search hits (JSON results) to `/tmp`. Read them and verify if they correspond to the user query
 
--- a/skills/vss-search-archive/SKILL.md
+++ b/skills/vss-search-archive/SKILL.md
@@ -75,14 +75,36 @@ When using this skill, ALWAYS follow this high-level workflow:
    `POST http://.../generate` request until the user has supplied an
    endpoint. Respond to the user with a single question asking for
    `HOST_IP` / the VSS agent endpoint and wait.
-2. Run the search(es) via approach chosen
-3. Present the results to the user query. Format response as a professional inspection report but name it `Video Search Results`:
+2. **Resolve the source — HARD STOP before any `/generate` call.**
+   If the user query references a specific video / sensor name
+   (e.g. "the airport video", "warehouse_cam_3", "sample warehouse"),
+   verify it's actually registered in VIOS **before** firing
+   `POST .../generate`:
+
+   ```bash
+   curl -s "http://${HOST_IP}:30888/vst/api/v1/sensor/list" | jq '.[].name'
+   ```
+
+   Then:
+   - **If the named source (or a clearly substring-matching name) IS in the list** → proceed to step 3.
+   - **If the named source is NOT in the list** → STOP. Do NOT fire `/generate` as a probe. Respond to the user with the registered source names and ask whether they meant one of those, want to ingest the missing video, or want to abandon the query. Wait for clarification.
+   - **If the query names no specific source** ("find forklifts in the ingested videos", "search across all sources") → skip the substring check, but `sensor/list` must still return non-empty (otherwise no sources are ingested → HARD STOP).
+
+   Why this is a hard rule: `/generate` will happily accept any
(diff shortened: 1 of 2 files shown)


skills/vss-search-archive/SKILL.md
```diff
@@ -75,14 +75,36 @@ When using this skill, ALWAYS follow this high-level workflow:
...
+   - **If the named source is NOT in the list** → STOP. Do NOT fire `/generate` as a probe. Respond to the user with the registered source names and ask whether they meant one of those, want to ingest the missing video, or want to abandon the query. Wait for clarification.
+   - **If the query names no specific source** ("find forklifts in the ingested videos", "search across all sources") → skip the substring check, but `sensor/list` must still return non-empty (otherwise no sources are ingested → HARD STOP).
+
+   Why this is a hard rule: `/generate` will happily accept any
+   natural-language `input_message` and return a polite "no results
+   / no sensors ingested" message regardless of whether the source
+   exists. Firing first and recovering from that response forces a
+   second pivot and pollutes the trajectory — listing sensors up
+   front is cheap and lets the agent give the user a precise
+   answer the first time.
```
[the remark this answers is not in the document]
zac-wang-nv: Good catch — agreed. Codex flagged the same edge case independently, and the deployment-prerequisite probe (ports 8000/9200) doesn't include VIOS at 30888, so VIOS being unreachable will silently look like an empty sensor list. Will follow up with an explicit error branch ("if `curl` returns non-zero or non-JSON: stop and report VIOS unreachable, do not infer 'no sources ingested'") and likely add 30888 to the prerequisite probe so the failure is caught earlier.

skills/vss-search-archive/SKILL.md
```diff
@@ -75,14 +75,36 @@ When using this skill, ALWAYS follow this high-level workflow:
...
+   (e.g. "the airport video", "warehouse_cam_3", "sample warehouse"),
+   verify it's actually registered in VIOS **before** firing
+   `POST .../generate`:
+
+   ```bash
+   curl -s "http://${HOST_IP}:30888/vst/api/v1/sensor/list" | jq '.[].name'
+   ```
+
+   Then:
+   - **If the named source (or a clearly substring-matching name) IS in the list** → proceed to step 3.
```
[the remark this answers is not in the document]
zac-wang-nv: Agreed — codex flagged the same ambiguity. The suggestion (exact case-insensitive match → unique substring → otherwise ambiguous, ask) is the right shape. Will fold this in along with the failure-handling fix from Thread 1, since both tighten step 2's decision boundaries.

skills/vss-search-archive/SKILL.md
```diff
@@ -75,14 +75,36 @@ When using this skill, ALWAYS follow this high-level workflow:
...
+   - **If the named source is NOT in the list** → STOP. Do NOT fire `/generate` as a probe. Respond to the user with the registered source names and ask whether they meant one of those, want to ingest the missing video, or want to abandon the query. Wait for clarification.
```
zac-wang-nv, line 90: Internal contradiction with the rewritten Gotcha at line 122: that Gotcha says "It is NOT the skill's job to ingest, find local files, or upload... don't trigger an ingest handshake on their behalf." But the menu offered here invites the user to answer "yes, ingest it" — at which point the skill has no authorized next step. Suggest dropping "want to ingest the missing video" from the menu and keeping just (a) "did you mean one of these names?" and (b) "abandon the query". If ingest is a legitimate follow-on, route the user to a separate ingest skill explicitly rather than implying this skill will do it.

skills/vss-search-archive/SKILL.md
```diff
@@ -75,14 +75,36 @@ When using this skill, ALWAYS follow this high-level workflow:
    `POST http://.../generate` request until the user has supplied an
    endpoint. Respond to the user with a single question asking for
    `HOST_IP` / the VSS agent endpoint and wait.
-2. Run the search(es) via approach chosen
-3. Present the results to the user query. Format response as a professional inspection report but name it `Video Search Results`:
+2. **Resolve the source — HARD STOP before any `/generate` call.**
+   If the user query references a specific video / sensor name
```
zac-wang-nv, line 79: The example "the airport video" is the same shape as visual/scene queries already in the "When to Use" section ("people near the loading dock", "vehicles in the parking lot"). An LLM reading this could classify any noun-phrase scene reference as a source-name constraint and hard-stop on queries that were never about a specific sensor. Suggest making the trigger more mechanical: e.g. "if the user explicitly names a sensor (alphanumeric_with_underscores, or quoted), or uses the phrase 'the X video' / 'the X stream' / 'the X feed'" — and explicitly excluding general scene words ("loading dock", "parking lot", "entrance") as content terms, not source names. Reinforcing with one example of each side would also help.

skills/vss-search-archive/SKILL.md
zac-wang-nv: Medium confidence — verify. Step 2 only verifies the source *exists* in VIOS; it doesn't ensure `/generate` will actually scope its search to that source. The default examples (lines 135–159) pass natural-language `input_message` with no `video sources` filter — the `Advanced control knobs` table lists `video sources` as a string array but examples don't use it. Net effect: the agent confirms `warehouse_cam_3` exists, then sends "find forklifts in warehouse_cam_3" and relies on `/generate` to infer the scope from prose, which may or may not actually filter. If the eval check is "did the agent search the right source?", that gate isn't really enforced by this PR. Worth confirming whether `/generate` does NL-based source resolution or whether step 2 should also instruct the agent to pass `video sources: ["warehouse_cam_3"]`.

zac-wang-nv:
## Review notes (general)

A few items that don't fit inline because they're outside this PR's diff hunks or span multiple sections.

### 1. "Search via REST API" examples below the new workflow contradict the hard-stop (`SKILL.md` ~L135–159)

The default examples in `## Search via REST API` show direct `POST /generate` with no preflight `sensor/list` — exactly the behavior the hard-stop is meant to prevent. LLMs commonly copy the nearest executable example. Suggest either (a) prefixing every example block with the `sensor/list` curl, or (b) adding a one-line callback above the section:

> "Examples below assume step 2's `sensor/list` check has already passed — do NOT call `/generate` directly without it."

### 2. Step 2 phrasing reads as conditional but isn't

The opening line of step 2 ("If the user query references a specific video / sensor name…") suggests the whole step is skippable when no source is named, but case 3 ("If the query names no specific source… `sensor/list` must still return non-empty") shows step 2 is in fact always run. Consider rephrasing the lead so it's clear `sensor/list` is unconditional and only the *decision tree* branches on whether a source was named.

### 3. Prerequisite probe gap

The probe at `SKILL.md:31–35` hits ports 8000 (agent) and 9200 (Elasticsearch) but not 30888 (VIOS). Step 2 now depends on VIOS being reachable. Adding a third `curl -sf --max-time 5 "http://${HOST_IP}:30888/vst/api/v1/sensor/list"` check to the prerequisite block would catch VIOS-down at probe time instead of at step 2 (and dovetails with the curl-failure handling from Greptile Thread 1).

PR #583, opened 2026-05-20
chore: remove Arize Phoenix (telemetry / observability) from VSS

zac-wang-nv:
## Summary

Removes Arize Phoenix end-to-end from VSS. Phoenix was wired in as an optional OpenTelemetry tracing exporter for the agent (`arizephoenix/phoenix:14.15.0`, default port 6006), but it isn't part of the supported product surface and adds an extra container + dependency tree we don't want to maintain. Verified zero Python imports of `arize`, `phoenix`, or `openinference` under `services/agent/src/` and `tests/` — nothing depends on Phoenix at code-level, only at deploy/config level.

After this PR, `grep -rni phoenix` returns zero matches across the repo (except the transitive `openinference-semantic-conventions` dep pulled in by `nvidia-nat-core`, which is unrelated to Phoenix and stays).

## What this changes

Layered into five reviewable commits:

1. **`chore(agent-configs)`** — drop the `telemetry.tracing.phoenix` block from 13 vss-agent `config.yml` / `config_rag.yml` files (docker + helm sides; all 5 profiles). NAT tolerates a missing `telemetry:` block — tracing is simply disabled when no exporter is configured.

2. **`chore(docker)`** — delete the `phoenix` container service and `phoenix-data` volume from `services/infra/compose.yml`; drop the `PHOENIX_PORT` env passthrough in `services/infra/haproxy/compose.yml` + the `bk_phoenix_strip` backend and `/phoenix` ACL in `haproxy.cfg.template`; drop the `PHOENIX_ENDPOINT` passthrough in `services/agent/compose.yml`; delete the `PHOENIX TELEMETRY CONFIGURATION` section from 5 profile `.env` files; drop Phoenix mentions from the Brev deploy notebook.

3. **`chore(agent-deps)`** — drop the `phoenix` extra from `nvidia-nat[…]` in `services/agent/pyproject.toml` (keep `opentelemetry` so future exporters can be wired in). Regenerated `services/agent/uv.lock` with **uv 0.6.17** (revision = 2, same as PR 558 / matches CI-pinned uv 0.6.2's read support). Validated with `uv 0.6.2 uv lock --check` — resolves 395 packages cleanly. Packages removed: `arize-phoenix-otel`, `nvidia-nat-phoenix`, `openinference-instrumentation`.

4. **`chore(helm)`** — delete the entire `services/infra/charts/phoenix/` subchart; drop the `phoenix` dependency from `services/infra/Chart.yaml`/`Chart.lock` and the `phoenix:` defaults block in `services/infra/values.yaml`; for each developer profile (`alerts`/`base`/`lvs`/`search`): remove the top-level `phoenix:` config block, `vssIngress.phoenixHost`/`phoenixPort`, the `PHOENIX_ENDPOINT` env in `agent.vss-agent.env`, the `/phoenix` path rule + host rule in the ingress templates, the printed Phoenix URLs in `NOTES.txt`, and the docs table rows in `README.md`. Drops the `phoenixEndpoint` default and `PHOENIX_ENDPOINT` env entry from `services/agent/charts/agent/values.yaml`.

5. **`chore(docs)`** — drop dangling Phoenix references from `services/agent/README.md` (env table row + .env example), 7 `skills/vss-deploy-profile/references/*.md` files (service tables, URL tables, HAProxy route rows, bp_wh extras prose), and 3 `skills/vss-deploy-profile/eval/*.json` files (the `docker ps … grep -qx phoenix` evaluator checks).

## Test plan

- [x] CI runs green (`Check helm parity`, `Lint`, `Type Check`, `Test`, `License Check`)
- [x] `Trigger Downstream Pipeline` deploy tests verify all profiles come up without the `phoenix` container
- [x] OSRB lockfile review — the 3 phoenix-related packages should appear as REMOVALS in the License Diff
- [x] Manual: `helm dependency build deploy/helm/services/infra/` to regenerate `Chart.lock` digest before any helm install (digest was retained as-is; regen happens on first `helm install`)

## Notes

- The OpenTelemetry tracing core stays available via NAT's `opentelemetry` extra. Any future exporter (OTLP, Jaeger, etc.) can be wired in by adding a new entry to the agent's `telemetry.tracing` config without touching dependencies.
- `Chart.lock` digest in `services/infra/` is now stale relative to the updated dependency list — running `helm dependency build deploy/helm/services/infra/` regenerates it cleanly. No-op for users who already `helm install` the umbrella charts (which re-resolve deps).
- The `vss-agent` config.yml had a commented-out `weave` exporter block right next to phoenix — that was also removed since it was a child of the same `telemetry.tracing` block. If Weave is wanted later, it can be re-added cleanly without phoenix in the way.


zac-wang-nv:
Fixed up uv.lock to be a **surgical** removal instead of a full regeneration.

Previous commit `c4b37574` (now replaced by `5b451d27`) used `uv 0.6.17 uv lock` which re-resolved every transitive dep to the latest available version, producing 208 changed packages even though only 3 were really phoenix-related.

The new commit `5b451d27` keeps every other package's version, source URL, and hash bit-identical to develop and only removes:
- `arize-phoenix-otel`
- `nvidia-nat-phoenix`
- `openinference-instrumentation`
- The `phoenix = [{ name = "nvidia-nat-phoenix" }]` extra block under nvidia-nat
- `phoenix` from the extras list in the manifest section's two nvidia-nat references

`git diff develop..HEAD -- services/agent/uv.lock` is now 55 lines (three deleted `[[package]]` blocks + the extras-list tweaks). Validated with CI-pinned `uv 0.6.2 uv lock --check` — passes cleanly (`Resolved 378 packages`).

License Diff should now show the expected 3 removals (and only those).

zac-wang-nv:
closing it as not needed for OSRB

PR #588, opened 2026-05-21
Fix agent catalog deploy troubleshooting

maxervo:
## Description
Addresses litmus agentic review feedback:

- Agentic Q&A | Fair | Top-level README plus per-skill SKILL.md answer most questions, but no root-level orienting pointer means an agent can route incorrectly across the 14 skills x 6 profiles.

- Agentic Error Handling & Recovery | Fair | Per-skill troubleshooting docs are detailed, but cross-cutting agent-facing failure modes (NIM cold-start, edge OOM, remote 5xx, key misconfiguration) are not consolidated.

## Checklist


zac-wang-nv:
can you target develop branch?

AGENTS.md
zac-wang-nv, line 1: this seems adding to the root, can we remove it at this moment? we can consider it in future.

PR #615, opened 2026-05-21
skill-eval: pool hygiene — reset deploy state between runs + drop auto-provision

zac-wang-nv:
Two related cleanups to the `vss-eval-*` pool model. Independent file-wise (AGENTS.md + brev_env.py + adapter comment + README), bundled in one PR.

## Change 1 — Reset deployment state between CI runs (was #615)

The per-box marker `/tmp/skill-eval/active-deploy.txt` was treated as per-pool-instance state, so a deploy from one PR persisted across runs and any later PR with the same profile hot-skipped redeploy — inheriting whatever stale image set the prior PR left running. `vss-eval-rtx-1g` was reporting VSS `3.0.0` from `/v1/metadata` while the agent expected `3.2.0`, breaking `vss-search-archive`.

Make the marker per-CI-run: at end-of-run (§ 7), before releasing each box's flock, wipe live containers and the marker while preserving slow caches (image layers, named volumes, repo clone, sample-data). Next run sees an empty marker and redeploys from its own `PR_HEAD_SHA`. Warm reuse within a run is unchanged.

`AGENTS.md` updates:
- § 7 (lock release): now requires `brev exec` to `docker rm -f $(docker ps -aq); docker network prune -f; rm -f active-deploy.txt` on every locked box before flock release.
- Pool-lifecycle hard rule: explicitly permits `brev exec` for the § 7 cleanup.
- Marker file description: notes the per-CI-run lifetime.

No code changes — `brev_env.py::_ensure_prerequisite_deployed` already handles empty/missing marker correctly (reconciles via `/vss-deploy-profile -p <profile>`).

## Change 2 — Remove `BrevEnvironment` auto-provisioning (was #616)

`BrevEnvironment.start()`'s Mode 2 fallback would `brev create` a fresh `harbor-<uuid8>` whenever no `BREV_INSTANCE` was resolved. The pool was supposed to make this dead code, but:
- `uvx harbor` invoked outside the CI agent (or any code path forgetting to export `BREV_INSTANCE`) silently spun up paid GPUs.
- The wrapper-side cleanup that used to delete these on a 5-min cooldown was removed when the pool became long-running, so leaks were silent.

Three orphan `harbor-*` had accumulated (2× `g7e.2xlarge`, 1× `massedcompute_L40S` — ~\$9/hr combined). Deleted out-of-band before this PR.

Make the absence of a resolved instance a hard error pointing at the pool. Drop the now-unreachable helpers: `_record_started_instance`, `_find_cheapest_matching_type`, `_suggest_registered_devices`, `_wait_for_running`. Also drop the stale "wrapper deletes them after 5-min cooldown" claim from `README.md`, and fix a dangling `_find_cheapest_matching_type` reference in `adapters/vss-deploy-profile/generate.py`.

Behaviour for trials with `BREV_INSTANCE` set (i.e. every CI run) is unchanged.

## Test plan

- [x] `vss-eval-rtx-1g` hard-reset out-of-band; first post-merge run on it starts cold.
- [x] Re-run `vss-search-archive` skill-eval on a PR — confirm the trial does a full `/vss-deploy-profile -p lvs` against the PR's HEAD SHA and `/v1/metadata` returns `3.2.0`.
- [x] Run two consecutive PRs that target the same platform back-to-back — confirm the second redeploys instead of hot-skipping.
- [x] Run a multi-spec PR — confirm warm reuse is still in effect across the specs (only one deploy per box per run).
- [x] Manually run `uvx harbor run` without `BREV_INSTANCE` — confirm it errors at startup with the new pointer message instead of `brev create`-ing.

### DIFF
--- a/.github/skill-eval/AGENTS.md
+++ b/.github/skill-eval/AGENTS.md
@@ -347,13 +347,32 @@ template is in § Harbor invocation below.
    --body-file …`. Do NOT post a planning / "refresh" comment up
    front — comments carry results, not intent.
 
-7. **Release all locks. DO NOT tear down any Brev instance.** The
-   `vss-eval-*` boxes are a long-running pool managed by the operator;
-   they stay up across runs (warm caches, pre-deployed VSS profiles,
-   docker layer reuse). You release the per-box flock so the next
-   worker can grab it; you never `brev stop` / `brev delete`. The
-   wrapper script no longer runs cleanup either — pool lifecycle is
-   strictly an operator concern.
+7. **Reset each locked box, then release its lock. DO NOT tear down
+   any Brev instance.** The `vss-eval-*` boxes are a long-running
+   pool managed by the operator; instances stay up across runs, and
+   so do the slow caches (docker image layers, named volumes —
+   postgres / ES / kafka data, repo clone, sample-data extract). But
+   the *deployment state* — running containers plus the
+   `/tmp/skill-eval/active-deploy.txt` marker — is per-CI-run: warm
+   reuse within a run amortises deploy across `(spec, task, trial)`
+   tuples on the same box, but between runs each PR must redeploy
+   from its own `PR_HEAD_SHA` so it doesn't inherit a stale stack.
+   So for each `vss-eval-*` box you held a flock on, BEFORE releasing
+   that flock, reset its deployment state:
+
+   ```bash
+   brev exec "$INSTANCE_NAME" 'docker rm -f $(docker ps -aq) 2>/dev/null; \
+                               docker network prune -f; \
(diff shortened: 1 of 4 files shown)


zac-wang-nv:
Reopening: branch rebased on current develop. Conflicts with merged #652 (`upload_dir()` refactor) were verified zero — that PR touched a different region of brev_env.py. All four commits replayed clean; verified the four auto-provision helpers are gone and `#652`'s new tempfile+`brev copy` path is preserved.

zac-wang-nv:
Superseded by #695 — same four commits rebased on current develop. (GitHub refused `reopen` after the force-push.)

PR #616, opened 2026-05-21
skill-eval: remove BrevEnvironment auto-provisioning

zac-wang-nv:
## Summary

Remove the Mode 2 auto-provision fallback in `BrevEnvironment.start()`. Trials must now run on an operator-managed `vss-eval-*` pool member (resolved via `BREV_INSTANCE` env var or `task.toml [metadata].brev_instance`); if neither is set, the harness fails loud instead of silently `brev create`-ing a fresh `harbor-<uuid8>`.

## Why

The auto-provision path was designed as a fallback for the pre-pool world, but in the long-running pool model it had three failure modes:

1. **Silent paid leaks.** `uvx harbor` invoked outside the CI agent (or in any code path that forgets to `export BREV_INSTANCE`) would spin up a fresh GPU and leave it running. The wrapper-side cleanup that used to delete these on a 5-min cooldown was removed when the pool went long-lived, so the `/tmp/brev/started-by-<run_id>.txt` markers are still written but never read.
2. **Concrete cost today.** Three orphan `harbor-*` were running on this org (2× `g7e.2xlarge`, 1× `massedcompute_L40S`) — ~\$9/hr combined. Deleted out-of-band before this PR.
3. **Hides operator intent.** AGENTS.md is explicit that the pool is operator-managed and the agent must select+export `BREV_INSTANCE`; the auto-provision fallback gave the agent (and any other caller) a way to bypass that without an error.

## What changed

`envs/brev_env.py`:
- `BrevEnvironment.start()`'s `else:` branch is now a `RuntimeError` pointing the caller at the pool model.
- Removed the now-unreachable helpers: `_record_started_instance`, `_find_cheapest_matching_type`, `_suggest_registered_devices`, `_wait_for_running`.
- Docstring + `_resolve_instance_name` doc + post-validation comment updated to reflect "validate, don't provision."

`adapters/vss-deploy-profile/generate.py`:
- Comment dropped its dangling reference to `_find_cheapest_matching_type`.

`README.md`:
- Step 7 of the run lifecycle (about `started-by-<run_id>.txt` + wrapper cleanup) removed.
- Troubleshooting entry "Cancelled run leaves orphan Brev instances" rewritten to reflect the post-removal world.

## Test plan

- [ ] Run a normal CI eval (e.g. trigger skills-eval on an open PR). Confirm the agent sets `BREV_INSTANCE` and the trials run as before — no `harbor-*` instances appear in `brev ls`.
- [ ] Manually run `uvx harbor run` against a fresh dataset *without* `BREV_INSTANCE` set. Confirm it errors at startup with the new pointer message instead of `brev create`-ing.
- [ ] Confirm `mypy` / `ruff` still pass on `envs/brev_env.py` (no dangling references to removed names).

### DIFF
--- a/.github/skill-eval/README.md
+++ b/.github/skill-eval/README.md
@@ -6,11 +6,10 @@ Evaluation is **fully CI-driven**. [`.github/workflows/skills-eval.yml`](../work
 
 1. Diffs the PR against its base branch and picks out changed skills with an eval spec at `skills/<skill>/evals/<name>.json` or legacy `skills/<skill>/eval/<name>.json`.
 2. Generates Harbor datasets per `(skill, profile, platform, mode)` via the adapter at [`adapters/<skill>/generate.py`](adapters/).
-3. Acquires a per-instance `flock` on a Brev GPU host, reusing one that matches the target platform or creating one via the fallback chain in [`AGENTS.md`](AGENTS.md).
+3. Acquires a per-instance `flock` on an operator-managed `vss-eval-*` pool member matching the target platform, per the fleet-selection algorithm in [`AGENTS.md`](AGENTS.md) § 5a. The harness does **not** auto-provision — if no pool member matches, the run blocks until one appears (or times out).
 4. Runs `uvx harbor run` against each dataset, one trial at a time, with the canonical invocation captured in [`AGENTS.md § Harbor invocation`](AGENTS.md).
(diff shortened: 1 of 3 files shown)


zac-wang-nv:
Consolidated into #615 — both changes ship together as 'skill-eval: pool hygiene'.

PR #630, opened 2026-05-22
Fix VST replay thumbnails on Brev

freshyjmp:
## Summary
- inject the replay stream ID as a header in VST ingress for snapshot endpoints
- cover both /picture and /picture/url routes in vst and mms nginx configs
- fixes browser thumbnail requests that cannot set the required streamId header

## Validation
- git diff --check
- static inspection of modified VST nginx routes

NVBug: 6190639


freshyjmp:
Closing this PR because it was created from the wrong base branch and includes main-vs-develop noise. Replacement clean develop-based PR: #631.

PR #633, opened 2026-05-22
fix(agent ui): Alerts Tab: Modify sensor input logic for Alert rule creation to match with the logic via Agentic Chat

ashan-nv:
## Description

Fix for https://nvbugspro.nvidia.com/bug/6200965

 ####  The bug

  Creating an alert rule from the UI worked when you pasted a URL like rtsp://host/sample.mp4, but failed for a real live stream URL like rtsp://host/live/<uuid>. The chat could create alerts for either.

   ####  Why one worked and the other didn't

  The old UI tried to figure out the sensor name from the URL by grabbing the last path segment. Then it looked that name up in VST's sensor list to get the sensor's ID.

  - For rtsp://host/sample.mp4, the last segment is sample.mp4. NVStreamer happens to register file-backed streams under their filename, so VST's sensor list had an entry named sample.mp4. Match — worked by coincidence.
  - For rtsp://host/live/<uuid>, the last segment is the UUID. VST's sensor list is keyed by the friendly name (e.g. warehouse-cam-1), not by the UUID. No match — failed.

  The chat avoided this entirely: you give it the friendly name, and it asks VST /v1/live/streams for the URL.

 #### What we did

  Made the UI work like the chat. Instead of pasting a URL, you pick a sensor from a dropdown populated by VST's live-stream catalog. The URL comes straight from VST. No more parsing.

  Two follow-on tweaks:
  - Allowed typing a custom name (for streams not yet registered).
  - Dropped sensor_id from the request payload to exactly match what chat sends, since including it pushed Alert Bridge down a code path that produced an "orphaned rule" warning.

https://github.com/user-attachments/assets/33efb7c7-166c-4bfe-a180-3f57623d729d

## Checklist
- [ X] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/HEAD/CONTRIBUTING.md).
- [ X] New or existing tests cover these changes.
- [ X] The documentation is up to date with these changes.

### DIFF
--- a/deploy/docker/services/ui/compose.yml
+++ b/deploy/docker/services/ui/compose.yml
@@ -16,7 +16,7 @@
 services:
 
   vss-ui:
-    image: nvcr.io/nvstaging/vss-core/vss-agent-ui:3.2.0-26.05.2-3ab18ca37ffb
+    image: nvcr.io/nvstaging/vss-core/vss-agent-ui:3.2.0-26.05.2-0ab9d94a7fb9
     profiles: ["bp_wh_2d","bp_developer_base_2d","bp_developer_search_2d","bp_developer_lvs_2d","bp_developer_alerts_2d_cv", "bp_developer_alerts_2d_vlm"]
     container_name: vss-agent-ui
     ports:
--- a/deploy/helm/developer-profiles/dev-profile-search/values.yaml
+++ b/deploy/helm/developer-profiles/dev-profile-search/values.yaml
@@ -562,7 +562,7 @@ vss-agent-ui:
     imagePullPolicy: IfNotPresent
   image:
     repository: nvcr.io/nvstaging/vss-core/vss-agent-ui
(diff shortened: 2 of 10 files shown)


akshayanv (commented):
@ashan-nv 

What happens - 
if on your screen your manage rules view is in focus, someone else goes to VST and uploads a new video from VST UI which has a new sensor name. Now, will you see the new sensor name in the dropdown list?

ashan-nv:
> @ashan-nv
> 
> What happens - if on your screen your manage rules view is in focus, someone else goes to VST and uploads a new video from VST UI which has a new sensor name. Now, will you see the new sensor name in the dropdown list?

so it gets fetched when you click the "Create alert rule" button but not if the dropdown is already open. It also caches the sensor list for 60 seconds so new sensors won't show until the next button click which is after 1 minute

akshayanv (approved):

PR #647, opened 2026-05-22
AGENT UI: [Search Tab]: Remove streamid from header to get picture

ghost:
Fixes VST service /picture API image load issue


akshayanv (commented):
Before this MR, double check with VST team that streamID needs to be removed. I think they used to suggest in general to keep it for all of their APIs. IF it is a bug on their side then we dont want to make this change.

akshayanv:
https://gitlab-master.nvidia.com/L4TMM/vms_shim/-/blob/v2.1/doc/api/vst_replay_stream_ms/swagger.yaml?ref_type=heads shows streamId in header for /picture API is must

akshayanv:
https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/tree/dev-26.05.3-4 was working even without this change

PR #690, opened 2026-05-25
feat(skills/vss-deploy-detection-tracking-2d): publish in catalog and expose FPS via /api/v1/metrics

vishkumar56:
## Description

Two small, related improvements to the `vss-deploy-detection-tracking-2d` skill:

**1. Add the skill to the public catalog**

`skills/README.md` lists every shipped skill in a Catalog table, but `vss-deploy-detection-tracking-2d` was never added. This PR adds it next to the other deploy skills with a one-liner description consistent with the table's style.

**2. Surface real per-stream FPS via `/api/v1/metrics`**

Today, when the user runs `collect_metrics.sh`, `/api/v1/metrics` returns `stream-count=0` for static-mode deploys and the script falls back to log-parsing FPS (see `collect_metrics.sh:248-261`). Root cause: the metrics API only reports FPS when an `nvdslogger` element is attached to an enabled sink, and none of the four shipped reference configs (`warehouse-2d`, `warehouse-3d`, `smartcity-rtdetr`, `smartcity-gdino`) have `nvdslogger=1` under `[sink0]` — warehouse-2d/3d have it only on `[sink1]` (kafka), smartcity-rtdetr has it commented out.

`scripts/update_output_sink.sh` is where the skill canonically writes the sink config on every deploy. Two changes there:

- **Always write `nvdslogger=1` to `[sink0]`** — for every use case and every sink mode (`fakesink`, `eglsink`, `filedump`). For filedump (`[sink0] enable=0`) the key is dormant — no behavior change for that mode. For fakesink/eglsink the metrics API now reports real per-stream FPS out-of-the-box; the log-parse fallback in `collect_metrics.sh` stays only as a safety net.
- **Switch `[tiled-display] enable` from `0` to `3` for fakesink** — DS tiler perf-only mode. The element stays in the pipeline but skips compositing, so the bench path stays lean while per-source perf samples still flow to `nvdslogger`. `eglsink`/`filedump` stay at `1` (need the composited buffer to render the grid / feed the file-write path).
- Extend the script's final verification with `_check "sink0.nvdslogger"` so a silent write failure now triggers `SINK_UPDATE_FAIL` instead of a regression.

**3. Docs in sync**

`references/apply-config.md` (edit-list table, per-use-case Sink row counts, worked-example ✔ box, and the existing "Note on `[tiled-display] enable`") and `scripts/update_batch_size.sh`'s tile-grid comment all updated to match the new policy.

### DIFF
--- a/skills/README.md
+++ b/skills/README.md
@@ -26,6 +26,7 @@ The VSS 3.2 GA skill names replaced the pre-GA slash-command names:
 | [vss-manage-alerts](vss-manage-alerts/SKILL.md) | Skill to add, manage, and monitor alerts on streamed video. |
 | [vss-deploy-profile](vss-deploy-profile/SKILL.md) | Skills to deploy, debug, or tear down any VSS profile using a docker compose-centric workflow. |
 | [vss-deploy-dense-captioning](vss-deploy-dense-captioning/SKILL.md) | Skill for deploying and calling the RT-VLM dense captioning microservice API. |
+| [vss-deploy-detection-tracking-2d](vss-deploy-detection-tracking-2d/SKILL.md) | Skill for deploying, operating, and calling the RTVI-CV perception microservice for 2D detection and tracking (warehouse 2d/3d, sparse4d, smartcity rtdetr/gdino). |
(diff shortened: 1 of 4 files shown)


zac-wang-nv (approved):
LGTM

PR #609, opened 2026-05-21
release: merge develop into dev/3.2.0 (3.2.0 GA cut)

zac-wang-nv:
## What

Release-train sync: **227 commits** from `develop` flowing into `dev/3.2.0` for the 3.2.0 GA cut. Reverse of the usual feature-PR direction — this is the periodic "pull latest develop into the release-staging branch" step, not a feature PR.

```
develop                                                            227 commits ahead
  │
  ├── (this PR — merge into dev/3.2.0)
  │
dev/3.2.0  (= release/3.2.0, both at 0a56b6f0)
```

## Theme breakdown of the 227 incoming commits

```
  43  fix                  — bug fixes across agent, UI, deploy, helm, skills
  18  feat                 — new skills, new features, new deploy profiles
  16  docs                 — README + reference doc updates
  12  ci                   — license-check, skill-eval, NV-BASE, container-source checks
  12  chore                — version bumps, dep upgrades
  10  skill-eval           — harness improvements
   6  refactor
   5  skill / skills       — skill-spec edits
   3  eval                 — eval-spec changes
   2  VSS Agent UI         — node version bump, alert UI fixes
   2  video-analytics-api  — container tag bumps
```

## Notable changes worth surfacing for QA

- **CI hardening** — permissive-license allowlist + denylist + override system ([#570](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/570) + [#587](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/587)), container-source-tag fix-on-registry-error ([#554](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/554)), license-diff CSV workflow ([#461](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/461) + [#476](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/476)), CODEOWNERS dep-manifest routing to OSRB ([#455](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/455)).
- **Launchables** — Docker CE 29.4.3 pin + plugin pins + `apt-mark hold` in both `deploy_vss_launchable.ipynb` and `deploy_nemoclaw_vss.ipynb` ([#603](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/603)). Closes recurring compose/buildx incompatibility on fresh Brev instances.
[...]

PR #694, opened 2026-05-25
skill-eval/verifier: retry once when judge omits verdict JSON

zac-wang-nv:
## Summary

When the agent-style judge in `verifiers/generic_judge.py` runs a check, gathers evidence, and then forgets to close with the required `{\"pass\": ..., \"matched\": ..., \"rationale\": ...}` JSON, send one follow-up turn in the same SDK session asking for the verdict JSON only.

## Why

Live failure on PR #609 run `26297115943`, spec `vss-generate-video-report/base_profile_report` step-1:

| Check | Result | Reason |
|---|---|---|
| 1 — Agent probed VST before uploading | ✅ pass | Trajectory step 274 hit `/vst/api/v1/sensor/list`, no PUT after |
| 2 — `/vst/api/v1/sensor/list` contains the uploaded video's stem | ❌ **fail** | **Judge ran the curl, saw `warehouse_safety_0001` in the response, narrated it — never emitted the verdict JSON** |
| 3 — `/vst/api/v1/sensor/<id>/streams` returns a local-file URL | ✅ pass | Same data; judge produced verdict |

Same trial, same evidence — only check 2 wandered. The `judge.json` for the failure said:

```
"rationale": "judge returned no compliant verdict (saw_result_message=True text_chars=322 text_blocks=2); head: \"I'll investigate this check by first probing the live endpoint and then examining the trajectory...\""
```

Reward dropped to 0.667, blocking steps 2-3 (chained on `requires_previous_passed`). The agent under test did everything right; the judge LLM (sonnet) just forgot the output format. This is a known fragility of agent-style rubric judging, not a spec bug — the same prompt produced a verdict for checks 1 and 3.

## What changed

`verifiers/generic_judge.py::_judge_llm_agent::_run`:

- After the first response loop, if `saw_result_message=True` AND collected text has no parseable verdict, send a second `client.query()` with a strict format-only nudge. Stream the second response into the same `collected_text` buffer.
- Uses the **same** `ClaudeSDKClient` session — prior tool results (Bash curl outputs, trajectory reads) stay in context, so the follow-up costs one short turn instead of a full re-investigation.
- `retry_attempted` is plumbed through to the failure rationale's `signals` block so we can distinguish:
  - `retry_attempted=False` → first response had no text at all or stream didn't end cleanly (different bug class)
  - `retry_attempted=True` → judge tried twice and still couldn't format → spec-prompt issue worth tightening

Cost accounting: changed `cost_usd = …` to `cost_usd += …` on each `ResultMessage`. Equivalent in the single-response case (prior behaviour); additive across the retry.

Scope is intentionally narrow: timeouts, crashes, and empty-text responses are not retried — they're symptoms of larger problems and re-prompting won't help. One retry only, no loop.

## Test plan

- [ ] Re-run `vss-generate-video-report` skill-eval (the spec that exposed this). Confirm check 2 now passes via the retry path (judge.json shows `pass: true` and `signals.retry_attempted=True`).
- [ ] Spot-check a normal run: confirm a check that emits the verdict on the first turn doesn't trigger retry (no extra `ResultMessage`, `retry_attempted=False`).
- [ ] Confirm timeout / crash paths still surface the original error rationale (retry block is conditional on `saw_result` being true).

### DIFF
--- a/.github/skill-eval/verifiers/generic_judge.py
+++ b/.github/skill-eval/verifiers/generic_judge.py
@@ -286,10 +286,14 @@ async def _judge_llm_agent(check: str, traj_path: str | None, *, timeout_s: int)
     collected_text: list[str] = []
     cost_usd = 0.0
     saw_result = False
+    result_is_error = False
+    result_subtype: str | None = None
+    result_stop_reason: str | None = None
     retry_attempted = False
 
     async def _run() -> None:
         nonlocal cost_usd, saw_result, retry_attempted
+        nonlocal result_is_error, result_subtype, result_stop_reason
         async with ClaudeSDKClient(options=options) as client:
             await client.query(_assemble_judge_prompt(check, traj_path))
             async for message in client.receive_response():
@@ -298,8 +302,16 @@ async def _judge_llm_agent(check: str, traj_path: str | None, *, timeout_s: int)
                         if isinstance(block, TextBlock) and block.text:
                             collected_text.append(block.text)
                 elif isinstance(message, ResultMessage):
-                    cost_usd += getattr(message, "total_cost_usd", 0.0) or 0.0
+                    # `total_cost_usd` on ResultMessage is cumulative for
+                    # the SDK session (same field that carries cumulative
+                    # `num_turns`), so re-assign on every ResultMessage
+                    # rather than accumulating — the last value is the
+                    # whole-session total.
+                    cost_usd = getattr(message, "total_cost_usd", 0.0) or 0.0
                     saw_result = True
+                    result_is_error = bool(getattr(message, "is_error", False))
+                    result_subtype = getattr(message, "subtype", None)
+                    result_stop_reason = getattr(message, "stop_reason", None)
                     break
 
             # Retry once if the first response ended cleanly but didn't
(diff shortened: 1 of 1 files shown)


zac-wang-nv:
Pushed 245bd206 addressing the review:

1. **Retry guard now gates on `not result_is_error`** (the main P1 concern). When the first response ended in an SDK-level error (`is_error=True` — rate limit, content policy, tool-use abort, max-turns surfaced via the error flag), retry is suppressed. Avoids burying the real failure cause under a doomed second-pass rationale.

2. **Cost double-count fixed** (greptile-bot, line 302 inline). Verified against the SDK source — `claude_agent_sdk/_internal/message_parser.py:222-244` parses `total_cost_usd` alongside the cumulative `num_turns` field, so it *is* session-cumulative. Both ResultMessage handlers now assign (`cost_usd = ...`) instead of `+=`. Equivalent in the no-retry case (single ResultMessage), correct in the retry case.

3. **SDK termination reason surfaced in failure rationale.** The `signals` block now carries `is_error`, `subtype`, and `stop_reason` so a triager can immediately distinguish "judge wandered into prose" from "SDK errored out".

Net: +35/−5 on top of the original 50/3.

PR #652, opened 2026-05-22
fix: eval brev upload

maxervo:
## Description
Skill eval setup can fail before the agent runs because BrevEnvironment.upload_dir() sends the entire tarred skills/ bundle as base64 embedded in a single brev exec command argument. Larger bundles, like the search skill eval bundle, can exceed the OS argv limit and raise OSError: [Errno 7] Argument list too long: 'brev'.

```
  [tool] Bash :: # That was the FIRST failed attempt at step-1 (the OSError from skills upload).
```
https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/26293619017/job/77400176534

This PR changes the upload path to copy a tarball with brev copy and extract it remotely with a short brev exec command.

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #695, opened 2026-05-25
skill-eval: pool hygiene — reset deploy state between runs + drop auto-provision

zac-wang-nv:
Reopen of #615 (closed; GitHub refused `reopen` because the branch was rebased on current develop). Same four commits, replayed on top of develop's current tip — verified no overlap with the meanwhile-merged #652 (`upload_dir()` refactor in `brev_env.py` — different region).

Two related cleanups to the `vss-eval-*` pool model. Independent file-wise (AGENTS.md + brev_env.py + adapter comment + README), bundled in one PR.

## Change 1 — Reset deployment state between CI runs

The per-box marker `/tmp/skill-eval/active-deploy.txt` was treated as per-pool-instance state, so a deploy from one PR persisted across runs and any later PR with the same profile hot-skipped redeploy — inheriting whatever stale image set the prior PR left running. `vss-eval-rtx-1g` was reporting VSS `3.0.0` from `/v1/metadata` while the agent expected `3.2.0`, breaking `vss-search-archive`.

Make the marker per-CI-run: at end-of-run (§ 7), before releasing each box's flock, wipe live containers, networks, named volumes, and the marker while preserving slow caches (image layers, repo clone, sample-data extract on the host fs). Next run sees an empty marker and redeploys from its own `PR_HEAD_SHA` against empty data stores. Warm reuse within a run is unchanged.

`AGENTS.md` updates:
- § 7 (lock release): now requires `docker rm -f $(docker ps -aq); docker network prune -f; docker volume prune -af; rm -f active-deploy.txt` on every locked box before flock release.
- Pool-lifecycle hard rule: explicitly permits `brev exec` for the § 7 cleanup.
- Marker file description, § 5c trial loop, canonical Harbor invocation comment, and fleet-selection paragraph: all aligned with the no-auto-provision reality (three stale "auto-provisions a fresh harbor-*" claims were rewritten as "raises at start()").

No code changes — `brev_env.py::_ensure_prerequisite_deployed` already handles empty/missing marker correctly.

## Change 2 — Remove `BrevEnvironment` auto-provisioning

`BrevEnvironment.start()`'s Mode 2 fallback would `brev create` a fresh `harbor-<uuid8>` whenever no `BREV_INSTANCE` was resolved. The pool was supposed to make this dead code, but:
- `uvx harbor` invoked outside the CI agent (or any code path forgetting to export `BREV_INSTANCE`) silently spun up paid GPUs.
- The wrapper-side cleanup that used to delete these on a 5-min cooldown was removed when the pool became long-running, so leaks were silent.

Three orphan `harbor-*` had accumulated (2× `g7e.2xlarge`, 1× `massedcompute_L40S` — ~\$9/hr combined). Deleted out-of-band before #615.

Make the absence of a resolved instance a hard error pointing at the pool. Drop the now-unreachable helpers: `_record_started_instance`, `_find_cheapest_matching_type`, `_suggest_registered_devices`, `_wait_for_running`. Also drop the stale "wrapper deletes them after 5-min cooldown" claim from `README.md`, and fix a dangling `_find_cheapest_matching_type` reference in `adapters/vss-deploy-profile/generate.py`.

Behaviour for trials with `BREV_INSTANCE` set (i.e. every CI run) is unchanged.

## Test plan

- [x] `vss-eval-rtx-1g` hard-reset out-of-band; first post-merge run on it starts cold.
- [ ] Re-run `vss-search-archive` skill-eval on a PR — confirm the trial does a full `/vss-deploy-profile -p lvs` against the PR's HEAD SHA and `/v1/metadata` returns `3.2.0`.
- [ ] Run two consecutive PRs that target the same platform back-to-back — confirm the second redeploys instead of hot-skipping.
- [ ] Run a multi-spec PR — confirm warm reuse is still in effect across the specs (only one deploy per box per run).
- [ ] Manually run `uvx harbor run` without `BREV_INSTANCE` — confirm it errors at startup with the new pointer message instead of `brev create`-ing.

### DIFF
--- a/.github/skill-eval/AGENTS.md
+++ b/.github/skill-eval/AGENTS.md
@@ -347,13 +347,32 @@ template is in § Harbor invocation below.
    --body-file …`. Do NOT post a planning / "refresh" comment up
    front — comments carry results, not intent.
 
-7. **Release all locks. DO NOT tear down any Brev instance.** The
-   `vss-eval-*` boxes are a long-running pool managed by the operator;
-   they stay up across runs (warm caches, pre-deployed VSS profiles,
-   docker layer reuse). You release the per-box flock so the next
-   worker can grab it; you never `brev stop` / `brev delete`. The
-   wrapper script no longer runs cleanup either — pool lifecycle is
-   strictly an operator concern.
+7. **Reset each locked box, then release its lock. DO NOT tear down
+   any Brev instance.** The `vss-eval-*` boxes are a long-running
+   pool managed by the operator; instances stay up across runs, and
+   so do the slow caches (docker image layers, named volumes —
+   postgres / ES / kafka data, repo clone, sample-data extract). But
+   the *deployment state* — running containers plus the
+   `/tmp/skill-eval/active-deploy.txt` marker — is per-CI-run: warm
+   reuse within a run amortises deploy across `(spec, task, trial)`
+   tuples on the same box, but between runs each PR must redeploy
+   from its own `PR_HEAD_SHA` so it doesn't inherit a stale stack.
+   So for each `vss-eval-*` box you held a flock on, BEFORE releasing
+   that flock, reset its deployment state:
+
+   ```bash
+   brev exec "$INSTANCE_NAME" 'docker rm -f $(docker ps -aq) 2>/dev/null; \
+                               docker network prune -f; \
+                               rm -f /tmp/skill-eval/active-deploy.txt'
+   ```
+
+   Run this once per locked box at the end of the run (after § 6
+   posts the final comment), regardless of trial outcome — `DONE`,
+   `BLOCKED`, or partial. The image cache, named volumes, repo
+   clone, and sample-data extract survive; only live containers and
+   the marker are dropped. Then close the lock FD (or `flock -u
+   $LFD`) so the next worker can grab the box. You never `brev stop`
+   / `brev delete`. Pool lifecycle is strictly an operator concern.
 
 8. **Exit.** Print a last line starting with `DONE:` summarizing
    outcomes (e.g. `DONE: 3/3 specs passed; 0 blockers`). If any spec
@@ -385,12 +404,13 @@ template is in § Harbor invocation below.
 - **Never touch pool-instance lifecycle.** No `brev create`,
   `brev start`, `brev stop`, `brev reset`, or `brev delete` against
   any `vss-eval-*` box. The pool is operator-managed; instances stay
-  running across runs. The agent only reads (`brev ls`, `brev exec
-  -- cat …`) and acquires the per-box flock. If no hardware-matching
-  pool member exists for the trial's platform, follow the wait-for-
-  pool path in § 5a (5-min `brev ls` poll, 28800s budget, then
-  `BLOCKED: pool exhausted for <platform>`) — provisioning is the
-  operator's job.
+  running across runs. The agent's `brev` surface is limited to
+  `brev ls`, `brev exec` (reads + the end-of-run deployment-state
+  reset documented in § 7), and acquiring/releasing the per-box
+  flock. If no hardware-matching pool member exists for the trial's
+  platform, follow the wait-for-pool path in § 5a (5-min `brev ls`
+  poll, 28800s budget, then `BLOCKED: pool exhausted for
+  <platform>`) — provisioning is the operator's job.
 - **Never dispatch code from non-mirror branches.** You only ever
   process `pull-request/<N>` SHAs; those are CPR-bot vetted. If you
   notice the PR head on github.com is ahead of the mirror, note it
@@ -436,8 +456,12 @@ runner-side **flock** on `/tmp/brev/<INSTANCE_NAME>.lock`,
 checked separately via `flock -n` in step 5a. The two together
 let the scoring pick a warm-and-free box first, then fall back
 to warm-but-busy (queue on `flock -w`) or cold-and-free (redeploy).
-See `specs/stale-marker.spec` for verifying the marker against
-the actual running containers.
+The marker is *per-CI-run state*, not per-pool-instance state:
+§ 7's end-of-run cleanup wipes it (along with all containers)
+before releasing the flock, so the next run on the same box
+starts with an empty marker and redeploys from its own
+`PR_HEAD_SHA`. See `specs/stale-marker.spec` for verifying the
+marker against the actual running containers.
 
 With fleet=1, selection collapses to a single candidate. With
 fleet>1, two concurrent workflow runs land on different boxes
--- a/.github/skill-eval/AGENTS.md
+++ b/.github/skill-eval/AGENTS.md
@@ -350,29 +350,33 @@ template is in § Harbor invocation below.
 7. **Reset each locked box, then release its lock. DO NOT tear down
    any Brev instance.** The `vss-eval-*` boxes are a long-running
    pool managed by the operator; instances stay up across runs, and
-   so do the slow caches (docker image layers, named volumes —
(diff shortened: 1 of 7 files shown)


zac-wang-nv:
Pushed 95d1e5c5 addressing the 9/10+ confidence review comments:

| # | File | Comment | Fix |
|---|---|---|---|
| 1 | `brev_env.py:5` | Module docstring still advertised "Two modes" incl. deleted auto-provision | Rewrote header to single-mode (validate-or-raise); dropped `brev_search` from consumed-fields list; pointer to AGENTS.md § 5a |
| 2 | `brev_env.py:128` | `brev_search` key in `requirements` dict had no remaining consumers | Removed. (Adapters still emit it into task.toml — separate follow-up.) |
| 3 | `AGENTS.md:367` | §7 cleanup `; ;` chain could wipe marker on partial prune failure | Rewrote as `xargs -r … && && &&` matching `brev_env.py::_ensure_prerequisite_deployed`; added prose on load-bearing `&&` + "leave box flagged on non-zero exit" semantics |
| 5 | `skills_eval_agent.py:226` | user_prompt told the agent to `stop/delete any Brev instance you brought online` — contradicted §7 hard rule | Replaced with "reset per § 7 → release flock. Never `brev stop` / `brev delete` any pool member." |
| 6 | `README.md:26` | GPU-targets table + "Instance creation fails" troubleshooting described the deleted auto-provision/teardown model | Replaced table with operator-managed `vss-eval-*` pool (platform → pool member mapping, instance types, pointer to § 7); rewrote troubleshooting as "Pool exhausted for `<platform>`" (28800s wait + BLOCKED path) |

Net: +41 / −31 across the four files.

**Skipped per the review-pass instruction (`fixe 9/10`):** Comment 4 (cancel/crash/timeout bypass §7, 8/10), Comment 7 (host-wide prune blast radius, 6/10), Comment 8 (Docker 23.0 `volume prune -a` compatibility, 5/10), and the three suppressed low-confidence findings. Happy to revisit any of those as a follow-up if you want — Comment 4 in particular is a real reliability gap once §7 becomes a contract the next run depends on, but it's a separate change (`atexit` handler in `skills_eval_agent.py` or an operator-side periodic sweep).

zac-wang-nv:
Pushed 3f00e178 addressing Comment 4 (cancel/crash/timeout bypass §7).

**Approach: option 1 from the review** — move the cleanup contract into the Python harness so every exit path that doesn't reach §7 still resets state. Two pieces:

1. **Touched-boxes ledger** (AGENTS.md §5b). Agent appends `$INSTANCE_NAME` to `/tmp/skill-eval/touched-boxes-${GITHUB_RUN_ID}.txt` right after each successful `flock -w`. Append, not overwrite — multiple locked boxes per run is normal.

2. **Python harness failsafe** (`skills_eval_agent.py`). New `_failsafe_reset_touched_boxes()` reads the ledger and runs the same `xargs -r docker rm -f && && && rm -f marker` chain §7 uses. Wired into:
   - `atexit` → normal exits, exceptions, **max-turns** (rc=3)
   - `signal.SIGTERM` → **cancel-in-progress** from GitHub's `concurrency` group
   - `signal.SIGINT` → local ^C

   Re-entry-guarded so atexit + signal-handler firing both is idempotent. Per-box failures log but don't block cleanup of the remaining boxes. The signal handler re-raises the default (`signal.SIG_DFL` + `os.kill`) after cleanup runs so the runner records the cancellation with exit code `128+signum`, not a misleading `sys.exit()`.

**Not covered: SIGKILL** (the workflow's 8h `timeout-minutes` terminator). The kernel offers no userspace signal for that — needs an operator-side periodic sweep, explicitly out of scope. AGENTS.md §7 now states this gap.

**AGENTS.md §7 contract is now:** agent is the primary path (runs §7 while results are fresh, can log per-box failures into the run summary); harness is the backstop for paths the agent never reaches. The agent must still run §7 itself on the happy path.

**Workflow YAML comment fixed.** The stale "released by the agent's trap" line at `skills-eval.yml:64-67` now correctly references the SIGTERM handler in the Python harness.

**Test plan** (cannot fully validate in CI itself — needs manual smoke):
- `kill -TERM <pid>` mid-trial → confirm `[failsafe]` stderr lines + `docker ps` empty + marker gone on the touched box.
- Fresh push on in-flight PR → confirm cancel-in-progress'd run's boxes come back clean.
- `AGENT_MAX_TURNS=10` on a real PR → confirm rc=3 + failsafe ran.

Net: +147 / −3 across three files.

zac-wang-nv:
Pushed fbc7d21e — the redesign you asked for. Replaced push-side end-of-run cleanup + Python failsafe with a **pull-side reconcile** on the next run's lock acquisition.

## Mechanism

The active-deploy marker is now tagged `<profile_tag>|<run_id>` (e.g. `lvs|26500001234`), where `<run_id>` is `$GITHUB_RUN_ID` (or `local-<pid>` outside CI). `BrevEnvironment._ensure_prerequisite_deployed` builds its desired marker the same way, so a marker from a prior run NEVER matches the current run's desired marker → next worker always reconciles (tear-down + redeploy from its own `PR_HEAD_SHA`).

| Exit path | Old design (push-side + failsafe) | New design (pull-side reconcile) |
|---|---|---|
| happy path | agent runs §7 reset | next worker reconciles on lock acquisition |
| `BLOCKED:` | agent runs §7 reset | next worker reconciles |
| cancel-in-progress (SIGTERM) | failsafe SIGTERM handler runs reset | next worker reconciles |
| max-turns (rc=3) | failsafe atexit runs reset | next worker reconciles |
| agent crash | failsafe atexit runs reset | next worker reconciles |
| **SIGKILL** (job timeout) | ❌ not covered, needed operator sweep | ✅ next worker reconciles |
| **host reboot** | ❌ not covered | ✅ next worker reconciles |

Within one run, multiple trials with the same profile still hot-skip (same profile + same run id → full match).

## Why this is cleaner

- **No "agent never reached §7" failure mode** — there's no §7 to reach.
- **No ledger / failsafe / atexit / signal handler / touched-boxes file** — entire class removed.
- **No conflict bug** between workers — there's no exit-side cleanup that can clobber another worker's box.
- **Covers SIGKILL and host reboot** for free — the reconcile is on the next run's start, no matter how the prior died.
- **Same warm-within-run reuse** — profile + run id match still triggers hot-skip.

## Diff

7 files changed, **+164 / −274** (net −110 LOC):

- `envs/brev_env.py` — marker format + reconcile, forward `GITHUB_RUN_ID` to box
- `adapters/vss-deploy-profile/generate.py` — test.sh writes tagged marker
- `skills_eval_agent.py` — drops failsafe (atexit, signal handlers, ledger)
- `AGENTS.md` — §5b dropped ledger append; §7 collapsed to "release locks"; marker file section and pool-lifecycle hard rule updated; stale claims about §7 cleanup removed
- `specs/stale-marker.spec` §1 — marker format updated
- `README.md` — pool hygiene description rewritten
- `workflows/skills-eval.yml` — `concurrency` comment updated

## Test plan

- Run two consecutive PRs on the same pool box. Confirm the second's `_ensure_prerequisite_deployed` log emits `prerequisite mismatch (active=… desired=…); reconciling` — i.e., it sees the prior run's marker and tears down before deploying.
- Run a multi-spec PR. Confirm second + later trials with the same profile log `prerequisite … already current … (run X); skipping reconcile` — warm reuse within the run is unchanged.
- Cancel a run mid-trial. Confirm the next run on the same box reconciles cleanly (no harness logs about "failsafe" since the failsafe is gone).
- Marker write on `vss-deploy-profile` trial pass should include `|$GITHUB_RUN_ID` — check by `cat /tmp/skill-eval/active-deploy.txt` on the box after a successful deploy trial.

PR #597, opened 2026-05-21
removed delete sensor for video file delete method , delete storage e…

abhishek-ak-nv:
## Description
removed delete sensor for video file delete method , delete storage endpoint clears both storage and sensor for file type

## Checklist
- [ ] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/HEAD/CONTRIBUTING.md).
- [ ] New or existing tests cover these changes.
- [ ] The documentation is up to date with these changes.

PR #719, opened 2026-05-26
fix(agent): restore VST sensor delete + retry truncated VLM responses

zac-wang-nv:
## Summary

Restore the `delete_vst_sensor` call removed by PR #597 in `DELETE /api/v1/videos/{video_id}`.

## Why

PR #597 ([diff](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/597/files)) dropped the sensor delete on the premise that `delete_vst_storage` already cleans up both storage and sensor for file-type assets. That premise is wrong:

- `delete_vst_storage` (utils.py:258) hits `GET /vst/api/v1/storage/timelines` then `DELETE /vst/api/v1/storage/...` — it only removes the stored video files.
- `delete_vst_sensor` (utils.py:227) hits `DELETE /vst/api/v1/sensor/{sensor_id}` — it removes the sensor registration (name, URL).

The shared utils docstring is explicit: `delete_vst_sensor` *"removes the sensor metadata (name, URL, etc.) but not the stored video files. Must be paired with delete_vst_storage to fully remove a video."*

After #597 merged, every video-file deletion via this endpoint leaves the VST sensor metadata behind. Concrete symptoms:

- `GET /vst/api/v1/sensor/streams` keeps returning entries for already-deleted videos.
- Re-uploading a video under the same filename fails because the sensor name is still registered.

## What changed

- Add `delete_vst_sensor` back to the imports.
- Call it once in `delete_video()` in the **same position as before #597** (between RTVI-CV removal and storage delete), with the same `TimeMeasure` / results-append / log format that #597 removed.
- Inline comment documents the docstring contract so future cleanup passes don't re-remove it.

Diff: +10 lines, single file (`services/agent/src/vss_agents/api/video_delete.py`).

## Test plan

- [ ] Upload a file-type video, then `DELETE /api/v1/videos/{video_id}` — confirm `GET /vst/api/v1/sensor/streams` no longer lists the deleted sensor.
- [ ] Re-upload a video under the same filename after delete — confirms the sensor slot is reusable.
- [ ] Existing pytest suite for `video_delete` passes (touched paths are exercised via mocked VST endpoints).

### DIFF
--- a/services/agent/src/vss_agents/api/video_delete.py
+++ b/services/agent/src/vss_agents/api/video_delete.py
@@ -293,6 +293,12 @@ def create_video_delete_router(
                 results.append(success)
                 logger.info(f"Remove from RTVI-CV: {'OK' if success else msg}")
 
+            # --- Delete VST storage (using shared vst utils) ---
+            with TimeMeasure("video_delete: delete VST storage"):
+                success, msg = await delete_vst_storage(vst_url, video_id)
+            results.append(success)
+            logger.info("Delete VST storage: %s", "OK" if success else msg)
+
             # --- Delete VST sensor (using shared vst utils) ---
             # Required: delete_vst_storage only removes stored files, not the
             # sensor registration — the two must be paired to fully remove a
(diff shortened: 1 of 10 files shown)


services/agent/src/vss_agents/api/video_delete.py
```diff
@@ -292,6 +293,15 @@ async def delete_video(video_id: str) -> DeleteVideoResponse:
...
 
+            # --- Delete VST sensor (using shared vst utils) ---
+            # Required: delete_vst_storage only removes stored files, not the
+            # sensor registration — the two must be paired to fully remove a
+            # video. Without this, sensors are orphaned in VST.
+            with TimeMeasure("video_delete: delete VST sensor"):
+                success, msg = await delete_vst_sensor(vst_url, video_id)
+            results.append(success)
+            logger.info("Delete VST sensor: %s", "OK" if success else msg)
+
```
abhishek-ak-nv, line 310: @zac-wang-nv can we place the delete storage first then delete sensor.

abhishek-ak-nv:
@zac-wang-nv ..this was merged in https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/734

please close this PR after review

PR #751, opened 2026-05-27
chore: record main Brev secure-link fix on develop

freshyjmp:
This is a record-only merge of main commit 6efeef65 (fix(brev): use direct secure-link port prefixes) into the develop line.

Current develop already has the equivalent direct secure-link behavior in the relocated deploy/docker and vss-deploy-profile paths, so this PR intentionally has zero file diff against develop. Its purpose is to make 6efeef65 an ancestor of develop so a future main-to-develop merge does not replay the old-path launchable patch.

Expected oddity: GitHub's commit tab shows the main-side ancestry behind 6efeef65. The files changed tab should be empty. That ancestry is required for this to work.

Important: merge this PR with a merge commit. Do not squash or rebase it, because preserving the merge parent is the point of the PR. A squash merge would turn this into a meaningless empty commit and would not prevent the future conflict.


zac-wang-nv (approved):
LGTM

PR #769, opened 2026-05-27
skill-eval: bump per-trial agent timeout to 1h

zac-wang-nv:
## Summary
- Raise `--agent-timeout-multiplier` from `3.0` → `6.0` in `.github/skill-eval/AGENTS.md`, lifting harbor's per-trial agent-exec ceiling from 1800s (30 min) to 3600s (1 hour).
- Update the accompanying notes: 1h budget rationale, and rewrite the prior "all three multipliers match at 3.0" sentence (no longer true — only the agent multiplier is bumped; env-build and verifier stay at 3.0).

## Why
Cold-box trials that combine `/deploy` (NIM pulls, NGC auth, container start) with multi-step adapter work were exceeding the 30-min agent-exec ceiling and surfacing as `NonZeroAgentExitCodeError` (exit 124, SIGTERM mid-trial). One hour gives margin for the longest observed cold-box trials without uncapping retries.

## Test plan
- [ ] Trigger skills-eval on a PR touching `skills/**` and confirm `uvx harbor run` is invoked with `--agent-timeout-multiplier 6.0`.
- [ ] Spot-check that the workflow's 480-min job cap still comfortably holds the longest expected queue (~13 trials × ≤1h each is the new worst case).
- [ ] Verify a known-slow trial (`lvs` / `alerts_*` cold box) completes without exit-124.

### DIFF
--- a/.github/skill-eval/AGENTS.md
+++ b/.github/skill-eval/AGENTS.md
@@ -10,7 +10,7 @@ You run **once per push**, from start to finish, on the
 `vss-skill-validator-v2` self-hosted runner. Your workspace is already
 checked out at the mirror head. You have `Bash`, `Read`, `Edit`,
 `Write`, `Glob`, `Grep`; no human is in the loop while you work. The
-workflow runs your invocation with an 8-hour hard timeout.
+workflow runs your invocation with a 12-hour hard timeout.
 
 ## Startup hygiene (do this first, before step 1)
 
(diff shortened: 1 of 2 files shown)


zac-wang-nv:
Addressed in 5e5590e6 — bumped `timeout-minutes: 480 → 720` (12 h) and `flock -w 28800 → 43200` everywhere they appear so the job cap and lock-hold stay in lockstep with the new 1 h per-trial agent ceiling.

Sizing rationale (now baked into the comment above `timeout-minutes`):
- Typical: 14 × 30 m = 420 m (well inside 720).
- Mixed-cold (a few trials hit the ceiling): (4 × 60 m) + (10 × 30 m) = 540 m.
- Pure worst case (14 × 60 m + env-build) is ~14 h; covering that would require either a larger job cap or capping concurrent ceilings — left out intentionally to keep the cap pragmatic.

Touched files:
- `.github/workflows/skills-eval.yml` — `timeout-minutes` 480 → 720; comment rewritten.
- `.github/skill-eval/AGENTS.md` — every `28800s` budget and `flock -w 28800` → `43200`; "8 h"/"8-hour" prose → "12 h".

PR #781, opened 2026-05-28
ci(skill-eval): drop harness-side profile pre-deploy + instance reuse

zac-wang-nv:
## Summary

Each trial now deploys whatever VSS profile it needs inside its own first agent turn, rather than the harness pre-deploying via a side channel before the agent starts. The deploy step is therefore visible in the trial's trajectory, reward, and `claude-code.txt` — debuggable the same way as every other agent action.

## What's removed

- **`BrevEnvironment._ensure_prerequisite_deployed()`** and its call site. Was the harness-side pre-deploy hook that ran `/vss-deploy-profile` via a sub-claude before the trial agent started. Gone — the trial's own first query handles deployment.
- **The `/tmp/skill-eval/active-deploy.txt` marker**, end-to-end:
  - writer in `adapters/vss-deploy-profile/generate.py::generate_test_script` (on reward=1.0, wrote `<profile_tag>|<run_id>`);
  - reader in `_ensure_prerequisite_deployed` (warm-reuse check) and the `AGENTS.md` § 5a fleet-selection scoring;
  - tests in `envs/tests/test_registered_node.py`;
  - `specs/stale-marker.spec` design doc — deleted.
- **Adapter emission of `profile`, `prerequisite_deploy_mode`, and `requires_deployed_vss`** into `task.toml [metadata]`. Nothing in the harness reads those fields anymore — they were inputs to `_ensure_prerequisite_deployed` only.

## What's kept and updated

- The **`profile`** field on eval specs is still meaningful as adapter input — it names the profile the adapter embeds into the instruction.md "deploy first via `/vss-deploy-profile -p X`" prose.
- **AGENTS.md § 5a** fleet-selection scoring loses the active-deploy marker dimension; ranking is now `(free-lock, name)`.
- **Adapter instruction.md prose** updated across `vss-ask-video`, `vss-deploy-detection-tracking-2d`, `vss-generate-video-report`, `vss-manage-video-io-storage`, `vss-search-archive`, `vss-summarize-video` from "profile already running" to "deploy it first via `/vss-deploy-profile`, then …". `vss-manage-alerts` already had a two-step structure where step 1 is the deploy — no change needed.

## Consequences

- Every spec's first `expects[]`'s reward now bakes in the success of the deploy. A broken deploy fails `reward=1.0` the same way a broken use-skill query does, in one place.
- **Across-run reuse is gone** — every trial that needs VSS up pays the full deploy time. Within-spec reuse (multiple `expects[]` on the same box) is unchanged because they share one harbor invocation.
- No more "stale marker on cold reboot" / "marker mismatch" / "marker wiped mid-run" failure modes. The PR #650 dataset-wipeout class of bug also disappears, since there's no shared mutable state on the coordinator for concurrent runs to step on.

## Test plan

- [ ] Run a deploy-skill eval (`vss-deploy-profile/eval/alerts_vlm.json`) — should pass end-to-end; the trial's first query *is* the deploy, no harness pre-deploy required.
- [ ] Run a use-skill eval (`vss-ask-video/eval/base_profile_video_understanding.json`) — the agent should read the updated instruction.md, deploy `base` first via `/vss-deploy-profile -p base`, then proceed with the spec's first query.
- [ ] Two concurrent dispatches on different boxes — should no longer fight over `/tmp/skill-eval/datasets/` because no shared marker reconcile races.
- [ ] Confirm AGENTS.md § 5a selection picks the same `vss-eval-*` box twice in a row when free — still works.


.github/skill-eval/adapters/vss-manage-video-io-storage/generate.py
```diff
@@ -189,7 +189,7 @@ def generate_task(platform: str, spec: dict, output_root: Path,
             PREAMBLE,
             "",
             f"Use the `/vss-manage-video-io-storage` skill against the VSS base profile "
-            f"already running on this `{platform}` host "
+            f"on this `{platform}` host (deploy it first via this skill's standalone runbook or `/vss-deploy-profile -p {profile}`, then "
```
zac-wang-nv: `{profile}` is undefined here — `generate_task(platform, spec, output_root, skill_dir, deploy_skill_dir, video_url)` has no `profile` parameter and never assigns to it. AST scan confirms only a Load reference at this line, no Store anywhere in the function. Reproduced locally with a stub spec: `NameError: name 'profile' is not defined`. This crashes the adapter on *every* invocation, not just specs with `deploy_mode` set.

Pull `profile = spec.get("profile", "base")` from the spec at the top of `generate_task` (matching the docstring's "profile-bound spec" / "profile-less spec" dichotomy) and gate the `/vss-deploy-profile -p {profile}` clause on whether the spec actually has a profile — vios_ops.json deliberately doesn't, and `references/deploy-vios-service.md` is the runbook for the profile-less path.

.github/skill-eval/adapters/vss-ask-video/generate.py
```diff
@@ -158,7 +158,7 @@ def generate_task(
             PREAMBLE,
             "",
             f"Use the `/vss-ask-video` skill against the VSS **{profile}** "
-            f"profile already running on this `{platform}` host "
+            f"profile (deploy it first via `/vss-deploy-profile -p {profile}` on this `{platform}` host, then "
```
zac-wang-nv: The outer `(deploy it first …` paren opens but never closes — the closing `)` after "below" closes the inner health-check paren. Rendered, the agent sees:

```
profile (deploy it first via /vss-deploy-profile -p base on this l40s host, then (http://localhost:8000/docs must respond, and a sample warehouse video must already be uploaded per the env notes below).
```

— one unclosed paren. Same defect in `vss-generate-video-report/generate.py:164-165`, `vss-summarize-video/generate.py:149-150`, and `vss-manage-video-io-storage/generate.py:192-193`. `vss-search-archive/generate.py:147-153` is the gold rewrite — adopt that structure (single paren, no nested clause) across the four adapters.

.github/skill-eval/adapters/vss-ask-video/generate.py
```diff
@@ -196,21 +196,16 @@ def generate_task(
...
             f'min_vram_gb_per_gpu = {pspec["min_vram_per_gpu"]}',
-            "requires_deployed_vss = true",
             "# Deploy mode is FULL-REMOTE (LLM + VLM both remote) — vss-ask-video",
             "# exercises POST /generate only, so there is no benefit to local NIMs.",
-            # prerequisite_deploy_mode is alerts-only — the deploy marker
             # is profile-name only for base/lvs/search; the consumer
-            # (envs/brev_env.py::_ensure_prerequisite_deployed) matches
             # on profile alone when this field is absent. Set it only if
             # this spec needs a specific alerts stack (verification vs
             # real-time).
```
zac-wang-nv: Orphan comment fragments left behind after the `*([f'prerequisite_deploy_mode = …'] …)` line and its lead-in comment were removed. The fragment now reads as if "is profile-name only" has no subject and "the consumer / on profile alone" is a sentence fragment referencing a deleted consumer (`envs/brev_env.py::_ensure_prerequisite_deployed`).

Same pattern in `vss-generate-video-report/generate.py`, `vss-search-archive/generate.py`, `vss-summarize-video/generate.py`, and `vss-manage-video-io-storage/generate.py`. Also `vss-deploy-video-embedding/generate.py` left a stranded "# IS the deploy, and the spec has no profile field." after its lead-in was removed, and `vss-manage-alerts/generate.py:223` left "# injects before this task. Read from spec, fall back to `real-time`." dangling.

Delete all of these as a sweep — the surviving lines aren't worth keeping without their lead-ins.

.github/skill-eval/adapters/vss-deploy-dense-captioning/generate.py
zac-wang-nv, line 1: **Stale prose at lines 100-106 — missed by the bulk substitution.**

This branch was missed by the bulk substitution that updated the other adapters. After this PR removes `_ensure_prerequisite_deployed`, the harness no longer predeploys anything — so "already-deployed full VSS profile" is false, "harness predeploys that prerequisite" is false, and forbidding `/vss-deploy-profile` strands the agent with no way to bring VSS up. Profile-spec dense-captioning trials will run against an empty box and fail their first query.

Match the new model the way `vss-search-archive/generate.py:147-153` did (cleanest rewrite — single closed paren): "Use `/vss-deploy-dense-captioning` against the VSS **{profile}** profile (deploy it first via `/vss-deploy-profile -p {profile}` …, then …)".

.github/skill-eval/adapters/vss-manage-video-io-storage/generate.py
zac-wang-nv, line 1: **Stale module docstring at lines 30-50.**

After this PR, `requires_deployed_vss` and `profile` are no longer emitted into `task.toml [metadata]`, and the coordinator no longer arranges pre-deploy chaining (that whole machinery is gone with `_ensure_prerequisite_deployed`). Future maintainers reading this will think the metadata is load-bearing.

Replace with: profile-less path → agent stands VIOS up via the bundled `references/deploy-vios-service.md` runbook in its first turn; profile-bound path → agent deploys via `/vss-deploy-profile -p <profile>` in its first turn. Same mental model as the AGENTS.md § 4 update.

zac-wang-nv:
Stacked a small SDK-side hardening commit on this branch (`f491f0ef`): closes the polling-loop / backgrounding anti-pattern at the harness boundary instead of leaving it to prose in AGENTS.md.

Two pieces, both in `.github/skill-eval/skills_eval_agent.py`:

- **`disallowed_tools`** added to `ClaudeAgentOptions` — explicit blocklist for `BashOutput`, `KillShell`, and the `Task*` family. `allowed_tools` is an allowlist for primary tools but the SDK passes these auxiliary tools through, which is how a drifting agent can reach `TaskCreate` even though it's not in the allow list.
- **`PreToolUse` hook on `Bash`** — `_block_bash_background()` rejects `run_in_background=True` and shell-level `&` / `nohup` / `disown` patterns. The `deny` decision surfaces back to the LLM as a tool-result error so the model retries synchronously.

The orchestrating agent on at least one observed run dispatched `uvx harbor run` as a background process and polled the output file, burning its turn budget before harbor finished and exiting with the protocol-failure exit-code-4 path. AGENTS.md § "No polling — block on harbor" forbids that in prose; this commit makes it structural.

If reviewers prefer this in a follow-up PR rather than stacked here, easy to peel back into its own branch.

zac-wang-nv:
Superseded by #819. #819 is stacked on this branch and now contains the full no-prereq-deploy refactor (these 9 commits) **plus** the matrix-dispatch work, rebased onto develop with the 2 P1 review fixes (per-leg tarball path race + flock-vs-job-timeout). Consolidating to a single PR on develop — closing this one. All of this PR's changes ship via #819.

PR #829, opened 2026-05-29
Addressing the "issues worth fixing" in the FVR report

prisrivastav-nv:
## Description
Addressing the "issues worth fixing" in the FVR report - 

> Two NemoClaw-side issues worth fixing before GA, both Needs Improvement (not release-blockers):
> 
> (A) nemoclaw doctor has a stale gateway-container-name check. It probes for openshell-cluster-nemoclaw (the pre-dev-26.05.3-* name), which doesn't exist anymore — the new container is nemoclaw-openshell-gateway. The doctor's summary says "attention needed (1 failed, 1 warning(s))" even on a fully working deploy because of this stale check. Fix: update the container-name check in nemoclaw doctor's Gateway section.
> 
> (B) Notebook prerequisites are under-documented. Three settings the notebook silently depends on are only surfaced via downstream error messages: (i) Python 3.11+ required because orchestrator_mcp_helper.py imports from enum import StrEnum; users on Ubuntu 22.04's system Python 3.10 get an ImportError from inside Cell 13. (ii) VSS_REPO_DIR override is needed if the repo isn't at ~/video-search-and-summarization; users who clone elsewhere get the same StrEnum error against the wrong tree. (iii) NEMOCLAW_ACCEPT_EXPERIMENTAL_OPENSHELL_UPGRADE=1 is required when upgrading from any pre-0.0.37 OpenShell state — the installer's error message names it but the notebook doesn't. Fix: add a prereq block at the top of the notebook listing these three settings. Optionally fail fast in the preflight cell.

## Checklist


zac-wang-nv (commented):
can you please format the notebook it so that only the delta is displayed from the diff?

prisrivastav-nv:
Closed in favour of #850

PR #827, opened 2026-05-29
chore(skills): trigger skill-eval for vss-deploy-profile

hugoverjus:
## Summary

- Adds a single trailing blank line to `skills/vss-deploy-profile/SKILL.md` so the PR diff touches `skills/vss-deploy-profile/**`.
- The only purpose is to trigger the **Tier-3 skills-eval** benchmark run for `vss-deploy-profile` (via `.github/workflows/skills-eval.yml`) and collect the Harbor eval results (reward, duration, turns, token usage) across its eval specs.
- No functional change to the skill.

## Coverage

`vss-deploy-profile` ships 5 eval specs, all exercised by this run:
- `base.json`
- `alerts_cv.json`
- `alerts_vlm.json`
- `search.json`
- `lvs.json`

## How the benchmark gets produced

Once a maintainer comments `/ok to test <sha>`, copy-pr-bot mirrors this PR to `pull-request/<N>` and skills-eval runs each spec on the declared platforms, posting one Harbor Eval results comment per spec.

## Test plan

- [ ] Maintainer issues `/ok to test` to kick off the skills-eval workflow
- [ ] Harbor Eval comments are posted with per-step reward/duration/turns/token tables for each spec

PR #833, opened 2026-05-29
fix(skill-eval): unique per-(spec,platform) harbor output dir (fixes PR #827 trial loss)

hugoverjus:
## Summary

Fixes the skills-eval coordinator failure seen on the **"Eval changed skills against PR"** check for PR #827, where the deploy-skill benchmark sweep silently lost trials (`base`, `lvs`, `alerts_cv`).

Root causes (from the run log):
- The coordinator fanned out several `uvx harbor run` invocations that **shared one `-o /tmp/skill-eval/results/$GITHUB_RUN_ID` root**. Harbor creates a `<date>/<trial>` subdir keyed to the *second*, so trials starting in the same second collided with `FileExistsError` and never ran.
- Those failures were **masked by `; echo "<spec> done"`** (exit code ignored), so the agent believed they had passed.
- `uvx` was **not on `PATH`** in the agent's subshells (`~/.local/bin` not sourced), stalling dispatch.

## Changes (`.github/skill-eval/AGENTS.md` — the coordinator's system prompt)

- **Unique per-(spec, platform) output root (`$TRIAL_OUT`)** for *every* harbor invocation — canonical, multi-step loop, and fan-out. Makes serial retries, stale dirs from a failed attempt, and concurrent fan-out all collision-proof.
- **Fan-out example** now keys `-o` by `(spec, platform)` (not just skill), captures each job's pid, and checks exit codes after `wait`; added a cheap duplicate-`-o` pre-flight assertion.
- **New forbidden patterns**: sharing one `-o` across concurrent runs, and masking a harbor failure with `; echo done`. A missing `reward.txt` must be recorded as a failure, not skipped.
- **`export PATH="$HOME/.local/bin:$PATH"`** in the canonical block so `uvx` resolves in subshells.
- Updated all downstream read paths (reward, `trial.log`, trajectory, per-trial metrics, viewer migration, suggestions) to the new per-(spec, platform) root.

Docs-only change to the agent prompt; no executable harness code touched.

## Test plan

- [ ] Re-run the skills-eval sweep for `vss-deploy-profile` (6 specs across RTX 1g/2g boxes) and confirm all specs produce a reward with no `FileExistsError`.
- [ ] Confirm the Harbor viewer URLs resolve for each `(spec, platform)`.

### DIFF
--- a/.github/skill-eval/AGENTS.md
+++ b/.github/skill-eval/AGENTS.md
@@ -369,9 +369,12 @@ Intentionally shared paths (do NOT scope these under `$SCRATCH`):
       trial cleanly. Spend turns on the next trial's setup or on
       reading already-completed trial logs instead.
    d. After each trial, parse
-      `/tmp/skill-eval/results/<run_id>/<date>/<trial>/verifier/reward.txt`
+      `$TRIAL_OUT/<date>/<trial>/verifier/reward.txt` (i.e.
+      `/tmp/skill-eval/results/<run_id>/<spec_stem>-<platform>/<date>/<trial>/verifier/reward.txt`)
       and `test-stdout.txt`. Record `(spec, platform, reward,
-      checks_passed/total, duration_s, trace_url)` for the comment.
+      checks_passed/total, duration_s, trace_url)` for the comment. A
+      missing `reward.txt` means the trial errored (e.g. non-zero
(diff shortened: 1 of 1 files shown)


hugoverjus:
Closing this standalone fix PR because its changes have been folded into #827, so we keep one combined PR for the deploy benchmark + harness fix.

PR #853, opened 2026-05-29
Feature/scbp smartcity alert verification

tnoumessi-nv:
## Description
Adds a Smart City industry profile that deploys on top of the existing alert-verification Compose stack, plus a launchable notebook that walks a user through the full deployment.

What's new
Industry profile — deploy/docker/industry-profiles/smartcities/

compose.yml + .env — Smart City service stack layered on the alert-verification base
Calibration — calibration-microservice/ with sample Agnew-intersection calibration.json and road-network.json, plus calibration-import.sh init script and an import-calibration Dockerfile
Perception — deepstream-perception/ RT-DETR 960×544 label config
Verification — vlm-as-verifier/ configs (alert_type_config.json, config.yml)
Agent — vss-agent/ config, MCP server config, and incident-report template
Analytics + dashboards — Kibana ITS dashboard objects, Behavior-Analytics Kafka config, Video-Analytics API + UI configs

Launchable notebook — deploy/docker/scripts/deploy_smartcities_launchable.ipynb

- 12-step end-to-end guide: prereqs → env vars → data/model download from NGC (RT-DETR, GDINO) → compose configuration via sed → docker login nvcr.io → docker compose up → deployment verification (container status, perception FPS, UI/API endpoint probes) → optional custom-stream upload → UI/Kibana/Map exploration → teardown
- Supports H100 / L40S / RTX PRO 6000 Blackwell hardware profiles and local / local_shared / remote LLM-VLM modes


tnoumessi-nv:
@dineshvenkat @hugoverjus @zac-wang-nv Please review

hugoverjus:
Here are some comments about the current PR:

- The notebook fails on the second code cell because `USE_REMOTE_LLM`, `USE_REMOTE_VLM`, `LLM_ENDPOINT_URL`, `VLM_ENDPOINT_URL`, `LLM_NAME`, and `VLM_NAME` are referenced before being defined. Running it top-to-bottom currently raises `NameError`.

- `GOOGLE_MAPS_API_KEY` defaults to empty, but `cell 17` asserts it is set. This blocks deployment for users who do not need the map view; a warning would be more appropriate than a hard failure.

- Remote LLM/VLM mode is only partially implemented. The notebook validates those settings, but the deploy step does not pass them into `dev-profile.sh` or the environment. This should either be wired through or removed from the UX.

- `va_mcp_server_config.yml` hardcodes `meta/llama-3.1-70b-instruct` instead of using `${LLM_NAME}`, which can make the MCP server use a different model than the agent.

One broader concern: the Smart Cities profile only works after the notebook rsyncs files into `developer-profiles/dev-profile-alerts`. That makes the profile non-self-contained and mutates the alerts profile. Existing industry profiles such as `warehouse-operations` reference their own `industry-profiles/...` paths directly, so this pattern is worth reconsidering.

Additional cleanup items: remove the hardcoded private IP in `vss-behavior-analytics-kafka-config.json`, replace or remove the dev-looking `../ITS_new/...` OSM path, scope teardown volume cleanup to the VSS project instead of all Docker volumes, add a trailing newline to `.env`, and fix typos in `calibration-import.sh`.

Note: the Greptile comments about `set -e` breaking the calibration polling loop look like false positives. I tested the shell behavior, and the loop retries because `fetchstatus()` exits with status 0.

dineshvenkat (approved):

PR #857, opened 2026-05-30
Fix launchable Docker prereq pin flow

freshyjmp:
## Summary
- Fixes NVBug 6244624 by allowing the launchable prerequisite cell to continue when Docker is newer than the supported range.
- Keeps the lower-bound Docker failure, but reports Docker >=29.5.0 as a warning because Section 4.1 pins Docker before deployment image pulls.
- Updates prerequisite text to document the supported range and pinning flow.

## Validation
- jq empty deploy/docker/scripts/deploy_vss_launchable.ipynb
- jq -r '.cells[5].source|join("")' deploy/docker/scripts/deploy_vss_launchable.ipynb | bash -n
- git diff --check


deploy/docker/scripts/deploy_vss_launchable.ipynb
```diff
@@ -199,9 +199,12 @@
     "DOCKER_VERSION=$(docker version --format '{{.Server.Version}}')\n",
     "version_ge \"$DOCKER_VERSION\" \"$MIN_DOCKER_VERSION\" \\\n",
```
zac-wang-nv, line 200: should this also be removed?

zac-wang-nv (approved):
LGTM

PR #863, opened 2026-05-31
Fix(vss-generate-video-calibration): harden deploy, calibration, and result reporting

shubham050300:
## Description

Addresses reviews in the vss-generate-video-calibration skill, tightening robustness across deploy, calibration, and post-run reporting without changing the core workflow.

### Deploy
- Verify NGC access to the AMC images before up -d — derive image refs from the resolved compose and pull them, failing fast with a clear message when the key logs in but can't access the vss-core images. Split the auth vs image-access troubleshooting cases.
- Added a projects-directory writability preflight (Step 5): write-test as the container user (UID 1000) and, on failure, apply a narrow setfacl (with approval) instead of a broad chmod. Prevents the [Errno 13] Permission denied at first create_project.
- Collapsed the two near-identical deploy paths into one profile-parameterized flow.

### Calibration
- Fixed settings-file detection: added calibration_settings.json to the auto-scan candidates, so it's applied via /v1/config before /calibrate.
- Added a pre-calibration confirmation (detector, settings file or defaults, optional overrides) with resolved values pre-filled; clarified detector fallback when a settings file omits it.
- Corrected the upload field name in common-steps.md (files, not video_files).

### Progress & results
- Long-run feedback: surface project ID / UI URL / log endpoint at calibration start, emit a once-a-minute progress heartbeat during RUNNING, report final elapsed time, and print the tail of the log on ERROR. Raised the poll ceiling 60→90 min to match the existing "stuck" threshold.
- Post-completion review section always shown: UI Results navigation, on-disk overlay image path, and project files dir. A missing evaluation_statistics (no GT.zip) is now reported as a normal case rather than ending result reporting.

## Checklist


skills/vss-generate-video-calibration/SKILL.md
```diff
@@ -3,6 +3,7 @@ name: vss-generate-video-calibration
 description: Use to run AutoMagicCalib on local MP4s, RTSP, or the bundled sample dataset, and to deploy vss-auto-calibration when needed. Not for non-AMC calibration or runtime analytics.
 license: Apache-2.0
 metadata:
+  author: "NVIDIA Video Search and Summarization team"
```
zac-wang-nv: we don't need to have author for github(external) skills.

shubham050300: Thanks, removed with the latest commit

zac-wang-nv (approved):
LGTM

PR #879, opened 2026-06-01
Fix launchable Docker 29.5 NVCR workaround on develop

freshyjmp:
## Summary
- Port the Docker 29.5.x / NVCR workaround from the main-branch launchable notebook fix into the current develop notebook path.
- Set `features.containerd-snapshotter=false` in `/etc/docker/daemon.json` when Docker Engine is 29.5.x, then restart Docker if the daemon config changes.
- Keep the launchable default source path pointed at the Brev bringup clone instead of cloning from GitHub by default.

## Context
- Related main PR: https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/849
- Related NVBug: https://nvbugs/6244624
- Builds on the already-merged prereq warning flow from PR #857.

## Validation
- `jq empty deploy/docker/scripts/deploy_vss_launchable.ipynb`
- Python AST compile check for all non-bash notebook code cells
- Extracted helper-function smoke test for Docker 29.5.x workaround behavior and idempotency
- `git diff --check`


zac-wang-nv (approved):
LGTM

PR #892, opened 2026-06-02
fix: make deploy skill request for remote/local llm/vlm confirmation

maxervo:
## Description
Deploy skill never asks whether the user wants a remote vs local LLM endpoint. It silently uses whichever env var happens to be set. Make this decision as an explicit step in the skill flow.

Preserve the intent of non-autonomous / autonomous mode, but ensure the autonomous mode continues only if the user intent (in user prompt) aligns with the env variables defined.

Sometimes `$LLM_REMOTE_URL` / `$VLM_REMOTE_URL` are used but these can cause confusion because the real env names used by `dev-profile.sh` are `LLM_ENDPOINT_URL`, VLM_ENDPOINT_URL`. Ensure the naming is consistent so skill does not drift off.

## Checklist

### DIFF
--- a/skills/vss-deploy-profile/SKILL.md
+++ b/skills/vss-deploy-profile/SKILL.md
@@ -99,7 +99,7 @@ for the remediation tree.
 
 ## Model Selection
 
-- `$LLM_REMOTE_URL` / `$VLM_REMOTE_URL` if the user asks for remote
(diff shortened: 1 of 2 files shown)


zac-wang-nv:
@maxervo , it should already be addressed, https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/889

can you validate with 889?

zac-wang-nv:
closing as superseded

PR #904, opened 2026-06-02
fix(launchable): skip Docker pin when version already in tested range (nvbug 6257280)

zac-wang-nv:
## Problem (nvbug 6257280)

`[DGX Spark][NemoClaw VSS]`: the Docker version-pin cell hard-installed an exact epoch-versioned package set unconditionally:

```bash
sudo apt-get install -y docker-ce="5:29.4.3-1~ubuntu.${VERSION_ID}~${UBUNTU_CODENAME}" ...
```

On DGX Spark (DGX-OS / arm64) that exact `5:29.4.3-1~ubuntu…` string is not in the configured apt index, so the install aborts with *version not found* — the **"package version mismatch"** in the bug. And the box already ships **Docker 28.5.1**, which is inside the tested range, so the pin was both failing *and* unnecessary (per the bug's own dev comment: *"you should not need to run the Docker package pinning if the existing docker version met the requirement"*).

The two notebooks' pin cells are kept byte-identical (since #603) but neither used the supported range to gate the pin — the launchable's §3 prereq check only *prints* OK/WARNING, it never skips §4.1.

## Fix

Gate the pin on the installed Docker Engine version. The cell now reads `docker version --format '{{.Server.Version}}'` and:

- **skips** the pin when Docker is already in **`[28.3.3, 29.5.0)`** (the same range the launchable §3 prereq check validates),
- otherwise runs the existing hard-pin + `apt-mark hold` unchanged,
- falls back to pinning if the version can't be read.

Applied to both `deploy_nemoclaw_vss.ipynb` (§2.1) and `deploy_vss_launchable.ipynb` (§4.1) so the cells stay in sync.

## Verification

`bash -n` clean; decision matrix exercised against representative versions:

| Installed Docker | Result |
|---|---|
| `28.5.1` (DGX Spark) | **SKIP** ✅ |
| `29.4.3` (pin target) | **SKIP** |
| `28.3.3` (min, inclusive) | SKIP |
| `28.3.2` | PIN |
| `29.5.0` (max, exclusive) | PIN |
| `29.5.1` / `30.0.0` | PIN |
| unreadable / empty | PIN (safe fallback) |

## Notes

- This is notebook/markdown only — no runtime code paths change for hosts that were already pinning.
- Out of scope (flagged for a follow-up): the **nemoclaw notebook still lacks** the launchable's storage relocation and the §879 Docker 29.5.x NVCR (`containerd-snapshotter=false`) workaround. Comment #1 on the bug ("ran into other errors later in the flow") is consistent with a 29.5-class NVCR pull failure once past the pin.


freshyjmp (approved):

nvleeryan:
@zac-wang-nv i tried on my spark and it skipped the docker pinning so it’s fine.  Did you try it on a fresh machine with no docker install or out of range docker version?

zac-wang-nv:
merging as we have both approval from brev and spark.

PR #902, opened 2026-06-02
fix relative paths

prisrivastav-nv:
## Description
Fix relative path references in agent readme

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/HEAD/CONTRIBUTING.md).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #906, opened 2026-06-02
ci: skip container-source gate for docs/tests-only changes

zac-wang-nv:
## Description

The **Check Agent Container Source** / **Check UI Container Source** gates hash the *entire* `services/agent` and `services/ui` trees and compare that tree SHA against the `com.nvidia.vss.source_tree_sha` baked into the deployed image. A docs- or tests-only change never alters what the image build consumes, yet it changes the folder's tree SHA and trips the gate — forcing a needless container re-spin or NGC manifest re-stamp for a README edit (e.g. #902).

This PR adds `source_check_only_nonbuild.py`. Before the SHA comparison runs, the gate checks the PR's diff under the service folder: if **every** changed file is one the build provably does not consume, it skips the comparison. The job still runs and reports success, so the required status stays green — no branch-protection change needed.

What counts as "non-build":
- **agent** — `services/agent/docker/Dockerfile` uses *selective* `COPY`, so service-root docs (`README.md`, `AGENTS.md`, `LICENSE.md`), `tests/`, `stubs/`, and CI-metadata files (`.gitleaks.toml`, `.secrets.baseline`, …) never enter the build context. Patterns are anchored to the service root — `src/vss_agents/**/README.md` **is** copied and is correctly treated as build-relevant.
- **ui** — `services/ui/.dockerignore` excludes `**/*.md` and `*.test.*`/`*.spec.*`, so markdown and test/spec files never ship.

Anything build-relevant, an unresolved merge-base, or a push to `develop`/`main` runs the full check exactly as before.

### Scope / safety
- No change to `check_container_tag_source.py` or to the build pipeline.
- The job always executes → the required status always reports (no skipped-required-check hang).
- Verified locally: agent docs-only → skip; agent `src/**/README.md` → run; agent `.py` → run; UI docs-only → skip.

### Follow-up
This is a stop-gap. The durable fix (separate PR, draft) scopes `source_tree_sha` to **build inputs** on both the build-stamp (`ci-vss-oss/ci/tools/create_manifest.py`) and check sides, after which this helper can be removed.

## Checklist

### DIFF
--- a/.github/scripts/source_check_only_nonbuild.py
+++ b/.github/scripts/source_check_only_nonbuild.py
@@ -1,0 +1,171 @@
+#!/usr/bin/env python3
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""Decide whether a PR touches only non-build files under a service folder.
+
+The ``Check {Agent,UI} Container Source`` gates compare the git tree SHA of
+``services/agent`` / ``services/ui`` against the ``com.nvidia.vss.source_tree_sha``
+baked into the deployed image. That tree SHA covers the *whole* folder, so a
+docs- or tests-only change (which never enters the image — the Dockerfiles use
+selective ``COPY`` / a ``.dockerignore``) still trips the gate and forces a
+needless container re-spin or manifest re-stamp.
+
+This helper is the stop-gap: it prints ``true`` when **every** file the PR
+changed under the service folder is one the image build provably does not
+consume, so the caller can skip the SHA comparison for that run. It prints
+``false`` (run the real check) whenever it is unsure — unknown base, an
+integration-branch push, or any build-relevant file in the diff.
+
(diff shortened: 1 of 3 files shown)


zac-wang-nv:
Addressed both in 63bf4d3:

1. **UI test/spec patterns narrowed** — `vss-agent-ui` non-build patterns now mirror `services/ui/.dockerignore` exactly: `**/*.md` plus `*.test.{js,ts,tsx}` / `*.spec.{js,ts,tsx}` only, replacing the broader `**/*.test.*` / `**/*.spec.*`. The skip-list is now a strict subset of what the build context drops, so it can't skip something that ships (e.g. a `*.test.py` is now treated as build-relevant → full check runs).

2. **Filter step is fail-safe** — both workflows' detect steps now set `continue-on-error: true`, and the `run` block defaults `skip=false` on any helper error or non-`true` output (`if ! skip=$(...); then skip=false; fi` + normalize). A script crash now falls through to the full check instead of failing the job.

Verified locally: `.test.tsx/.ts/.js` & `.spec.*` → skip; `.test.py` → build-relevant; and the bash step yields `skip=false` on crash/garbage output, `true`/`false` otherwise.

PR #907, opened 2026-06-02
ci: scope container-source SHA to build inputs (root-cause fix for docs re-spins)

zac-wang-nv:
> **Draft — do not merge until the companion `ci-vss-oss` change lands (see below).** This PR is non-breaking on its own, but its benefit only activates once images are stamped with the new hash.

## Problem

The **Check Agent/UI Container Source** gates compared the *entire* `services/agent` / `services/ui` git tree (`git rev-parse <commit>:services/agent`) against the image's `com.nvidia.vss.source_tree_sha`. Any change to a file that the image build never consumes — `README.md`, `AGENTS.md`, `tests/`, CI metadata — changes that tree SHA and trips the gate, forcing a needless container re-spin or NGC manifest re-stamp (e.g. #902). PR #906 is the stop-gap; this is the root-cause fix.

## Change

**`source_tree_hash.py`** — `build_input_tree_sha()` hashes only the files that influence the image:
- git-tracked files under the service folder,
- **minus** anything matched by that folder's `.dockerignore` (mirrors the docker build context),
- **plus** the Dockerfile (docker strips it from the context, but it shapes the image),
- as a `sha256` over sorted `<blob_sha> <path>` lines — deterministic and reproducible identically at build time and check time.

Includes a stdlib `.dockerignore` matcher (`**`, `*`, `?`, leading `/`, trailing-`/` dirs, `!` negation). No third-party deps.

**`services/agent/.dockerignore`** — lists the service-root docs/tests/CI files the agent Dockerfile's *selective* `COPY` never consumes. Anchored to the service root so shipped `src/**/README.md` still counts. (`services/ui` already has a `.dockerignore` excluding `**/*.md` and test/spec files.) This also trims the agent build context.

**`check_container_tag_source.py`** — now accepts **either** the build-input hash (new stamps) **or** the legacy full-tree SHA (existing images). Backward-compatible by construction.

## Why this is safe to merge early

- On `develop` today, the agent image's stamp matches via the **legacy full-tree** branch → still PASS. Verified by running the script locally against the live `nvstaging` image.
- No image carries a build-input stamp yet, so the new branch is inert until `ci-vss-oss` produces one. Behavior is identical to today until then.
- Validated locally:
  - agent docs (`README.md`+`AGENTS.md`) edit → build-input hash unchanged; `src/*.py` edit → changed; shipped `src/vss_agents/orchestrator/README.md` edit → changed (correctly).
  - UI `README.md` edit → unchanged; UI `Dockerfile` edit → changed (force-included).
  - dual-accept: build-input match → PASS; legacy full-tree match → PASS; neither → FAIL.

## Required companion change (separate, `ci-vss-oss`)

`ci/tools/create_manifest.py` must stamp `com.nvidia.vss.source_tree_sha` using **this exact algorithm** — vendor `source_tree_hash.py` or reimplement it byte-for-byte. Once it does, newly promoted images carry the build-input hash and docs/tests-only PRs stop tripping the gate without a re-spin. Existing images continue to pass via the legacy path until re-promoted.

## Relationship to #906
#906 (path-filter stop-gap) gives immediate relief now; this PR makes the gate correct at the source. Once this is fully rolled out, #906's helper can be removed.

## Checklist

### DIFF
--- a/.github/scripts/check_container_tag_source.py
+++ b/.github/scripts/check_container_tag_source.py
@@ -29,6 +29,9 @@ import sys
 from dataclasses import dataclass
 from pathlib import Path
 
+# Sibling module in .github/scripts/ (sys.path[0] when run as a script).
+from source_tree_hash import build_input_tree_sha
+
 
 SOURCE_TREE_SHA_LABEL = "com.nvidia.vss.source_tree_sha"
 SOURCE_PATH_LABEL = "com.nvidia.vss.source_path"
@@ -522,6 +525,7 @@ def check_resolved_image(
     item: ResolvedImage,
     current_commit: str,
     current_tree: str,
+    current_build_inputs: str,
     idx: int,
     total: int,
 ) -> bool:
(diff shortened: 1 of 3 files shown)


zac-wang-nv:
Closing in favor of #913. Per the thread discussion (with @hverjus): rather than re-scoping the source_tree_sha (custom build-input hash / .dockerignore parsing), #913 keeps the existing full-subtree SHA check untouched and instead gates *when* it runs — only on build-relevant source changes or a per-service image-tag bump. That avoids the custom-hash/Docker-context concerns raised here and also fixes the broader issue of the gate failing on PRs that don't touch container source at all (e.g. #911). `protected_source_sha` can be revisited post-release if we want the check to log the exact shipped fingerprint.

PR #913, opened 2026-06-02
ci: run container-source gate only on build-relevant or tag changes

zac-wang-nv:
## Problem

The **Check Agent/UI Container Source** gates compare the full `services/agent` / `services/ui` git subtree SHA against the image's `com.nvidia.vss.source_tree_sha`. Once an allowed docs change merges to `develop` (via the #906 skip), the folder's subtree SHA drifts from the promoted image — and because the check re-evaluates HEAD's (drifted) folder, it then **fails on every later PR, even ones that touch neither the service nor the image tag** (e.g. #911, which only changes `.github/skill-eval/`). These are **ruleset-required** checks ("Protect develop"), so a red/missing status blocks the PR.

> Note: because they're required, we can't gate them with a workflow-level `on: push: paths:` filter — a required check that never runs leaves the PR hanging. So this is done **in-job**: the job always runs and reports success; it just skips the SHA comparison when it isn't relevant.

## Change

Extend the gate helper so a service's check **runs the SHA comparison only when the PR changed**:

1. a **build-relevant** file under that service folder (anything the image build consumes — not the allowlisted docs/tests/metadata), **or**
2. that service's **resolved deployable image tag** (a tag bump in `deploy/docker` compose/`.env`).

Otherwise it skips (passes) — a predecessor's docs-drift can no longer red-flag unrelated PRs.

The tag check is **per-service and value-based**: it resolves the image ref at HEAD vs the merge-base (reusing `check_container_tag_source.py`'s own compose/`.env` resolution), so an agent tag bump in a shared `.env` doesn't trip the UI gate and vice-versa.

The existing SHA comparison in `check_container_tag_source.py` is **untouched** — no custom hash, no `.dockerignore` parsing. This is purely about *when* the (unchanged) check runs.

## Verified locally (merge-base vs HEAD)

| PR touches | agent gate | ui gate |
|---|---|---|
| only `.github/skill-eval/` (#911) | skip | skip |
| `services/agent/README.md` (docs) | skip | skip |
| `services/agent/tests/…` | skip | skip |
| `services/agent/src/**` | **run** | skip |
| agent tag bump (`VSS_AGENT_VERSION`) | **run** | skip |
| ui tag bump (`services/ui/compose.yml`) | skip | **run** |

The job still always runs (required-status-safe; `continue-on-error` + fail-safe default from #906 retained).

## Relationship to #906 / #907
- Supersedes the one gap in #906 (it ran the full check when a PR touched *no* service files → the source of the #911-style failures).
- Makes #907 (build-input/`.dockerignore` SHA filtering) unnecessary — we no longer need the check to *ignore* docs; we just don't *run* it for docs/unrelated PRs. Closing #907.

## Unblocking already-open PRs
A PR picks up this logic from its own branch's workflow, so already-open PRs (#911) need to merge/rebase `develop` after this lands (or re-promote the images to clear the current drift).

## Checklist

### DIFF
--- a/.github/scripts/check_container_tag_source.py
+++ b/.github/scripts/check_container_tag_source.py
@@ -76,12 +76,12 @@ def strip_quotes(value: str) -> str:
     return value
 
 
-def read_env_file(path: Path) -> dict[str, str]:
+def parse_env_text(text: str) -> dict[str, str]:
+    """Parse ``KEY=value`` lines from .env content. Shared with the gate helper
+    so both resolve the same way regardless of which revision the text is read
+    from (working tree vs ``git show``)."""
     values: dict[str, str] = {}
-    if not path.exists():
-        return values
-
-    for raw_line in path.read_text().splitlines():
+    for raw_line in text.splitlines():
         line = raw_line.strip()
         if not line or line.startswith("#"):
             continue
@@ -94,6 +94,12 @@ def read_env_file(path: Path) -> dict[str, str]:
         if re.fullmatch(r"[A-Za-z_][A-Za-z0-9_]*", key):
             values[key] = strip_quotes(value)
     return values
+
+
+def read_env_file(path: Path) -> dict[str, str]:
+    if not path.exists():
+        return {}
+    return parse_env_text(path.read_text())
 
 
 def image_name(ref: str) -> str:
@@ -120,9 +126,11 @@ def commit_prefix_from_tag(tag: str | None) -> str | None:
     return matches[-1].group("sha").lower()
 
 
-def find_image_refs(compose_file: Path, expected_image_name: str) -> list[str]:
+def image_refs_in_text(text: str, expected_image_name: str) -> list[str]:
+    """Extract ``image:`` refs matching ``expected_image_name`` from compose
+    content. Shared with the gate helper (see ``parse_env_text``)."""
     refs: list[str] = []
-    for line in compose_file.read_text().splitlines():
+    for line in text.splitlines():
         match = IMAGE_LINE_RE.match(line)
         if not match:
             continue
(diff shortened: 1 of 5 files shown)


zac-wang-nv:
Addressed the review (commits `36c96ad`, `d5709c5`):

- **Comment 3 — unresolvable ref masked (real, fixed).** `resolve_service_refs` now returns `(resolved, unresolved)`; a raw ref matching the service that no `.env` can expand is recorded as `unresolved`. `service_tag_changed` runs the full check when HEAD introduces an unresolved ref the merge-base didn't have — so the gate no longer hides a failure `verify()` would raise (`return 1` on unresolved). Covered by a test.
- **Comment 2 — duplicated parsers (fixed).** Extracted `parse_env_text` / `image_refs_in_text` text-cores in `check_container_tag_source.py`; `read_env_file`/`find_image_refs` now wrap them, and the gate imports them. The "resolves identically to the check" invariant is now structural, not hand-maintained.
- **Comment 4 — `SOURCE_PATHS` duplication (fixed).** Now derived from `chk.IMAGE_CONFIGS`.
- **Comment 1 — no tests (fixed).** Added `test_source_check_only_nonbuild.py` (stdlib `unittest`, 11 cases): `is_nonbuild` per service incl. shipped `src/**/README.md` staying build-relevant and UI `*.test.py` staying build-relevant; `resolve_service_refs` resolved/unresolvable/other-service-isolation; `service_tag_changed` bump / no-change / conservative-on-empty / newly-unresolvable. The repo's pytest job runs from `services/agent` so it won't collect this — each gate workflow runs it as a step, so a helper regression fails the required check.
- **Empty `base_refs` (Greptile).** Not a correctness issue: an empty `base_refs` with non-empty HEAD means `head != base` → runs the check (safe direction); the diagnostic loop already prints the deltas. Left as-is.

Note: branch is `BEHIND` develop (which advanced); happy to update it before merge.

PR #965, opened 2026-06-05
fix(alerts): restore sensor_id on UI rule creation

ashan-nv:
## Summary

Real-time alert rules created from the **Manage Alerts** tab were not generating any alerts. The UI omitted `sensor_id` from the `POST /api/v1/realtime` request, so Alert Bridge persisted the rule (returning an `id`, so creation appeared to succeed) but never registered the stream with RTVI-VLM — so no captions/alerts were ever produced. Creating the same alert via the chat window first registered the stream, which is why a chat-then-UI sequence appeared to work.

## Root cause

This is a regression. PR #548 (`52cefeed`, merged 2026-05-19) added `sensor_id` to the create request to fix nvbug 6132128. PR #633 (`26ac93fb`, merged 2026-05-22) reworked the sensor input from URL-based to a name-based picker and, in swapping the resolver (`resolveSensorForLiveStreamUrl` → `resolveSensorByName`), dropped `sensor_id` — on the incorrect assumption that Alert Bridge could resolve the stream from `sensor_name` alone. The chat path actually sends both `sensor_name` and `sensor_id`.

## Fix

Restores `52cefeed`'s behavior while keeping #633's name-based picker:
- `resolveSensorByName` now resolves `sensor_id` from `/v1/sensor/list` (`name` → `sensorId`) — the same authoritative source #548 used and the same UUID the chat path forwards. When the sensor isn't `online` in `/v1/sensor/list`, `sensor_id` is intentionally left unset rather than falling back to the catalog's inner `streamId`: `streamId` is a different field and isn't guaranteed to equal the VST `sensorId` (e.g. remote sensors), so guessing it could register the wrong stream. In that case the rule is still created by name; Alert Bridge just won't auto-register the stream until an authoritative `sensor_id` is available.
- `CreateRealtimeRuleInput` carries `sensor_id` again; `createRule` forwards it in the POST body.

## Tests
- Deployed the backend on a brev machine, added a new stream to VST, then added the stream in video management in the UI, created a realtime alert, then confirmed alerts are showing up for that stream.

image

- `vstSensorList.test.ts`: asserts `sensor_id` is taken from `/v1/sensor/list` (returns a different value than the catalog `streamId` to prove which wins) + a test locking in that `sensor_id` stays unset when the sensor isn't online.
- `useRealtimeAlertRules.test.ts` / `CreateAlertRulesView.test.tsx`: assert the create POST body now includes the resolved `sensor_id`.

These act as a regression guard so `sensor_id` can't be silently dropped a third time.

Refs: nvbug 6132128, regressed by #633.


ashan-nv:
> `resolveSensorByName` needs a fallback to `match.streamId` when `sensorMap.get(match.name)` returns undefined.

Thanks — the observation is accurate (an offline-in-`/v1/sensor/list` sensor yields an unset `sensor_id`, so Alert Bridge won't auto-register), but leaving it unset is **intentional**, not an unfinished fallback.

The catalog's inner `streamId` (`/v1/live/streams`) and the VST `sensorId` (`/v1/sensor/list`) are distinct fields and aren't guaranteed to be equal — for remote sensors in particular they can differ. Falling back to `streamId` would risk registering the *wrong* stream, which is worse than not auto-registering. For plain RTSP sensors the two happen to coincide (verified against a live VST: `sample-warehouse-ladder.mp4` → both `f6870d08-…`), but the resolver deliberately doesn't rely on that coincidence.

Behavior when the sensor isn't online: the rule is still created by name, and it starts firing once an authoritative `sensor_id` becomes available (e.g. the sensor comes online / is registered via the chat path). This is a narrow, deliberate limitation — distinct from the original regression, which dropped `sensor_id` for *all* sensors including online ones, so no UI-created rule ever fired.

The code comment and the "stays unset when not online" test lock this in. I've also updated the PR description, which previously described a `streamId` fallback that was never the intended behavior. A UI hint for the offline-sensor case is tracked as a separate follow-up.

ashan-nv:
Closing as issue is no longer visible. The assumption is that alert bridge made a change to relax the need for a sensor id when creating an alert rule. This was included in the agent chat alert creation request but not in the UI flow which started the bug before. However, alert bridge seems to handle this now.

PR #979, opened 2026-06-08
fix(skills): use imperative voice in vss-manage-video-io-storage skill

nvprakhar:
Replaced the role-play opener with an imperative command

## Description
- Replaced the role-play opener with an imperative command ("Call the VIOS REST API to…"), which also mirrors the existing `description:` frontmatter ("Use to call the VIOS REST API…").
- Tightened the trailing instruction ("Always query…" → "Query…") to match the imperative voice.

## Checklist


zac-wang-nv (commented):
LGTM, can you sync it using /nvskills-ci ? it needs to be signed and synced

PR #1016, opened 2026-06-09
fix: search rtsp streams with proper message shape

maxervo:
## Description
The issue was that the vss-search-archive skill documented an invalid RTSP /generate request shape: 
- either it uses `input_message` with `search_source_type`, but this API only accepts extra options with the `messages` envelope.
- or it uses `input_message` only, with “rtsp” in text → accepted, but defaults to video_file internally and searches the wrong index.

We updated the skill and RTSP troubleshooting references accordingly to differentiate video file vs RTSP operations as well. We document the proper message shape for RTSP.

Tested on new Brev machine with search profile deployment. Reproduced and tested fix.

## Checklist

Screenshot 2026-06-09 at 1 31 00 PM

### DIFF
--- a/skills/vss-search-archive/SKILL.md
+++ b/skills/vss-search-archive/SKILL.md
@@ -241,6 +241,8 @@ curl -s -X POST http://${HOST_IP}:8000/generate \
 
 ### More Examples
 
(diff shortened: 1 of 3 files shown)


zac-wang-nv:
we should standardize into using 'messages' , and remove "input_messages", with some doc about why it should work this way(extra fields support). this can be done later.

PR #1121, opened 2026-06-29
Consolidate duplicate RT VLM realtime alert results

danh-nguyen-nvidia:
## Checklist

### DIFF
--- a/services/alert/src/realtime/services/incident_service.py
+++ b/services/alert/src/realtime/services/incident_service.py
@@ -103,7 +103,6 @@ class IncidentService:
             extra={
                 "es_enabled": es_client is not None,
                 "index_base": self._index_base,
-                "consolidation_enabled": bool(self._consolidation.get("enabled", False)),
             },
         )
 
@@ -185,12 +184,7 @@ class IncidentService:
                 doc["_index"] = hit.get("_index")
                 incidents.append(doc)
 
-            do_consolidate = (
-                bool(self._consolidation.get("enabled", False))
-                if consolidate is None
-                else bool(consolidate)
-            )
-            items = self._consolidate(incidents) if do_consolidate else incidents
+            items = self._consolidate(incidents) if consolidate else incidents
 
             logger.info(
                 "Incidents query completed",
@@ -198,7 +192,7 @@ class IncidentService:
                     **ctx,
                     "returned": len(items),
                     "raw_hits": len(incidents),
-                    "consolidated": do_consolidate,
+                    "consolidated": bool(consolidate),
                     "total": total_count,
                     "duration_s": round(duration, 3),
                 },
--- a/services/alert/src/realtime/services/incident_service.py
+++ b/services/alert/src/realtime/services/incident_service.py
@@ -50,9 +50,6 @@ except ImportError:
 
 
 _EPOCH = datetime(1970, 1, 1, tzinfo=timezone.utc)
-# Two adjacent chunks may skip at most this many indices and still count as the
-# same continuous event (tolerates an occasional dropped/non-positive chunk).
-_MAX_CHUNK_SKIP = 2
 
 
 def _parse_ts(value) -> Optional[datetime]:
@@ -277,9 +274,10 @@ class IncidentService:
     def _is_continuous(current: List[dict], doc: dict, gap_seconds, max_duration) -> bool:
         """Whether ``doc`` extends the open event ``current``.
 
-        Continuation requires the event to stay within the duration cap and
-        either belong to the same caption session with consecutive chunk
-        indices, or fall within the inter-alert gap of the previous chunk.
+        Continuation requires the event to stay within the outer duration cap
+        and the next positive to arrive within ``max_inter_alert_gap_seconds``
+        of the previous chunk. The time gap is authoritative — there is no
+        per-session bypass. Unparseable timestamps end the current event.
         """
         first = current[0]
         prev = current[-1]
@@ -290,25 +288,11 @@ class IncidentService:
             if start and doc_end and (doc_end - start).total_seconds() > max_duration:
                 return False
 
-        prev_info = _info_of(prev)
-        doc_info = _info_of(doc)
-        prev_idx = _to_int(prev_info.get("chunkIdx"))
-        doc_idx = _to_int(doc_info.get("chunkIdx"))
-        same_session = bool(
-            prev_info.get("requestId")
-            and prev_info.get("requestId") == doc_info.get("requestId")
-            and prev_idx is not None
-            and doc_idx is not None
-            and 0 < (doc_idx - prev_idx) <= _MAX_CHUNK_SKIP
-        )
-
-        within_gap = False
         prev_end = _parse_ts(prev.get("end")) or _parse_ts(prev.get("timestamp"))
         doc_start = _parse_ts(doc.get("timestamp"))
-        if prev_end and doc_start:
-            within_gap = (doc_start - prev_end).total_seconds() <= gap_seconds
-
-        return same_session or within_gap
+        if prev_end is None or doc_start is None:
+            return False
+        return (doc_start - prev_end).total_seconds() <= gap_seconds
 
     @staticmethod
     def _build_event(chunks: List[dict], representative: str) -> dict:
--- a/services/alert/src/realtime/services/incident_service.py
+++ b/services/alert/src/realtime/services/incident_service.py
@@ -20,6 +20,7 @@ Service for querying incidents from Elasticsearch.
 
 import asyncio
 import copy
+import hashlib
 import logging
 import time
 from datetime import datetime, timezone
@@ -239,7 +240,7 @@ class IncidentService:
 
         Operates only on the documents already returned for this page; the
         underlying store is never modified and raw documents stay available
-        (``consolidate=false`` or the per-event ``info.rawIds``).
+        via a ``consolidate=false`` query.
         """
         cfg = self._consolidation
         gap_seconds = cfg.get("max_inter_alert_gap_seconds", 60)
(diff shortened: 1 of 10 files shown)


services/alert/src/web/api/realtime_routes.py
```diff
@@ -582,14 +585,23 @@ async def list_incidents(
         ge=0,
         description="Number of incidents to skip (for pagination)",
     ),
+    consolidate: Optional[bool] = Query(
```
dineshvenkat, line 610: The endpoint defaults to consolidated mode via the config block's `enabled: true`. The opt-in flag should flip the behavior the other way — keep today's raw chunk response as the default and let callers ask for the consolidated view explicitly. Otherwise every existing caller (UI, scripts, partners) changes shape silently on merge.

services/alert/src/realtime/services/incident_service.py
```diff
@@ -181,3 +235,132 @@ async def list_incidents(
...
+            start = _parse_ts(first.get("timestamp"))
+            doc_end = _parse_ts(doc.get("end")) or _parse_ts(doc.get("timestamp"))
+            if start and doc_end and (doc_end - start).total_seconds() > max_duration:
+                return False
+
+        prev_info = _info_of(prev)
+        doc_info = _info_of(doc)
+        prev_idx = _to_int(prev_info.get("chunkIdx"))
+        doc_idx = _to_int(doc_info.get("chunkIdx"))
+        same_session = bool(
```
dineshvenkat: `_is_continuous()` returns `same_session or within_gap`. When two chunks share an RT-VLM session and have consecutive `chunkIdx`, they merge regardless of the time gap between them. So a session that emits chunk 13 at 10:00 and chunk 14 at 10:15 still folds into one event. The unit test at `test/unit/api/test_incident_service.py:279` codifies that behavior as intended, which makes the time gap effectively non-binding for same-session chunks.

services/alert/src/realtime/services/incident_service.py
```diff
@@ -181,3 +235,132 @@ async def list_incidents(
...
+
+        merged_queries: list = []
+        for chunk in chunks:
+            llm = chunk.get("llm")
+            if isinstance(llm, dict) and isinstance(llm.get("queries"), list):
+                merged_queries.extend(llm["queries"])
+        if merged_queries and isinstance(event.get("llm"), dict):
+            event["llm"]["queries"] = merged_queries
+
+        return event
```
dineshvenkat, line 502: `_build_event` deep-copies the representative chunk wholesale. The consolidated event ends up carrying the chunk's top-level `Id` (the stable fingerprint the UI reads at `useAlerts.ts:151`) plus ES-internal `_id` and `_index`. So the event and one of its underlying raw chunks share the same identity — a UI that keys on `Id` will treat them as the same row. Same `_build_event` also stores `info.chunkCount` as a string ("5"), `info.isConsolidated` as a string ("true"), and `info.rawIds` as a comma-delimited string. Any consumer has to parse those, and `rawIds.split(",")` breaks if a chunk id ever contains a comma.

services/alert/src/realtime/services/incident_service.py
```diff
@@ -140,20 +185,29 @@ async def list_incidents(
...
+            items = self._consolidate(incidents) if do_consolidate else incidents
+
             logger.info(
                 "Incidents query completed",
                 extra={
                     **ctx,
-                    "returned": len(incidents),
+                    "returned": len(items),
+                    "raw_hits": len(incidents),
+                    "consolidated": do_consolidate,
```
dineshvenkat: The ES query applies `offset` and `limit` to raw chunks before the consolidator runs. An event whose chunks straddle a page boundary becomes two events — page N shows the head, page N+1 shows the tail as a separate event. Under realistic load (more than `limit` chunks in the requested window), the same physical incident shows up multiple times across adjacent pages, which is the duplication this feature is meant to remove.

services/alert/test/functional/p1/test_realtime_alerts/run.sh
```diff
@@ -582,6 +582,40 @@ else
...
+
+if [ "$RAW_CNT" = "3" ] && [ "$CONS_CNT" = "1" ] && [ "$CONS_TOTAL" = "3" ] && [ "$CONS_FLAG" = "true" ]; then
+    print_status "ok" "PASS: 3 raw chunks -> 1 consolidated event (total stays raw=3)"
+    ((PASSED++))
+else
+    print_status "fail" "FAIL: consolidate (raw=$RAW_CNT cons=$CONS_CNT total=$CONS_TOTAL flag=$CONS_FLAG)"
+    ((FAILED++))
+fi
+curl -s -X DELETE "${ES_HOST}/${CONSOL_IDX}" >/dev/null 2>&1 || true
+
```
dineshvenkat, line 621: Only one functional test is added (3 chunks → 1 event). The bound-breach cases (outer-window breach, continuity break), grouping-isolation cases (different alert types on same camera, different cameras on same alert type), multi-rule shared-stream isolation, the negative cases that should return 400 when the new flag is set without filters, and the mixed-source fixture that would prove verifier-path docs don't leak in — none of these are exercised. CI will go green while those paths stay unprotected.

services/alert/src/realtime/services/incident_service.py
```diff
@@ -181,3 +235,132 @@ async def list_incidents(
...
+                    current = [doc]
+            if current:
+                events.append(self._build_event(current, representative))
+
+        # Newest first, matching the raw query's sort order.
+        events.sort(
+            key=lambda e: _parse_ts(e.get("end")) or _parse_ts(e.get("timestamp")) or _EPOCH,
+            reverse=True,
+        )
+        return events
```
dineshvenkat, line 404: Events are sorted by `event.end` descending. The expected order is by event start descending — matching today's chunk-listing order in the existing query. With the current sort, a long-running event that started earlier but ended later ranks above a newer short event, so a fresh single-chunk alert can get buried under an older event that just happened to wrap up recently.

services/alert/src/realtime/services/incident_service.py
```diff
@@ -47,6 +48,37 @@
     INCIDENT_QUERY_FAILURES = None
 
 
+_EPOCH = datetime(1970, 1, 1, tzinfo=timezone.utc)
+# Two adjacent chunks may skip at most this many indices and still count as the
+# same continuous event (tolerates an occasional dropped/non-positive chunk).
+_MAX_CHUNK_SKIP = 2
```
dineshvenkat: `_MAX_CHUNK_SKIP = 2` is a module-level constant with no comment explaining the value and no config knob to change it. It silently widens the same-session merge window by allowing a chunkIdx gap of up to 2 to still count as continuous. An operator tuning `max_inter_alert_gap_seconds` would not realize that the same-session branch carries its own implicit skip tolerance.

services/alert/test/unit/api/test_incident_service.py
```diff
@@ -201,3 +201,263 @@ def test_without_client(self):
...
+        chunks = [
+            _chunk(idx=1),
+            _chunk(idx=2, start="2025-01-01T00:00:25.000Z", end="2025-01-01T00:00:55.000Z"),
+        ]
+        mock_es_client.client.search.return_value = {
+            "hits": {"total": {"value": 2}, "hits": [_as_hit(c) for c in chunks]}
+        }
+        svc = IncidentService(es_client=mock_es_client, consolidation={"enabled": False})
+        data, _ = await svc.list_incidents(consolidate=True)
+        assert data["count"] == 1
```
dineshvenkat, line 676: The unit tests cover the happy paths but miss the boundary-equality cases (a gap exactly equal to `max_inter_alert_gap_seconds`; a total span exactly equal to `max_event_duration_seconds`), the defensive cases for chunks missing `sensorId`/`category`/`timestamp`, and the bypass when `consolidate=false`. Those are the most likely places for an off-by-one or a malformed-input regression to slip through.

dineshvenkat:
services/alert/src/web/api/realtime_routes.py (general)

The route's `description=` for `/incidents` still describes only the original sensor/category/time-range filters, and `response_model=IncidentListResponse` is unchanged. The new `consolidate` query parameter and the consolidated event payload shape are not in the OpenAPI surface, so anyone reading the docs or generating a client SDK won't see them.

services/alert/src/web/api/realtime_routes.py
danh-nguyen-nvidia: Updated by using opt-in flag only for consolidation on 5f94a23f20e084a1cb6efff225c75d1be2ebf7dc

services/alert/src/realtime/services/incident_service.py
danh-nguyen-nvidia: For now each consolidated event will have a distinct ID with prefix "evt-". Keep chunkCount, isConsolidated and remove rawIds. Commit: f40ac4c65f6c1078461918ff59374e601a4b2c87

services/alert/src/realtime/services/incident_service.py
danh-nguyen-nvidia: Fixed in this commit a7fd11996bbffa3938838d70f274979661169e91

services/alert/src/realtime/services/incident_service.py
danh-nguyen-nvidia: Fixed in 03e6718277ce6bf262ac19bba456f8488d8cddbf

services/alert/src/realtime/services/incident_service.py
danh-nguyen-nvidia: Fixed in 03e6718277ce6bf262ac19bba456f8488d8cddbf

services/alert/src/realtime/services/incident_service.py
danh-nguyen-nvidia: Fix in 42f171d25810a84b989614a8bde3ec9a35488c28
Description: Scan the whole matched window and group before paginating so a consolidated event is never split across page boundaries: offset/limit now apply to the events and total is the event count.

services/alert/test/functional/p1/test_realtime_alerts/run.sh
danh-nguyen-nvidia: Added more functional tests in ec47d2a2d976da97f571c94eb47372a79888327d

services/alert/test/unit/api/test_incident_service.py
danh-nguyen-nvidia: Added unit tests in ec47d2a2d976da97f571c94eb47372a79888327d

PR #1162, opened 2026-07-01
Fix(agent): add runtime vlm verified selection for incident reports

liamy-nv:
## Description
- Fixes agent returning regular MDX incidents when user asks for verified / VLM verified incidents
- Adds runtime vlm_verified parameter through the incident lookup tool chain
- Updates top-agent prompt so vlm_verified is set from the current message only, not inherited from prior turns

## Problem
- get_incidents always used static MCP config (vlm_verified: false) to pick the Elasticsearch index
- Verified queries hit mdx-incidents-* instead of mdx-vlm-incidents-* if asking for verified incidents
- Follow-up queries like "list all incidents" could incorrectly inherit vlm_verified=true from earlier verified requests

## Checklist


liamy-nv:
/ok to test [2342bcc](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1162/commits/2342bcc7ef803735756455aad7c2962912db4fa6)

liamy-nv:
/ok to test [c2358a1](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1162/commits/c2358a1fd755e42ee980bf40d38e9e35d41e1e64)

deploy/docker/industry-profiles/warehouse-operations/warehouse-2d-app/deepstream/configs/ds-pgie-config.yml
```diff
@@ -49,7 +49,7 @@ property:
   net-scale-factor: 0.00392156862745098
   network-mode: 2
   network-type: 0
-  model-engine-file: /opt/storage/rtdetr_warehouse_v1.0.2.fp16.onnx_b3_gpu0_fp16.engine
+  model-engine-file: /opt/storage/rtdetr_warehouse_v1.0.2.fp16.onnx_b4_gpu0_fp16.engine
```
zac-wang-nv: Engine filename bumped `_b3_` → `_b4_` to match the new batch-size 4. Just confirming DeepStream regenerates this engine from `onnx-file` on first start (it will if the `_b4_` engine isn't present) rather than expecting a prebuilt `_b4_` artifact shipped with the profile — otherwise the PGIE fails to load at startup. If the engine is prebuilt/baked, make sure the b4 variant is the one shipped.

zac-wang-nv:
The core fix here is solid — `vlm_verified` is threaded cleanly from the agents down to the incident tools with a proper runtime-override-with-config-fallback (`effective_vlm_verified`), the always-`WithVLM` registration is safe (verdict filter only fires when `effective_vlm_verified` is true), and I confirmed the warehouse profile defaults `vlm_verified: false` so the prompt's "omit → regular incidents" behavior holds. One process nit: this commit also bundles unrelated warehouse pipeline tuning that isn't mentioned in the commit message or PR description — deepstream `batch-size` 3→4 (+ engine file `_b3_`→`_b4_`), pgie `pre-cluster-threshold` 0.5→0.6, VST `max_devices_supported` 16→49, and `nv_streamer_sync_file_count` 3→4. Those change pipeline capacity/detection behavior, so it'd help to call them out in the PR description (ideally split from the vlm_verified fix). Separately, most of the +400/−181 is JSON whitespace reformatting, which buries the ~6 real value changes and makes the diff hard to review — worth isolating formatting from semantic edits in future.

zac-wang-nv:
Note on [`services/agent/src/vss_agents/video_analytics/tools.py:802`](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/c2358a1fd755e42ee980bf40d38e9e35d41e1e64/services/agent/src/vss_agents/video_analytics/tools.py#L802) (leaving here rather than inline since that line isn't part of this diff):

Minor consistency note (non-blocking): `_analyze` still keys the index off `_config.vlm_verified` directly, while `get_incident`/`get_incidents` now honor the runtime override. This is fine for the warehouse profile — `analyze` isn't in the `video_analytics_mcp` include list and `AnalyzeInput` has no `vlm_verified` field — but if `analyze` ever gets exposed with the override elsewhere, this path would silently ignore it. Flagging so it isn't forgotten.

deploy/docker/industry-profiles/warehouse-operations/warehouse-2d-app/deepstream/configs/ds-pgie-config.yml
liamy-nv: Confirmed: the profile does not ship a prebuilt _b4_ engine. ds-pgie-config.yml sets both onnx-file and model-engine-file under /opt/storage ($VSS_DATA_DIR/models/mtmc/). Only the ONNX is downloaded (NGC mdx-mtmc / rtdetr_warehouse_v1.0.2.fp16.onnx)

liamy-nv: Pushed a commit to remove these changes unrelated to this PR

liamy-nv:
> Note on [`services/agent/src/vss_agents/video_analytics/tools.py:802`](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/c2358a1fd755e42ee980bf40d38e9e35d41e1e64/services/agent/src/vss_agents/video_analytics/tools.py#L802) (leaving here rather than inline since that line isn't part of this diff):
> 
> Minor consistency note (non-blocking): `_analyze` still keys the index off `_config.vlm_verified` directly, while `get_incident`/`get_incidents` now honor the runtime override. This is fine for the warehouse profile — `analyze` isn't in the `video_analytics_mcp` include list and `AnalyzeInput` has no `vlm_verified` field — but if `analyze` ever gets exposed with the override elsewhere, this path would silently ignore it. Flagging so it isn't forgotten.

ACK - will align _analyze with effective_vlm_verified if we add verified analyze support

liamy-nv:
/ok to test [b8c02f8](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1162/commits/b8c02f8aa0cea484b04668dd1dae3e6fb45598d7)

zac-wang-nv (dismissed):

zac-wang-nv (dismissed):

zac-wang-nv (approved):

zac-wang-nv (dismissed):

simings-nv:
Nice fix — the runtime override with config fallback is a clean pattern. 👍

One small thought: would it be worth adding a couple of unit tests around the runtime `vlm_verified` selection — e.g. asserting which index `get_incident`/`get_incidents` end up querying for input `True` / `False` / omitted? It'd help protect this behavior from regressing later.

PR #1170, opened 2026-07-02
refactor(search): NAT-free search_core library + unified vss-cli + updated vss-search-archive skill

ayyappa-dev:
## Description
Introduces `lib.search_core`: a self-contained, NAT-free video-search library
(embed / attribute / fusion / critic) with a single `vss-cli` entrypoint.

- Primitives orchestrate injected protocol clients; pure, unit-tested helpers
  for fusion math, ES filters, coercion, and result mapping.
- Library error hierarchy (InvalidInputError / IndexNotFoundError /
  BackendUnreachableError) with agent-friendly CLI exit codes; no framework
  exceptions leak across boundaries.
- Per-item resilience (one corrupt doc never sinks a search), field coercion,
  and scrub_log on data-controlled log values.
- Single CLI: `vss-cli search` (folds in the former `search-archive`) with
  --json/flags/--stream and json|jsonl|table output; `vss-search-archive`
  skill updated to match.
- Comprehensive unit/e2e tests (496) green; ruff + mypy clean.

## Checklist


services/agent/src/lib/search_core/primitives/_search_helpers.py
```diff
@@ -0,0 +1,898 @@
...
+        search_input=search_input,
+        embed_search=embed_search,
+        config=config,
+        attribute_search_fn=attribute_search_fn,
+        critic_agent=critic_agent,
+        behavior_es=behavior_es,
+    ):
+        if isinstance(update, SearchOutput):
+            return update
+    return SearchOutput(data=[])
```
prisrivastav-nv: We're returning `SearchOutput(data=[])` as the fallback return value if the generator yields nothing. In contrast, for the streaming path for the same use case in [search.py (L312)](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1170/changes?diff=unified#diff-a781b68f86477e22161c2ffe0b008184ab049d480b3b5d7569f8985abfc4f080R309), we're returning a terminal `ErrorEvent(error_code="NoFinalResult")`.

The two should ideally agree on the return value; the non-streaming wrapper here should raise an error instead of an empty success response, matching the streaming path.

services/agent/src/lib/search_core/primitives/embed_search.py
```diff
@@ -0,0 +1,233 @@
...
+        """Run the ES search, mapping a missing index to IndexNotFoundError.
+
+        Returns the raw ES response (an ``ObjectApiResponse``-like mapping, typed
+        ``Any`` to match the ``ElasticIndex`` protocol). Other transport/API
+        failures are already raised as ``BackendUnreachableError`` by the
+        ``ElasticClient`` wrapper.
+        """
+        try:
+            return await self._es.search(index=search_index, body=search_query)
+        except ESNotFoundError as e:
```
prisrivastav-nv: [`ElasticClient.search()`](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1170/changes?diff=unified#diff-104becfa933e5aefd1368a05317a814e42bc1e9473d3cb5ade48b608d09084b9R195) already catches ESNotFoundError and raises `IndexNotFoundError`; the `except` block here is unreachable/dead code, which is doing the same conversion again.

Also, we're catching the raw ES exception type directly, which violates the library's intent/convention that primitives should see the library error hierarchy, not the framework exception itself.

services/agent/src/lib/search_core/clients/rtvi_cv_embed.py
```diff
@@ -0,0 +1,111 @@
...
+                return cached
+            embedding = await self._fetch_text_embedding(text)
+            self._text_cache.put(text, embedding)
+            return embedding
+
+    async def _fetch_text_embedding(self, text: str) -> list[float]:
+        payload = {"text_input": text, "model": ""}
+        try:
+            timeout = httpx.Timeout(connect=30.0, read=120.0, write=120.0, pool=30.0)
+            async with httpx.AsyncClient(timeout=timeout) as client:
```
prisrivastav-nv: Critical - this method creates a new `httpx.AsyncClient` every time; it's in the critical search path and invoked for every attribute-search call, ie, _once per embed result per attribute_.

For reference, `CosmosEmbedClient` uses a lazy `self._client` pattern and keeps the connection pool alive across requests.

services/agent/src/lib/search_core/primitives/_search_helpers.py
```diff
@@ -0,0 +1,898 @@
...
+
+        # Critic verification.
+        #
+        # DELIBERATE DEVIATION from the hybrid error policy used elsewhere in
+        # this module: the embed / attribute / fusion paths re-raise any
+        # ``SearchError`` so callers get precise exit codes, but the critic block
+        # catches *everything* (including ``SearchError``) and soft-fails with a
+        # ``search_messages`` note. The critic is EXPERIMENTAL, additive
+        # verification layered on top of already-valid search results — a dead
+        # VLM/VST backend must NOT downgrade an otherwise-successful search into
```
prisrivastav-nv: The behavior observed in this code comment is true for VLM related runtime issues (dead backend).

But `ConfigurationError`, which is a subclass of `SearchError`, is also swallowed here and returns results without critic verification and without a distinct error message about misconfiguration for example invalid `vlm_base_url`. `ConfigurationError` represents a misconfiguration that must be allowed to propagate from the critic block (re-raise path alongside the deliberate soft-catch); or at minimum we should add an error log.

services/agent/src/lib/search_core/clients/elastic.py
```diff
@@ -0,0 +1,230 @@
...
+    Mirrors the lifecycle of services/agent/src/vss_agents/utils/es_client.py
+    (VSSESClient) but without coupling search_core to it.
+    """
+
+    # Keyed by (endpoint, event loop): an AsyncElasticsearch is bound to the
+    # loop it was created on (its httpx pool captures that loop). Reusing one
+    # across a fresh ``asyncio.run`` — a new loop — raises "event loop is
+    # closed". Keying by the running loop object (not its id, which CPython
+    # reuses after a closed loop is freed) gives each loop its own client and
+    # never returns a client bound to a different, already-closed loop.
```
prisrivastav-nv, line 94: The mitigation of the problem identified in the comments is incomplete: it prevents reuse of a wrong-loop client but does not clean up the stale entry.

In [`from_endpoint()`](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1170/changes#diff-104becfa933e5aefd1368a05317a814e42bc1e9473d3cb5ade48b608d09084b9R136), prune entries whose loop is closed (`loop.is_closed()`) before inserting the new one (at the top of the method).

prisrivastav-nv (changes requested):
Overall - requested a few changes (thread below). Additionally - 

1. Please address the [greptile review comment](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1170#discussion_r3528598129).

2. Packaging changes as discussed offline - either `lib.*` as slim distribution or two variants of `vss-agent` distribution with NAT and without NAT (with `lib` packages). cc: @zac-wang-nv

services/agent/src/lib/search_core/primitives/_search_helpers.py
```diff
@@ -0,0 +1,898 @@
...
+        seen: dict[str, AttributeSearchResult] = {}
+        for r in all_results:
+            key = _coerce_str(r.metadata.object_id)
+            if key not in seen or _coerce_float(r.metadata.behavior_score) > _coerce_float(
+                seen[key].metadata.behavior_score
+            ):
+                seen[key] = r
+        attr_results = sorted(seen.values(), key=lambda r: _coerce_float(r.metadata.behavior_score), reverse=True)[
+            :top_k
+        ]
```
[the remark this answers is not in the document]
zac-wang-nv: Verified fixed at head: `_search_helpers.py:456-458` now treats missing/`"unknown"` object ids as non-mergeable and keeps each row individually (`if not key or key == "unknown": unmergeable.append(r)`), merging only real keys by max `behavior_score` — consistent with `_attribute_helpers.deduplicate_by_object`. Resolving.

services/agent/src/lib/vst/client.py
```diff
@@ -0,0 +1,468 @@
...
+    aiohttp.ClientConnectionError,
+    aiohttp.ServerTimeoutError,
+    TimeoutError,
+)
+
+
+# ---------------------------------------------------------------------- types
+
+
+class VSTError(BackendUnreachableError):
```
zac-wang-nv, line 67: **(P2, breaks the PR's exit-code contract)** `VSTError` subclasses `lib._foundation.errors.BackendUnreachableError`, but the CLI catches `lib.search_core.errors.BackendUnreachableError` (`cli/search.py:65`, handler at `:1187`) — a *subclass* of the foundation one. So `VSTError` falls through to the generic `except Exception` → a dead VST during `--video-source` resolution (`_resolve_named_sources` → `get_name_to_stream_id_map`, `cli/search.py:951-1010`) exits **1 "unexpected error"** instead of the documented **3 "backend unreachable"**, which is exactly the case the exit-code contract exists for (agent branches 3=retry-backend). Same gap in the stream path: `Search.stream()`'s `except SearchError` misses `VSTError`, so `_STREAM_BACKEND_ERROR_CODES = frozenset({"VSTError"})` (`cli/search.py:132`) is dead code and the stream emits `UnexpectedError`. Since `lib.vst` can't import `search_core` (layering), the clean fix is at the top layer: catch `lib._foundation.errors.BackendUnreachableError` (the common base) in `_invoke` and in the stream error translation. A regression test that kills a fake VST and asserts exit 3 would lock the contract.

services/agent/src/lib/critic/critic.py
```diff
@@ -0,0 +1,327 @@
...
+                    if end_offset > clip_end_offset:
+                        end_offset = clip_end_offset
+                    vlm_response = await self._vlm.analyze(
+                        sensor_id=video.sensor_id,
+                        start_timestamp=str(start_offset),
+                        end_timestamp=str(end_offset),
+                        prompt=formatted_prompt,
+                        time_format="offset",
+                    )
+            except BackendUnreachableError as e:
```
zac-wang-nv, line 313: **(P2, resilience regression vs the pre-split code)** The per-video catch narrowed from `except Exception` (at `41980f9`) to `except BackendUnreachableError`, but the VST helpers this path depends on can still raise **raw framework exceptions**: in `get_video_clip_url`/`get_name_to_stream_id_map`/`get_streams_info` the `session.get` sits inside `with retry:` with `reraise=True`, so after 3 exhausted attempts the raw `aiohttp.ClientConnectionError`/`TimeoutError` re-raises unwrapped — escaping `asyncio.gather` in `CriticAgent.run()` and crashing the whole verification run instead of yielding per-video UNVERIFIED. This contradicts `vst/client.py`'s own docstring ("no raw aiohttp/stdlib exception leaks to callers"). Suggest converting retry-exhausted transport errors to `VSTError` at the helper boundary (fixes every consumer), which combined with Comment 1's base-class catch also keeps the critic degrading gracefully.

services/agent/src/lib/search_core/runtime.py
```diff
@@ -0,0 +1,632 @@
...
+            # VST
+            "vst_internal_url": vst_internal_url,
+            "vst_external_url": vst_external_url,
+            # Embed clients
+            "cosmos_embed_endpoint": cosmos_embed_endpoint,
+            "cosmos_embed_model": env.get("RTVI_EMBED_MODEL", "cosmos-embed1-448p"),
+            "rtvi_cv_endpoint": rtvi_cv_endpoint,
+            # Search orchestrator profile knobs — the whole point of this builder
+            "embed_confidence_threshold": search_cfg.get("embed_confidence_threshold", 0.1),
+            "default_max_results": search_cfg.get("default_max_results", 10),
```
zac-wang-nv, line 511: **(P2, silent behavior change)** The embed-search default cap silently dropped from 100 to 10. At `41980f9` the runtime had `embed_default_max_results = 100` ("a bare from_kwargs() runtime should not silently return only a handful of embed hits") and `from_config_file` read `functions.embed_search.default_max_results`. Now `EmbedSearch.from_runtime` reuses the orchestrator-level `default_max_results` (10), and `from_config_file` never reads the `functions.embed_search` block — yet the docstring at `runtime.py:392` still lists `functions.embed_search.default_max_results` as a profile setting this builder exists to reproduce. Deployed profiles that set it are silently ignored, and `vss-cli search embed` without `top_k` returns 10 hits instead of 100. Either read the embed block back in, or document the intentional change and fix the docstring.

services/agent/src/lib/vst/client.py
```diff
@@ -0,0 +1,468 @@
...
+                        start = timeline_list[0].get("startTime")
+                        end = timeline_list[0].get("endTime")
+                        start_dt = iso8601_to_datetime(start)
+                        end_dt = iso8601_to_datetime(end)
+                        if (end_dt - start_dt).total_seconds() < 1:
+                            raise VSTError(f"Timeline duration is too short for stream {stream_id}")
+                        return start, end
+                except VSTError:
+                    raise
+                except Exception as e:
```
zac-wang-nv: **(P2, retry defeated)** `get_timeline` still defeats its own retry: unlike the three sibling helpers (which keep `session.get` outside the parse-only `try`), here the network call at line 343 sits inside the broad `try` whose `except Exception` converts transient `aiohttp.ClientConnectionError`/`TimeoutError` into `VSTError` **before** tenacity's `retry_if_exception_type` predicate sees them — and `VSTError` is deliberately non-retryable (`client.py:51-55`). A single TCP blip fails the timeline path immediately instead of retrying up to 3×, contradicting the module docstring ("retries are limited to connection/timeout errors"). Move the `try` to start after `response.text()`, matching `get_name_to_stream_id_map`.

services/agent/src/lib/search_core/primitives/_attribute_helpers.py
```diff
@@ -0,0 +1,1073 @@
...
+        existing_start = existing_result.metadata.start_time or existing_source.get("timestamp")
+        existing_end = existing_result.metadata.end_time or existing_source.get("end")
+        new_start = new_source.get("timestamp")
+        new_end = new_source.get("end")
+
+        earliest_start = _earlier(new_start, existing_start)
+        latest_end = _later(new_end, existing_end)
+        if earliest_start != existing_start or latest_end != existing_end:
+            merge_count += 1
+            existing_result.metadata.start_time = earliest_start
```
zac-wang-nv, line 314: **(P2, ranking quality — your call on intended semantics)** In append/multi-attribute mode, `deduplicate_by_object` merges a duplicate `(sensor_id, object_id)` by widening the time range only — it keeps the first-seen row's `behavior_score` and discards a higher-scoring later match, then the append path ranks and top-k slices by that stale score. Concrete case: object 42 matches "person" @0.4 (first attribute) and "walking" @0.9 (later); dedup keeps 0.4, so with `top_k=1` and another object @0.5, object 42 is truncated out despite having the best true relevance. This is inconsistent with the object-ids path (`_search_helpers.py:459-462`), which deliberately keeps `max(behavior_score)`. If the intent is "best relevance per object," this path should keep the max too.

skills/vss-search-archive/SKILL.md
```diff
@@ -150,149 +196,259 @@ curl -s -X POST "http://${HOST_IP}:8000/api/v1/rtsp-streams/add" \
...
+     the user meant one of them, wants to ingest the named source using the
+     agent-backed workflow above, or wants an unrestricted search.
+   - If several sources match, stop and ask the user to choose.
+   - Never remove a requested source constraint, substitute a different video,
+     or run a broad search as a probe.
+   - If the source is absent, do not test or invoke the search CLI; clarification
+     is the final action for that request.
+
+3. Decompose the request into explicit fields using
+   [Query decomposition](references/query_decomposition.md). The CLI does not
```
zac-wang-nv, line 311: **(P2, skill navigation)** 4 of the 5 `references/*.md` files are never linked from SKILL.md — only `query_decomposition.md` is. `cli_usage.md`, `deployment_resolution.md`, `troubleshooting.md`, and `discovery_modes.md` are unreachable during skill use (the latter two are linked only from skill-card.md, which is the compliance card, not agent-loaded content). Since references are loaded just-in-time by following links from SKILL.md, orphaned files are effectively dead weight. Suggest linking each from the relevant section: Host CLI → cli_usage.md, deployment prerequisite → deployment_resolution.md, a Troubleshooting cue → troubleshooting.md, query examples → discovery_modes.md. (This also matches 3 of the skills-review bot's high findings.)

skills/vss-search-archive/SKILL.md
```diff
@@ -1,144 +1,190 @@
...
----
+`$HOME/video-search-and-summarization` is only the Harbor/default workspace;
+set `VSS_REPO_ROOT` for a checkout elsewhere. If validation or the command
+fails, report the error and stop. Do not substitute an agent runtime route or
+manually call search backends.
 
 ## Deployment prerequisite
 
-This skill requires the VSS **search** profile running on the host at `$HOST_IP`. Before any request:
+This skill requires the VSS **search** profile. Resolve `AGENT_URL` and `ES_URL`
```
zac-wang-nv, line 54: **(P2, skill executability)** `${AGENT_URL}` and `${ES_URL}` gate the mandatory health probe (lines 77-78) and every ingestion/deletion recipe, but the skill never says *how* to resolve them — the merge-base version had concrete endpoints (`http://${HOST_IP}:8000`, `:9200`), and `references/deployment_resolution.md` covers only the CLI's internal discovery, not the curl flows. An agent can't execute step 1 of the deployment prerequisite deterministically. Add a concrete resolution recipe (e.g. read the selected profile's published ports / ingress) or link a reference that contains one. (Longer-term this becomes moot if endpoints ride the HAProxy front-end — see the design comment.)

zac-wang-nv:
Re-reviewed the full PR at head `599fa37ff` (previous pass was at `41980f9`; the `de6dc734→599fa37f` rebase only touched `cli/deployment.py` env layering + deployment tests + skill docs in PR content — all code citations below re-verified). The restructure is genuinely good: clean layering (`_foundation` → `vst`/`vlm` → `critic` → `search_core` → `cli`, grep-verified no back-edges), the standalone packaging story works end-to-end (built the wheel locally: the `src/lib` symlink dereferences correctly in sdist/wheel, `lib.knowledge` is excluded exactly as the `lib/__init__.py` docstring promises, and the CI smoke test enforces no NAT/torch leakage), and the boundary tests assert the properties that matter. The Sonar consolidation introduces no scan gap, and the skill-eval adapter change fixes a real prior bug (`exit 0` swallowing the judge verdict). This also settles the packaging question from [the review](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1170#pullrequestreview-4642726488) as the slim-distribution option — the wheel is genuinely NAT-free.

**@prisrivastav-nv — all five of your comments plus the Greptile P1 you flagged are addressed at head** (verified in code, for your re-review):
- Greptile P1 unknown-id dedup → `_search_helpers.py:456-458` keeps sentinel rows individually
- empty-`SearchOutput` fallback → now raises `NoFinalResultError` (`_search_helpers.py:713`, `errors.py:88`), mirroring the stream's `NoFinalResult` terminal event
- dead ES `except` in `embed_search.py` → removed (conversion lives only in `ElasticClient`)
- per-call `httpx.AsyncClient` in `rtvi_cv_embed.py` → lazy pooled `self._client` (`:49,63-67,88`), matching `CosmosEmbedClient`
- `ConfigurationError` swallowed in the critic soft-catch → moot: commit `490996041` decoupled the critic from retrieval entirely; the catch block no longer exists
- `elastic.py` stale registry entries → `from_endpoint` now prunes closed-loop entries and best-effort closes their transports (`:153-164`)

**CI (must fix to merge):**
- ❌ `signature-check` — `skill.oms.sig` is stale: 6 of 7 signed file digests no longer match and the 3 new `references/*.md` aren't in the manifest. Needs a re-signing pass with the NVIDIA skills key — do it **after** the doc edits from the inline comments land, or it'll need signing twice.
- ❌ `Check vss-agent tag source` + `Trigger Downstream Pipeline` — standard for agent-source PRs: build+promote a new `vss-agent` image and repin the tag.

**P2s** — see inline: (1) `VSTError` misses the CLI's `BackendUnreachableError` handler → exit 1 instead of 3 (undercuts the PR's headline exit-code contract; stream path has the same hole, leaving `_STREAM_BACKEND_ERROR_CODES` dead), (2) critic crashes on retry-exhausted raw `aiohttp` errors instead of per-video UNVERIFIED, (3) embed default cap silently 100→10 with `functions.embed_search.default_max_results` ignored, (4) `get_timeline` retry defeated, (5) attribute-dedup keeps first-seen score.

**P3 (nits, non-blocking):**
- `embed_translation.py:64` — non-string JSON array elements are stringified (`null`→`"None"`, `true`→`"True"`) into ES source-name filters instead of dropped
- `_foundation/sanitize.py:37` — `scrub_log` keeps U+0085/U+2028/U+2029/U+007F, which several log viewers treat as line terminators (CWE-117 defense is CR/LF-only)
- `services/agent/pyproject.toml` — agent-side `[project.dependencies]` still misses direct `httpx`/`pyyaml` declarations for wheel-shipped lib code (satisfied only transitively via NAT); the vss-cli pyproject declares both correctly
- `cli/search.py` — no `BrokenPipeError` handling: `vss-cli search --output jsonl | head` prints an "unexpected error" + exit 1, undercutting the jsonl piping story
- `critic/critic.py:189-201` — comment says an explicit `"confirmed"` is trusted "even when a stray criterion parses False", but the code now demotes it to UNVERIFIED in that case; update the comment (or the behavior, if unintentional)
- `lib/vlm/__init__.py:5` — eager `from .openai import ...` makes a bare `import lib.vlm` load `cv2`+`httpx`, contradicting the lazy-import discipline used everywhere else (use module `__getattr__` like `lib.search_core`)
- `.github/workflows/ci.yml:325,366` — duplicate "Job 5" banner comments after inserting the vss-cli job
- `SKILL.md:93` — editing artifact: dropped "+" ("the VIOS upload RTVI-CV registration")
- `SKILL.md` — `--pretty` and `--include-embedding` (real flags, `cli/search.py:309,319`) still undocumented
- `vss-cli/src/lib` symlink — fine on Linux/macOS and in CI (verified), but a Windows checkout with `core.symlinks=false` materializes it as a text file and the build breaks; worth a one-line note in the vss-cli README if Windows dev matters

zac-wang-nv:
Stepping back from the line-level comments: the library is bigger than its job, and I'd like us to simplify before this becomes the foundation other teams extend. ~10k lines of runtime Python for "deploy-aware video search" — several layers can come out without losing anything the PR promises:

1. **Endpoint discovery should be the reverse proxy's job, not the CLI's.** `cli/deployment.py` is 808 lines of docker/k8s discovery + port-forward lifecycle, re-deriving per-service endpoints from deployment state. The stack already fronts services with HAProxy (`deploy/docker/services/infra/haproxy/haproxy.cfg.template`): `/api` → agent, `/vst` → VST ingress. The CLI should take a single `--base-url` (HAProxy origin / k8s ingress) and use fixed proxy paths. The proxy doesn't yet route Elasticsearch or the Cosmos/RTVI embed NIMs — adding those three backends to the template (+ helm ingress parity) is a far smaller, mostly-static change than maintaining discovery code, and it makes the CLI behave identically local vs remote with no port-forwards. (Exposing raw ES through the public front-end deserves a conscious security call — method/path restrictions or an internal-only frontend.)
2. **Collapse the dual error hierarchy.** `lib._foundation.errors` vs same-named subclasses in `lib.search_core.errors` isn't just extra code — it's the root cause of two P2s in this review (Comments 1–2). One hierarchy, fewer bugs.
3. **Drop the `ElasticClient` loop-keyed registry/refcount machinery** for the CLI's one-shot lifecycle — create the client per run, close it. If agent integration later needs pooling, add it in that PR where it can be justified.
4. **Commit to or drop the protocol/DI layer.** `clients/protocols.py` claims to be the only import surface for primitives, but `_search_helpers.py` imports `ElasticClient` directly and `_attribute_helpers.py` bypasses the VST protocol for `lib.vst` free functions. Half-applied DI is the worst of both — pick one.
5. **Shrink `runtime.py`** (632 lines, largely a NAT-config parity builder). With proxy-based endpoints (#1) most of it reduces to a handful of env vars/flags; the profile-parity question belongs to the deferred agent-integration PR.

**Concrete target CLI shape** (where I'd like this to converge — three commands, pipe-composable):
```
vss-cli search "query"                      # default: embed + attribute legs, fused, JSONL out
vss-cli search "query" -m embed             # single method (-m repeatable: -m embed -m attr → fuse)
vss-cli search "query" -f camera=cam1 -f class=forklift   # generic data filters → ES term/range clauses
... -m embed | ... -m attr | vss-cli fusion # fusion as a pure stdin-JSONL stage (RRF is already pure math in _fusion.py)
vss-cli search -m embed --list-filters      # discovery: allowlisted fields+types from that method's index _mapping
vss-cli search -m attr --list-filters class # + sample real values via terms agg ("forklift", "person", …)
```
For scale: `search` currently accepts **50 flags** — 13 endpoint/deployment wiring (absorbed by `--base-url` via #1), 6 index-selection (defaultable by convention), ~6 data filters (absorbed by `-f`), 7 fusion/ranking knobs (move onto `vss-cli fusion`), plus query/output/behavior flags. The reshape lands `search` at roughly 10. Generic `-f field=value` replaces the bespoke filter flags and absorbs new metadata fields with zero new flags (validated against a known-field registry; behavior knobs like top-k stay as flags). **Requirement: filters must be discoverable from the CLI itself, per method** — `search -m <method> --list-filters` reads the live `_mapping` of the index that method actually queries, intersected with a curated allowlist (internal ES fields don't leak); `--list-filters <field>` additionally samples real values via a terms aggregation. Both honor `--output json`, so a host agent's loop is: list filters → compose `-f` pairs → search. Echo the allowlisted registry in `search --help` for humans. `vss-cli fusion` reading tagged JSONL records decouples fusion from orchestration — most of the `search run` orchestrator layer evaporates. Two boundaries to preserve: query decomposition stays host-agent-owned (the attribute leg consumes caller-supplied phrases; the CLI never calls an LLM — the current PR gets this right), and the one-shot `search "query"` convenience path should remain for agents that prefer one process over shell pipelines (sharing the same pure fusion function). Note pipes becoming first-class makes the BrokenPipeError nit above worth fixing properly.

Rough impact: ~1.5–2k lines and two bug classes removed, plus a materially smaller surface for whoever extends search next. #2 I'd ask for in this PR (it's implicated in real bugs); #1/#3–5 and the CLI reshape can be a fast-follow if we agree on direction — #1 touches deploy configs beyond this PR's scope. Happy to sync offline. cc @prisrivastav-nv (fits the packaging/distribution discussion).

zac-wang-nv:
**Search-eval diagnosis: screenshot URLs are structurally broken for file-ingested sources — fix in draft PR #1323**

After repairing the eval infrastructure (runner SSH config + Skybridge link migration for the `vss-eval-*` fleet — the `*.brevlab.com` per-port scheme is deprecated/NXDOMAIN), the `vss-search-archive · search · RTXPRO6000BW` leg now runs end-to-end: **step-1 and step-2 reward=1.0** including the public-origin check. The one remaining failure (step-3, 5/6) is a product bug, not infra:

- Search hits index file-ingested sources on the synthetic epoch (`startTime=2025-01-01T00:01:00Z`), but VST anchors its replay timeline at ingest wall-clock (`/storage/timelines` → `2026-07-18T04:15:21→04:18:51`), so every screenshot URL gets `500 VMSInternalError: no valid stream found for given timestamps`.
- The old NAT `vst.snapshot` tool translated via `get_timeline`; the NAT-free direct-URL path in `lib.search_core` skips that step.

Draft fix: #1323 (timeline mapping at all three screenshot build sites, best-effort, with unit tests). Happy to hand it over or iterate — details in the PR description.

zac-wang-nv (approved):
LGTM

PR #1176, opened 2026-07-02
SOP ingestion profile for vision agent skill

danh-nguyen-nvidia:
## Checklist


zac-wang-nv:
given deploy-sop and build-sop are all consumed by build-vision-agent, can you put them under,  `build-vsion-agent/references/services/sop/*` ?

skills/vss-build-vision-agent/references/services/sop.md
```diff
@@ -0,0 +1,170 @@
...
+curl -fsSL $REPO/$BASE/patches/tools.py                 -o /tmp/sop-release-tools.py
+curl -fsSL $REPO/$BASE/configs/va_mcp_server_config.yml -o /tmp/sop-release-va-mcp.yml
+```
+
+**Adapt to the running image** — do NOT mount the 3.1 `tools.py` verbatim (it would regress
+the image's other `video_analytics` tools and targets the old package layout):
+
+1. **Resolve the package + site-packages path** on the image (see `sop/deploy-sop-report.md`
+   § Deploy Steps). The current base uses `agent/video_analytics/` (an earlier layout used
+   `vss_agents/`).
```
zac-wang-nv: this was again updated back to vss_agents, 

please check and sync with latest from develop.

sorry for the back and forth

skills/vss-build-vision-agent/references/services/sop/integrate-sop-report.md
zac-wang-nv, line 1: do we still need this reference? it seems most of the contents are already in sop.md

skills/vss-build-vision-agent/references/services/sop/integrate-ds-sop.md
```diff
@@ -0,0 +1,164 @@
...
+
+Use this service when the workflow requires **SOP compliance monitoring of a procedural task** (e.g. assembly / installation steps) on a live camera or stored video — structured, deterministic "did the operator perform step N, in order" events, as opposed to free-form dense captions (RT-VLM's job). DS-SOP occupies the same perception/inference slot as RT-VLM or RT-CV, but bundles a CV action model + a VLM in one DeepStream container **and (unlike RT-VLM) produces an annotated RTSP output VIOS records**. The image is `ds-sop:1.0.0`, built via the `vss-build-ds-sop` skill.
```
zac-wang-nv: vss-build-ds-sop skill is absorbed inside build-vision-agent, this could be a reference to build-ds-sop.md

skills/vss-build-vision-agent/references/services/sop/integrate-ds-sop.md
```diff
@@ -0,0 +1,164 @@
...
+- **DS-SOP → VIOS wiring is deploy-time, not composed.** build-vision-agent composes DS-SOP + VIOS into one profile but does **not** wire the video flow (it never does — even RT-VLM's stream registration is a post-boot API call). So registering DS-SOP's `:8554/ds-out` output as a VIOS sensor (`POST /vst/api/v1/sensor/add`) is a **mandatory deploy-time step** build-vision-agent's deploy flow (or the operator) performs — exactly like the JSON Logstash pipeline. See `deploy-ds-sop.md`.
+- **Single perception per profile.** DS-SOP and RT-VLM both target `mdx-vlm-captions` and both want the GPU — select exactly one per profile.
+- **Models + configs are host-staged, not in the image** — `/opt/models/...` (DDM + cosmos-reason) and `/opt/sop/configs/...` must exist before bring-up (see `deploy-ds-sop.md`).
+
+## Scope notes
+
+- **Source:** built from `microservices/sop-inference-bp/` in `NVIDIA/sop-monitoring-blueprints` (branch `main`), which ships the `:8554` annotated RTSP output the DS-SOP→VIOS flow uses. See the `vss-build-ds-sop` skill.
```
zac-wang-nv: same here

skills/vss-build-vision-agent/references/services/sop/deploy-sop-report.md
zac-wang-nv, line 1: do we still need this file? it seems most of its contents are covered by sop.md

zac-wang-nv (dismissed):
LGTM

skills/vss-build-vision-agent/references/services/sop/build-ds-sop.md
```diff
@@ -0,0 +1,124 @@
...
+- Verify: `docker images ds-sop:1.0.0 --format '{{.Repository}}:{{.Tag}} {{.Size}}'` (~50 GB).
+
+## Step 2 — Standalone smoke test (NO models)
+`API_DUMMY_TEST=true` skips `SOPProcessManager`/model load and boots only the API server —
+ideal to prove the image is valid (all heavy imports OK + server serves) without any model.
+```bash
+docker run --rm -d --gpus all --network host \
+  -e API_DUMMY_TEST=true -e API_SERVER_PORT=8300 \
+  --name dssop-smoke ds-sop:1.0.0
+# wait ~30–60 s for the heavy imports (torch / vLLM / DeepStream / GstRtspServer), then:
```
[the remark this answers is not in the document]
danh-nguyen-nvidia: Fix on 9ae0e750a4a8edaa67dd682b9495c71dbba2ae73

skills/vss-build-vision-agent/references/services/sop/deploy-sop-report.md
```diff
@@ -0,0 +1,89 @@
...
+**3. `generated.env`** — union the running stack's `generated.env` with
+`sop-report.env.fragment`, **append the build's invented flag** to `COMPOSE_PROFILES`
+(keep DS-SOP/ELK flags). No credentials needed.
+
+**4. Dry-run + normalize** (apply the patch overlay as an extra `-f`):
+```bash
+docker compose --env-file <generated.env> -f <BUILD_DIR>/compose.yml \
+  -f $DST/sop-report-override.yml config > resolved.sop-report.yml
+grep -c 'sop-report/video_analytics' resolved.sop-report.yml   # expect 2 (patch mounts on vss-va-mcp)
+uv run skills/vss-deploy-profile/scripts/normalize_resolved_yml.py resolved.sop-report.yml
```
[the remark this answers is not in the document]
danh-nguyen-nvidia: Fix on 9ae0e750a4a8edaa67dd682b9495c71dbba2ae73

skills/vss-build-vision-agent/references/services/sop/sop-report/patches/video_analytics/sop_tools.py
```diff
@@ -0,0 +1,774 @@
...
+    if sensor_id:
+        query["query"]["bool"]["must"].append({"wildcard": {"sensor_id.keyword": f"*{sensor_id}*"}})
+    if start_time:
+        query["query"]["bool"]["must"].append({"range": {"@timestamp": {"gte": start_time}}})
+    if end_time:
+        query["query"]["bool"]["must"].append({"range": {"@timestamp": {"lte": end_time}}})
+
+    return query
+
+
```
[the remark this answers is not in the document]
danh-nguyen-nvidia: This file is removed so I will close the conversation

skills/vss-generate-video-report/SKILL.md
```diff
@@ -65,18 +68,19 @@ Never route reports through VSS-agent `POST /generate`.
 
 **Mode A** needs the VSS **base** profile (VST + VLM NIM).
 **Mode B** needs the VSS **alerts** profile (VA-MCP + Elasticsearch).
+**Mode C** needs a **VA-MCP that exposes the SOP tools** (`get_sop_*`) over Elasticsearch `mdx-vlm-captions-*` — deployed by the SOP profile (compose via `/vss-build-vision-agent`; see its `references/services/sop/deploy-sop-report.md`).
```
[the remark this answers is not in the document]
danh-nguyen-nvidia: Fix on 9ae0e750a4a8edaa67dd682b9495c71dbba2ae73

zac-wang-nv (dismissed):
LGTM

danh-nguyen-nvidia:
/ok to test [ccb3551](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1176/commits/ccb3551789d71629a34707c91cdbbad6e006e222)

skills/vss-build-vision-agent/references/services/sop.md
```diff
@@ -0,0 +1,210 @@
...
+RTSP_PORT=8554
+SW_ENCODER=true
+VLLM_GPU_MEMORY_UTILIZATION=0.6            # 0.3 is fine on ≥80 GB GPUs
+
+# --- SOP report tools (vss-va-mcp patch, see § Patch specifics) ---
+# VSS_VA_MCP_CONFIG_FILE → the SOP va_mcp config staged into the container (selects
+#   get_sop_*). VSS_AGENT_SITE_PACKAGES → mount root; verify vs the ghcr.io image.
+VSS_VA_MCP_CONFIG_FILE=<staged: adapted from the downloaded SOP release VA-MCP config — see § Patch specifics>
+VSS_AGENT_SITE_PACKAGES=/vss-agent/.venv/lib/python3.13/site-packages
+VSS_AGENT_PKG=agent                 # resolved from the image (the pkg that holds video_analytics) — see § Patch specifics
```
zac-wang-nv: `agent` is the wrong package name for this tree, and the three places this PR states it disagree with each other:

- here (and line 105: *"`agent` on develop-latest; `vss_agents` ships only `api`"*) → `agent`
- `sop/sop-report/sop-report-override.yml:27` → *"`vss_agents` on develop, `agent` on some builds"*
- `eval/profile_sop_1_compliance_monitoring.json:64` → *"`vss_agents` on develop, `agent` on some builds"*

On `develop` there is no `agent` package at all. `services/agent/packages/` contains exactly `vss_agents`, `vss_cli`, `vss_core`, and `video_analytics` lives at `services/agent/packages/vss_agents/src/vss_agents/video_analytics/` — with the `es_client.py` / `tools.py` / `utils.py` this section describes. So line 105's parenthetical is inverted: `vss_agents` is the package that holds `video_analytics`.

This matters beyond wording. `sop-report-override.yml` mounts to `${VSS_AGENT_SITE_PACKAGES}/${VSS_AGENT_PKG}/video_analytics/tools.py`; with `VSS_AGENT_PKG=agent` Docker happily creates the missing path and bind-mounts the patch into a package Python never imports — `get_sop_*` silently never registers, and the § Patch specifics step 6 verify (`import ${VSS_AGENT_PKG}.video_analytics.sop_tools`) fails with the value this file hands you.

The probe in step 1 is correct and would resolve `vss_agents`; it's the hardcoded example and the parentheticals that are wrong. Suggest flipping both to `vss_agents` and making the override/eval wording match. (This is the same issue raised earlier in the thread on this file — it looks like only one of the three sites got updated.)

skills/vss-build-vision-agent/references/services/sop.md
```diff
@@ -0,0 +1,210 @@
...
+  `vss-va-mcp`); they add only the `get_sop_*` patch mount, no new service.
+
+## Delta composition
+
+SOP is not a Foundation; compose it as a **delta** off the closest Foundation —
+`alerts` in `2d_vlm` mode (it already carries `vss-va-mcp`, Elasticsearch, Kafka,
+Redis, VIOS, and a VLM-perception slot). Against alerts `COMPOSE_PROFILES_VLM`:
+
+- **add** `ds-sop` — a genuinely new service → emit `patches/ds-sop.yml` from the
+  DS-SOP Compose block in `sop/deploy-ds-sop.md § Example Compose Snippet`; and
```
zac-wang-nv: Wrong file. `## Example Compose Snippet` exists only in `sop/integrate-ds-sop.md:75` — `deploy-ds-sop.md` has no such section. This is the instruction the agent follows to emit `patches/ds-sop.yml`, so it sends it to a file that does not contain the block it needs.

`integrate-ds-sop.md:154` already refers to the section correctly ("§ Example Compose Snippet above"), so only this pointer needs fixing.

skills/vss-build-vision-agent/references/services/sop/deploy-ds-sop.md
```diff
@@ -0,0 +1,124 @@
...
+  cp <sop-ref-bundle>/sop-vlm-captions-json-logstash.conf \
+     <BUILD_DIR>/patched/services/infra/elk/logstash/pipelines/kafka/
+  # 2. register a SEPARATE pipeline-id in the build-local pipelines-kafka.yml (do NOT merge into
+  #    mdx-lvs). Guard the append so a re-run does not duplicate the entry (a duplicate pipeline.id
+  #    makes Logstash refuse to start):
+  PK=<BUILD_DIR>/patched/services/infra/elk/logstash/configs/pipelines-kafka.yml
+  grep -q 'sop-vlm-captions-json' "$PK" || cat >> "$PK" <<'YML'
+  - pipeline.id: sop-vlm-captions-json
+    path.config: "/usr/share/logstash/pipelines/sop-vlm-captions-json-logstash.conf"
+  YML
```
zac-wang-nv: This heredoc never terminates. `<<'YML'` requires the delimiter at column 0; `YML` here is indented two spaces (the block is nested under a list item, so the indent travels with any copy-paste). Switching to `<<-` would not help either — `<<-` strips leading **tabs**, not spaces.

Result when run: the shell keeps consuming until EOF, so step 3's `docker restart logstash` and the verify `curl` get appended into `pipelines-kafka.yml` instead of executing, and the operator sees no error until Logstash won't start.

Second issue in the same block: even terminated, the appended lines carry the 2-space indent, while the existing entries in `deploy/docker/services/infra/elk/logstash/configs/pipelines-kafka.yml` are at column 0 (`- pipeline.id: mdx-kafka`). Mixing those breaks the YAML sequence.

Dedenting the heredoc body and terminator to column 0 fixes both.

skills/vss-build-vision-agent/references/services/sop/sop-report/sop-report-override.yml
```diff
@@ -0,0 +1,34 @@
...
+#   docker compose --env-file <generated.env> \
+#     -f <BUILD_DIR>/compose.yml \
+#     -f <staged>/sop-report-override.yml \
+#     up -d vss-va-mcp
+# (target the service directly — never invent a COMPOSE_PROFILES entry)
+#
+# The patch adds the SOP tools (get_sop_status / get_sop_report /
+# get_sop_as_incidents / get_sop_as_incident) to the `video_analytics` group; they
+# only register when the va_mcp config's `include:` lists them (the adapted VA-MCP
+# config, from the downloaded SOP release config — see ../sop.md § Patch specifics).
```
zac-wang-nv: `../sop.md` is a dead link from here. This file sits at `references/services/sop/sop-report/`, so `../sop.md` resolves to `references/services/sop/sop.md`, which doesn't exist; the target is `references/services/sop.md`, i.e. `../../sop.md`. Same on lines 21 and 28.

Worth noting the sibling files get this right by accident of depth — `integrate-ds-sop.md` is one level up, so its `../sop.md` (lines 17, 73, 153, 162) is correct. Only the `sop-report/` subdirectory needs `../../`.

skills/vss-generate-video-report/SKILL.md
```diff
@@ -337,19 +346,58 @@ If `get_incidents` returns zero results, STOP and return exactly a one-line empt
...
+
+### Step 3 — Fill the SOP Compliance Report template
+
+Copy [`assets/sop-compliance-report.md`](assets/sop-compliance-report.md), fill every placeholder from the Step 2 result (message count, compliance status, cycle counts, the missing / mis-ordered step tables, actions observed), and return the rendered markdown. For placeholders `get_sop_report` does not carry: generate `{report_id}` + `{report_date}`, set `{agent_version}` to `vss-generate-video-report (Mode C)`, and set `{video_analysis_details}` / `{snapshot_image}` to `N/A` (Mode C runs no report-time VLM and fetches no media). Fill `{notes}` with the data provenance and snapshot caveats (source/scope, the bounded `end_time` used, the doc count vs the 1000-doc `get_sop_report` cap, and that a live stream never reaches EOS so `final_*` counts stay 0 and every violation is per-chunk); fill `{recommendations}` with the compliance interpretation (recurring missing / mis-ordered steps and whether they reflect the source clip rather than an operator fault). Keep the source asset unchanged; never leave a placeholder, and never include template instructions in a filled cell.
```
zac-wang-nv: `{total_actions}` ("Total Actions Recorded" in `assets/sop-compliance-report.md`) has no documented source. Step 2 describes `actions_observed` as carrying *"unique actions, latest action"* — no total — and `{total_actions}` isn't in this sentence's list of placeholders the tool doesn't carry.

That leaves two instructions in direct conflict: *"never leave a placeholder"* here, versus *"Reproduce the tool's numbers ... verbatim ... never paraphrase, rename, or invent them"* at line 376. An agent hitting this will invent a count.

Either add `{total_actions}` to the not-carried list with an explicit rule (derive from `report_summary.total_messages`? set `N/A`?), or document the response field it comes from.

Minor, same paragraph: `final_*` is referenced in the `{notes}` guidance but never defined anywhere in the skill.

skills/vss-generate-video-report/SKILL.md
```diff
@@ -337,19 +346,58 @@ If `get_incidents` returns zero results, STOP and return exactly a one-line empt
...
 ---
 
+## Mode C — SOP compliance report
+
+Use for "generate an SOP compliance report" over a sensor + time range. Data comes from the SOP tools on VA-MCP (added by the SOP profile); this skill aggregates and renders the template itself.
+
+### Step 1 — Resolve the sensor + time range
+
+- Capture the named sensor as `sensor_id`. `start_time` / `end_time` are ISO 8601 UTC (`YYYY-MM-DDTHH:MM:SS.sssZ`); resolve relative phrases ("last hour", "today") against the host clock.
+- Confirm the SOP tools are present (once): `/vss-query-analytics` `tools/list` must include `video_analytics__get_sop_report`. If absent, the deployment lacks the SOP tools — hand off to `/vss-build-vision-agent` to compose the SOP profile.
```
zac-wang-nv: Mode C hands off to `/vss-query-analytics` for two things that skill doesn't document:

1. **`tools/list`** — `vss-query-analytics/SKILL.md` documents only the two-step `initialize` → `tools/call` pattern. There is no `tools/list` recipe anywhere in it (grep: zero hits), so "run `tools/list` via that skill" has no procedure to follow.
2. **`get_sop_report`** — its `## Tool Reference` section enumerates `get_incidents`, `get_incident`, `get_sensor_ids`, `get_places`, `get_fov_histogram`, `analyze`, `vst_sensor_list`. None of the four `get_sop_*` tools appear. Mode B works because `get_incidents` *is* documented there; Mode C has no such anchor.

Either add a `tools/list` snippet + the `get_sop_*` entries to `vss-query-analytics/SKILL.md`, or inline the two-step curl in Mode C Step 1 the way the *Deployment prerequisite* block does for Modes A and B.

Related, in the same file: the Mode C probe at line 79 is comment-only while Modes A and B each get an executable `curl` — and `## Instructions` step 2 (line 26) still says "hand off to `/vss-deploy-profile` if its probe fails", which is the wrong skill for Mode C (line 83 correctly routes it to `/vss-build-vision-agent`).

skills/vss-build-vision-agent/references/services/sop/sop-vlm-captions-json-logstash.conf
```diff
@@ -0,0 +1,77 @@
...
+      target   => "@timestamp"
+      timezone => "UTC"
+    }
+    mutate { remove_field => ["event_time_ms"] }
+  }
+
+  # checker_result arrives as a nested object; keep it but also surface a flat compliance
+  # flag the dashboard can filter on.
+  if [checker_result] {
+    if [checker_result][cycle_completed] {
```
zac-wang-nv: The stated intent is *"surface a flat compliance flag the dashboard can filter on"*, but `if [field]` in Logstash tests truthiness, so this only fires when `cycle_completed` is `true`. Docs with `cycle_completed: false` get no `sop_cycle_completed` field at all — the dashboard can filter for completed cycles but not for incomplete ones, which is the case you'd actually want to find.

Using an explicit `if [checker_result][cycle_completed] == true { ... } else { ... }`, or just checking for field presence, gives both values.

Also in this file (line 72): `document_id => "%{req_id}"` has no guard. If a chunk ever arrives without `req_id`, the id becomes the literal string `%{req_id}` and every such doc overwrites the same ES document — silently, since there's no failure. Given the eval explicitly asserts `doc _id equals req_id`, a `if ![req_id] { drop {} }` (or a fingerprint fallback) would make the failure visible instead of lossy.

skills/vss-build-vision-agent/references/services/sop/integrate-ds-sop.md
```diff
@@ -0,0 +1,164 @@
...
+    restart: "no"
+    profiles: ["sop-kibana-init"]    # self-named profile
+    depends_on:
+      kibana:
+        condition: service_healthy
+    entrypoint: ["/bin/sh", "-c"]
+    command:
+      - |
+        curl -fsSL --retry 5 --retry-delay 5 --max-time 60 -o /tmp/sop-kibana-objects.ndjson \
+          "${SOP_KIBANA_OBJECTS_URL:-https://raw.githubusercontent.com/NVIDIA/sop-monitoring-blueprints/main/agentic/vss-sop-skills/vss-sop-build/references/deployments/sop/sop-app/kibana-dashboard/sop-kibana-objects.ndjson}"
```
zac-wang-nv: Inconsistent pinning against the same upstream repo. `sop.md` deliberately pins `0dd472f` for the `tools.py` / VA-MCP config download (lines 115, 128, 208), but this `sop-kibana-init` default fetches from unpinned `main` — so an upstream dashboard change silently alters what a rebuilt SOP profile imports.

`build-ds-sop.md` has the larger version of the same gap: Step 0 does a bare `git clone ... && cd microservices/sop-inference-bp` with no checkout of the pin, and that tree builds the ~50 GB `ds-sop:1.0.0` image. An upstream commit there changes the image with no signal.

Suggest pinning both to `0dd472f` (or whatever the validated commit is) for consistency with § Patch specifics — the `SOP_KIBANA_OBJECTS_URL` override already gives an escape hatch, and `build-ds-sop.md` just needs a `git checkout <pin>` after the clone.

zac-wang-nv:
Reviewed at `079723def` against `develop` (merge-base `f9edf5cd4`): 10 files, +951/-9.

Overall this is a well-researched contribution — the DS-SOP reference bundle captures a lot of hard-won operational detail (the vLLM `0.3` → `0.6` VRAM footgun on 48 GB cards, the protobuf-vs-JSON Logstash mismatch, the `camera_streaming` Redis event VST never publishes itself, the `@timestamp`-from-`pipeline_chunk_end_timestamp` trap). The adapter change is clean and I verified its blast radius is contained: `generate_task` is per-adapter, `wants_report` matches only the SOP spec across all 10 eval specs, and the `gpu_count` override is a no-op for every other spec since they all declare `RTXPRO6000BW: 2` — identical to the PLATFORMS default. The `es_client` whitelist caveat in § Patch specifics also checks out against `vss_agents/video_analytics/es_client.py` (`INDEXES` + `get_index` raising `ValueError`, prefix-aware).

I've left 8 inline comments. The one that most needs attention is the `VSS_AGENT_PKG` inconsistency in `sop.md` — three sites in this PR disagree on whether the package is `agent` or `vss_agents`, and the value `sop.md` hardcodes would make the SOP patch mount into a package Python never imports.

Three process notes:

1. **The `profile_sop_1_compliance_monitoring` eval this PR adds is still `IN_PROGRESS`** on both H100 and L40S (started 00:54Z). The other 47 checks are green, but the one that actually exercises this profile hasn't reported. Worth holding merge for it.

2. **Nothing in CI validates intra-skill reference links.** Two of the inline findings are dead pointers in agent-followed instructions and neither is caught by `skills-check` or the playbook-compliance bot. With `references/services/sop/` now two levels deep, relative links are easy to get wrong — a small link linter over `skills/**/*.md` would pay for itself here.

3. **The `skills-review-bot` advisory (26 findings) ran against `059195cf3f` and was never re-run** against the current head, so its comment reads as stale. I re-verified its three criticals and six highs against `079723def`; the inline comments here are the ones that still reproduce.

danh-nguyen-nvidia:
/ok to test [57828c7](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1176/commits/57828c7bf053c5bd837a528ab7c5d282bd74efe4)

skills/vss-build-vision-agent/references/services/sop/integrate-ds-sop.md
danh-nguyen-nvidia: Fix on [9a1b4c5](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1176/commits/9a1b4c53e93ee58a3f273269d295b0c6022418f5)

skills/vss-build-vision-agent/references/services/sop/sop-vlm-captions-json-logstash.conf
danh-nguyen-nvidia: Fix on [9a1b4c5](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1176/commits/9a1b4c53e93ee58a3f273269d295b0c6022418f5)

skills/vss-generate-video-report/SKILL.md
danh-nguyen-nvidia: Fix on [9a1b4c5](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1176/commits/9a1b4c53e93ee58a3f273269d295b0c6022418f5)

skills/vss-generate-video-report/SKILL.md
danh-nguyen-nvidia: Fix on [9a1b4c5](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1176/commits/9a1b4c53e93ee58a3f273269d295b0c6022418f5)

skills/vss-build-vision-agent/references/services/sop/sop-report/sop-report-override.yml
danh-nguyen-nvidia: Fix on [9a1b4c5](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1176/commits/9a1b4c53e93ee58a3f273269d295b0c6022418f5)

skills/vss-build-vision-agent/references/services/sop/deploy-ds-sop.md
danh-nguyen-nvidia: Fix on [9a1b4c5](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1176/commits/9a1b4c53e93ee58a3f273269d295b0c6022418f5)

skills/vss-build-vision-agent/references/services/sop.md
danh-nguyen-nvidia: Fix on [9a1b4c5](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1176/commits/9a1b4c53e93ee58a3f273269d295b0c6022418f5)

skills/vss-build-vision-agent/references/services/sop.md
danh-nguyen-nvidia: Fix on [9a1b4c5](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1176/commits/9a1b4c53e93ee58a3f273269d295b0c6022418f5)

danh-nguyen-nvidia:
/ok to test [3de4202](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1176/commits/3de4202cdc7dc463981f4ccb52bad8c91e43e579)

danh-nguyen-nvidia:
/ok to test [fdb98f5](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1176/commits/fdb98f577e5385ae85e33a350df461dc4820b421)

danh-nguyen-nvidia:
/ok to test [e2e37fd](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1176/commits/e2e37fdfdd14c69612c03ef2e4d76d3bdf52ffb3)

skills/vss-build-vision-agent/eval/profile_sop_1_compliance_monitoring.json
```diff
@@ -0,0 +1,32 @@
+{
+  "skills": [
+    "vss-build-vision-agent",
+    "vss-generate-video-report"
+  ],
+  "profile": "sop-1",
+  "resources": {
+    "platforms": {
+      "RTXPRO6000BW": { "gpu_count": 1 }
```
zac-wang-nv, line 11: This spec is missing `"runtime_deploy": false`, so it still generates a **deploy** task — which is the one thing `bcfa65430` set out to remove.

`adapters/vss-build-vision-agent/generate.py:234` reads `bool(spec.get("runtime_deploy", True))`, and line 248 turns that into the instruction preamble. I ran the adapter against this spec to be sure:

```
$ python3 .github/skill-eval/adapters/vss-build-vision-agent/generate.py \
    --spec .../profile_sop_1_compliance_monitoring.json --platform RTXPRO6000BW ...

# instruction.md, line 3:
Use the `/vss-build-vision-agent` skill to build and deploy the `sop-1` profile on `RTXPRO6000BW`.

# task.toml:
description = "Build+deploy sop-1 profile (1/1) on RTXPRO6000BW"
runtime_deploy = true
```

So the agent is told to "build and deploy" three lines above a query that ends *"Stop after validation; do NOT deploy"*, and six lines above an env note that says *"It does NOT deploy: `ds-sop:1.0.0` is validated as an image reference (no image, models, or GPU required)"*. On a box with no `ds-sop:1.0.0` (local-build-only, ~50 GB) that resolves as burned turns on a deploy the spec deliberately excluded — on the leg whose entire purpose is to run *without* that image.

The sibling precedent is exact: `profile_an_1_stored_video_summarization_api.json` is the same shape (no-deploy variant of a `*_runtime_harbor` spec, query ends *"stop without deploying"*) and it declares `"runtime_deploy": false`. Generating that one emits `build the an-1 profile` / `runtime_deploy = false`, as intended.

Worth noting this is latent in `profile_in_2_..._harbor.json` and `profile_in_3_..._harbor.json` too — both say "do not deploy" in the query and both omit the key. That's a separate cleanup; I'd rather this new spec just be right from the start.

skills/vss-build-vision-agent/references/services/sop/sop-report/sop-report-override.yml
```diff
@@ -0,0 +1,35 @@
...
+#  under the tracked deploy/docker/ tree).
+#
+# ${VSS_AGENT_SITE_PACKAGES} defaults to /vss-agent/.venv/lib/python3.13/site-packages and
+# ${VSS_AGENT_PKG} to the image's video-analytics package (vss_agents on develop; a published
+# image may differ) — resolve both against the image (../../sop.md § Patch specifics step 1).
+
+services:
+  vss-va-mcp:
+    volumes:
+    - ${SOP_STAGE_DIR}/video_analytics/tools.py:${VSS_AGENT_SITE_PACKAGES}/${VSS_AGENT_PKG}/video_analytics/tools.py:ro
```
zac-wang-nv: Moving staging build-local is the right call, but it changed the failure mode of an unset variable, and none of the three variables on this line has a Compose default.

`VSS_APPS_DIR` was safe by construction — every Foundation ships it (`developer-profiles/*/overrides.env`, `industry-profiles/*/overrides.env`). `SOP_STAGE_DIR` is new in this PR and I can't find a single producer: grepping `*.py`, `*.sh`, `*.env`, `*.yml`, `*.yaml` across the repo returns only the four reference-doc sites this PR adds. It exists solely if the build writes it into `_builds/<name>/override.env`.

When it's absent, Compose substitutes empty and the mount source becomes the host path `/video_analytics/tools.py`. Docker creates that as an empty **directory** and binds it over the module — `vss-va-mcp` comes up, imports nothing, and fails with `Unknown included functions`. That is precisely the round-1 finding #1 failure class, re-entered through a different door.

The same applies to the other two: the header comment at lines 27–29 says `VSS_AGENT_SITE_PACKAGES` *"defaults to `/vss-agent/.venv/lib/python3.13/site-packages`"* and `VSS_AGENT_PKG` *"to the image's video-analytics package"* — but there is no `:-` in the YAML, so those defaults are prose only. Unset, the destination collapses to `//video_analytics/tools.py`.

Cheapest fix is the idiom already in this tree — `deploy/docker/services/infra/sdrc/docker-compose.yaml:66` uses `${HOST_IP:?HOST_IP must be set in .env or shell before running compose}`:

```yaml
- ${SOP_STAGE_DIR:?SOP_STAGE_DIR must point at the build-local staging dir (_builds/<name>/sop-report)}/video_analytics/tools.py:${VSS_AGENT_SITE_PACKAGES:-/vss-agent/.venv/lib/python3.13/site-packages}/${VSS_AGENT_PKG:?resolve from the image — see sop.md § Patch specifics}/video_analytics/tools.py:ro
```

`config` then fails loudly at validation instead of the stack coming up subtly broken — and it would be caught by this PR's own `docker compose config --quiet` check.

skills/vss-build-vision-agent/eval/profile_sop_1_compliance_monitoring_runtime_harbor.json
```diff
@@ -0,0 +1,35 @@
+{
+  "skills": [
+    "vss-build-vision-agent",
+    "vss-generate-video-report"
+  ],
+  "profile": "sop-1-runtime",
+  "resources": {
+    "platforms": {
+      "RTXPRO6000BW": { "gpu_count": 1 }
```
zac-wang-nv: Two eval-config notes on the split, neither blocking.

**1. `gpu_count: 1` changes pool routing, not just accounting.** These are now the only two specs in the family at 1 — the other nine are all 2. In `run_leg.py:504`, `exact = 0 if hint == required_count else 1` sorts exact matches *first*, so a `*-1g*` node is now preferred over `vss-eval-rtx-2g`. Over-provisioned boxes stay eligible (`run_leg.py:484` only rejects `hint < required`), so this won't strand the leg — but if a 1-GPU RTX PRO 6000 node ever joins the pool, both SOP legs go there. The rationale in `2130bc03b` holds for dedicated allocation; note that `streamprocessing-ms` and `sensor-ms` both carry `runtime: nvidia`, so they share the GPU with DS-SOP's vLLM rather than needing none.

**2. For the build-validate spec, `0` is the more precise value.** Its env says *"no image, models, or GPU required"*, and the harness has a first-class encoding for that: `run_leg.py:463` — *"gpu_count == 0 (remote-all / GPU-independent) accepts any RUNNING box"* — with `gpu_type` enforcement skipped at lines 466/482/486. `vss-setup-behavior-analytics/generate.py:13` documents the pattern. That leg is a `docker compose config` run; letting it land on any free box frees RTX PRO capacity for the runtime legs.

Separately: this runtime spec carries **15 checks** and omits `judge_max_turns`, so it defaults to 60. The only other 15-check runtime spec in this family, `profile_an_1_stored_video_summarization_runtime_harbor.json`, sets `90`. This one's checks look heavier than an-1's — `docker exec` + a two-step MCP JSON-RPC handshake, `ffprobe` against RTSP, ES sampling, and a full Mode C render. `in-2`/`in-3` runtime do sit at the default, so this may well be fine; flagging it in case the an-1 bump was made for exactly this reason.

skills/vss-build-vision-agent/references/services/sop/deploy-ds-sop.md
```diff
@@ -0,0 +1,163 @@
...
+Run these after bring-up to confirm the deployment is healthy end-to-end (the profile eval
+automates the same checks). Stop at the first failure and see § Known Deployment Issues.
+
+```bash
+# 1. DS-SOP up
+curl -sf --max-time 10 http://localhost:8300/v1/ready                      # → 200
+curl -sf --max-time 10 http://localhost:8300/v1/models | grep -q ds_sop_model
+
+# 2. Kafka topic exists and has messages (SOP JSON is being published)
+docker exec kafka kafka-topics.sh --bootstrap-server localhost:29092 \
```
zac-wang-nv: `kafka-topics.sh` does not exist on this stack's Kafka image. `deploy/docker/services/infra/compose.yml:154` pins `confluentinc/cp-kafka:8.3.0`, and the Confluent images ship the CLI as `kafka-topics` — no `.sh`, that's the Apache Kafka distribution's name. The strongest evidence is 34 lines below the image pin: the service's own healthcheck at `compose.yml:188` runs `kafka-topics --bootstrap-server localhost:29092 --list`, bare, against this exact container.

So this step fails with `executable file not found in $PATH` regardless of Kafka's actual health. Because the section opens with *"Stop at the first failure and see § Known Deployment Issues"*, an operator or agent following it stops at step 2 and starts debugging a Kafka problem that isn't there — on a stack where steps 3–6 would have passed.

The rest of the line is right, for what it's worth: `container_name: kafka` is confirmed at `compose.yml:179`, and `KAFKA_LISTENERS` binds `INTERNAL://0.0.0.0:29092` so `localhost:29092` resolves inside the container. Just drop the `.sh`.

skills/vss-build-vision-agent/references/services/sop.md
```diff
@@ -0,0 +1,217 @@
...
+   package, and `vss-va-mcp` crash-loops with `Unknown included functions`. Pick the real on-disk
+   directory (has `video_analytics/`, not a symlink) instead:
+   ```bash
+   docker run --rm --entrypoint /vss-agent/.venv/bin/python3 \
+     ghcr.io/nvidia-ai-blueprints/vss/vss-agent:develop-latest -c '
+   import os, sysconfig
+   sp = sysconfig.get_paths()["purelib"]
+   for pkg in ("vss_agents", "agent"):
+       d = os.path.join(sp, pkg)
+       if os.path.isdir(os.path.join(d, "video_analytics")) and not os.path.islink(d):
```
zac-wang-nv: The probe itself is right — checking for a real on-disk `video_analytics/` and rejecting symlinks is the correct discriminator, and preferring `vss_agents` in the iteration order is the fix. But when **neither** candidate matches it prints nothing and exits `0`.

That reproduces the exact failure class this rewrite exists to prevent. The next line says *"Set `VSS_AGENT_PKG` and `VSS_AGENT_SITE_PACKAGES` from the output"*; with empty output an agent sets both to empty, and — since `sop-report-override.yml` has no Compose defaults either (Comment 2) — the mount destination becomes `//video_analytics/tools.py`. Docker creates the path, the patch never loads, and `vss-va-mcp` fails the same silent way, just with a different empty directory.

A loop `else` that exits non-zero with the candidates it tried would make the failure loud, and would also catch the case where the package lives in `platlib` rather than `sysconfig.get_paths()["purelib"]`.

skills/vss-build-vision-agent/references/services/sop.md
```diff
@@ -0,0 +1,217 @@
...
+curl -fsSL --max-time 60 $REPO/$BASE/patches/tools.py                 -o /tmp/sop-release-tools.py
+curl -fsSL --max-time 60 $REPO/$BASE/configs/va_mcp_server_config.yml -o /tmp/sop-release-va-mcp.yml
+```
+
+**Adapt to the running image** — do NOT mount the 3.1 `tools.py` verbatim (it would regress
+the image's other `video_analytics` tools and targets the old package layout):
+
+1. **Resolve which package holds `video_analytics` + the site-packages path** on the image —
+   **probe the filesystem, NOT `import`.** `import agent.video_analytics` succeeds even where no
+   `agent/` directory exists, because `vss_agents/__init__.py` registers `agent` as a runtime
```
zac-wang-nv: I can't substantiate the alias mechanism from this tree, and I think it's misattributed. At `2130bc03b`, `services/agent/packages/vss_agents/src/vss_agents/__init__.py` is a 14-line SPDX licence header and nothing else — no `sys.modules` assignment, no `__path__` manipulation. Grepping `services/agent/packages/*/src/` for `sys.modules` and `__path__` returns nothing, there is no `agent` package or `.pth` file anywhere, and `vss_agents/pyproject.toml` declares `packages = ["src/vss_agents"]` only.

The observed crash-loop has a simpler and sufficient explanation: the round-1 text hardcoded `VSS_AGENT_PKG=agent` in the env block, so anyone copying it mounted under `agent/` without the probe ever running. That alone produces the empty-directory shadow and the `Unknown included functions` failure.

This is only a doc-accuracy point — the guidance is right either way, and I'd keep the filesystem probe regardless of mechanism. But it's stated as a definite cause in an agent-executed reference (and repeated at line 155, *"`agent` is only the alias"*), so the next person debugging a package-resolution problem will go hunting for an alias that isn't there. Either cite the evidence (the `docker run ... -c 'import agent.video_analytics'` output from the image that showed it) or soften to *"`import`-based probing is unreliable here — prefer the on-disk check"*. The published `ghcr.io` image is built outside this repo, so if you observed it there, that's worth saying explicitly.

skills/vss-generate-video-report/SKILL.md
```diff
@@ -337,19 +346,67 @@ If `get_incidents` returns zero results, STOP and return exactly a one-line empt
...
+```
+
+(No bash arrays — POSIX-`sh` safe; the session id is guarded, and the tool check exits non-zero when `get_sop_report` is missing.)
+
+### Step 2 — Fetch the aggregated SOP report from VA-MCP
+
+Call `video_analytics__get_sop_report` on the same endpoint (reuse `$MCP` / `$SID` / `$CT` / `$AC` from Step 1; `tools/call` after `initialize`):
+
+```bash
+curl -s --max-time 30 -X POST "$MCP" -H "$CT" -H "$AC" -H "mcp-session-id: $SID" \
```
zac-wang-nv, line 572: The `[ -n "$SID" ]` guard added at line 364 is a good addition — but it protects the block that *creates* `$SID`, not the one that *consumes* it. Line 371 asserts *"the session id is guarded"*, and that reads as covering the whole flow; Step 2 here is a separate fenced block and reuses `$MCP` / `$SID` / `$CT` / `$AC` with no guard of its own. Line 375 says so explicitly: *"reuse `$MCP` / `$SID` / `$CT` / `$AC` from Step 1"*.

That matters because of how this skill is actually executed. An agent runs each fenced block as its own shell invocation, and shell state does not persist between them — Claude Code's Bash tool documents exactly this. So in Step 2 all four variables are empty: the URL collapses to `curl -s -X POST ""`, and the header becomes a bare `mcp-session-id:`. You get a curl usage error or an opaque server-side session rejection, not the clear "is vss-va-mcp up?" message Step 1 would have given.

Cheapest fix is to make Step 2 self-contained (re-`initialize` inline), or to state explicitly that Steps 1 and 2 must run in a single shell and repeat the `[ -n "$SID" ]` guard at the top of Step 2 — and narrow the line-371 claim to Step 1 either way.

skills/vss-build-vision-agent/references/services/sop/sop-vlm-captions-json-logstash.conf
```diff
@@ -0,0 +1,89 @@
...
+
+  # checker_result arrives as a nested object; keep it but also surface a flat compliance flag
+  # the dashboard can filter on. Set a real BOOLEAN via ruby — `mutate add_field` only produces
+  # strings, which break Kibana boolean filters and clash with the field's type across daily
+  # indices. Always emit it (true OR false) so an incomplete cycle is filterable, not just absent:
+  if [checker_result] {
+    ruby {
+      code => '
+        cc = event.get("[checker_result][cycle_completed]")
+        event.set("sop_cycle_completed", cc == true || cc.to_s == "true")
```
zac-wang-nv: Right fix — `mutate add_field` really does only produce strings, the boolean matters for Kibana filters, and using `ruby` is consistent with the `event_time_ms` filter already at lines 41–48 of this same file. Two small notes, neither blocking:

1. **The predicate is narrower than what it replaced.** `if [checker_result][cycle_completed]` was a truthiness test, so a non-boolean truthy value (say `1`) counted as complete; `cc == true || cc.to_s == "true"` sends that to `false`. That's fine if the checker always emits a JSON boolean — but if it can emit `1`/`0`, this silently mislabels completed cycles, which is the same class of silent miss the change is fixing. Worth a quick confirmation of what DS-SOP actually puts in that field.
2. **Absent and false are still indistinguishable.** `event.get` returns `nil` when `cycle_completed` is missing, and `nil` lands in the `false` branch — so a chunk whose checker never populated the field is indexed as an incomplete cycle rather than as unknown. Fine if the field is guaranteed present; if not, `cc.nil?` deserves its own outcome.

zac-wang-nv:
Re-reviewed at `2130bc03b`. **`bcfa65430` is the best commit in this PR** — moving the adapted patch out of `${VSS_APPS_DIR}/services/agent/sop-report/` into a build-local `${SOP_STAGE_DIR}` is the correct architectural call, and it's a complete migration rather than a half one: `grep VSS_APPS_DIR` across `sop.md`, `references/services/sop/`, and `vss-generate-video-report/` now returns zero hits. Splitting the eval is equally right — a leg that only needs `docker compose config` shouldn't have been gated behind a 50 GB locally-built image and 14.7 GB of models.

All eight round-1 findings remain genuinely fixed, verified against the files rather than the commit messages, and `88a7f9c0e`'s adapter fallbacks are correct (both discovered directories exist, both are `None`-guarded and `.exists()`-gated).

Nine comments, none of them product bugs. The one I'd fix before merge is **Comment 1**: `profile_sop_1_compliance_monitoring.json` is missing `"runtime_deploy": false`, so the adapter still emits *"build and deploy the `sop-1` profile"* and `runtime_deploy = true` — I ran it to confirm — three lines above the query's *"Stop after validation; do NOT deploy"*. That's a two-word change, and without it the no-deploy leg this commit created still asks for a deploy on a box that intentionally has no `ds-sop:1.0.0`. `profile_an_1_stored_video_summarization_api.json` shows the intended form.

**Comment 2** is the one worth a moment's thought beyond this PR. `SOP_STAGE_DIR` has no producer in the repo and no Compose default, so an unset value silently degrades to a host-root bind that shadows the real module — the same failure class as round-1 finding #1, re-entered through a new door. `${VAR:?message}` is already the house idiom (`services/infra/sdrc/docker-compose.yaml:66`), and it would make this PR's own `docker compose config --quiet` check catch it.

The rest are carried over from a round-3 pass whose comments were never posted; I re-verified every one at this head and updated the line numbers. Two small leftovers not worth their own threads: `final_*` is still referenced in the `{notes}` guidance at `SKILL.md:389` without being defined anywhere in the skill, and the standing round-1 item — nothing in CI validates intra-skill reference links. The latter stays moot while the links are correct, but `references/services/sop/` is two levels deep and relative-path errors were the dominant defect class in round 1; a link linter over `skills/**/*.md` would have caught two of those for free.

CI-wise, the full matrix is running at this head, including both new SOP legs on `RTXPRO6000BW`. Same ask as before: let them go green before merging — the last green SOP eval was at `079723def`, and everything since has rewritten what it exercises. With the split, the build-validate leg should give a fast signal once Comment 1 lands.

skills/vss-build-vision-agent/eval/profile_sop_1_compliance_monitoring_runtime_harbor.json
```diff
@@ -0,0 +1,37 @@
...
+  "expects": [
+    {
+      "query": "Use `/vss-build-vision-agent` to build and deploy under `_builds/sop-1-runtime/`: \"Compose and deploy an SOP compliance-monitoring application with SOP compliance reports.\" `ds-sop:1.0.0` is already built. Bring the stack up end-to-end, feed DS-SOP the staged video DIRECTLY (on-demand `POST http://localhost:8300/v1/chat/completions`, max_length_sec=10), then verify SOP detection in Elasticsearch and the skill-driven Mode C compliance report. Leave the stack running.",
+      "checks": [
+        "Deployment uses exactly `docker compose -f _builds/sop-1-runtime/resolved.yml up -d`; it does not re-read Foundation env files, use an invented Compose profile, or deploy a copied Compose project. No file under `{{repo_root}}/deploy/docker/` is modified (patch + Logstash overlay are build-local under `_builds/`).",
```
zac-wang-nv: this test seems too rigid.

skills/vss-build-vision-agent/eval/profile_sop_1_compliance_monitoring_runtime_harbor.json
zac-wang-nv, line 1: we used to have two eval specs for all skills which is not efficient. can we combine them into one eval?

danh-nguyen-nvidia:
/ok to test [5d0dc07](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1176/commits/5d0dc0749feba278904de74904de77ac5b7ea0d7)

skills/vss-build-vision-agent/eval/profile_sop_1_compliance_monitoring_runtime_harbor.json
danh-nguyen-nvidia: Fix on [fe5e44d](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1176/commits/fe5e44da540c8fa220b6564c193d7d57d14fb5b8)

skills/vss-build-vision-agent/eval/profile_sop_1_compliance_monitoring_runtime_harbor.json
danh-nguyen-nvidia: Fix on [fe5e44d](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1176/commits/fe5e44da540c8fa220b6564c193d7d57d14fb5b8)

skills/vss-build-vision-agent/eval/profile_sop_1_compliance_monitoring.json
```diff
@@ -0,0 +1,35 @@
...
+        "After feeding the staged video on-demand, `curl -sf 'http://localhost:9200/mdx-vlm-captions-*/_count'` returns `.count > 0`; sampled docs carry FLAT fields (chunk_idx, response, sensor_id, cv_boundary_score, pipeline_chunk_end_timestamp, req_id, checker_result) -- NOT protobuf-nested; doc `_id` equals req_id (idempotent upsert) and `@timestamp` derives from the wall-clock pipeline_chunk_end_timestamp (docs land on the real date, not 1970).",
+        "Detection is real: the `response` field shows REAL SOP steps in sequence (e.g. '(1) installing the first fan...' -> power supplies -> cover), NOT every doc being '(10) ... not belong', and cv_boundary_score across docs is a healthy mix (at least some >= 0.5, not all approximately 0).",
```
zac-wang-nv: This is the one runtime failure that survived the jump to 13/15, and its root cause has changed since the last report.

At `2130bc03b` check 9 failed as *"all `cv_boundary_score` values are 0"*, and the report attributed that to the test video decoding as a solid blue frame. That explanation no longer holds: at `9a6b68455` **check 8 passed** — `response` carried real SOP steps in sequence — so the fixture is producing real detections. What the verifier reports instead is:

> The Elasticsearch index `mdx-vlm-captions-2026.07.29` does not contain a `cv_boundary_score` field **at all** (confirmed via `_mapping` and `_search`). The field is absent from every document.

Absent from the mapping is a schema gap, not a data-quality one. And nothing in this PR produces that field: `sop-vlm-captions-json-logstash.conf` has **zero** references to `cv_boundary_score`. That pipeline is a straight passthrough of DS-SOP's JSON plus exactly three synthesised fields — `@timestamp` (from `pipeline_chunk_end_timestamp`), `sop_cycle_completed`, and the `req_id` fingerprint fallback. So the field can only reach Elasticsearch if DS-SOP emits it under `SOP_MESSAGING_SCHEMA=JSON`, and the evidence says it does not.

That conf is **byte-identical** between `9a6b68455` and `5d0dc0749`, so the run queued at this head will hit this again.

The merge also raises the cost of leaving it: `cv_boundary_score` is now named on **both** L27 (the FLAT-fields list) and this check, so one absent field is exposed to two of the thirteen checks. In the split spec it appeared in old checks 7 and 9, and **7 passed** — the judge graded that one on flat-vs-nested rather than strict field presence — so the blast was contained to one check. Whether L27 stays that lenient is now a coin-flip.

Worth settling before another 1-hour leg: does DS-SOP emit `cv_boundary_score` in JSON mode at all? One `docker exec` reading a raw message off `mdx-vlm-captions` answers it. Then either drop the field from both L27 and this line, or — if it ships under a different key — map it in the pipeline. As written, the spec asserts a field the delta never produces.

danh-nguyen-nvidia: Docs with chunk_idx = -1 won't have cv_boundary_score and response, while with id >= 0 we still have them normally (verified on my own deployment). Two different results depend on what docs were read, so I add a condition to look at chunk_idx first before asserting other values on 1f03d75dadc1f8c282fd3f56bd59aaa15053f68b

danh-nguyen-nvidia:
/ok to test [13d2a70](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1176/commits/13d2a7084c66d5c287d150ed790fd38b3da13508)

danh-nguyen-nvidia:
/ok to test [5f96055](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1176/commits/5f96055b8a70134a2cc05ec0259603cb6caee31c)

danh-nguyen-nvidia:
/ok to test [f1307ad](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1176/commits/f1307ad4eef427f030879e376a43da8dc2476f1a)

zac-wang-nv (approved):
LGTM

PR #1190, opened 2026-07-03
feat(deploy/ci): container dev & release workflow migration

hugoverjus:
Introduce a single source of truth for first-party container coordinates and the CI scaffolding to make dev testing "just work" while keeping develop release-ready.

- deploy/docker/containers.env: one place for VSS_CONTAINER_REGISTRY / VSS_CONTAINER_TAG plus per-image overrides; all env-overridable.
- Refactor every first-party vss-core/* compose image: line to consume the variables with byte-identical inline defaults (docker compose config unchanged; gate resolves identical refs).
- check_container_tag_source.py: resolve inline ${VAR:-default} before name matching so parameterized registries still verify.
- Phase 1: build-dev-images.yml builds changed first-party images to GHCR with pr-<N>-<sha> tags and stamps the source_tree_sha OCI label.
- Phase 2: auto-bump-container-tag.yml + bump_container_tag.py commit the published tag (single line) back to the PR branch, signed-off, on green container tests; the source-check gate becomes a verification.
- Phase 3: qa-release-promote.yml + release_promote.py copy GHCR -> NGC dev -> NGC staging keeping the same tag (Nspect via promotion); staging eval is a non-blocking signal.
- Docs: HTML presentation (now vs after, Mermaid diagrams) and the proposal.
- Unit tests for the bump and promote scripts.

## Checklist


hugoverjus:
ok to test

.github/workflows/build-dev-images.yml
```diff
@@ -0,0 +1,291 @@
...
+      - name: Compute immutable candidate coordinates
+        id: meta
+        run: |
+          python3 .github/scripts/container_build_plan.py metadata \
+            --ref-name "$GITHUB_REF_NAME" \
+            --commit-sha "$GITHUB_SHA" \
+            --owner "$GITHUB_REPOSITORY_OWNER" \
+            --image-name "${{ matrix.name }}" \
+            --tree-sha "$(git rev-parse "HEAD:${{ matrix.source_path }}")" \
+            --github-output "$GITHUB_OUTPUT"
```
simings-nv, line 112: Thanks for this, the immutable tag plus source_tree_sha stamping is really nice.
 
One thing I wanted to check, since I'm still learning the CI trust boundaries here. A few matrix values get interpolated straight into the shell:

- line 101: --include="${{ matrix.lfs_include }}"
- line 110: --image-name "${{ matrix.name }}"
- line 111: HEAD:${{ matrix.source_path }} inside git rev-parse
 
If those come from container-inventory.json (which a PR can edit), I think a crafted source_path could break out of the command substitution and run on the build runner. GitHub's hardening guide suggests passing context through env: and referencing "$SOURCE_PATH" instead. 

Two questions I'd value your take on:
1. Given how /ok to test and the mirror push work, is PR input actually reachable here, or is there a gate I'm missing?
2. If it is, would routing these through env: be a reasonable fix?

Probably not a big issue, given only vetter can run /ok to test

Happy to be wrong. Thanks!

.github/workflows/nightly-promote-ghcr.yml
```diff
@@ -0,0 +1,123 @@
...
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  release:
+    if: vars.VSS_NIGHTLY_PROMOTE_ENABLED == 'true'
+    runs-on: [self-hosted, downstream-pipeline]
+    timeout-minutes: 185
```
simings-nv, line 39: Would it make sense to pull these timing values (the job timeout-minutes and the two MAX_POLL_DURATION_SECONDS) up into config variables, similar to how VSS_NIGHTLY_PROMOTE_ENABLED is handled? It'd let us tune them without a code change when we see real pipeline durations.

The one thing I'd want to be careful of: the two polls (lines 71 and 90) run in sequence in this job, each up to 180 min, so the job needs timeout-minutes to cover their sum, but right now it's 185. If we make both tunable, I'd worry about them drifting out of that relationship later. Could we either split the two polls into separate jobs (so each timeout only covers one poll), or add a small preflight check that fails fast if the job timeout is less than the combined poll budget? Curious whether you'd lean one way or the other, still getting familiar with how we manage this kind of thing here.

.github/workflows/build-dev-images.yml
```diff
@@ -0,0 +1,291 @@
...
+  # Assemble ONE complete release set for this commit: freshly built images
+  # from the fragments, everything else carried forward explicitly as
+  # reuse-pinned entries. Runs even when nothing was built (docs-only push):
+  # the resulting set is still complete and testable.
+  assemble:
+    name: Assemble release set
+    needs: [changes, build]
+    if: >-
+      always() && needs.changes.result == 'success' &&
+      (needs.build.result == 'success' || needs.build.result == 'skipped')
```
simings-nv, line 223: Wanted to check how a single failed build leg surfaces downstream. Since build is fail-fast: false (line 90), one failed image makes the aggregate needs.build.result = failure, so assemble is skipped and no release-set artifact is produced.

The handoff (select_release_set_run in update_pr_ghcr_candidates.py) only accepts conclusion == "success", so it never matches, polls the full budget, and raises "no successful GHCR build run found". If I'm reading it right, on-call sees that generic message in ci.yml an hour later with no pointer to which image actually failed.

Could the handoff treat a build run that finished failure as terminal and fail fast with a link, instead of polling for a success that can't arrive? Same "key on terminal state" idea as the timeouts. 

Still learning the flow here, so happy to be corrected.

zac-wang-nv (approved):
LGTM

PR #1271, opened 2026-07-13
feat(alert service): Add Prometheus metrics to the on-demand workflow

hfarooq-nv:
## Description
Adding Prometheus metrics to the on-demand workflow rather than only supporting the Kafka based workflow. 

## Checklist


dineshvenkat:
Media-download failures get labeled "unknown" instead of "media_download" — because the error code isn't passed when the download fails. So the media_download failure metric always shows zero, even when downloads are actually failing.

dineshvenkat:
**[MINOR] `ondemand_vlm_duration_seconds` includes media download / base64 prep — not just VLM inference**

`src/handlers/direct_media/direct_media_handler.py:300-318`, `:353`

The injected `vlm_duration_observer` wraps `analyze_single_media` / `analyze_multiple_images`, which perform media download and base64 preparation *before* the VLM API call when `vlm_media_source_using_base64=true`. The histogram is named/documented as VLM inference duration, so in that mode it conflates media I/O with model latency — benchmarks will attribute slow downloads to the VLM. There is also no separate media-download duration series, so download time isn't observable on its own.

---

**[MINOR] Not zero-overhead when `PROMETHEUS_METRICS_ENABLED=false`**

`src/web/api/verification_routes.py:128`, `src/web/service/ondemand_verification_service.py:156`, `src/handlers/direct_media/direct_media_handler.py:300`

With metrics disabled the recorder functions correctly no-op, but the request/background paths still capture `time.monotonic()` (route + service + handler `finally` blocks) and still invoke the observer callback on every request. The parent feature's acceptance criterion is "zero measurable overhead" when disabled; this leaves a small residual on the hot path. Consider gating the `monotonic()` spans and the `vlm_duration_observer` injection on the enabled flag so nothing runs when metrics are off.

hfarooq-nv:
Addressed the "unknown" media download failure and non zero overhead issues in latest commit

dineshvenkat (dismissed):

PR #1286, opened 2026-07-14
fix: send x-stream-id header on RTVI-CV /stream/remove

zac-wang-nv:
## Summary

The video delete endpoint (`DELETE /api/v1/videos/{video_id}`) called RTVI-CV `POST /api/v1/stream/remove` without the `x-stream-id` routing header. On multi-replica RTVI-CV deployments behind HAProxy Ingress with consistent-hash routing, the ingress injects a random `uuid()` when the header is missing, so the remove request lands on a random pod instead of the pod that owns the stream — silently orphaning streams (and their GPU / DeepStream pipeline resources) on the original pod.

The `/stream/add` path already sends this header (`video_ingest.py`), as do all RTSP ingest/delete paths. This change makes the remove path consistent.

Fixes NVBUG 6455296.

## Changes
- `services/agent/src/vss_agents/api/video_delete.py`: send `headers={"x-stream-id": sensor_id}` on the RTVI-CV remove POST.
- `services/agent/tests/unit_test/api/test_video_delete.py` (new): unit tests for `_remove_from_rtvi_cv`, pinning the URL/payload/header contract.

## Testing
- `uv run pytest tests/unit_test/api` — 178 passed.
- Pre-commit hooks (ruff, mypy, SPDX, DCO, TruffleHog) pass.

### DIFF
--- a/services/agent/src/vss_agents/api/video_delete.py
+++ b/services/agent/src/vss_agents/api/video_delete.py
@@ -120,7 +120,10 @@ async def _remove_from_rtvi_cv(
     logger.info(f"Removing from RTVI-CV: POST {url}")
 
     try:
-        response = await client.post(url, json=payload)
+        # `x-stream-id` is the consistent-hash routing key for multi-replica
+        # RTVI-CV deployments — it must match the header sent on /stream/add
+        # so the remove lands on the pod that owns the stream.
+        response = await client.post(url, json=payload, headers={"x-stream-id": sensor_id})
         if response.status_code in (200, 201, 204):
             logger.info("RTVI-CV stream removed: %s", scrub_log(sensor_id))
             return True, "OK"
--- a/services/agent/tests/unit_test/api/test_video_delete.py
+++ b/services/agent/tests/unit_test/api/test_video_delete.py
@@ -1,0 +1,91 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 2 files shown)


zac-wang-nv:
ok to test

zac-wang-nv:
**Downstream CI status — blocked by eval-infra regression, not this PR**

Diagnosis of the `Trigger Downstream Pipeline` / `test-base` failures (ci-vss-oss pipeline 57997119):

- `test-base` fails deterministically in report generation: the VLM rejects the frame batch with 422 `At most 5 image(s) may be provided in one prompt` (`video_understanding.py:693`).
- Timeline on the shared eval VLM (cosmos3-reasoner @ 10.86.6.50:32086): branches **without** any cosmos frame-limit handling passed `test-base` at 09:43/09:54/10:18/10:43 PDT today, and failed from 10:51 PDT onward (2/2 identical). The NIM started enforcing the 5-image limit around ~10:45 PDT.
- Current `main` has no cosmos handling in `video_understanding.py`, so **every branch based on main now fails `test-base`**. PR #1274 carries the fix (`_is_cosmos_model` → full-MP4 base64 path, bypassing the per-prompt image cap) and its pipelines pass.
- Separately, `eval-warehouse-bp-wh-2d` fails GPU preflight on runner `runner-jnsnbxejq` (host `hverjus-9m7vec`): `no GPUs visible / NVML Unknown Error` — unhealthy node, needs drain. Both retries landed on the same node.

Fixes already made in this PR while getting the downstream build green:
- libssl3 security-patch pin 3.0.19→3.0.20 (old deb removed from debian-security archive; broke all agent image builds)
- alert service base image nvcr.io/nvidian → public nvcr.io/nvidia (anonymous pulls of the private org 403, broke alert-ms builds)

Will re-run once #1274 merges (then merge main into this branch) or the eval VLM is rolled back.

zac-wang-nv:
Rebased onto `develop` and retargeted the PR (`main` was the wrong base — its eval config still points at cosmos-reason2-8b, which now fails `test-base` report generation with the VLM 5-image limit; `develop` already carries the cosmos3 config and the libssl3/alert-image fixes I had duplicated). Branch is now a single commit with the x-stream-id fix + tests.

/ok to test a9573e32a78cece377732efb8b1fcc74bd0d4e5f

PR #1292, opened 2026-07-14
fix(skill-eval): move fleet selection into run_leg.py try-lock cascade

zac-wang-nv:
## Problem

Fleet selection is agent-side prose (AGENTS.md § 5a): each leg's agent snapshots `brev ls --json`, scores candidates, checks "lock appears free" with an advisory `flock -n`, and hands one instance to `run_leg.py`. Two legs starting together both see the same "best" lock-free box — **neither has acquired the real lock yet** — and converge on it (check-then-act TOCTOU).

Observed on run [29373239241](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/29373239241): `lvs_profile_summarize` and `lvs_api_ops` both picked `vss-eval-rtx-1g-2` within seconds of each other; the flock held correctness (no overlap on the box), but the second leg blocked for **16 minutes** while `vss-eval-rtx-1g-3` sat idle and eligible.

## Fix

`run_leg.py` now owns selection — the pick and the reservation are one atomic step:

- Reads `gpu_type` / `gpu_count` from the dataset's `task.toml` `[metadata]`.
- Snapshots `brev ls --json` (with the same retry + trailing-banner parsing as `brev_env`), filters to RUNNING `vss-eval-*` boxes by loose GPU-token match — including the `gpu="-"` catalog-flake fallback via `instance_type`.
- Orders exact name-hinted `gpu_count` matches (`*-1g*` → 1, `*-2g*` → 2) first, preserving pool partitioning; over-provisioned boxes remain as fallback (brev_env's `>=` validation and box reset are unchanged).
- Walks candidates with **non-blocking** flock attempts and claims the first success — concurrent legs fan out across the pool instead of herding onto one box.
- When all candidates are held or none is eligible: re-snapshot every 60s up to the existing 21000s budget, then exit 75 (`BLOCKED: lock timeout`).

The eval agent no longer picks a box or exports `BREV_INSTANCE`. `--instance` stays as an operator override for manual debugging (still lock-guarded); `task.toml [metadata].brev_instance` still pins per-task. AGENTS.md § 5a/5b and all cross-references rewritten to the new contract.

## Tests

Added to `test_run_leg.py`: candidate filtering (pool prefix, RUNNING, gpu-type token match, flake fallback), exact-count-first ordering, `gpu_count=0` behavior, and `hold_pool_lock` claim-first-free / skip-held / timeout / release-on-exit. 15/15 pass.

Follow-up to the PR #1241 triage (cross-coordinator lock hole) — with selection now structural, the remaining reliance on agent prose for box handling is gone.

### DIFF
--- a/.github/skill-eval/AGENTS.md
+++ b/.github/skill-eval/AGENTS.md
@@ -308,78 +308,49 @@ The canonical harbor command is in § Harbor invocation.
    mirror sync. If every changed skill is parked, you exit BLOCKED
    without reaching step 5.
 
-5. **Pick a fleet member, lock it, and run harbor trials.** For each
-   target platform:
+5. **Run harbor trials via the leg wrapper — it picks and locks the
+   fleet box itself.** For each target platform:
 
-   a. **Select an instance from the `vss-eval-*` fleet for this
-      platform.** The harness is a worker-pool: one skill-eval agent =
-      one serial worker. Concurrency comes from multiple workflow runs
-      each grabbing a different box. Don't hardcode `vss-eval-l40s` —
-      score and pick:
+   a. **Do NOT select an instance and do NOT export `BREV_INSTANCE`.**
+      `run_leg.py` owns fleet selection: it reads the leg's hardware
+      requirements from the dataset's `task.toml` `[metadata]`
+      (`gpu_type`, `gpu_count`), snapshots `brev ls --json`, filters to
+      RUNNING `vss-eval-*` boxes whose GPU matches, and walks the
+      candidates best-first with **non-blocking** `flock` attempts —
+      claiming the first box it can actually lock. Selection and
+      reservation are one atomic step inside the wrapper, so two
+      concurrent legs fan out to different boxes instead of both
+      "choosing" the same lock-free-looking one and serialising
+      (check-then-act TOCTOU — the failure mode that motivated this).
 
-      ```bash
-      # Candidates: RUNNING+READY ^vss-eval-* boxes whose gpu_type matches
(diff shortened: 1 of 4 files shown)


zac-wang-nv:
Pushed d4ad8af3f — companion fix for the other half of the lock-vs-box-state gap, found while debugging PR #1281's two red eval legs:

**Symptom:** base and search legs both lost SSH mid-deploy on `vss-eval-rtx-2g-2` (❌ 0.71 / ❌ 0.67), minutes after a superseded run's legs were **cancelled** at 13:58 while mid-deploy. Both passed on retry.

**Gap:** cancellation kills the runner-side tree — releasing the flock, which dies with `run_leg.py` — but never reaps the trial agent launched *on the box* over SSH. `_reset_docker_runtime()` removes the orphan's containers, not the orphan itself, so it keeps driving `compose up`/model pulls, contends with the next trial, and can resurrect containers after the wipe. The try-lock cascade in this PR fixes selection herding but by design cannot cover this: the lock's lifetime is the process, the deployment's is not.

**Fix:** `brev_env.start()` now SIGKILLs anything matching the exact on-box trial invocation (whole process groups, so detached Bash-tool children die too) before the /logs wipe and docker reset. Bracket-pattern + SELF_PGID guard prevent self-reaping; fails loud if the reap itself breaks. Validated locally with a decoy agent in its own session: group-reaped, caller survives; no-match path prints `[stray-agent-reap] none`.

hugoverjus (approved):

PR #1293, opened 2026-07-15
fix(skill-eval): preserve VIOS bind mounts across multi-step trials

sarathm-nvidia:
## Problem

On a multi-step eval spec, `BrevEnvironment.start()` re-ran the repo sync on **every** step. Its `git clean -fdx -e data/ -e .env` deletes untracked paths — including `deploy/docker/data-dir/` (the default VIOS data location), the host source of the LVS/VIOS bind mounts (`clip_storage`, `vst_data`, …) — while step-1's containers are still running and mounted onto them. The containers are left pinned to an unlinked inode, and VIOS uploads then fail with `Failed to open output file: No such file or directory` (the `lvs_profile_summarize` step-2 failure; see #1227).

## Fix

Gate the repo sync to a spec's **first trial** (a single-step spec, or `step-1`), matching the existing docker-reset / host-purge gating. On `step-2+` the checkout is already at `PR_HEAD_SHA` from step-1 (one `run_leg.py` flock holds the box across the chain), so the re-sync is redundant as well as destructive.

## Guardrail (permanent, non-fatal)

Adds a bind-mount probe that runs before/after the sync on every step and logs mount liveness to `<trial>/artifacts/logs/artifacts/mount-probe.log`. It self-discovers VIOS mounts and reports `healthy` / `stale`, so a regression is caught loudly.

## Evidence

Same leg (`lvs_api_ops`) and deploy root (`data-dir`), step-2 `after-sync`:

| | verdict |
|---|---|
| without this fix | **stale** — host inode gone, link count 0 (mount deleted mid-chain) |
| with this fix | **healthy** — inode unchanged, mount preserved |

## Scope

Fixes the mount deletion. A separate `lvs_profile_summarize` step-2 failure (summarize contract checks) is tracked in #1227.

## Verification (run locally)

These are not CI-gated (the skill-eval tests run locally / via the manual sweep).

- `envs/tests/test_step_gate.py` — sync gated on `step-2+`, run on `step-1` / single-step.
- `envs/tests/test_mount_probe.py` — probe verdict + parse.
- `envs/tests/prove_mount_probe.sh` — docker probe self-test (`healthy → stale` on a deleted mount).


zac-wang-nv (approved):
LGTM

PR #1313, opened 2026-07-17
refactor(agent): fold vss-cli into a single vss distribution with extras

zac-wang-nv:
## What

Folds the standalone `nvidia-vss-cli` nested project (separate `pyproject.toml`, `uv.lock`, and a `src/lib` symlink back into the agent tree) into a single **`vss`** distribution at `services/agent/`, organized by install profile:

| Install | What you get |
|---------|--------------|
| `vss` | NAT-free libraries (`src/lib`) + the host CLI (`src/cli`) with its `vss` console script |
| `vss[cli]` | Stable alias for CLI installs (holds CLI-only deps if any appear; the script ships with the base wheel) |
| `vss[agent]` | The NAT-based agent application (`src/agent`, plus `lib.knowledge` deps) |

The source tree flattens to one top-level package per profile — `src/lib`, `src/cli` (was `src/lib/cli`), `src/agent` (was `src/vss_agents`) — and the console command is renamed `vss-cli` → `vss`. Host CLI invocation:

```bash
uv run --project "${VSS_REPO_ROOT}/services/agent" --no-dev vss search run ...
```

> **Deployment note:** helm/docker `config.yml` `runner_class` moves to `agent.api.custom_fastapi_worker.CustomFastApiFrontEndWorker` (updated in-repo). External compose repos (met-blueprints) pinning the old `vss_agents.…` path must update when adopting an image built from this layout.

## Why

The nested project existed to give host users a NAT-free install, but it cost a second lockfile, a symlink hack, duplicated packaging/lint/mypy config, and special-casing in CI, Sonar, and the license tooling. Extras express the same isolation from one `pyproject.toml`: dependency-level guarantees replace wheel-content guarantees (the code always ships; the NAT/torch stack only installs with `[agent]`).

## Changes

- **pyproject**: dist renamed `vss_agents` → `vss`; NAT stack moved to the `agent` extra; `vss-cli` script + NAT-free `cli-dev` dependency group (included by `dev`); one `uv.lock` (no dependency version drift — diff is structural only)
- **tests**: `vss-cli/tests/{cli,critic,search_core}` → `services/agent/tests/unit_test/lib/`; packaging-boundary tests rewritten for the extras model; the elastic cross-loop rebind test now drives explicit event loops (the `nvidia-nat` test plugin patches `asyncio.run` to reuse one persistent loop, which invalidated the test's premise in the unified env)
- **CI**: the `vss-cli-test` job becomes a NAT-free lane at `services/agent` (base + `cli-dev` sync → mypy on `lib`, lib tests with 80% coverage gate, wheel build, clean-env smoke asserting no `nat`/`torch`/`langchain-core` distributions); typecheck/test/sonar jobs sync with `--extra agent`
- **License tooling**: `check_python_licenses.sh` scans base + `agent` extra; `license_diff_csv.py` now follows the **root project's extras** so shipping deps stay in the OSRB inventory. Note: this also brings the pre-existing knowledge-backend extras (`frag_lib`, `llama_index`, `langchain`) into the inventory — they were previously never followed.
- **Docker**: `uv sync` gains `--extra agent`
- **Skills/adapter**: `vss-search-archive` docs, eval specs, and the Harbor adapter all use the new invocation
- **mypy**: `python_version` 3.11 → 3.13 (matches `requires-python >=3.13`); strict checking now covers `src/lib`

## Verification

- Full env (`--group dev --extra agent`): **2952 passed**
- NAT-free lane (base + `cli-dev`): **601 passed**, coverage 86.18% (gate: 80%)
- `mypy` clean on both `src/lib/*` (51 files) and `src/vss_agents` (129 files)
- `ruff check` / `ruff format --check` / `yamllint` clean
- Wheel smoke in a clean venv: 27 light dists, `vss-cli --help` works, no NAT/torch
- Lock regenerated and verified consumable by **uv 0.6.2** (the CI pin)
- License-diff unit tests pass, including a new root-extras regression test


zac-wang-nv:
**TODO — follow-up PR (out of scope for #1313): single-origin runtime resolution**

- Add restricted `/elasticsearch` (method + index allowlist), `/cosmos-embed`, and `/rtvi-cv` routes to `deploy/docker/services/infra/haproxy/haproxy.cfg.template` **and** the helm ingress examples, mirroring the existing `/vst` + `/storage` routes.
- Give the CLI a `--base-url <origin>` fast path: every service endpoint derives from the single ingress origin via path routing; index names come from config/env/flags.
- **Remove** the container/pod discovery machinery in `services/agent/src/cli/deployment.py` (docker inspect + k8s port-forward lifecycle). No discovery fallback — deployments are expected to run the ingress routes. We should not maintain logic that discovers pods/containers. *(Decision: zacw, 2026-07-17)*
- Skill + eval contract updates follow (`--deployment docker --profile search` → `--base-url`).

PR #1314, opened 2026-07-17
LVS CI Pipeline

david-duu:
## Checklist


services/video-summarization/tests/unit/test_via_server.py
```diff
@@ -0,0 +1,1068 @@
...
+                "events": ["event1"],
+                "stream": True,
+            },
+        )
+        assert resp.status_code == 400
+
+    def test_summarize_excluded_api_type_is_stripped(self, test_client, mock_via_server):
+        mock_via_server._stream_handler.get_models_info.return_value = self._model_info()
+        mock_via_server._stream_handler._ctx_mgr = None
+        with patch("via_server.validate_required_prompts", return_value=[]):
```
zac-wang-nv, line 317: These 7 `TestSummarizeRoute` tests (lines 317, 339, 370, 386, 441, 473, 515) patch `via_server.validate_required_prompts`, but that symbol doesn't exist anywhere in `src/` (0 grep hits repo-wide), so `patch` raises `AttributeError` and every one fails at setup. I ran the suite to confirm: **806 passed, 7 failed** — so CI running `pytest tests/unit` will be red on merge. These look written against a different `via_server.py`; please either port the missing validation helper or rewrite them against the actual `/summarize` validation path.

services/video-summarization/ci/pipeline-helpers.groovy
```diff
@@ -0,0 +1,6503 @@
...
+ */
+def fetchSSACredsFromVault() {
+    withCredentials([string(credentialsId: 'met-vss-prod-jwt-issuer', variable: 'JOB_JWT')]) {
+        container('jenkins-shared-lib-base') {
+            sh "vault --version"
+            echo "Fetching SSA credentials from vault"
+            getVaultCredentials(
+                secrets: [
+                    [secretPath: 'nvidia/services/ssa/clients/nvssa-prd-eVx9BDYG4DiQ6rhhyAerZz2YpZveLueD3SrT64-Jvx0/kv/secret', secretData: 'id', envVar: 'SSA_ID'],
+                    [secretPath: 'nvidia/services/ssa/clients/nvssa-prd-eVx9BDYG4DiQ6rhhyAerZz2YpZveLueD3SrT64-Jvx0/kv/secret', secretData: 'secret', envVar: 'SSA_SECRET']
```
zac-wang-nv, line 1198: This embeds a **production** SSA/Starfleet client ID (`nvssa-prd-…`) plus its exact Vault path in a public repository. The secret value isn't exposed, but publishing a prod client identifier and the path to its credential is exactly the reconnaissance an attacker needs to target it. Please move this into the internal Jenkins job/shared-library config (not the public tree), and consider rotating the SSA client since the identifier is now public.

services/video-summarization/ci/pipeline-helpers.groovy
```diff
@@ -0,0 +1,6503 @@
...
+}
+
+/**
+ * Fetches NGC and API credentials from HashiCorp Vault and puts them into the designated environment variables.
+ * Runs in the container named jenkins-shared-lib-base, which must be present in the wrapped pod YAML context.
+ *
+ * @return void
+ * @throws Exception if Vault credentials cannot be fetched
+ */
+def fetchCredsFromVault() {
```
zac-wang-nv, line 1162: `fetchCredsFromVault`/`fetchSSACredsFromVault` disclose the complete internal Vault auth flow on the public repo: endpoint (`prod.vault.nvidia.com`), namespaces (`swtegra-metropolis-apps`, `metropolis-lab`), JWT role (`cloudbees-met-vss-cicd-role`), mount path (`auth/jwt/cloudbees/met-vss-cicd/`), and KV secret paths (`cloudbees/met-vss-cicd/kv2/creds`, `blossom/metropolis-mdx-cicd/kv2/nv-one-click/common`). This should live in the internal shared library, not the public tree — please externalize it.

services/video-summarization/compose/upload_perf_results.sh
```diff
@@ -0,0 +1,133 @@
+#!/usr/bin/env bash
+######################################################################################################
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: LicenseRef-NvidiaProprietary
```
zac-wang-nv, line 4: This file carries the NVIDIA **proprietary** SPDX header ("Any use, reproduction, disclosure or distribution … is strictly prohibited") while every sibling file — and the repo — is Apache-2.0. Publishing it on the public repo directly contradicts its own license text. Please relicense to Apache-2.0 or drop it from the public PR.

services/video-summarization/tests/integration/test_kafka_logstash_e2e.py
```diff
@@ -0,0 +1,1153 @@
...
+# media-server.
+KAFKA_E2E_FILE_URL = os.environ.get("KAFKA_E2E_FILE_URL", "http://media-server/2min.mp4")
+
+# Known-good NVIDIA-internal RTSP source (looped MP4 served via Wowza). RTVI
+# pulls real frames from this URL, the VLM produces real captions, and
+# Logstash writes real raw_events docs to ES. Override via env if needed.
+# Matches run_sanity_kafka.sh's SANITY_RTSP_URL default.
+RTSP_URL = os.environ.get(
+    "KAFKA_E2E_RTSP_URL",
+    "rtsp://nv-wowza-pdc.nvidia.com:1935/vod/warehouse_1.mp4",
```
zac-wang-nv, line 144: This default leaks an internal NVIDIA Wowza streaming host onto the public repo, and the test is unusable off the corp network. Please make the RTSP URL a required env var (skip when unset) or point the default at the in-stack media-server RTSP source, and drop the hostname from the docstring/comments.

services/video-summarization/ci/pipeline-helpers.groovy
```diff
@@ -0,0 +1,6503 @@
...
+        'coverage combine --data-file=/workspace/.coverage.kafka-e2e || true && ' +
+        'coverage xml --data-file=/workspace/.coverage.kafka-e2e -o /workspace/coverage_reports/coverage-kafka-e2e.xml && ' +
+        'coverage html --data-file=/workspace/.coverage.kafka-e2e -d /workspace/htmlcov-kafka-e2e && ' +
+        'coverage report --data-file=/workspace/.coverage.kafka-e2e > /workspace/coverage_reports/coverage-kafka-e2e-summary.txt && ' +
+        'python3 /workspace/ci/utils/convert_junit_to_csv.py /workspace/test_kafka_logstash_e2e-report.xml /workspace/kafka-e2e-test-results.csv ' +
+        '|| RC=1; exit \\$RC"'
+
+    def testRC = 0
+    try {
+        testRC = runTestsInDocker(useSudo, dockerRunArgs, testCommand)
```
zac-wang-nv, line 2651: `runTestsInDocker()` only *builds and returns a shell-script string* (line 2134) — it never runs `sh`. So the Kafka→Logstash E2E pytest never executes and this stage would "pass" unconditionally. This is the exact bug your own comment at ~3050 documents fixing in `runEsShardLimitTest`; that fix wasn't applied here. It's latent today only because the call-site is hard-disabled with `if (false && …)` (line 5471 on the current head) — please wrap the command in `sh(script: …, returnStatus: true)` before that gate is ever re-enabled (and note `Jenkinsfile.nightly.multiarch:140` already forwards `RUN_KAFKA_E2E=true`, so nightly claims coverage it can't run).

services/video-summarization/ci/pipeline-helpers.groovy
```diff
@@ -0,0 +1,6503 @@
...
+    def targetComposeDir = composeFilePath.replaceAll(/\/[^\/]+$/, '')
+    def targetComposeFile = composeFilePath.replaceAll(/^.*\//, '')
+    def imageReplacementCmd = ""
+    def manifestCheckCmd = ""
+
+    if (builtImageTag != null) {
+        echo "Will replace docker-compose.yml image with built image: ${builtImageTag}"
+        imageReplacementCmd = """
+            echo "Replacing LVS image with: ${builtImageTag}"
+            sed -i 's|image: nvcr.io/.*/vss-video-summarization:.*|image: ${builtImageTag}|g' ${targetComposeFile}
```
zac-wang-nv, line 3464: `params.LVS_IMAGE` flows through `getLatestImage()` (returns user input verbatim) into `prepareDockerComposeDeployment`, where `builtImageTag` is Groovy-interpolated into a `sed`/`docker` `sh` block. A value with a single quote or `sed` metacharacter escapes the quoting and runs arbitrary shell on the GPU node (which holds NGC/Vault creds in env). The standalone path validates via `resolveImageTagForArch` (`^[A-Za-z0-9._:\/-]+$`); this path doesn't. Please apply the same charset validation (and consider validating `GPU_TYPE` before splicing it into the pod YAML at `Jenkinsfile.dind_with_gpu:86`).

services/video-summarization/tests/integration/test_rtvi_e2e.py
```diff
@@ -0,0 +1,500 @@
...
+
+import pytest
+import requests
+
+logger = logging.getLogger(__name__)
+
+# -- Skip gate ---------------------------------------------------------------
+
+pytestmark = pytest.mark.skipif(
+    os.environ.get("RTVI_E2E_TEST", "1") != "1",
```
zac-wang-nv, line 79: This skip gate is **opt-out** (`RTVI_E2E_TEST` defaults to on) while every other e2e module (`KAFKA_E2E_TEST`, `ES_SHARD_LIMIT_TEST`) is opt-in. A plain `pytest tests/integration/` on a fresh public checkout runs this module, and the `lvs_ready`/`rtvi_ready` fixtures poll for 300s/600s before `pytest.fail` — so it hangs ~5–10 min then reports 9 failures on any machine without the compose stack. Please make it opt-in for consistency, or add a fast connectivity pre-flight that `pytest.skip`s immediately when LVS/RTVI are unreachable.

services/video-summarization/tests/conftest.py
```diff
@@ -0,0 +1,193 @@
...
+    required_env_vars = [
+        "OPENAI_API_KEY",
+        "NGC_API_KEY",
+        "NVIDIA_API_KEY",
+        "VIA_VLM_API_KEY",
+        "HF_TOKEN",
+    ]
+    missing = [var for var in required_env_vars if not os.environ.get(var)]
+    if missing:
+        pytest.exit(
```
zac-wang-nv, line 146: `integration_test_setup` calls `pytest.exit(..., returncode=1)` when API keys are unset — this aborts the **entire** pytest session, so a public user running `pytest tests/` loses the unit tests that would otherwise pass. Please use `pytest.skip(...)` inside the fixture so only the tests requesting it are skipped, and avoid hard-coding `/opt/nvidia/...` download paths (not writable on most hosts).

services/video-summarization/tests/integration/test_rtvi_integration.py
```diff
@@ -0,0 +1,790 @@
...
+        ),
+        "audio_transcript": "",
+    }
+
+
+# ── Fixtures ──────────────────────────────────────────────────────────────────
+
+
+@pytest.fixture(scope="module")
+def mock_rtvi(free_tcp_port_factory):
```
zac-wang-nv, line 251: `free_tcp_port_factory` isn't defined in any conftest, nor provided by a declared test dependency (CI installs only `pytest-timeout coverage sseclient-py`; pytest-asyncio's equivalent is `unused_tcp_port_factory`). As written, all 23 tests in this module error at setup with "fixture not found". If it comes from an image-only plugin, please pin/document it; otherwise add a small socket-based implementation to `tests/conftest.py`.

services/video-summarization/tests/conftest.py
```diff
@@ -0,0 +1,193 @@
...
+                pytest.skip(
+                    f"Artifactory returned {e.code} for {url}. "
+                    "Set ARTIFACTORY_USER and ARTIFACTORY_TOKEN to run integration tests."
+                )
+            raise
+
+
+# Base URL for lmm/streams (used by asset_manager unit tests and integration tests).
+ARTIFACTORY_LMM_STREAMS_BASE = (
+    "https://artifactory.nvidia.com/artifactory/sw-ds-generic-bld-local/lmm/streams"
```
zac-wang-nv, line 68: The shared unit-test conftest bakes NVIDIA-internal Artifactory URLs and `/opt/nvidia/...` write paths into the public repo's top-level test config, and `artifactory_lmm_streams_mp4` does live network I/O from an otherwise hermetic setup. External contributors can't use any of it. Consider moving the download fixtures into `tests/integration/conftest.py` and parameterizing the asset base URL via an env var. (Same internal-Artifactory default recurs across ~10 functional/integration test files as the only test-asset source — worth a single env-var knob.)

services/video-summarization/tests/conftest.py
```diff
@@ -0,0 +1,193 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
zac-wang-nv, line 1: On a clean checkout `pytest tests/unit` aborts at collection with `FileNotFoundError: /tmp/via-logs/via_engine.log` — importing `via_logger` (`src/via_logger.py:135`) opens a log file in a directory it never creates. Rather than relying on the README's `mkdir` line, add `os.makedirs(os.environ.get('VIA_LOG_DIR', '/tmp/via-logs'), exist_ok=True)` at the top of this conftest so the suite runs out of the box.

services/video-summarization/compose/run_benchmark.sh
```diff
@@ -0,0 +1,475 @@
...
+    set +u
+    source vss-perf-env/bin/activate
+    set -u
+fi
+info "Activated virtualenv: vss-perf-env"
+export VIA_BACKEND="http://localhost:$LVS_PORT"
+[ -n "$VLM_GPUS" ] && export VIA_VLM_GPUS="$VLM_GPUS"
+[ -n "$LLM_GPUS" ] && export VIA_LLM_GPUS="$LLM_GPUS"
+eval "$BENCHMARK_CMD"
+benchmark_exit=$?
```
zac-wang-nv, line 458: With `set -euo pipefail`, `eval "$BENCHMARK_CMD"` aborts the script on a nonzero exit before `benchmark_exit=$?` runs — so the failure-log dump (462–466) and the `-d` teardown (469–473) are dead code, leaving GPU stacks running after failed CI runs. Use `benchmark_exit=0; eval "$BENCHMARK_CMD" || benchmark_exit=$?`. Same pattern in `upload_perf_results.sh:125`.

services/video-summarization/compose/run_benchmark.sh
```diff
@@ -0,0 +1,475 @@
...
+    error "ARTIFACTORY_USER and ARTIFACTORY_TOKEN must be set"
+    echo "  Example:"
+    echo "  export ARTIFACTORY_USER='<username>'"
+    echo "  export ARTIFACTORY_TOKEN='<token>'"
+    echo "  # or set them in compose/.env"
+    exit 1
+fi
+
+# --- Optional: override LVS image in compose file (e.g. for CI or local -I) ---
+if [ -n "$IMAGE_TAG_OVERRIDE" ]; then
```
zac-wang-nv, line 272: The `-I` override does `sed -i "s|image: nvcr.io/.*/vss-video-summarization:.*|…|g"`, but every compose file here writes `image: ${LVS_IMAGE:-nvcr.io/...}`, so the pattern never matches and `-I` silently no-ops (and where it could match, it dirties a tracked file). Since the variable already exists, `export LVS_IMAGE="$IMAGE_TAG_OVERRIDE"` is both correct and avoids mutating the checkout.

services/video-summarization/compose/BlueprintBuilderGenerated/docker-compose.yml
```diff
@@ -0,0 +1,600 @@
...
+      LVS_MCP_PORT: '38112'
+      CA_RAG_CONFIG: /opt/nvidia/via/config/default_config.yaml
+      DISABLE_CA_RAG: ${DISABLE_CA_RAG:-false}
+      ENABLE_AUDIO: ${ENABLE_AUDIO:-false}
+      ENABLE_DENSE_CAPTION: ${ENABLE_DENSE_CAPTION:-}
+      ENABLE_VIA_HEALTH_EVAL: ${ENABLE_VIA_HEALTH_EVAL:-false}
+      LVS_ENABLE_MCP: ${LVS_ENABLE_MCP:-true}
+      GRAPH_DB_HOST: graph-db
+      GRAPH_DB_USERNAME: neo4j
+      GRAPH_DB_PASSWORD: passneo4j
```
zac-wang-nv, line 144: The `lvs` service hardcodes `GRAPH_DB_PASSWORD`/`ARANGO_DB_PASSWORD`/`MINIO_PASSWORD` as literals (144, 149, 154) while the DB services read `${GRAPH_DB_PASSWORD:-passneo4j}` etc. (447, 463). Overriding a password in `.env` changes the database but leaves the `lvs` client on the old literal → auth failure. The defaults are template-normal; the client/server inconsistency is the bug. Make the `lvs` entries `${…:-default}` so both sides stay in sync.

services/video-summarization/compose/otel-stack.yaml
```diff
@@ -0,0 +1,66 @@
...
+    restart: unless-stopped
+
+volumes:
+  phoenix_data:
+    driver: local
+
+networks:
+  app-network:
+    external: true
+    name: compose-files_app-network  # Match your existing network
```
zac-wang-nv, line 66: This external network name looks like a leftover from an old directory name — none of the stacks in this tree create it (e.g. the 4-GPU stack produces `x86-rtvi-cr-nemo3-nano-4gpu_app-network`), so the otel stack can't attach without editing the file. Consider parameterizing it (`name: ${APP_NETWORK:?set to your compose project network}`) so `docker compose up` fails with a clear message.

services/video-summarization/compose/overlays/gpu-h100.yaml
```diff
@@ -0,0 +1,21 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# H100 single-GPU dense layouts prefer x86-rtvi-cr-nemotron-9b_1gpu.yaml.
+# ─────────────────────────────────────────────────────────────────────────────
+services:
+  rtvi-vlm:
+    environment:
+      MODEL_PATH: "ngc:nim/nvidia/cosmos3-nano-reasoner:modelopt-fp8-final_format_fix"
```
zac-wang-nv, line 21: Several default artifacts across the compose tree look internal/pre-GA and may not be publicly pullable: the `cosmos3-nano-reasoner:*final_format_fix` NGC tags (all overlays), `nvidia-nemotron-nano-9b-v2-dgx-spark:1.0.0-variant` (a literal "variant" placeholder tag, spark/thor stacks), and HF `nvidia/Nemotron-Nano-V3-Omni-GA0420-FP8`. Please confirm each is published to public NGC/HF at merge time and swap dev tags for released ones — otherwise every stack here fails at pull for external users.

services/video-summarization/tests/scripts/get_summary_qa_results_into_xlsx.py
```diff
@@ -0,0 +1,1521 @@
...
+
+            # Store QA pairs if not already stored for this video
+            if video_data[video_name]["qa_pairs"] is None:
+                try:
+                    video_data[video_name]["qa_pairs"] = df[["Question", "Answer"]].copy()
+                except Exception:
+                    try:
+                        video_data[video_name]["qa_pairs"] = df[["question", "answer"]].copy()
+                    except Exception as e:
+                        return e
```
zac-wang-nv, line 162: `process_results` returns the exception *object* mid-loop (also lines 175, 624) instead of raising or continuing, so one malformed CSV discards all other folders and hands an `Exception` to `create_excel_report`, which then crashes with an unrelated `AttributeError`. Either `raise`, or log and `continue` to the next folder. (Separately — this 1,521-line script consumes the internal BYOV harness outputs that are `--ignore-glob`'d out of this repo, so no public user can produce its inputs. Along with `ask_on_files.py`, worth asking whether these internal QA-reporting tools belong in the public tree at all — see the split proposal in Comment 24.)

services/video-summarization/tests/functional/test_api_endpoints.py
```diff
@@ -0,0 +1,634 @@
...
+
+    script_dir = Path(__file__).parent
+
+    # Build pytest arguments
+    pytest_args = [
+        str(script_dir / "test_api_endpoints.py"),
+        "-m",
+        marker_map[args.category],
+        f"--base-url={args.base_url}",
+        f"--timeout={args.timeout}",
```
zac-wang-nv, line 606: `--timeout` is consumed by **pytest-timeout** as a per-test kill switch (and, with `timeout=240` set in pyproject, overrides it down to 30s) — killing the summarization tests this file itself says take 30–60s — while the intended HTTP timeout (`--http-timeout` in the conftest) silently stays at its default. Please pass `--http-timeout={args.timeout}` here.

services/video-summarization/.flake8
```diff
@@ -0,0 +1,31 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+[flake8]
+max-line-length = 110
+extend-ignore = E203, W503, E501
```
zac-wang-nv, line 18: `E501` in `extend-ignore` makes `max-line-length = 110` dead config, and it conflicts with this service's ruff `line-length = 100` in `pyproject.toml` — one tree, two lint tools, disagreeing (and one unenforced) limits. The repo's GitHub CI already standardizes on ruff; consider dropping `.flake8` and consolidating there (note the new Jenkinsfile does invoke flake8, so they'd need to align on the same tool).

zac-wang-nv:
**Scope / public-repo fit.** The PR is titled "LVS unit tests and CI," but roughly two-thirds of the +42.7k lines are internal infrastructure that can't run outside NVIDIA and that discloses internal topology: `ci/**` (Jenkins pipelines wired to internal GitLab `gitlab-vss-lvs`, K8s cloud `blsm-prod-cloud`, private registries `nvcr.io/nv-metropolis-dev/…` and `gitlab-master.nvidia.com:5005/…`, ~15 Jenkins credential IDs, service-account/pull-secret names, NSPECT ID, an internal Slack link), and `compose/BlueprintBuilderGenerated/` (a checked-in generated stack + ~4k lines of sanity harnesses, incl. an NCCL debug README for a `docker-compose.yml` that isn't in the folder, and two ~95%-identical 1k-line kafka scripts). The unit-test suite itself is genuinely good and hand-tuned (~95% meaningfully test behavior). The recent commits ("rm old files", "rem some refs") show this trimming is already underway — the DASHBOARD_API_URL internal IP, the `ask_on_files.py` internal IP/model, and `tests/.env` have all been cleaned up. Recommend finishing the job: keep `ci/` and the internal harnesses in the internal mirror only, or sanitize what remains (vault paths, credential IDs, internal hosts, Jira IDs like `METLVSMS-500`, internal-repo references). See the concrete split proposal in the next comment.

zac-wang-nv:
**Proposed split: move the internal infrastructure to `metromind/ci-vss-oss` instead of the public repo.** There's already an internal home for all of this, built around the exact pattern that keeps internal CI *out* of the public tree: `ci-vss-oss` pulls this repo in as a **git submodule** (`.gitmodules` → `video-search-and-summarization`, tracking `develop`) and drives internal pipelines against the public deployment assets without duplicating them — e.g. `.gitlab-ci.yml` already sets `BLUEPRINT_DEPLOYMENTS_DIR = video-search-and-summarization/deploy/docker` and `VSS_PARENT_CHARTS_DIR = video-search-and-summarization/deploy/helm/developer-profiles`. So the internal CI can consume anything this repo publishes under `deploy/` via the submodule — it does not need those files copied into the public tree.

Concretely, most of this PR maps onto directories `ci-vss-oss` already has:

| This PR (public) | Suggested target in `ci-vss-oss` | Why |
|---|---|---|
| `services/video-summarization/ci/Jenkinsfile.*`, `pipeline-helpers.groovy`, `pod-templates/`, `ci/utils`, `ci/scripts` | `ci/jenkins/`, `ci/tools/` | Jenkins/GitLab pipeline code + internal node/registry topology already live here |
| `ci/Jenkinsfile.perf`, `perf-configs.yaml`, `run_benchmark.sh`, `upload_perf_results.sh`, the missing `perf/benchmark/` harness | `eval/` (has `compose.yml`, `run_eval.sh`, `configs/`, `monitoring/`, `scripts/`) | eval/perf harness with its own compose + node config already exists |
| `compose/BlueprintBuilderGenerated/` (generated stack + `run_sanity*.sh`, NCCL debug README) | `eval/` or `tests/` | internal sanity/e2e harness, not a user-facing profile |
| `DASHBOARD_API_URL` (already fixed to read `env.DASHBOARD_API_URL`) | `dashboard/` (the Dev Dashboard service lives here) | the dashboard *is* an internal service in `ci-vss-oss`; the env-var indirection you just added is the right pattern |
| `fetchCredsFromVault` inline vault paths (Comments 2–3) | `vault-secrets.json` manifest | `ci-vss-oss` declares secrets as `{secretPath, secretData, envVar}` entries, never inlined into pipeline code — adopting this removes the leak by construction |
| `tests/scripts/get_summary_qa_results_into_xlsx.py`, `ask_on_files.py` | `ci-vss-oss` (internal tooling) | consume the internal BYOV/nvdev harness, unusable publicly |

**What stays in this public PR:** the unit suite (`tests/unit/**` — genuinely good, keep it), the `.flake8`/`pyproject`/`sonar` config, and any *sanitized, publicly-runnable* compose profiles that belong next to the existing `deploy/docker/` tree (note the public repo already ships `deploy/docker/{compose.yml,developer-profiles,industry-profiles}`, which the new `compose/` tree partly overlaps — worth reconciling rather than adding a parallel one). Net effect: the public repo gets the tests + a runnable profile, `ci-vss-oss` gets the pipelines/perf/dashboard/secrets via the submodule it already tracks, and none of the vault/credential/host leakage in Comments 2–6 ever reaches the public tree. Happy to pair on the migration — the submodule bump on the `ci-vss-oss` side is a one-line pointer update.

zac-wang-nv:
**Merge blockers, please address before merge:**
1. **CI red** — the 7 `validate_required_prompts` unit tests fail deterministically (Comment 1); `pytest tests/unit` won't be green.
2. **PR description** — the checklist is now ticked and the title reads "LVS CI Pipeline", but the **Description section is still empty**. For a 119-file / 42.7k-line change please add a real description and, given the scope drift above, an explicit note on why the CI/compose/perf trees are included. (DCO is now satisfied — all commits on the rebased branch carry a `Signed-off-by` trailer. 👍)

zac-wang-nv:
**Doc staleness roundup** — several new guides describe things that don't exist in this repo or contradict the code they document, which will mislead external readers:
- `perf/benchmark/` (referenced by `Jenkinsfile.perf`, `run_benchmark.sh`, `upload_perf_results.sh`, `PERF_PIPELINE_GUIDE.md`) is **not in the repo** — the entire perf path dangles. Either include the (sanitized) perf harness or drop the perf pipeline + docs.
- `PERF_PIPELINE_GUIDE.md` — its config-ID inventory, node labels, and parameter defaults don't match `perf-configs.yaml` / `Jenkinsfile.perf`.
- `tests/README.md` — references `test_perf`, `test_qa_video_file_2_accuracy`, `src/models/vila15/VILA`, and `pytest.ini`, none of which exist here; never documents how to run the new unit suite.
- `CI_FAST_REBUILD.md` / `DEVELOP_MULTIARCH_PIPELINE_GUIDE.md` / `compose/README.md` — each disagrees with the code (app-only prefix set, the gate implementation, and Cosmos-Reason2 vs the `cosmos3-nano-reasoner`/`cosmos-reason3` the configs actually deploy).

## Instructions
- Change VERDICT to RESOLVE / ACK / FOLLOW-UP / SKIP as needed (bot threads are all informational SKIPs)
- Edit any reply or comment text directly
- Delete any comment block to skip it
- Save the file, then reply **"post"** to execute

PR #1323, opened 2026-07-18
fix(search): map screenshot timestamps onto the VST replay timeline

zac-wang-nv:
## What

Search hits for **file-ingested** sources are indexed on a synthetic, midnight-anchored epoch (`2025-01-01T00:01:00Z` = 60 s into the file), while VST anchors its replay timeline at **ingest wall-clock**. Screenshot URLs built from raw ES timestamps point outside the recording, so VST rejects every one:

```
GET /vst/api/v1/replay/stream/<id>/picture?startTime=2025-01-01T00:01:00Z
→ 500 {"error_code":"VMSInternalError","error_message":"no valid stream found for given timestamps"}
```

Observed live on the `vss-search-archive` Harbor eval (run 29573058835, PR #1170): step-3 media validation failed 5/6 on **every** attempt once the public origin was reachable. The old NAT `vst.snapshot` tool did this translation (`get_timeline` + offset) for its `offset` mode; the NAT-free direct-URL contract skipped it.

## How

- `lib.vst.map_timestamp_to_timeline` — pure mapper: in-timeline wall-clock (live RTSP) passes through; out-of-timeline timestamps get their midnight-anchored offset re-based onto the timeline start, clamped to the end; unparseable input passes through.
- `lib.vst.get_timelines_map` (+ `VSTClient` method) — one `/storage/timelines` call for all streams, collapsed to envelopes.
- Wired at all three build sites (embed `_hit_to_item`, `enrich_attribute_results`, `_attach_screenshots`), best-effort everywhere — timeline-fetch failure degrades to unmapped URLs, never drops hits. `getattr` guard keeps existing `VSTSnapshot` fakes working.

## Review notes

- The midnight-anchor assumption for the synthetic epoch matches the `mdx-*-2025-01-01` fixture convention; if ingestion records an explicit per-video ES epoch somewhere, deriving the offset from that would be stricter — pointer welcome.
- Draft until the eval re-runs green on top of this; diagnosis thread on PR #1170.

## Verification

NAT-free lane: 610 passed (incl. new mapping tests); mypy (51 files) and ruff clean.

### DIFF
--- a/services/agent/src/lib/search_core/primitives/_attribute_helpers.py
+++ b/services/agent/src/lib/search_core/primitives/_attribute_helpers.py
@@ -44,9 +44,12 @@ from lib._foundation.time import datetime_to_iso8601
 from lib._foundation.time import iso8601_instants_match
 from lib._foundation.time import iso8601_to_datetime
 from lib._foundation.time import safe_iso8601_to_datetime
+from lib.vst import VSTError
 from lib.vst import build_screenshot_url
 from lib.vst import get_stream_id
 from lib.vst import get_timeline
+from lib.vst import get_timelines_map
+from lib.vst import map_timestamp_to_timeline
 
 from .._internal.coerce import _coerce_str
 from .._internal.es_filters import build_video_sources_filter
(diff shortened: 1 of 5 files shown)


zac-wang-nv:
Hardened after full-suite validation: the timelines fetch is now **conditional** (only when a result actually needs a screenshot) and **single-attempt with a 5s timeout** — best-effort enrichment must not add retry backoff to searches when VST is slow/down. Full unit suite: **2961 passed** (agent + lib + cli); NAT-free lane 610; mypy/ruff clean.

PR #1348, opened 2026-07-21
feat(spatialai-data-utils): scope AICity MTMC eval to GT frames + fix missing-warehouse HOTA weighting

yizwang-nv:
## Description

Brings the `spatialai-data-utils` sub-package in line with the canonical repo, rebased cleanly onto the latest `develop` — a **9-file diff** with no unrelated commits.

**Supersedes #1295**, which is based on a stale, since-force-pushed `develop` (297 commits / 1,117 files / `mergeable_state: dirty`) and only carried the frame-scoping change. This PR is the same feature plus the participant-reported fix, on a clean base.

Two evaluation fixes for the AICity MTMC (Multi-Camera 3D People Tracking) HOTA evaluator:

1. **Scope evaluation to GT-annotated frames** — predictions on frames the ground truth does not annotate are ignored instead of being counted as false positives, so a full submission evaluated against a partial/truncated GT is not unfairly penalised. (This is the change from #1295.)

2. **Fix missing-warehouse HOTA weighting** (reported by a Track-1 participant) — a warehouse present in the evaluation GT but missing from a submission is now scored `HOTA=0` at its full GT weight, instead of being dropped from the GT-count-weighted mean. Previously a partial submission covering only a subset of warehouses could out-score a complete one. A scene truncated to 0 GT rows still contributes nothing (weight 0).

Smaller self-contained fixes to the same sub-package:
- Narrow the optional `cv2` import guard to `ModuleNotFoundError` so a broken OpenCV build no longer masquerades as "not installed".
- Derive SHA-256 staging filenames in the cloud download path to avoid collisions.
- Require `GCS_ENDPOINT_URL` when `STORAGE_PROVIDER=gcs`; doc fixes (cluster-map output filename, docker entrypoint invocation).

### Verification
- New regression tests: a partial submission can no longer beat — or exceed a perfect score above — an equivalent full submission; plus unit coverage of the weighted-average aggregator.
- Non-regression: the bundled full sample submission still scores weighted **HOTA 20.15** (unchanged).
- Tracking test suite passes (50 tests).

> Follow-up: the `spatialai-data-utils` package version bump (the open TODO from #1295) is still pending a decision.

## Checklist

- [x] I have installed and run pre-commit hooks locally.

### DIFF
--- a/libs/analytics/spatialai-data-utils/README.md
+++ b/libs/analytics/spatialai-data-utils/README.md
@@ -116,14 +116,14 @@ index:
 pip install 'torch>=2.10.0' --index-url https://download.pytorch.org/whl/cpu   # or CUDA build
 pip install 'pytorch3d @ git+https://github.com/facebookresearch/pytorch3d.git@33824be' \
     --no-build-isolation
-pip install spatialai-data-utils==2.0.1 \
+pip install spatialai-data-utils==2.0.2 \
     --extra-index-url=https://edge.urm.nvidia.com/artifactory/api/pypi/sw-metropolis-pypi/simple
 # add optional extras as needed, e.g. the evaluation stack:
-#   pip install 'spatialai-data-utils[eval]==2.0.1' --extra-index-url=...
+#   pip install 'spatialai-data-utils[eval]==2.0.2' --extra-index-url=...
 ```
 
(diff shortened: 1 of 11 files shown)


yunHe95 (approved):
LGTM

zac-wang-nv (approved):

PR #1352, opened 2026-07-22
fix(ci): retry transient failures when triggering the downstream pipeline

simings-nv:
## What
Retry transient failures in `trigger-downstream-pipeline.sh`. `request_json()` (the shared helper for every downstream GitLab call, including the pipeline trigger) did a single `urlopen` with no retry, so any transient connection error or HTTP 429/5xx failed the whole bridge job.

## Why
The bridge job holds a self-hosted runner for up to 245 minutes. A single transient blip on the trigger or a GitLab API lookup would fail that job outright and force a full re-trigger (another held slot plus a re-vet).

## Honest scope
This is **preventive hardening, not a fix for an observed failure.** In recent CI history the `Trigger Downstream Pipeline` job is the top red source, but its failures are:
- ~65 x `Poll pipeline` (the downstream pipeline genuinely failed; retry cannot and should not mask that)
- ~12 x `Set up runner` (self-hosted runner provisioning, which happens before this script runs)
- ~1 x other

None were transient trigger/connection failures, so there is no example run of the exact failure this fixes. It closes a real fragility (a 245-min runner held on a zero-retry call) that simply has not bitten in the recent window.

## Change
Bounded retry loop (4 attempts, exponential backoff 2/4/8s) on connection errors and HTTP 429/500/502/503/504; fails fast on non-transient 4xx. Emits a `::warning::` per retry. The existing secret-safe error extraction on final failure is unchanged.

### DIFF
--- a/.github/scripts/trigger-downstream-pipeline.sh
+++ b/.github/scripts/trigger-downstream-pipeline.sh
@@ -6,6 +6,7 @@ import re
 import socket
 import ssl
 import sys
+import time
 from typing import Any
 from urllib.error import ContentTooShortError
 from urllib.error import HTTPError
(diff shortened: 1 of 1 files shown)


simings-nv:
Closing to keep the review queue reflecting work that is actively wanted. This has been open since July without review; the change is small and self-contained, so reopening or re-raising it later is cheap if the need is still there.

PR #1353, opened 2026-07-22
chore(ci): add yamllint pre-commit hook mirroring the CI step

simings-nv:
## What
Add yamllint as a pre-commit hook, mirroring the CI "Lint (Python)" yamllint step, so a YAML error is caught locally in ~1s instead of failing a CI run.

## Why
The "Lint (Python)" CI job runs two linters: **ruff** (Python) and **yamllint** (YAML). ruff is already a pre-commit hook, but yamllint is not, so a stray YAML change (e.g. an extra blank line producing an `empty-lines` error) is not caught until it burns a CI run and forces a re-push.

This is not hypothetical. In recent history it has failed CI on **4 PRs (6 runs, including re-runs)**, every one on the same trivial `too many blank lines` error:

- PR #1254 (failed 3x): [run 1](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/29761458387/job/88417254931), [run 2](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/29622349214/job/88019940118), [run 3](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/29780787899/job/88481894327)
- PR #1319: [run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/29616593416/job/88429462099)
- PR #1320: [run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/29616441981/job/88003030585)
- PR #1348: [run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/29864691739/job/88750687313)

Each is a full CI round-trip (and often a maintainer re-trigger) spent on a blank line. Shifting the check left removes the round-trip. Python lint (ruff) already never gates because it is in pre-commit; this closes the same gap for YAML.

## Change
A `repo: local` hook running `uvx yamllint` with the same config and file scope as the CI step. Run ephemerally via `uvx` so yamllint (GPL-3.0-or-later) never lands in the project venv or lockfile, matching the CI step's rationale. It lints only changed in-scope files; warnings are non-blocking (yamllint exits 0 on warnings), so it fails only on real YAML errors, exactly like CI.

PR #1354, opened 2026-07-22
feat(ci): skip downstream pipeline on docs/CI-only PRs

simings-nv:
## What
Skip the downstream blueprint pipeline on PRs whose changes cannot affect the built blueprint (docs- or CI-only changes).

## Why
The downstream pipeline runs on a self-hosted GPU runner and holds it for up to 245 minutes, and it triggers on every PR regardless of whether the PR can affect the blueprint.

## Evidence (recent CI history)
- **Self-hosted pool contention:** the `downstream-pipeline` runner failed at `Set up runner` **12 times** in the recent window, before the job could even start. Not triggering PRs that cannot affect the blueprint relieves that contention.
- **Docs/CI-only PRs needlessly trigger the full 245-min eval.** Concrete examples: this very stack of CI-only PRs, which touch only `.github/` and `.pre-commit-config.yaml` and cannot change the blueprint, each trigger the eval: #1352, #1353, #1354 (this PR).

## Change
In `trigger-downstream-pipeline.sh`, before triggering, check the PR's changed files (reusing the existing changed-file helpers) and skip when every changed file is clearly irrelevant (`.github/`, `docs/`, `*.md`, `.pre-commit-config.yaml`, license files). The job exits 0, so its check goes green without holding a runner.

Fail open: any change not clearly irrelevant, or any inability to determine the changed set, still runs the eval, so a real change is never skipped. Applies to PR runs (`pull-request/<N>`) only, never `main`/`develop` pushes. Also single-sources the changed-file computation shared with the launchable-notebook check.

## Live example (this PR stack)
#1353, a `.pre-commit-config.yaml`-only PR, has every one of its ~20 checks green except one: the downstream `Trigger Downstream Pipeline`, which is red solely because the downstream `helm static warehouse charts` gate fails. That gate fails **main-wide, on every PR, regardless of content**: it renders warehouse Helm charts from `deploy/helm/industry-profiles/warehouse-operations/` in the resolved VSS checkout, but that path does not exist on `develop` (the warehouse profile ships as docker-compose assets, not Helm charts — there are zero warehouse `Chart.yaml`). The gate has a "chart root missing -> skip" path, but it escalates to a hard failure whenever an explicit VSS ref is set, which every PR's downstream trigger does. So a `.pre-commit-config.yaml` change cannot possibly affect it, yet it blocks the PR. This PR would skip the downstream entirely for such a CI-only change, so a main-wide downstream failure would no longer block a trivial CI/docs PR. (The gate itself also needs a separate fix in ci-vss-oss; see notes.)

## Note
Draft for discussion. This is the surgical, safe half of "decouple the bridge" (it removes docs/CI-only PRs from the eval and relieves the self-hosted runner pool). It intentionally does NOT change the blocking semantics for code PRs; making the downstream advisory for code PRs is a separate, policy-level decision.

### DIFF
--- a/.github/scripts/trigger-downstream-pipeline.sh
+++ b/.github/scripts/trigger-downstream-pipeline.sh
@@ -288,16 +288,46 @@ def push_event_changed_files() -> set[str]:
     return filenames
 
 
-def launchable_notebook_changed() -> bool:
-    """Return true when this run's VSS changes touch the launchable notebook."""
+# Changes that cannot affect the built blueprint (services/ + deploy/), so the
+# downstream blueprint pipeline would only re-eval identical inputs. Fail OPEN:
+# run the eval on anything not clearly in this set, so a real change is never
+# skipped.
+_IRRELEVANT_PREFIXES = (".github/", "docs/")
+_IRRELEVANT_SUFFIXES = (".md", ".rst", ".txt")
+_IRRELEVANT_EXACT = frozenset({".pre-commit-config.yaml", "LICENSE", "NOTICE", ".gitignore"})
+
+
+def run_changed_files() -> set[str]:
+    """Changed filenames for this run (PR synthetic branch or push); empty if unknown."""
     ref_name = os.environ.get("GITHUB_REF_NAME", "").strip()
     pr_match = re.fullmatch(r"pull-request/(\d+)", ref_name)
     if pr_match:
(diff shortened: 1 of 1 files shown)


simings-nv:
Closing to keep the review queue reflecting work that is actively wanted. This has been open since July without review; the change is small and self-contained, so reopening or re-raising it later is cheap if the need is still there.

PR #1364, opened 2026-07-22
fix(rtvi): make codec installation resilient

hugoverjus:
Resolve codec packages from signed HTTPS Ubuntu metadata with retries so security updates cannot invalidate hard-coded URLs, and validate both RTVI installers in CI.

## Checklist


services/rtvi/rt-vlm/src/scripts/install_codecs_nonroot.sh
```diff
@@ -0,0 +1,235 @@
...
+    echo "ERROR: Failed to download one or more codec packages" >&2
+    exit 1
+fi
+
+shopt -s nullglob
+DEBS=("$DEBS_DIR"/*.deb)
+shopt -u nullglob
+if [ ${#DEBS[@]} -ne ${#PACKAGES[@]} ]; then
+    echo "ERROR: Expected ${#PACKAGES[@]} codec packages, downloaded ${#DEBS[@]}" >&2
+    exit 1
```
zac-wang-nv, line 211: `dpkg -x` failures aren't checked (the script has no `set -e`), and `.installed` is touched unconditionally at L228 — so a partial extraction (disk full, corrupt archive) gets permanently masked: the next container start hits the early-exit guard and never repairs the install. This behavior predates the PR, but since the theme here is resilience, worth fixing while the file is being rewritten:

```bash
dpkg -x "$deb" "$INSTALL_DIR/" || {
    echo "ERROR: Failed to extract $deb" >&2
    exit 1
}
```

(mirrored into the rt-embed copy — the CI `cmp` gate will insist anyway).

.github/workflows/ci.yml
```diff
@@ -134,6 +134,32 @@ jobs:
...
+      - name: Validate codec installers
+        run: |
+          EMBED_INSTALLER=services/rtvi/rt-embed/src/scripts/install_codecs_nonroot.sh
+          VLM_INSTALLER=services/rtvi/rt-vlm/src/scripts/install_codecs_nonroot.sh
+          bash -n "$EMBED_INSTALLER"
+          bash -n "$VLM_INSTALLER"
+          cmp "$EMBED_INSTALLER" "$VLM_INSTALLER"
+          CODEC_VALIDATE_ONLY=true bash "$EMBED_INSTALLER"
+
   # ---------------------------------------------------------------------------
```
zac-wang-nv, line 164: This only exercises the amd64/archive.ubuntu.com path — the arm64/ports.ubuntu.com resolution used by the SBSA image (`vss-rt-vlm:*-sbsa`) is never validated, because `DEB_ARCH` is hard-derived from `dpkg --print-architecture`. Cross-arch validation works without qemu: I verified locally that with `APT::Architecture=arm64` and the ports sources, an amd64 host resolves arm64 .debs fine. Suggest making the arch overridable in the installer (`DEB_ARCH=${CODEC_DEB_ARCH:-$(dpkg --print-architecture)}`) and adding a second invocation here with `CODEC_DEB_ARCH=arm64`, so a ports-side or arm64-package regression can't slip through CI green.

.github/workflows/ci.yml
```diff
@@ -134,6 +134,32 @@ jobs:
             -d "{extends: default, rules: {line-length: {max: 200}, document-start: disable, indentation: {spaces: 2, indent-sequences: consistent}}}" \
             $YAML_FILES
 
+  codec-installer-test:
```
zac-wang-nv, line 137: `trigger-downstream-pipeline` gates on the other fast checks (lint, typecheck, security, copyright-headers, …) but not on this job. Since it's cheap (~1 min: the local validate run took ~6s plus setup) and directly guards image behavior, consider adding `codec-installer-test` to that `needs` list so a broken installer fails fast instead of after the multi-hour downstream pipeline. Related: if merges are gated by branch-protection required checks, this new check presumably needs to be added to that list too.

zac-wang-nv:
Verified this locally: ran the installer in `CODEC_VALIDATE_ONLY` mode as non-root on Ubuntu 24.04 amd64 — apt metadata refresh over HTTPS and resolution of all 60 signed artifacts succeeded in ~6s; `bash -n`, `shellcheck`, and `cmp` of the two copies are all clean. Package parity with the old pinned lists checks out (60 amd64 / 59 arm64, `libvpl2` amd64-only preserved), and the canonical path/invocation/outputs are unchanged so `start_rtvi_embed.sh` and the base image's `start_rtvi_vlm.sh` need no edits.

The design looks right: fully isolated apt dirs, `signed-by`-pinned keyring, `APT::Update::Error-Mode=any`, retry/timeout knobs, and mirror-override escape hatches for air-gapped/internal-mirror deployments. The rt-embed/rt-vlm duplication is unfortunate but the CI `cmp` gate is a pragmatic mitigation given the separate Docker build contexts.

One trade-off worth stating explicitly in the PR description: versions are no longer pinned — what a container installs now tracks noble-updates/noble-security at startup time, so two starts of the same image on different days can extract different codec builds. That's the intended fix (resilience over reproducibility), but deployment/debugging docs should know installs are no longer byte-reproducible. Minor related nit: validate-only mode counts only `https://` URIs, so a `CODEC_UBUNTU_MIRROR` override to an `http://` internal mirror works for real installs but fails validate mode — fine as HTTPS enforcement in CI, just noting the asymmetry.

PR #1368, opened 2026-07-23
removed mdx names from env

romalakar:
## Description
removed mdx names from env and replace it with vss

## Checklist


romalakar:
/ok to test [cf39893](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1368/commits/cf3989340103425276cf666210656e51f5e78014)

kaushikc-nvidia:
/ok to test [0d054fd](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1368/commits/0d054fd0e8df001ba4c2424588ea3a1d63a78950)

deploy/docker/scripts/dev-profile.sh
```diff
@@ -1753,11 +1753,12 @@ function state_down() {
     fi
   done
 
-  echo "[INFO] Bringing down docker compose project 'mdx' (with volumes)..."
+  local _compose_project_name="${COMPOSE_PROJECT_NAME:-vss}"
```
zac-wang-nv: This reads `COMPOSE_PROJECT_NAME` from the **caller's shell**, but after this PR the value lives in the profile's `overrides.env` — and the new comment there actively invites changing it ("Change this to run multiple profiles on one host"). I checked: neither script ever sources the profile env into the environment (`set -a; source …` at line 1688 covers `containers.env` only, which doesn't define this key), so the shell variable is unset in normal use and this always resolves to `vss`.

Meanwhile `state_up` passes `--env-file developer-profiles/dev-profile-${profile}/generated.env` (a copy of `overrides.env`), so Compose *does* honour the override on the way up. The two halves disagree:

- `COMPOSE_PROJECT_NAME=vss-alerts` in `overrides.env` → `up` creates project `vss-alerts`
- `down` runs `docker compose -p vss down -v --remove-orphans` → tears down the wrong project (or nothing), leaving every container, the network, and all named volumes orphaned

It's aggravated by ordering: `state_down` deletes `generated.env` for all four profiles (lines 1742–1754) *before* computing the project name, so the script destroys its own record of what was deployed.

Worth noting the notebooks in this same PR get it right — `deploy_vss_launchable.ipynb:1759` does `read_env_value(env_file, "COMPOSE_PROJECT_NAME", os.environ.get("COMPOSE_PROJECT_NAME", "vss"))`. Suggest the same shape here, using the helper this file already uses ~15 times, and reading it before the `generated.env` cleanup:

```bash
local _profile_env="${deployment_directory}/developer-profiles/dev-profile-${profile}/.env"
local _profile_overrides_env="${deployment_directory}/developer-profiles/dev-profile-${profile}/overrides.env"
local _compose_project_name="$(get_env_value_from_files "COMPOSE_PROJECT_NAME" "${_profile_env}" "${_profile_overrides_env}")"
_compose_project_name="${_compose_project_name:-vss}"
```

deploy/docker/scripts/blueprint-deploy.sh
```diff
@@ -1189,11 +1189,12 @@ function state_down() {
     fi
   done
 
-  echo "[INFO] Bringing down docker compose project 'mdx' (with volumes)..."
+  local _compose_project_name="${COMPOSE_PROJECT_NAME:-vss}"
```
zac-wang-nv: Same issue as `dev-profile.sh:1756` — the value now lives in `industry-profiles/warehouse-operations/overrides.env`, not the shell, so a user who changes it gets an `up` under their project name and a `down` against `vss`. Same fix applies, using `get_env_value_from_files` against the deployment's `.env` + `overrides.env`, evaluated before the `generated.env` cleanup at lines 1178–1190.

skills/vss-deploy-profile/references/teardown.md
```diff
@@ -75,17 +76,17 @@ docker ps -a --format '{{.Names}}' \
...
+docker network rm "${project}_default" 2>/dev/null || true
 
 # `down` (no -v) also leaves every named volume. Remove the stale DATA volumes
 # that poison a fresh deploy — ES indices, Kafka offsets, Postgres, logstash
 # libs, nvstreamer recordings — while KEEPING model caches (rtvi-*, *_cache).
-# Names are <project>_<vol>; match on the volume-name suffix.
+# Names are <project>_<vol>; match the selected project plus known stale data volumes.
 docker volume ls -q \
-  | grep -E '(mdx-elastic-(data|logs)|mdx-kafka|mdx-logstash-libs|phoenix-data|vios_pg_data|mdx-nvstreamer-(data|videos))$' \
+  | grep -E "^${project}[-_](elastic-(data|logs)|kafka|logstash-libs|nvstreamer-(data|videos))$|(${project}[-_])?(phoenix-data|vios_pg_data)$" \
```
zac-wang-nv: This regex matches nothing for the ELK/Kafka/nvstreamer volumes, which makes the cache-preserving teardown stop doing the one thing it exists for.

Compose names a volume `<project>_<declared-name>`, and this PR doesn't rename the declared names — they still carry a literal `mdx-` prefix:

```
services/infra/compose.yml:            mdx-elastic-data, mdx-elastic-logs, mdx-kafka, mdx-logstash-libs, phoenix-data
dev-profile-alerts/compose.yml:        mdx-nvstreamer-data, mdx-nvstreamer-videos
```

So with `project=vss` the real volumes are `vss_mdx-elastic-data`, `vss_mdx-kafka`, … but the first alternative anchors `^vss[-_]elastic-data$` — no match. The old suffix-anchored regex (`(mdx-elastic-(data|logs)|mdx-kafka|…)$`) matched these correctly; this rewrite regresses it. Stale ES indices and Kafka offsets then survive into the next deploy — exactly the contamination the surrounding comment warns about.

Second issue in the same expression: the `(${project}[-_])?` prefix is optional *and* the alternative has no `^` anchor, so `(vss[-_])?(phoenix-data|vios_pg_data)$` also matches another project's `otherproj_phoenix-data`. That's cross-project data deletion from a doc an agent executes verbatim.

Suggested form — anchor both ends and keep the `mdx-` declared names:

```bash
| grep -E "^${project}_(mdx-(elastic-(data|logs)|kafka|logstash-libs|nvstreamer-(data|videos))|phoenix-data|vios_pg_data)$" \
```

This independently confirms the Skills Review bot's 🔴 critical ("Volume grep regex misses mdx-prefixed volume names", agreement 1/6) and its 🟠 high ("Optional project prefix lets regex match foreign volumes") — both are real despite the low agreement scores.

skills/vss-deploy-profile/references/teardown.md
```diff
@@ -1,24 +1,25 @@
...
+`-v --remove-orphans`. Two flavors, depending on whether you keep model caches.
 
 ## Full teardown — reclaim the host
 
 Removes containers, the project network, **and all named volumes** (including
 multi-GB NIM/RTVI model caches). This is the canonical `dev-profile.sh` teardown.
 
 ```bash
-docker compose -p mdx down -v --remove-orphans
+docker compose -p "${COMPOSE_PROJECT_NAME:-vss}" down -v --remove-orphans
```
zac-wang-nv, line 15: Same shell-vs-file gap as the scripts: `COMPOSE_PROJECT_NAME` is defined in the profile's `overrides.env`, not exported into the reader's shell, so this expands to `vss` unconditionally — including for the user who changed it, which is precisely the case the prose above now tells them about.

Since this doc is executed by an agent, it should say where to read the value from, e.g.:

```bash
project="$(grep -E '^COMPOSE_PROJECT_NAME=' <profile-dir>/overrides.env | cut -d= -f2- | tr -d "\"'")"
docker compose -p "${project:-vss}" down -v --remove-orphans
```

Line 62 and the `project=` assignment on line 81 have the same problem.

deploy/docker/services/analytics/video-analytics-api/compose.yml
```diff
@@ -19,7 +19,7 @@ services:
     command: node index.js --config /opt/mdx/vss-video-analytics-api/configs/vss-video-analytics-api-config.json
     restart: always
     ports:
-      - ${MDX_PORT:-8081}:8081
+      - ${VIDEO_ANALYTICS_API_HOST_PORT:-8081}:${VIDEO_ANALYTICS_API_PORT:-8081}
```
zac-wang-nv: Splitting host from container port is the right call, and the matching `haproxy/compose.yml` change genuinely fixes a bug — the old `${VIDEO_ANALYTICS_API_PORT:-${MDX_PORT:-8081}}` let a host-port override leak into the container-side port HAProxy dialled. Good catch.

One caveat on the new container-side knob: the app's listen port is hardcoded in `services/analytics/video-analytics-api/configs/vss-video-analytics-api-config.json` (`"port":8081`), and no profile defines `VIDEO_ANALYTICS_API_PORT`. So the variable is inert today, but anyone who sets it silently breaks the service — the mapping and HAProxy both retarget while the process keeps listening on 8081. Either template the port into that config JSON, or leave the container side as a literal `8081` and keep only `VIDEO_ANALYTICS_API_HOST_PORT` tunable.

deploy/docker/developer-profiles/dev-profile-base/.env
```diff
@@ -63,7 +63,7 @@ STREAM_TYPE=kafka
 # =============================================================================
 
 # Project name for volume naming (DO NOT MODIFY)
-COMPOSE_PROJECT_NAME=mdx
+COMPOSE_PROJECT_NAME=vss
```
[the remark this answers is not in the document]
zac-wang-nv: Confirming this one still stands at head `24b668c3` — it's the main unaddressed risk in the PR.

One correction to the specifics: the volume names cited don't exist. Compose names volumes `<project>_<declared-name>`, and the declared names in this repo still carry a literal `mdx-` prefix (`mdx-elastic-data`, `mdx-elastic-logs`, `mdx-kafka`, `mdx-logstash-libs` in `services/infra/compose.yml`; `mdx-nvstreamer-data/videos` in `dev-profile-alerts/compose.yml`). So the real before/after is `mdx_mdx-elastic-data` → `vss_mdx-elastic-data`, not `mdx_elasticsearch-data` → `vss_elasticsearch-data`.

The substance is right, and it's worse than described: the orphaned `mdx_*` volumes are *also* invisible to the cleanup path. `references/teardown.md` now sweeps `^${project}[-_]...`, which never matches `mdx_*` once the project is `vss` — so an upgrading user gets silently-empty ES/Kafka/Postgres **and** no documented way to reclaim the old volumes. Please add a migration note (or a one-shot `docker volume rm mdx_*` step) before merge.

deploy/docker/developer-profiles/dev-profile-base/overrides.env
```diff
@@ -21,9 +27,9 @@ HARDWARE_PROFILE=H100
 # MODEL PLACEMENT AND RUNTIME MODE
 # =============================================================================
 # Local LLM GPU device. Edge hardware may require device 0.
-LLM_DEVICE_ID='0'
+LLM_DEVICE_ID=1
 # Local VLM GPU device. Remote VLM and edge hardware may use device 0.
-VLM_DEVICE_ID='0'
+VLM_DEVICE_ID=1
```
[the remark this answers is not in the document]
zac-wang-nv: Stale — this comment is anchored to `0d054fd0e`, which has since been superseded by the rebase to `24b668c3`. At the current head `dev-profile-base/overrides.env` reads `LLM_DEVICE_ID='0'` / `VLM_DEVICE_ID='0'` (lines 30 and 32), so the 0→1 change is no longer in the diff. Resolving.

zac-wang-nv:
## Review — PR #1368 (`removed mdx names from env`)

Reviewed at head `24b668c31939`. **Confidence 2/5** — the rename is incomplete in ways that break teardown, and the PR's own headline feature (a user-tunable `COMPOSE_PROJECT_NAME`) isn't honoured by either teardown script.

Five inline comments are on the diff. The items below are things **outside** the diff that need updating in the same change — otherwise they break silently once the compose project stops being `mdx`.

### 1. `deploy_warehouse_launchable.ipynb` was left behind by the rename

This PR removes `COMPOSE_PROJECT_NAME=mdx` and `MDX_PORT` from [`warehouse-operations/.env`](deploy/docker/industry-profiles/warehouse-operations/.env) and updates the smartcities and VSS launchables — but the warehouse launchable, which is the notebook for the profile whose env files actually changed, still hardcodes the old values:

- `deploy_warehouse_launchable.ipynb:348` — `COMPOSE_PROJECT = "mdx"   # COMPOSE_PROJECT_NAME from the warehouse .env` (that key no longer exists in the warehouse `.env`; it's `vss` in `overrides.env`)
- lines 1405, 1417, 1939, 1961, 1968, 1982 — `docker compose -p mdx down/start/stop`, all now targeting a project that won't exist
- line 1942 — prose: "All warehouse services run under the Compose project **`mdx`**"
- line 1771 — `"mdx": env_value("MDX_PORT", "8081")`; survives only because the default happens to match, and drifts the moment a user retunes `VIDEO_ANALYTICS_API_HOST_PORT`
- lines 1411, 1921 — stale volume/index name references

Porting the `read_env_value(...)` helper already added to the other two notebooks would cover this.

### 2. The skill updates are incomplete — 15 files across 3 skills

The PR updates `vss-deploy-profile/references/{teardown,warehouse}.md`, but the rename's blast radius through `skills/` is much wider. I swept the whole tree at head and filtered out the `mdx` occurrences that are *legitimately* unchanged (container paths `/opt/mdx/…`, declared volume names `mdx-*`, the `mdx-bev` ES index, container-name grep prefixes). What's left is genuinely broken:

**a) `vss-deploy-profile` — the skill this PR already edits, 3 files missed.** Compose names the default network `<project>_default`, and the PR's own `teardown.md` edit now says "such as `vss_default`" — but three sibling files still say `mdx_default`:
- `references/prerequisites.md:181,193,198,200` — including a UFW rule comment and a `docker network inspect mdx_default` command
- `references/troubleshooting.md:128`
- `references/warehouse.md:641` — **this file is in the PR diff** (lines 35 and 173 were updated); line 641 was missed

**b) `vss-build-vision-agent` — 6 files.** Every `docker compose -p mdx` and every `--filter label=com.docker.compose.project=mdx` now matches zero objects, so an agent following these tears down nothing:
- `references/teardown.md:10,25,30,34,78,96,99,106`
- `references/deployment.md:95` — `docker compose -p mdx -f "$BUILD_DIR/resolved.yml" down -v --remove-orphans`
- `references/data-directory.md:131` — label-filtered volume sweep
- `references/prerequisites.md:178,190,195,197` and `references/troubleshooting.md:123` — `mdx_default`
- `references/services/alerts.md:30` — still documents `MDX_PORT`

**c) `vss-deploy-detection-tracking-3d` — 4 doc files.** These cite *project-prefixed* volume names, which do change (`mdx_mdx-kafka` → `vss_mdx-kafka`, `mdx_vios_pg_data` → `vss_vios_pg_data`):
- `references/deploy-rtvi-cv-3d-stack.md:121,140`, `references/troubleshooting.md:31`, `references/configure-cameras.md:259`
- `references/teardown.md:47` — `docker volume rm $(docker volume ls -q | grep '^mdx_')`, which now selects nothing, and the surrounding text specifically warns that prune won't do it

The eval specs under (c) fail differently and more quietly — see the next section.

### 3. Two eval specs become unconditional passes

These aren't prose — they're executable assertions:

```
evals/deploy.json            /expects[2]/checks[1]
evals/calibration-chain.json /expects[1]/checks[3]

  docker volume ls -q | grep -qE '^mdx_mdx-kafka$|^mdx_vios_pg_data$' && exit 1 || exit 0
```

The check asserts the MV3DT named volumes are *absent* after teardown. Both specs deploy via `--env-file industry-profiles/warehouse-operations/.env --env-file …/generated.env`, and `generated.env` is a straight copy of `overrides.env` — so after this PR the project is `vss` and the real volumes are `vss_mdx-kafka` / `vss_vios_pg_data`. The anchored `^mdx_` pattern can no longer match, `&& exit 1` never fires, and `|| exit 0` always runs. **The check passes whether or not teardown removed anything** — it can no longer fail.

The `query` blocks in both specs (`/expects[0]/query`) also instruct the agent that "the named volumes `mdx_mdx-kafka` and `mdx_vios_pg_data` must be absent" as a precondition, so the prompt is stale in the same direction.

This is easy to miss because `skills/vss-deploy-detection-tracking-3d/**` isn't in this PR's file list, so the Skills Eval gate for that skill won't be triggered by these changes — the specs will keep reporting green until someone edits that skill for an unrelated reason. Both patterns need `^vss_` (or better, to derive the project from `COMPOSE_PROJECT_NAME` like the rest of this PR does).

### 4. Test coverage gap on the behaviour this PR introduces

`test-dev-profile.sh:1805` asserts `[DRY-RUN] docker compose -p vss down -v --remove-orphans`, which passes only because `COMPOSE_PROJECT_NAME` is unset in the test shell — it exercises the default, never the "user changed the project name" path that the new `overrides.env` comment invites. A case that writes a non-default `COMPOSE_PROJECT_NAME` into `overrides.env` and asserts `down` targets that value would have caught Comments 1 and 2.

### 5. Unrelated changes bundled in

and per the repo's surgical-changes guideline would be cleaner split out (or at least called out in the description):

1. `RTVI_VLM_MODEL_PATH='ngc:…'` → unquoted, across `dev-profile-{alerts,base,lvs}/overrides.env` plus 4 expectation strings in `test-dev-profile.sh`. Note the adjacent `RTVI_VLM_ENDPOINT` expectations still assert `"''"`, so the quoting convention is now mixed.
2. `VSS_AGENT_VERSION` dropped from the `services/agent/agent.env` shared-env spec (`test-dev-profile.sh:1181`). I verified `agent.env` has no `VSS_AGENT_VERSION` at the merge-base either, so this is a legitimate fix for an already-red assertion — it just belongs to whoever removed the key, not to this PR.
3. The new `_allowed_duplicate_keys=(RTVI_VLM_MAX_MODEL_LEN)` exemption, which relaxes the stable-vs-override split invariant for the `base` profile.

None are wrong; they just make the diff harder to attribute and widen the blast radius of a rename PR.

### 6. CI: the gate sweep has not run on this head

The last `/ok to test` was for `0d054fd` (Jul 24) and the branch has since been rebased to `24b668c3` (pushed today 12:31 UTC). `gh run list --branch pull-request/1368` shows zero runs for the new head — the only checks on it are SonarQube (queued) and Greptile (in progress). Everything green further up the page (CI, Skills Eval, Helm Sync, License Diff, playbook compliance) was measured against a commit that is no longer the head. A fresh `/ok to test 24b668c3` is needed before this is mergeable.

Two related notes: the Greptile summary comment is also stale — its headline concern (`LLM_DEVICE_ID`/`VLM_DEVICE_ID` 0→1 in `dev-profile-base/overrides.env`) no longer exists at this head, the rebase dropped it. And the Jul 24 Skills Eval leg was red on 3 of 5 specs (`base`, `lvs` `NonZeroAgentExitCodeError`, `warehouse` 0.846) — the bot attributes all three to SSH/infra transients, which looks right, but it means the last *clean* eval signal for this skill is the Jul 23 run against `cf39893`.

kaushikc-nvidia:
/ok to test [514f74b](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1368/commits/514f74b7806fb0f239214b59b38d8ca85b8c7865)

skills/vss-deploy-profile/references/teardown.md
zac-wang-nv: The volume rename in `355b55a34` fixed most of this, but it renamed the Kafka volume to **`kafka-data`** while this regex still says **`kafka`** — so `vss_kafka-data` no longer matches and survives the cache-preserving teardown.

Verified against the declared names at this head:

```
$ printf 'vss_elastic-data\nvss_elastic-logs\nvss_kafka-data\nvss_logstash-libs\nvss_phoenix-data\nvss_vios_pg_data\n' \
  | grep -E "^vss[-_](elastic-(data|logs)|kafka|logstash-libs|nvstreamer-(data|videos))$|(vss[-_])?(phoenix-data|vios_pg_data)$"
vss_elastic-data
vss_elastic-logs
vss_logstash-libs
vss_phoenix-data
vss_vios_pg_data          # vss_kafka-data missing
```

Two things make this worth fixing rather than shrugging at:

1. **You already wrote the correct version in this same PR.** `skills/vss-build-vision-agent/references/teardown.md:109` — also touched by `b9f9afdd7` — reads `'(elastic-(data|logs)|kafka-data|logstash-libs|phoenix-data|vios_pg_data)$'` and scopes with `--filter "label=com.docker.compose.project=${COMPOSE_PROJECT_NAME}"`. That form is correct on every count; this one is the odd sibling.
2. **A lingering volume is the exact failure mode another file in this PR documents.** `skills/vss-manage-video-io-storage/references/deploy-vios-service.md:171` (in this diff) describes `docker compose up -d` hanging forever on an invisible "Volume X exists but doesn't match configuration… Recreate?" prompt when a stale volume's `driver_opts` bind path drifts — and names `vss_kafka-data` as one of the volumes to remove first.

Two smaller things in the same expression:
- `nvstreamer-(data|videos)` is now dead — `355b55a34` deleted those declarations from `dev-profile-alerts/compose.yml` and `smartcities/compose.yml`, so nothing can ever match it.
- `calibration-toolkit-alerts` (renamed from `mdx-calibration-toolkit-alerts`) isn't in the list. Pre-existing — it wasn't in the merge-base list either — so ignore unless you want it swept.

Suggested form, matching the sibling skill:

```bash
docker volume ls -q \
  --filter "label=com.docker.compose.project=${project}" \
  | grep -E '(elastic-(data|logs)|kafka-data|logstash-libs|phoenix-data|vios_pg_data)$' \
  | xargs -r docker volume rm
```

The `--filter` also retires the unanchored `(${project}[-_])?` alternative that can reach another project's `phoenix-data` / `vios_pg_data`.

deploy/docker/scripts/dev-profile.sh
```diff
@@ -1657,10 +1657,33 @@ function state_up() {
...
 function state_down() {
-  local _profile_dir_names _profile_dir_name _generated_env
+  local _profile_dir_names _profile_dir_name _profile_dir _source_env _overrides_env _generated_env
 
-  echo "[INFO] Cleaning up generated.env files from all profiles..."
   _profile_dir_names=('base' 'lvs' 'search' 'alerts')
+
+  local _compose_project_name="${COMPOSE_PROJECT_NAME:-}"
+  if [[ -z "${_compose_project_name}" ]]; then
+    for _profile_dir_name in "${_profile_dir_names[@]}"; do
```
zac-wang-nv: The resolution logic itself is right and fixes what I raised — this is about the case it can't express.

`overrides.env` now tells users "Change this to run multiple profiles on one host", but `down` has no profile selector (`Usage: dev-profile.sh (up|down) [options]`), so this loop picks the **first** profile with a `generated.env` and tears down only that project — while the loop below deletes **every** profile's `generated.env`.

Reproduced at this head with `base` → `vss-base` and `alerts` → `vss-alerts`, both with a `generated.env`:

```
$ ./deploy/docker/scripts/dev-profile.sh down --dry-run
[INFO] Cleaning up generated.env files from all profiles...
[DRY-RUN] rm -f .../dev-profile-base/generated.env
[DRY-RUN] rm -f .../dev-profile-alerts/generated.env
[INFO] Bringing down docker compose project 'vss-base' (with volumes)...
[DRY-RUN] docker compose -p vss-base down -v --remove-orphans
```

`vss-alerts` keeps running, and the `generated.env` that recorded it is gone. A second `down` doesn't recover it either — the fallback loop also hits `base` first and returns `vss-base` again, so the alerts stack is permanently unreachable through this script.

Options, cheapest first:
- **Narrow the docs.** If one-profile-at-a-time is the intended contract, drop "run multiple profiles on one host" from the five `overrides.env` files — the current text promises something `down` can't deliver.
- **Tear down every distinct project found.** Collect the names across all four profiles, dedupe, and `down` each one before the `generated.env` cleanup.
- **Add `down --profile <name>`** to mirror `up`.

Happy for this to be a follow-up issue — just don't want the invitation to ship without one of the three.

kaushikc-nvidia:
/ok to test [9532cda](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1368/commits/9532cda56acd28dc73a7099c7ceed07e4c651117)

deploy/docker/services/analytics/video-analytics-api/compose.yml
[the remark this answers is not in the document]
kaushikc-nvidia: Addressed via commit: https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1368/changes/1173d48e2e7f9716618f8ed69eb343e784115633

deploy/docker/scripts/dev-profile.sh
kaushikc-nvidia: Used option 2 and addressed in commit: https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1368/changes/9532cda56acd28dc73a7099c7ceed07e4c651117

skills/vss-deploy-profile/references/teardown.md
kaushikc-nvidia: Addressed in commit: https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1368/changes/4ae854370098607599a242584418316e1ed3a195

kaushikc-nvidia:
/ok to test [9532cda](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1368/commits/9532cda56acd28dc73a7099c7ceed07e4c651117)

kaushikc-nvidia:
/ok to test https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1368/changes/1173d48e2e7f9716618f8ed69eb343e784115633

kaushikc-nvidia:
/ok to test [8b430bd](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1368/commits/8b430bdb7a461eb7162afcac672c74899b240641)

zac-wang-nv (approved):
LGTM.

TODO: fix the readme file from deploy/docker

kaushikc-nvidia:
/ok to test [4e0ff57](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1368/commits/4e0ff57cfec351c8f5fbba86b8bb2bb69247c502)

PR #1366, opened 2026-07-23
Feat/spit nemoclaww and vss mcp notebook

nvleeryan:
## Description
Split the old combined notebook into deploy_nemoclaw.ipynb (sandbox / agent harness) and deploy_vss_orchestrator.ipynb (host prep + MCP + VSS deploy); remove deploy_nemoclaw_vss.ipynb.
Make NemoClaw setup harness-neutral (OpenClaw and Hermes) and drop VSS-only knobs from the NemoClaw notebook.
Use Brev /etc/brev/environment-context.json as the sole source of truth for secure-link FQDNs and env id; stop inventing NetBird / brevlab.com domains.
Add brev_util.py and slim network_util.py to IP helpers only; wire notebooks and compose through context-backed FQDNs.
MCP helper: drop duplicated Brev domain logic; sync with --extra agent so NAT MCP runs.
Orchestrator robustness: skip broken mdx/VIOS symlinks in permission setup; always pull latest vss-agent tag; remove hardcoded orchestrator module checks.
Reorganize notebook steps / docs and add/update unit tests for brev, network, compose, and storage.

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/HEAD/CONTRIBUTING.md).
- [x] I have installed and run [pre-commit hooks](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/HEAD/CONTRIBUTING.md#1-enable-pre-commit-hooks) locally (`uv run pre-commit install` once, then hooks run on every `git commit`).
- [x] Every commit on this PR is [DCO sign-off](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/HEAD/CONTRIBUTING.md#developer-certificate-of-origin-dco)'d (`git commit -s` adds a `Signed-off-by` trailer that certifies you have the right to submit the change under Apache-2.0).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #1375, opened 2026-07-23
Feat/https toggle vss orchestrator mcp

nvleeryan:
## Description
Allow VSS Orchestrator to run with either HTTP or HTTPS mode.  It was initiated by the MCP HTTPS requirement from NemoClaw, but the same mechanism also be used to allow other NAT MCP to run with HTTPS if needed.  
User can select to run with HTTP or HTTPS mode by specifying the ENV ORCHESTRATOR_ENABLE_HTTPS to true / false from the NemoClaw notebook.
Users select HTTP or HTTPS via ORCHESTRATOR_ENABLE_HTTPS (true/false) in the NemoClaw notebook. In HTTPS mode, if the cert/key paths aren't set they default to assets/orchestrator_mcp_{cert,key}.pem, and if either file doesn't exist a self-signed pair is auto-generated (requires openssl).

## Checklist


nvleeryan:
/ok to test https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1375/commits/1a73c2e908d68ee39d1ca1924a028cad8adf7c4d

nvleeryan:
/ok to test https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1375/commits/1b5832d02c786156db236ea930dea46729b4ffef

services/agent/packages/vss_agents/src/vss_agents/orchestrator/mcp_ssl_worker.py
```diff
@@ -0,0 +1,170 @@
...
+
+
+class _SSLFastMCP(FastMCP):
+    """``FastMCP`` variant that enables TLS when ``ORCHESTRATOR_ENABLE_HTTPS=true``."""
+
+    async def _run_uvicorn(self, app: ASGIApp, *, transport: str, path_suffix: str = "") -> None:
+        import uvicorn
+
+        certfile, keyfile = resolve_ssl_paths()
+        scheme = "https" if https_enabled() else "http"
```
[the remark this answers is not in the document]
zac-wang-nv: Still open at `1b5832d02` — `_run_uvicorn` calls `resolve_ssl_paths()` on line 114 and `https_enabled()` again on line 115. Agreed with the suggestion: `scheme = "https" if certfile else "http"` reuses the value already resolved on the previous line and reads more directly. Cosmetic rather than a correctness risk (both reads happen microseconds apart at startup), but worth folding in.

services/agent/packages/vss_agents/src/vss_agents/orchestrator/mcp_ssl_worker.py
```diff
@@ -0,0 +1,170 @@
...
+            token_verifier = IntrospectionTokenVerifier(self.front_end_config.server_auth)
+
+        return _SSLFastMCP(
+            name=self.front_end_config.name,
+            host=self.front_end_config.host,
+            port=self.front_end_config.port,
+            debug=self.front_end_config.debug,
+            auth=auth_settings,
+            token_verifier=token_verifier,
+        )
```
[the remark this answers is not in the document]
zac-wang-nv: Addressed — the delegation landed in `7579ea0a` and is the right call: NAT keeps ownership of the auth wiring and only the server class changes. I verified the substitution against the pinned `nvidia-nat-mcp` 1.6.0: the parent resolves `FastMCP` through the module global at `front_end_plugin_worker.py:400` and never awaits inside, so the patch window can't interleave with another coroutine.

The delegation did make mypy unhappy for one head (`nvidia-nat-mcp` ships no `py.typed`, so the `super()` call is `Any` and `warn_return_any` rejected returning it), and that is fixed in `88a31118`. `Type Check (mypy)` is green at `1b5832d02` and I reproduced it locally. Resolving this one; I've left a separate note on line 158 about hardening the same spot at runtime.

zac-wang-nv:
CI is green now — nice turnaround on the mypy break, `88a31118` is a clean one-line fix and I confirmed `Type Check (mypy)` passes locally as well as in CI. All 40 checks pass at `1b5832d02`, and both new suites pass locally (8 + 7).

One thing left, carried over from the last head and unchanged by it: **the HTTPS toggle is still inert.** Every consumer of the new code is missing, so merging as-is ships dead code — `ORCHESTRATOR_ENABLE_HTTPS=true` has no effect, the stock worker still runs, still plain HTTP.

What happened: the three original feature commits (`bfde7f68`, `29927dd6`, `7a28a806`) edited `deploy/docker/scripts/deploy_nemoclaw_vss.ipynb`. On `develop`, #1366 (`11f73ac2`) deleted that notebook and split it into `deploy_nemoclaw.ipynb` + `deploy_vss_orchestrator.ipynb`. Merge `652a0203` resolved the delete/modify conflict by taking the deletion, so the notebook side of the feature went away while the library side stayed. Re-confirmed by grep at `1b5832d02` — `ensure_mcp_tls_certs`, `SSLMCPWorker` and all three `ORCHESTRATOR_*` env vars have zero references outside their own definitions and tests.

The wiring that needs re-applying to `deploy_vss_orchestrator.ipynb` (and, for the registration line, `deploy_nemoclaw.ipynb`):
- `ORCHESTRATOR_ENABLE_HTTPS` / `ORCHESTRATOR_CERTFILE` / `ORCHESTRATOR_KEYFILE` / `ORCHESTRATOR_SSL_SAN` settings + derived defaults
- the `ensure_mcp_tls_certs(...)` call and the `SSL_CERT_FILE` export, so `nat mcp client` health checks trust the generated cert
- `--runner_class vss_agents.orchestrator.mcp_ssl_worker.SSLMCPWorker` on the `nat mcp serve` invocation (or `runner_class:` in `vss_orchestrator_mcp_config.yml`, which may be the more durable home now that the notebook has been split once already), plus the `env[...]` exports for the subprocess
- `MCP_SCHEME` / `MCP_CLIENT_HOST` / `SANDBOX_MCP_URL` and the preflight cert-exists checks

Two of those fixed bugs that are still live on `develop` today: both split notebooks build `MCP_URL = f"http://{MCP_HOST}:{MCP_PORT}/mcp"` from `MCP_HOST = "0.0.0.0"` (`deploy_vss_orchestrator.ipynb:161`, `deploy_nemoclaw.ipynb:227`), and `deploy_nemoclaw.ipynb:791` registers that bind-all URL with `nemoclaw sandbox mcp ... add vss_orchestrator --url {MCP_URL}`. The branch's `MCP_CLIENT_HOST` (dial loopback when the bind address is `0.0.0.0`) and `SANDBOX_MCP_URL` (register the alias the sandbox can actually route to) both addressed that and are now lost — worth re-landing deliberately rather than letting them disappear with the notebook. (The `AGENT_RUNTIME` knob from `7a28a806` did survive in `develop`'s notebooks, so that one needs nothing.)

If you'd rather land the library half now and re-apply the notebook wiring in a follow-up, that's a reasonable call — but it's worth saying so explicitly in the PR description and filing the follow-up, because as it stands nothing in the tree records that this code is waiting on a caller.

services/agent/packages/vss_agents/src/vss_agents/orchestrator/mcp_ssl_worker.py
```diff
@@ -0,0 +1,170 @@
...
+    :class:`_SSLFastMCP` for the module-level ``FastMCP`` name during
+    ``super().create_mcp_server()`` and then aligns the auth resource URL scheme
+    when HTTPS is enabled.
+    """
+
+    async def create_mcp_server(self) -> FastMCP:
+        original = mcp_worker.FastMCP
+        mcp_worker.FastMCP = _SSLFastMCP
+        try:
+            server: FastMCP = await super().create_mcp_server()
```
zac-wang-nv: This does fix the CI break — verified locally, `Success: no issues found`. Worth being explicit about what it does though: annotating the target makes mypy accept the `Any` from the untyped `super()` call as a `FastMCP`. It silences the checker by assertion, not by checking, so the runtime gap that the `no-any-return` warning was pointing at is still open.

That gap is real. The substitution depends on two undocumented NAT implementation details — that the parent resolves `FastMCP` through the module global, and that it doesn't await while doing so. Both hold in the pinned 1.6.0 (I checked the wheel: `front_end_plugin_worker.py:400`), but when they stop holding the failure is silent: `super()` returns a stock `FastMCP`, the transport overrides never run, and the server comes up on plain HTTP while the notebook, the health check and the sandbox registration all assume `https://` — surfacing as a TLS handshake error several layers from the cause.

An `isinstance` guard satisfies mypy the same way the annotation does (real narrowing this time) *and* turns that silent downgrade into a loud startup failure:

```python
server = await super().create_mcp_server()
...
if not isinstance(server, FastMCP):
    raise RuntimeError(f"NAT returned a non-FastMCP server: {type(server)!r}")
if https_enabled() and not isinstance(server, _SSLFastMCP):
    raise RuntimeError(
        "NAT did not construct the TLS-capable FastMCP subclass; the FastMCP "
        "substitution in SSLMCPWorker is incompatible with this NAT version.",
    )
```

I ran this exact block against the repo's mypy config — clean, no `cast` needed, `warn_redundant_casts` stays happy. Non-blocking now that CI is green, but it converts a NAT-version landmine from "silently serves HTTP" into "fails at startup with the reason."

Same place is also the natural home for the `base_path` caveat currently only in the module docstring: `front_end_plugin.py` builds its own `uvicorn.Config` when `base_path` is set, so `ORCHESTRATOR_ENABLE_HTTPS=true` + a `base_path` silently serves plain HTTP. The orchestrator config doesn't set one today, but rejecting the combination here is a one-line guard against a config trap that is otherwise only findable in a docstring.

services/agent/packages/vss_agents/tests/unit_test/orchestrator/test_mcp_ssl_worker.py
```diff
@@ -0,0 +1,123 @@
...
+
+    async def fake_super_create(self: Any) -> FakeParentServer:
+        captured["fastmcp_during_super"] = ssl_mod.mcp_worker.FastMCP
+        return FakeParentServer()
+
+    worker = ssl_mod.SSLMCPWorker.__new__(ssl_mod.SSLMCPWorker)
+    worker.front_end_config = SimpleNamespace(host="localhost", port=9901)
+    original = ssl_mod.mcp_worker.FastMCP
+
+    with patch.object(ssl_mod.MCPFrontEndPluginWorker, "create_mcp_server", fake_super_create):
```
zac-wang-nv, line 101: Good that the restore-on-exit path is covered. The catch is that this test replaces the parent with `fake_super_create`, which is precisely the thing the substitution depends on — it asserts that the module global is patched *while the parent runs*, but not that the real parent actually reads that global when constructing. A NAT release that moves to a local import, a factory, or a different module would leave this test green while HTTPS silently stops working.

Exercising the real `MCPFrontEndPluginWorker.create_mcp_server` would close that. It only needs a config whose `general.front_end` carries `name`/`host`/`port`/`debug`/`server_auth=None`, and then `assert isinstance(server, ssl_mod._SSLFastMCP)` — one test that fails loudly on the next NAT bump, which is exactly when you want to hear about it. It also pairs naturally with the runtime guard suggested on line 158.

services/agent/packages/vss_agents/tests/unit_test/orchestrator/test_mcp_ssl_worker.py
```diff
@@ -0,0 +1,123 @@
...
+    async def fake_super_create(self: Any) -> FakeParentServer:
+        return FakeParentServer()
+
+    worker = ssl_mod.SSLMCPWorker.__new__(ssl_mod.SSLMCPWorker)
+    worker.front_end_config = SimpleNamespace(host="0.0.0.0", port=9901)
+
+    with patch.object(ssl_mod.MCPFrontEndPluginWorker, "create_mcp_server", fake_super_create):
+        await ssl_mod.SSLMCPWorker.create_mcp_server(worker)
+
+    assert auth.resource_server_url == AnyHttpUrl("https://0.0.0.0:9901")
```
zac-wang-nv: The remaining untested piece is `_SSLFastMCP._run_uvicorn` — the code that does the actual TLS work. Everything covered here is env parsing and class substitution; nothing asserts that `ssl_certfile` / `ssl_keyfile` reach `uvicorn.Config`, which is the one behaviour the whole PR exists for.

It's cheap to cover: patch `uvicorn.Config` and `uvicorn.Server` inside the module's import, call `run_streamable_http_async()` on an `_SSLFastMCP` with `ORCHESTRATOR_ENABLE_HTTPS=true` and a tmp cert/key pair, and assert the kwargs. A second case with HTTPS off asserting both SSL kwargs are `None` also pins the "identical to the stock worker" claim in the docstring.

services/agent/packages/vss_agents/src/vss_agents/orchestrator/mcp_ssl_worker.py
```diff
@@ -0,0 +1,170 @@
...
+binding while calling ``super().create_mcp_server()``, so auth wiring stays in
+NAT and only the server class differs. When HTTPS is enabled, the stock
+``http://`` ``resource_server_url`` is rewritten to ``https://``.
+
+Toggle via environment:
+
+- ``ORCHESTRATOR_ENABLE_HTTPS`` selects the scheme: ``true`` enables TLS, ``false``
+  (the default) serves plain HTTP. Only ``true``/``false`` are accepted (case-
+  insensitive). HTTPS is never inferred from cert/key presence.
+- When ``ORCHESTRATOR_ENABLE_HTTPS=true``, ``ORCHESTRATOR_CERTFILE`` and
```
zac-wang-nv, line 42: This contract disagrees with the PR description, which says "if the cert/key paths aren't set they default to `assets/orchestrator_mcp_{cert,key}.pem`". Both were true before the merge — the defaulting happened in the notebook, so by the time the env reached this module the paths were always populated and `resolve_ssl_paths()` raising on unset was a correct backstop. With the notebook gone, `git grep orchestrator_mcp_cert` has no hits anywhere in the tree, so the module is the only surviving definition of the contract and it's the stricter one.

When the wiring is re-applied, worth deciding whether the defaulting goes back in the notebook or moves in here next to the env parsing — the latter would make the module usable standalone and keep the docstring honest. Either way the PR description needs updating to match what actually ships.

deploy/docker/scripts/orchestrator_mcp_helper.py
```diff
@@ -140,6 +141,76 @@ def check_mcp_health(mcp_url: str, agent_dir: str | Path, timeout_s: int = 15) -
...
+            a partial custom pair by auto-generating).
+        ValueError: if ``san`` is empty when generation is required.
+        RuntimeError: if ``openssl`` is not available when generation is required.
+        subprocess.CalledProcessError: if ``openssl`` fails.
+    """
+    cert_path = Path(certfile).expanduser().resolve()
+    key_path = Path(keyfile).expanduser().resolve()
+    cert_exists = cert_path.is_file()
+    key_exists = key_path.is_file()
+    if cert_exists and key_exists:
```
zac-wang-nv, line 182: The partial-pair fix reads well. Separate case that this early return still misses: an **expired** cert is never regenerated. With `days=365`, a long-lived deployment keeps serving the stale pair a year later, and the failure surfaces as a TLS handshake error out of `nat mcp client` during the health check — a confusing place to debug from.

A cheap `openssl x509 -checkend <seconds> -noout -in <cert>` before the early return (regenerate when it returns non-zero) keeps this self-healing. Non-blocking, but the failure mode is unpleasant enough to be worth the few lines. Regenerating would need to be conditional on the pair being self-generated, so as not to fight a user's CA-signed cert — checking the subject against the `subject` argument, or dropping a marker file next to the pair, both work.

deploy/docker/scripts/orchestrator_mcp_helper.py
```diff
@@ -140,6 +141,76 @@ def check_mcp_health(mcp_url: str, agent_dir: str | Path, timeout_s: int = 15) -
...
+
+    san_value = san.strip()
+    if not san_value:
+        raise ValueError("san is required to auto-generate MCP TLS cert/key.")
+    if not shutil.which("openssl"):
+        raise RuntimeError("openssl is required to auto-generate MCP TLS cert/key.")
+
+    cert_path.parent.mkdir(parents=True, exist_ok=True)
+    key_path.parent.mkdir(parents=True, exist_ok=True)
+    subprocess.run(
```
zac-wang-nv: `openssl req -nodes` writes an unencrypted RSA private key and respects the process umask, so on a default `umask 022` the key lands world-readable (0644). Worth a `key_path.chmod(0o600)` right after the `subprocess.run` — or wrapping the call in a temporary `os.umask(0o077)` if you'd rather not have the file exist world-readable even briefly.

Related, for when the notebook wiring comes back: the old default key path was `<VSS_REPO_DIR>/assets/orchestrator_mcp_key.pem`, inside the repo working tree, and `.gitignore` still has no `*.pem` entry (`git check-ignore assets/orchestrator_mcp_key.pem` returns not-ignored at `1b5832d02`). That puts an unencrypted private key one `git add -A` away from a public repo, and the `trufflehog` pre-commit hook wouldn't reliably catch it since the file only exists on the user's machine at deploy time. Adding `assets/orchestrator_mcp_*.pem` to `.gitignore` in this PR is a cheap guard for whenever the default path returns.

deploy/docker/scripts/tests/test_orchestrator_mcp_helper.py
```diff
@@ -34,5 +36,86 @@ def test_returns_none_when_no_containers(self) -> None:
             self.assertIsNone(helper.resolve_openshell_gateway_container("demo"))
 
 
+
+class EnsureMcpTlsCertsTests(unittest.TestCase):
```
zac-wang-nv, line 39: These five tests — including the two new ones guarding the partial-pair overwrite, which is the security-relevant behaviour in this PR — don't run anywhere in CI. `services/agent/pyproject.toml` sets `testpaths` to the three `packages/*/tests/unit_test` dirs, and the `Test (pytest)` job runs pytest with `working-directory: services/agent`, so nothing ever collects `deploy/docker/scripts/tests/`. Same for lint: `Lint (Python)` runs ruff with `working-directory: services/agent`, and the pre-commit hooks are gated on `files: ^services/agent/`, so this directory is unlinted too (hence the three blank lines above this class surviving — harmless, just an artifact of the same gap). Re-checked against the CI changes the `develop` merge brought in; still true.

They do pass — I ran them locally at `1b5832d02`, 7 passed — which is exactly the problem: nothing in CI would have told you if they didn't.

The cheapest fix is the existing "Video summarization contracts" step in `ci.yml` (line 316), which already runs `uv run --project services/agent pytest <explicit paths>` from the repo root — adding `deploy/docker/scripts/tests/test_orchestrator_mcp_helper.py` to that list is a one-line change. Alternative is moving the helper + tests under `services/agent/` where the existing tooling picks them up for free. Given the PR checklist ticks "New or existing tests cover these changes", worth making the coverage real.

zac-wang-nv:
One TLS question that isn't answered anywhere in the diff: **does the NemoClaw sandbox actually verify the certificate, and if so, how does it come to trust it?**

`ensure_mcp_tls_certs` generates a self-signed leaf — `openssl req -x509 ... -subj /CN=vss-orchestrator-mcp` with no CA in the picture. A self-signed cert only works if every client is explicitly told to trust it. The only trust distribution anywhere in the original design was `os.environ["SSL_CERT_FILE"] = ORCHESTRATOR_CERTFILE` in the notebook process, which covers `check_mcp_health()` (`orchestrator_mcp_helper.py:109` runs `nat mcp client` via `subprocess.run` and inherits the notebook env). That is the *local* client.

The client I can't account for is the one inside the sandbox. `deploy_nemoclaw.ipynb:791` registers the server with `nemoclaw sandbox mcp ... add vss_orchestrator --url {MCP_URL}`, and with HTTPS on that URL becomes `https://{HOST_INTERNAL_ALIAS}:{PORT}/mcp`. The SAN derivation deliberately added `DNS:{HOST_INTERNAL_ALIAS}` so the hostname matches — but hostname match is not trust. Nothing I can find copies the cert into the sandbox container, mounts it, or sets `SSL_CERT_FILE` / `NODE_EXTRA_CA_CERTS` in the sandbox environment. I can't settle it from this repo, since the MCP client lives in NemoClaw.

Two possibilities, and they want different follow-ups:

1. **The sandbox client verifies TLS.** Then this path fails at the first tool call with a verification error, and the feature needs a cert-distribution step (bind-mount the PEM into the sandbox, or set the trust env var in the sandbox spec) alongside the wiring that has to be re-applied anyway.
2. **The sandbox client doesn't verify** (or is configured not to). Then it works, but what the feature buys is transport encryption with no server authentication — anything that can win the race for that host:port can impersonate the orchestrator MCP. That's still a real improvement over plain HTTP for the NemoClaw requirement, but it's worth stating plainly in the docs rather than leaving readers to assume "HTTPS" means authenticated.

Either way, worth confirming which one it is and writing the answer down when the notebook wiring comes back. If it turns out to be (1), it's better to find that now than during the first `https` deploy.

nvleeryan:
/ok to test https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1375/commits/f5bdfc03ab554385d6c53bd871583a4873b9d0b3

nvleeryan:
/ok to test https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1375/commits/3d174ec58f14741c0766939c2c98e222bc61ee5a

zac-wang-nv (approved):
LGTM

Issue #47, opened 2025-09-22
Saving Chunks and their related summary , metadata


samz5320:
Hi , has anyone tried saving chucks of live-stream/uploaded video along with their metadata and summary.

PR #1399, opened 2026-07-24
ci: ensure BDD test image is available (pull, else build) before compose up

gous-75:
## Problem
Deployment-test run #47 failed at `docker compose up`:
```
test  Error  pull access denied for bdd_tests, repository does not exist ...
ERROR: Test execution failed with exit code: 1
```
The generated `docker-compose.test.yaml` runs the test container from `${BDD_TEST_IMAGE:-bdd_tests:v1.10.0_x86}` — a **bare local tag**. Nothing builds or pulls it, and the Jenkins node is `docker system prune -a`'d between runs, so the image is absent → compose tries to pull it from Docker Hub → denied → the whole `up` fails. The tests never ran.

## Change
Add `ensure_bdd_test_image()`, called from `run_docker_compose` before `docker compose up`:
1. use the image if already present locally,
2. else `docker pull` it (works when `BDD_TEST_IMAGE` points at a registry image, e.g. `nvcr.io/.../bdd_tests:<tag>`),
3. else build it from `services/vios/test/bdd_tests/Dockerfile`.

It exports `BDD_TEST_IMAGE` so `docker compose up` resolves the identical reference. To use the fast pull path, set `BDD_TEST_IMAGE` to a published registry image in the pipeline; otherwise it builds locally as a fallback.

## Verification
Unit-tested the three branches with a stubbed `docker` (present → no pull/build; pull-ok → no build; pull-fail → build) — all pass. `bash -n` clean.

### DIFF
--- a/services/vios/cicd_files/docker-compose-test/start_test.sh
+++ b/services/vios/cicd_files/docker-compose-test/start_test.sh
@@ -742,10 +742,44 @@ run_build_commands() {
 }
 
 # Function to run docker compose
+# Ensure the BDD test-runner image referenced by docker-compose.test.yaml exists.
+# Order: use it if already local -> else try to pull (works when BDD_TEST_IMAGE
+# points at a registry image) -> else build it from the Dockerfile. The compose
+# `test` service uses ${BDD_TEST_IMAGE:-bdd_tests:v1.10.0_x86}, so export the same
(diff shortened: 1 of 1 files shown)


gous-75:
Folded into #1395, which now carries all VIOS CI changes (base-image pull-first, BDD test-image pull-else-build, toolchain pull-first, and background module pushes).

PR #1420, opened 2026-07-27
Alert MS - Enable multi CPU core scaling

danh-nguyen-nvidia:
## Checklist


services/alert/enhance_alert_with_vlm.py
```diff
@@ -2432,6 +2483,187 @@ def _mark_prometheus_process_dead(process: Optional[Process]) -> None:
...
+    except Exception:
+        logger.error("Pipeline process %d failed", index, exc_info=True)
+        raise
+    logger.info("Pipeline process %d stopped", index)
+
+
+def _start_pipeline_process(config_path: str, index: int, process_count: int,
+                            ready_event: Optional[Any] = None,
+                            seeded_event: Optional[Any] = None) -> Process:
+    process = Process(
```
dineshvenkat: Pipeline and API children use the default Linux multiprocessing context after the parent has started background threads, so they fork inherited thread, lock, socket, and logging state. Can we use an explicit `spawn` context instead, consistent with Behavior Analytics v3.3.0?

services/alert/src/utils/process_scaling.py
```diff
@@ -0,0 +1,140 @@
...
+    ``auto`` would otherwise start 248 processes that consume memory and never
+    receive a partition. An explicit integer is an operator instruction and is
+    left alone; the caller warns instead.
+    """
+    raw = (config or {}).get("alert_agent", {}).get("processes", DEFAULT_PROCESS_COUNT)
+
+    if raw is None:
+        return DEFAULT_PROCESS_COUNT
+
+    if isinstance(raw, str):
```
dineshvenkat, line 202: `alert_agent.processes` currently accepts `auto` and numeric strings, and an explicit count can start when Kafka capacity is unknown or below N. Can we restrict it to a positive integer, require `event_loop` for N > 1, and fail startup unless at least N partitions are available?

services/alert/src/mdx/kafka_message_broker.py
```diff
@@ -53,6 +59,27 @@ def get_consumer(self, topic: str, group_id: str) -> Consumer:
...
 
+    def await_group_join(self, consumer: Consumer, timeout: float) -> bool:
+        """Poll until the coordinator admits ``consumer`` to its group.
+
+        ``subscribe`` only starts the join; it completes on a later poll. Until
+        then the member holds no partitions, so a producer that starts in the
+        meantime writes past a ``latest`` offset the consumer has not reached
+        yet and those records are never delivered.
+
+        Membership is read from ``memberid`` rather than ``assignment`` because
```
dineshvenkat: Kafka partition identity is dropped before async work is tracked, and there is no revoke callback to stop work for a partition that Kafka is moving. Can we preserve partition identity through dispatch, stop new work for a revoked partition, and drain its in-flight work before reassignment completes?

dineshvenkat: Readiness is currently latched from consumer-group membership, so a worker with no assigned partition can appear ready and remain ready after revoke or exit. Can we set readiness only after a non-empty assignment and clear it when assignment is lost or the worker exits?

services/alert/src/utils/process_supervisor.py
```diff
@@ -0,0 +1,159 @@
...
+            while not self._shutdown.is_set():
+                self._reap_and_restart()
+                self._shutdown.wait(self._poll_interval)
+        finally:
+            self.stop()
+
+    def request_shutdown(self) -> None:
+        self._shutdown.set()
+
+    def _reap_and_restart(self) -> None:
```
dineshvenkat: The supervisor currently restarts only the dead child, leaving the container alive while in-flight work and process state may be incomplete. Can we terminate and reap all children and exit the parent nonzero on any unexpected worker exit so Kubernetes restarts the complete instance?

services/alert/enhance_alert_with_vlm.py
```diff
@@ -2432,6 +2483,187 @@ def _mark_prometheus_process_dead(process: Optional[Process]) -> None:
...
+
+    Only child 0 seeds, and it seeds while building its pipeline. A follower
+    builds faster because it skips that write, so without this it can consume
+    its partitions against a store the leader has not filled yet and fail
+    every lookup as having no prompt.
+    """
+    if seeded_event is None or seeded_event.is_set():
+        return
+    logger.info("Pipeline process %d waiting for the prompt store to be seeded", index)
+    if not seeded_event.wait(SEED_TIMEOUT_SECONDS):
```
dineshvenkat: A follower currently times out waiting for prompt seeding and then continues consuming without confirmed prompts. Can we initialize prompts successfully in the parent before starting workers and make any initialization failure or timeout fatal?

services/alert/src/utils/process_supervisor.py
dineshvenkat: The multiprocess path has no aggregate metrics for configured, alive, and ready workers, assigned partitions, worker failures, or rebalance-drain outcomes. Can we add bounded-cardinality parent-owned metrics for these states so degraded capacity is visible?

services/alert/test/functional/capability/run_multiprocess_scaling.sh
```diff
@@ -0,0 +1,753 @@
...
+# returns and before anything is dispatched, so a message that reached the
+# worker pool is already committed under either setting and dies with its
+# process. The redelivery window batching opens is one poll batch, entered only
+# if the crash lands inside the poll loop itself.
+#
+# Deterministic and therefore asserted: batch_commit=false never replays;
+# everything that clears dedup is persisted; the restarted child resumes its
+# partitions. The per-mode loss counts are measured evidence for the window
+# above, not pass criteria — the crash lands where it lands.
+ts_033() {
```
dineshvenkat, line 694: This functional scenario still asserts child replacement and batch-commit behavior. Can we replace it with coverage for fail-all, assignment/revoke drain, readiness transitions, prompt ownership, and the multiprocess metrics?

services/alert/test/unit/test_process_supervisor.py
```diff
@@ -0,0 +1,293 @@
...
+        supervisor.stop()
+        assert len(seen) == 2
+
+    def test_count_must_be_positive(self):
+        with pytest.raises(ValueError):
+            ProcessSupervisor(count=0, spawn=Spawner())
+
+
+class TestRestart:
+    def test_dead_child_is_replaced_in_place(self):
```
dineshvenkat: This test still pins dead-child replacement. Can we update the unit suite to cover fail-all, explicit spawn context, strict process and partition validation, assignment/revoke readiness, and parent-owned prompt initialization?

services/alert/enhance_alert_with_vlm.py
danh-nguyen-nvidia: Done in 78cc62c85. Children and the FastAPI process now start from an explicit `spawn` context. The fork hazard was real: by the time the pipeline children start, the parent is already running the Prometheus HTTP server thread, the readiness thread and a child of its own. Verified against Kafka — four children start and join cleanly.

services/alert/src/utils/process_scaling.py
danh-nguyen-nvidia: Done in 98be91893, with one deviation worth flagging. `processes` is a positive integer only (`"auto"` removed), `event_loop` is required above 1, and startup fails unless the source partitions cover the count.

The check is on the **total** across source topics, matching the wording here. Note the total can pass while individual processes still idle: 8 + 1 partitions with 4 processes totals 9, yet three processes hold nothing on `mdx-alerts`. Failing per-topic would break any deployment with an asymmetric partition layout, so the short topic is named in a startup warning instead. Happy to make it strict — it is about ten lines.

services/alert/src/mdx/kafka_message_broker.py
danh-nguyen-nvidia: Done in 659bb5f36, on current-develop commit semantics as you asked.

`read_data` now emits one batch per partition instead of merging by kind, the partition travels to the single point every pipeline mode passes through, and the revoke callback drains what that partition still owes before the rebalance completes. Bounded at 15s so it cannot overrun `max.poll.interval.ms` and cost the member its place.

Scoped as you specified: this closes the dedup overlap window, and makes no claim about crash loss. Verified against a real rebalance — four drains, all completed, roughly 0.45s each.

**Caveat:** that was measured against a stubbed backend answering in 0.5s. A real VLM is slower by orders of magnitude, so `alert_bridge_rebalance_drains_total{outcome="timed_out"}` is the signal to watch before trusting the bound in production.

danh-nguyen-nvidia: Done in 34f15f0df, with the follow-up fix in 47e6b6ad7.

Readiness now hangs off `on_assign`/`on_revoke`/`on_lost` and follows the assignment lifecycle, cleared on revoke or loss. You were right that it was latched and never cleared — `set()` was the only operation on it anywhere in the file, so a child whose partitions moved, or that died, left the instance still reporting itself ready.

Gated on "assignment decided" rather than non-empty, per your reply. 47e6b6ad7 fixes a deadlock the first attempt introduced: waiting on one consumer's assignment left the others unpolled, so the group rebalance never completed and startup failed on any deployment with two source topics. Every consumer is now polled on every pass while waiting.

services/alert/src/utils/process_supervisor.py
danh-nguyen-nvidia: Done in 1fb5e56c1. Any unexpected child exit now terminates and reaps the rest, and the parent exits non-zero so the orchestrator restarts the instance whole.

Worth noting the old guard could never have fired: a child takes longer than the 60s rapid-restart window just to build its pipeline and join its group, so every restart looked like a healthy one and the counter reset each time.

Verified: killing one child leaves no orphan holding a consumer-group slot.

services/alert/enhance_alert_with_vlm.py
danh-nguyen-nvidia: Done in f448d8253. The supervising process seeds before forking anything, and a failure takes the container down.

You were right to reject the bounded wait — "consume anyway" on expiry was exactly the outcome the wait existed to prevent. The wait, its timeout and the cross-process flag behind it are all gone.

One thing this surfaced: `AlertTypeConfigLoader` never assigned `alert_configs` when the config file was absent, and three of its five readers had no `hasattr` guard. Moving seeding to the parent turned that from one broken child into a startup failure of the whole instance. Fixed in 47e6b6ad7, with a regression test.

services/alert/src/utils/process_supervisor.py
danh-nguyen-nvidia: Done in d0b2cd8c5, with a follow-up in 34e3ede81.

From the supervising process: configured, alive and ready process counts, plus exits labelled by whether a shutdown had been asked for. From each pipeline process: the partitions it holds (summed across the instance by the multiprocess collector) and each drain of revoked partitions labelled drained or timed_out. Cardinality is fixed — no per-process or per-partition labels — so the series count does not grow with the fleet.

34e3ede81 closes a race the first version had: the gauges refreshed on the supervisor's poll while readiness fires the moment the last child signals, so a scrape at startup read `ready=0` from an instance that was already serving. They are now published before the readiness line, so the log line and the numbers behind it agree.

services/alert/test/functional/capability/run_multiprocess_scaling.sh
danh-nguyen-nvidia: Done across 0e9bd4eb7, 134875899 and 4b725f481.

TS-031 and TS-033 rewritten for fail-all — they asserted child replacement, which no longer happens. TS-034 added for readiness transitions and prompt ownership; TS-035 for the fleet gauges, checked against the processes and partitions that actually exist rather than merely being present.

4b725f481 fixes three harness defects found running the suite against a real Kafka. The one worth flagging beyond this suite: the CPU sampler identified pipeline children by command line, which spawn breaks — a saturated four-process instance reported as 0.1% CPU. Anything else that identifies Alert Bridge children by command line has the same problem.

Left to an exploratory check rather than gated: the revoke drain. Forcing a rebalance needs a second member joining mid-load, and a scenario that has never run is a poor place for a hard assertion. It was exercised manually — four drains, all completed.

services/alert/test/unit/test_process_supervisor.py
danh-nguyen-nvidia: Done, spread across the commits that changed each behaviour rather than as a separate pass — a commit that changes a contract updates its own tests.

All five areas are covered: fail-all (1fb5e56c1), explicit spawn context (78cc62c85), strict process and partition validation (98be91893), assignment and revoke readiness (34f15f0df, 47e6b6ad7), and parent-owned prompt initialization (f448d8253). The suite is now 3000 tests.

Both bugs found running against Kafka have their own regression test rather than a silent fix: the assignment wait starving its own group, and the alert-type loader raising on a config file it treats as optional.

dineshvenkat (dismissed):

danh-nguyen-nvidia:
/ok to test [4f5a2ff](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1420/commits/4f5a2ff5f973d67bd81fd33b51a0c5103bc18f69)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32800586800)

danh-nguyen-nvidia:
/ok to test [e1a736e](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1420/commits/e1a736ec2ec517cd881212707dd05f50bf7887ab)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32815026513)

trongp-nvidia (dismissed):

danh-nguyen-nvidia:
/ok to test [f6528a2](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1420/commits/f6528a20725298b0523b3ae9d889da012097b257)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32832954033)

trongp-nvidia (approved):

danh-nguyen-nvidia:
/ok to test [116e087](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1420/commits/116e087d5917181b39dcaed1135cfb220e537da7)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32838838323)

PR #1351, opened 2026-07-21
feat(agent): Move base profile to RTVI-VLM 

liamy-nv:
## Description
- Migrated the base profile from standalone Cosmos NIM to RTVI-VLM
- Updated Docker Compose and Helm to deploy/configure rtvi-vlm
- Added local and remote RT-VLM routing, model IDs, and GPU settings
- Updated dev-profile.sh to map supported --vlm models to RTVI-VLM checkpoints
- Updated deployment tests and README documentation for the new configuration

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/HEAD/CONTRIBUTING.md).
- [x] I have installed and run [pre-commit hooks](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/HEAD/CONTRIBUTING.md#1-enable-pre-commit-hooks) locally (`uv run pre-commit install` once, then hooks run on every `git commit`).
- [x] Every commit on this PR is [DCO sign-off](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/HEAD/CONTRIBUTING.md#developer-certificate-of-origin-dco)'d (`git commit -s` adds a `Signed-off-by` trailer that certifies you have the right to submit the change under Apache-2.0).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #1301, opened 2026-07-15
Expand alerts skill for VSS 3.3

danh-nguyen-nvidia:
## Checklist
- [ ] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/HEAD/CONTRIBUTING.md).
- [ ] I have installed and run [pre-commit hooks](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/HEAD/CONTRIBUTING.md#1-enable-pre-commit-hooks) locally (`uv run pre-commit install` once, then hooks run on every `git commit`).
- [ ] Every commit on this PR is [DCO sign-off](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/HEAD/CONTRIBUTING.md#developer-certificate-of-origin-dco)'d (`git commit -s` adds a `Signed-off-by` trailer that certifies you have the right to submit the change under Apache-2.0).
- [ ] New or existing tests cover these changes.
- [ ] The documentation is up to date with these changes.

PR #1254, opened 2026-07-10
feat(skills): Refactor skill based report generation

liamy-nv:
## Summary
- Refactored `vss-generate-video-report` to a skill-first flow and removed reliance on legacy `/generate` behavior.
- Made Mode A profile-agnostic with two input paths.
- Kept Mode B as incident-range reporting via analytics MCP (`/vss-query-analytics`).
## Mode A Paths
- **A1:** VST clip URL + VLM endpoint.
- **A2:** Local video file on runtime disk or base64 payload + explicit VLM endpoint/model.
## Key Changes
- Moved report templates from `assets/` to `references/report-templates/` and updated all references.
- Added centralized default prompt at `references/default-vlm-prompt.md`.
- Clarified strict per-mode prerequisites, including runtime-local file location requirements for Mode A/A2.
- Improved endpoint/discovery guidance and explicit failure behavior.
- Added required long-video rule: hand off to LVS for clips/videos longer than 2 minutes.
## HITL Behavior
- Resolve HITL using legacy runtime flag first: `video_report_gen.hitl_enabled`.
- Use harness fallback only if runtime config is unavailable: `HITL_ENABLED=true|false`.
- Default to HITL off when unset, with first-run note to user.
- Enforced explicit approval loop with `APPROVE | EDIT | NEW` until `APPROVE`, including anti-stall guardrails.
## Evals / Validation
- Updated evals for new template paths and skill contract behavior.
- Added coverage for HITL on/off/fallback.
- Added negative checks for prior stall patterns (`yes/confirm` and empty-string assumptions).
- Added coverage for profile-agnostic Mode A routing.
- Added coverage for local-file/base64 input path.
- Added coverage for long-video LVS handoff.
- Added coverage for VLM ambiguity resolution.

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/HEAD/CONTRIBUTING.md).
- [x] I have installed and run [pre-commit hooks](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/HEAD/CONTRIBUTING.md#1-enable-pre-commit-hooks) locally (`uv run pre-commit install` once, then hooks run on every `git commit`).
[...]

PR #1422, opened 2026-07-27
fix(ci): grant statuses:read so NVSkills CI stops failing at startup

zac-wang-nv:
## The `/nvskills-ci` slash command is dead

This is not a cosmetic-red-check PR. `Team Request NVSkills CI` is what dispatches NVSkills CI when someone comments `/nvskills-ci` on a PR — and since ~15:10 UTC on 2026-07-27 that dispatch has done **nothing**, silently.

**Every `issue_comment`-triggered run since the break: 72 runs, 100% `startup_failure`, zero successes** (most recent 2026-07-28T00:32Z).

The command is in active use — recent invocations on #1254, #1397, #1301, #1227, #1351. Each of those users commented `/nvskills-ci` and got silence: no dispatch, no error, no comment. `startup_failure` produces no jobs, no logs and no annotations, so there is nothing for them to look at. `gh run view --log` returns `log not found`.

## Root cause

`NVIDIA/skills` added `statuses: read` to the called workflow's permissions in [`bc8b450a`](https://github.com/NVIDIA/skills/commit/bc8b450a) (2026-07-20).

| | |
|---|---|
| This caller grants | `contents: read`, `pull-requests: read` |
| `NVIDIA/skills/.github/workflows/team-request.yml@main` declares | `contents: read`, `pull-requests: read`, **`statuses: read`** |

A reusable workflow cannot request a permission its caller withheld, so the run is rejected at startup. Permission validation happens *before* job-level `if:` evaluation, which is why even pushes the gate would have skipped come back red.

## Note: we are granting this for a job that cannot run here

Worth stating plainly, since it looks odd. The called workflow has two jobs:

| Job | Gate | Uses `statuses: read`? |
|---|---|---|
| `require-nvskills-ci` | `if: github.event_name == 'pull_request'` | **yes** — it is the only consumer (reads `/commits/{sha}/statuses`) |
| `request` | `issue_comment` with `/nvskills-ci`, or signature push | no |

This caller has no `pull_request` trigger, so `require-nvskills-ci` never executes for this repo. The permission is therefore required **purely to pass startup validation** for a job that is inert here.

Granting it is still the correct minimal fix: GitHub validates the called workflow's entire declared permission set at startup, regardless of which jobs will actually run. The alternative — adopting the upstream `pull_request` trigger so that job becomes live — is a real behaviour change and belongs in its own PR (see below).

## Confirmed by controlled A/B

Two branches pushed 16 seconds apart, identical with respect to this workflow except for the permission block:

| Time (UTC) | Branch | Result |
|---|---|---|
| 19:21:37 | `fix/nvskills-ci-statuses-permission` (**this PR**) | **`skipped`** ✅ |
| 19:21:53 | `fix/ghcr-anonymous-token-auth` (unchanged caller) | `startup_failure` ❌ |

`skipped` is the correct outcome for those pushes: the workflow resolves, loads the called workflow, and the `if:` gate correctly declines. That is the difference between "the gate evaluated and said no" and "the run died before the gate was read".

## Not specific to this repo

Other consumers of the same reusable workflow are startup-failing identically — worth flagging upstream:

| Repo | Latest `Request NVSkills CI` |
|---|---|
| `NVIDIA/Megatron-LM` | `startup_failure` |
| `NVIDIA-AI-Blueprints/aiq` | `startup_failure` |
| `NVIDIA/physicsnemo` | `startup_failure` |

The canonical caller template at [`NVIDIA/nvskills-ci:templates/team-request-workflow.yml`](https://github.com/NVIDIA/nvskills-ci/blob/main/templates/team-request-workflow.yml) already lists all three permissions; consumer repos drifted after `bc8b450a`.

## Deliberately out of scope

That template has also grown `pull_request` (path-filtered) and `status` triggers with matching `if:` clauses. This PR does **not** adopt them. Doing so would activate `require-nvskills-ci` and change which events gate a PR — a separate decision, worth taking on its own merits now that we know that job is currently inert here.

## Verification

`team-request.yml` parses as valid YAML; the A/B above is the functional proof.

### DIFF
--- a/.github/workflows/team-request.yml
+++ b/.github/workflows/team-request.yml
@@ -14,9 +14,14 @@ jobs:
       (github.event_name == 'push' &&
         github.actor == (vars.NVSKILLS_SIGNATURE_PUSH_ACTOR || 'nv-skills-ci[bot]') &&
         startsWith(github.event.head_commit.message, vars.NVSKILLS_SIGNATURE_COMMIT_TITLE || 'Attach NVSkills validation signatures'))
+    # Must cover every permission the called workflow declares, otherwise the
+    # run dies with startup_failure before any job executes. NVIDIA/skills
+    # added `statuses: read` in bc8b450a (it reads the commit status list to
+    # find the NVSkills CI verdict), so the caller has to grant it too.
     permissions:
       contents: read
       pull-requests: read
+      statuses: read
     uses: NVIDIA/skills/.github/workflows/team-request.yml@main
     secrets:
       NVSKILLS_CI_DISPATCH_TOKEN: ${{ secrets.NVSKILLS_CI_DISPATCH_TOKEN }}


zac-wang-nv:
## Confirmed by controlled A/B

Both fix branches were pushed 16 seconds apart. They are identical with respect to this workflow except that this branch adds `statuses: read`:

| Time (UTC) | Branch | `Team Request NVSkills CI` |
|---|---|---|
| 19:21:37 | `fix/nvskills-ci-statuses-permission` (**this PR**) | **`skipped`** ✅ |
| 19:21:53 | `fix/ghcr-anonymous-token-auth` (unchanged caller) | `startup_failure` ❌ |

`skipped` is the correct outcome for these pushes: the workflow now resolves and loads the called workflow, and the job's `if:` gate correctly declines to run (the commit is neither a `/nvskills-ci` comment nor a signature commit from `nv-skills-ci[bot]`).

That is the difference between "the gate evaluated and said no" and "the run died before the gate was ever read" — which is exactly the bug.

PR #1423, opened 2026-07-27
fix(ci): stop sending the NGC key to ghcr.io in the container source check

zac-wang-nv:
## Problem

`Check Agent Container Source` and `Check UI Container Source` both fail on `develop`:

```
tag:      develop-latest
manifest: UNREACHABLE (token endpoint returned 403: Forbidden)
[FAIL] could not fetch the image manifest for
       ghcr.io/nvidia-ai-blueprints/vss/vss-agent:develop-latest.
       The image is missing from the registry, the credentials are wrong,
       or the network is broken — fix the upstream promotion rather than
       falling back to the tag's git-SHA suffix.
```

The image is fine. The credential is the bug.

Deployable tags now resolve to `ghcr.io`, but `_fetch_bearer_token()` guarded the NGC credential on `if ngc_key:` alone, with no registry check. The workflow sets `NGC_CLI_API_KEY` on every run, so ghcr.io always received a Basic header carrying `$oauthtoken:<NGC key>` — and rejected it.

Reproduced against the live registry:

```
ghcr + NGC key set  -> token endpoint returned 403: Forbidden
ghcr + no creds     -> OK (anonymous pull token)
```

The credential was strictly worse than sending nothing.

## Two defects, one root cause

1. **NGC creds leak to non-NGC registries.** `nvcr.io` needs `$oauthtoken` + NGC key; ghcr.io answers that Basic header with 403. The public package resolves anonymously without it.
2. **The GHCR credential branch was dead code.** `elif registry_username and registry_password:` sits directly below the NGC branch. Since `NGC_CLI_API_KEY` is set on every run, the `elif` was unreachable — the `GHCR_USERNAME`/`GHCR_TOKEN` support already in the script had never once executed. The workflows did not pass those secrets either.

The failure message also actively misdirects: it blames a missing image or broken promotion, so the natural next step is to go hunting through the promotion pipeline rather than the auth header.

## Fix

- Gate the NGC credential on `registry.endswith("nvcr.io")`, so ghcr.io falls through to anonymous (public) or to the registry-specific branch (private).
- Pass the existing `GHCR_USERNAME` / `GHCR_TOKEN` secrets through both workflows, making that branch reachable.

## Verification

Against the live registries, using the patched function:

| Case | Before | After |
|---|---|---|
| `ghcr.io` + NGC key set | `403 Forbidden` | **OK**, token len 76 |
| `ghcr.io`, no creds | OK | OK, token len 76 |
| `nvcr.io` + NGC key | Basic auth sent | Basic auth still sent (unchanged) |

The `nvcr.io` case still returns 403 under test only because the test key is fake — that confirms the NGC path is preserved rather than bypassed.

`.github/scripts/test_source_check_only_nonbuild.py`: **26 tests, OK**.

## Related, not fixed here

Both workflows skip this step on `pull-request/*` refs (`if: ... && !startsWith(github.ref_name, 'pull-request/')`), so the check only ever runs on `main`/`develop` — where it has been failing 100% of the time. It therefore cannot block a PR today. Worth a separate look at whether that skip is still wanted.

### DIFF
--- a/.github/scripts/check_container_tag_source.py
+++ b/.github/scripts/check_container_tag_source.py
@@ -515,7 +515,14 @@ def _fetch_bearer_token(
     token_url = f"{realm}?service={urllib.parse.quote(params.get('service', ''))}&scope={urllib.parse.quote(scope)}"
 
     req = urllib.request.Request(token_url)
-    if ngc_key:
+    # The NGC key is only a credential for nvcr.io. Sending it to any other
+    # registry is worse than sending nothing: ghcr.io answers a Basic header it
+    # cannot resolve with 403, whereas an unauthenticated request to the same
+    # endpoint yields a perfectly good anonymous pull token for a public
+    # package. Gating on the registry also keeps the registry-specific branch
+    # below reachable — NGC_CLI_API_KEY is set on every run of this check, so an
+    # unguarded `if ngc_key` made those credentials dead code.
+    if ngc_key and registry.endswith("nvcr.io"):
         basic = base64.b64encode(f"$oauthtoken:{ngc_key}".encode()).decode()
         req.add_header("Authorization", f"Basic {basic}")
     elif registry_username and registry_password:
--- a/.github/workflows/check-agent-container-source.yml
+++ b/.github/workflows/check-agent-container-source.yml
@@ -67,4 +67,9 @@ jobs:
         if: steps.filter.outputs.skip != 'true' && !startsWith(github.ref_name, 'pull-request/')
         env:
           NGC_CLI_API_KEY: ${{ secrets.NGC_CLI_API_KEY }}
+          # Deployable tags now resolve to ghcr.io, which the NGC key cannot
+          # authenticate. Public packages read fine anonymously; these let the
+          # check keep working if a package is ever made private.
+          GHCR_USERNAME: ${{ secrets.GHCR_USERNAME }}
+          GHCR_TOKEN: ${{ secrets.GHCR_TOKEN }}
         run: python3 .github/scripts/check_container_tag_source.py --image-name vss-agent
--- a/.github/workflows/check-ui-container-source.yml
+++ b/.github/workflows/check-ui-container-source.yml
@@ -68,4 +68,9 @@ jobs:
         if: steps.filter.outputs.skip != 'true' && !startsWith(github.ref_name, 'pull-request/')
         env:
           NGC_CLI_API_KEY: ${{ secrets.NGC_CLI_API_KEY }}
+          # Deployable tags now resolve to ghcr.io, which the NGC key cannot
+          # authenticate. Public packages read fine anonymously; these let the
+          # check keep working if a package is ever made private.
+          GHCR_USERNAME: ${{ secrets.GHCR_USERNAME }}
+          GHCR_TOKEN: ${{ secrets.GHCR_TOKEN }}
         run: python3 .github/scripts/check_container_tag_source.py --image-name vss-agent-ui


zac-wang-nv:
## End-to-end verification

Ran the real check script against the live registry with `NGC_CLI_API_KEY` set, exactly as CI does — same command, same env, same `develop-latest` tags. Only the code differs:

**Before (unpatched worktree, same env):**
```
tag:      develop-latest
manifest: UNREACHABLE (token endpoint returned 403: Forbidden)
[FAIL] could not fetch the image manifest for
       ghcr.io/nvidia-ai-blueprints/vss/vss-agent:develop-latest.
Result: 1 failure(s) out of 1 unique ref(s).
```

**After (this branch):**
```
[1/1] ghcr.io/nvidia-ai-blueprints/vss/vss-agent:develop-latest
  manifest: com.nvidia.vss.source_tree_sha=0c76a712909d498a6c32aa911bc64a58082a7c23
  comparing services/agent/:
    at HEAD:           0c76a712909d498a6c32aa911bc64a58082a7c23
    in image manifest: 0c76a712909d498a6c32aa911bc64a58082a7c23
    → identical
  [PASS]
Result: all 1 unique ref(s) match.
```

`--image-name vss-agent-ui` likewise goes green (`65e327e4…` identical, `[PASS]`).

So the gate does not merely stop erroring — it resumes doing its actual job: reading `com.nvidia.vss.source_tree_sha` from the manifest annotation and confirming the published image matches the source tree. That comparison has been dead on `develop` for as long as the tags have pointed at ghcr.io.

## Note on package visibility

`ghcr.io/nvidia-ai-blueprints/vss/vss-agent` currently reads **anonymously** end-to-end (token *and* manifest), so the fix works today with no credentials at all. That is worth stating explicitly given the org-wide "members may only create private packages" policy — these particular package names have evidently been flipped public.

The `GHCR_USERNAME`/`GHCR_TOKEN` plumbing is therefore not load-bearing right now; it is there so the check keeps working if a package is ever made private again, and so the credential branch that was previously unreachable is genuinely wired.

zac-wang-nv:
## Superseded by #1425 — recommend closing

Further digging says this fix is aimed at code that should not exist.

I claimed above that the alert-ms check was the one still doing real work, since it lacks the `pull-request/` skip. That was wrong. Its comparison step carries `continue-on-error: true`, so it runs the check and discards the result — [run 30295132615](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/30295132615) logs the exact 403 this PR fixes, `Result: 1 failure(s)`, and still reports **success** at both step and job level.

So all three container-source workflows are non-functional gates, and they are the *only* callers that pair `NGC_CLI_API_KEY` with a `ghcr.io` ref. `ghcr_image_guard.py` — the one consumer that does real enforcement — passes `GHCR_USERNAME`/`GHCR_TOKEN` and no NGC key, so it never takes the broken branch.

Fixing the auth here would therefore restore a comparison that `build-dev-images.yml` already performs at push time, with working credentials, and actually enforces.

**Suggest closing this in favour of #1425**, which deletes all three (−795 lines) and keeps `check_container_tag_source.py` as the shared library it now is.

Worth reviving only if #1425 is rejected — the 403 diagnosis and the end-to-end verification above stand on their own either way.

zac-wang-nv:
Closing — superseded by #1425, which deletes the three workflows this was fixing.

Verified against the post-#1425 tree: those three are the only callers that pair `NGC_CLI_API_KEY` with a `ghcr.io` manifest read. Once they are gone, nothing reaches the `if ngc_key:` branch — `skills-eval.yml` and `skills-eval-daily.yml` set the NGC key but make no registry-manifest calls, `ci.yml`'s only hit is the hermetic `test_ghcr_image_guard.py`, and `ghcr_image_guard.py` itself passes `GHCR_USERNAME`/`GHCR_TOKEN` with no NGC key.

The diagnosis stands on its own if it is ever needed again: sending `$oauthtoken:<NGC key>` to ghcr.io's token endpoint returns 403, while the same request with no Authorization header returns a valid anonymous token — the credential was strictly worse than none. Full reproduction and end-to-end verification are in the comments above.

PR #516, opened 2026-05-18
skills: expand VIOS API reference with full endpoint coverage

nvprakhar:
## Description
Bring the VIOS skill's API documentation up to parity with the vms_shim vios-api skill. The previous reference covered only basic CRUD (sensor list/add/delete, timelines, clip extraction, snapshot, file upload). Add the remaining endpoint families needed for agent-driven control of a running VIOS/VST backend:

- Sensor lifecycle: credentials, network (GET/POST), reboot, replace, encode/image settings (GET/POST), with the ONVIF-only caveats.
- Sensor-MS timelines: /sensor/<id>/timelines and /sensor/timelines.
- Recorder microservice: status (aggregate + per-stream shape differences), stream add/remove, start/stop/event, schedule CRUD with 5-field CRON, timelines, /files segment listing.
- WebRTC live + replay session control: iceServers, stream/start, iceCandidate POST/GET, setAnswer, stop/pause/resume, query (peerid lowercase quirk), swap, settings, stats, status, plus replay-only trick-mode seek and live->VOD swap.
- RTSP proxy: /proxy/info, /proxy/streams, stream/add, idempotent delete + the /proxy/session alias.
- Storage file management beyond clip download: /storage/info, single-stream /storage/<streamId>, file/list, /file/<sensorId>/list, /file/<sensorId>/path, /file/path, mediainfo, file download by id, full-coverage clip download, delete by path or id, protect / protected.
- Service configuration: GET on all six microservices, POST writable subsets on sensor/live/replay (storage/proxy/recorder read-only).
- Service discovery: /version and /help per microservice with the per-service response-key differences called out.
- Cross-service rules: streamId header convention, mediaSessionId lifecycle, bearer auth, CRON schedule format, WebRTC payload sizing, RTSP port discovery, opaque numeric fields, 404 vs error-JSON.

Consolidate so references/api-reference.md is the single source of API truth. Remove the duplicate sections 4-8 that were embedded in SKILL.md; SKILL.md now keeps only setup, service map, a one-line pointer to the reference, and the sensor-name -> clip/snapshot workflow. Result: SKILL.md 366 -> 182 lines, api-reference.md 170 -> 1108 lines, zero endpoints lost in the move (verified by diffing the unique /vst/api/v1/... paths against the source skill).

[...]

PR #1424, opened 2026-07-27
fix(skill-eval): build Harbor trace URLs in run_leg.py

zac-wang-nv:
## Problem

Trace links in skills-eval PR comments are assembled by hand by the eval agent from the template in `AGENTS.md`. On [PR #1254 / run 30284131217](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/30284131217), all seven step links ended in the `--include-task-name` **filter**:

```
…/aws/anthropic%2Fbedrock-claude-opus-4-6/step-7          ← what shipped
…/aws/anthropic%2Fbedrock-claude-opus-4-6/nvidia-vss%2Fvss-generate-video-report-base-l40s-step-7   ← what the viewer indexes
```

The viewer resolves its task route by Harbor's fully-qualified `task_name`, so nothing matched. And because the viewer is a client-side SPA (`"ssr": false, "isSpaMode": true`) where *every* route returns the same HTTP 200 shell, each link opened as a **blank page** rather than 404ing — indistinguishable from missing trace data.

This matters more than a cosmetic link bug: the results tarball deliberately excludes `agent/` trajectories (secret-leak fix, #516 leaked `NGC_CLI_API_KEY` that way). The trace link is therefore the *only* route to an agent's reasoning from CI, and when it's wrong there's no fallback.

Not a regression in code — every earlier comment on that PR got it right (90 links across 07-20 → 07-24; only the newest 7 are broken). It's a hand-assembly step that the agent will occasionally get wrong, so this moves it into the wrapper.

## Worked example — run 30284131217, step-7

The viewer's route table binds the last path segment to `:taskName`:

```
routes/task   path=jobs/:jobName/tasks/:source/:agent/:modelProvider/:modelName/:taskName
```

and the route module puts it straight on the wire as a `task_name` filter:

```js
queryFn: () => A(s, N, $, {taskName: n, source: u, agentName: m, modelName: x})
…
r?.taskName && n.set("task_name", r.taskName)
```

Replaying that exact `GET /api/jobs/<job>/trials` request with each candidate tail:

| `task_name=` | result |
|---|---|
| *(no filter — is the data even there?)* | `total=1  ['step-7__E6dBECL']` |
| `step-7` — what the comment linked | `total=0  []` |
| `nvidia-vss/vss-generate-video-report-base-l40s-step-7` | `total=1  ['step-7__E6dBECL']` |

Zero rows is the blank page. Note the first line: the trial was present the whole time — still returning `total=1` hours after the run — so this is not the 12 h scratch GC either. That sweep carries `! -path '*/_viewer/*'` and only reaps `results/<slug>/<run_id>/`, the artifact-side copy, never the `_viewer/<job>/` path the trace links point at.

Diagnostic tell for next time: **if browsing the viewer and filtering finds the trial but the link opens blank, the link is wrong — not the data.** Missing data fails both paths.

The working URL, which differs from the shipped one in exactly one segment:

```
https://harbor-13xh5gpe7.brevlab.com/jobs/vss-generate-video-report__base_profile_report__L40S__30284131217__2026-07-27__17-16-47/tasks/l40s/claude-code/aws/anthropic%2Fbedrock-claude-opus-4-6/nvidia-vss%2Fvss-generate-video-report-base-l40s-step-7
```

`trace_url()` in this PR reproduces that string byte-for-byte from the trial's own `result.json` — that equality is asserted in `test_trace_url_matches_the_viewer_route`.

## Change

`run_leg.py` now owns both halves of the viewer publish:

- **`publish_trace()`** — copies each finished trial into the shared `_viewer` root (contents, flattened, copy-not-move) after every harbor invocation, including timed-out ones. This is work the agent previously did by hand with a `cp -a "$RES/<date>/."` form that had to be spelled exactly right (trailing `/.`) to stay idempotent across a multi-step spec.
- **`trace_url()`** — derives every segment from the trial's own `result.json` (`source`, `agent_info.name`, `agent_info.model_info.provider|name`, `task_name`). The link cannot drift from what the viewer indexes, and it's produced even when `harbor-view.service` is down.

Why this layer: `HarborInvocation.include_task_name` is the *filter* (`step-7`) — the wrapper never had the fully-qualified name in process, and the agent had to make a separate `GET /api/jobs/<job>/tasks` call to learn it while `step-7` was already in front of it in the harbor command, the trial dir name, and the reward path. Reading the named `task_name` field out of `result.json` removes the choice: six fields in that file contain a `step-7`-shaped string, and only one is the right one.

URLs are appended to `<results-root>/trace-urls.tsv` and echoed as `[run-leg] trace: <step> -> <url>`. `AGENTS.md` and `README.md` now instruct the agent to paste them verbatim and never construct one; a step with no row errored before producing `result.json` and should be reported without a link.

Publish failures are caught and logged, never fatal — the verdict comes from `reward.txt`, not the viewer.

## Verification

- `trace_url()` run against the real `result.json` from run 30284131217 reproduces the hand-verified working URL byte-for-byte.
- 5 new tests in `tests/test_run_leg.py` cover the exact regression (tail must not be the bare filter), `%2F` segment encoding, incomplete/missing `result.json`, and that publish flattens into the viewer job dir without disturbing the collector's copy. All pass.
- `ruff check` clean on both changed Python files.

Note: `python3 .github/skill-eval/tests/test_run_leg.py` reports 3 failures in `PoolCandidates` (RTX 4090 allowlist). Those are **pre-existing on `develop`** — the same 3 fail on a pristine checkout (22 tests / 3 failures there, 27 / 3 here) and are untouched by this change.

## Known limits

- **Not yet exercised by a live leg.** This PR's own diff plans zero eval legs (`plan_matrix.py` selects on `skills/**` and `adapters/<skill>/**`), so `/ok to test` runs everything except an eval. A `workflow_dispatch` of Skills Eval against this branch with a single-leg skill would exercise it end to end.
- **`test_run_leg.py` is not in `ci.yml`'s pytest list** — that job runs a hand-listed subset of `.github/skill-eval/tests/`, so these tests only run locally. Adding it requires first fixing the 3 pre-existing `PoolCandidates` failures above; worth a follow-up.
- **The agent still has to paste the URL.** This change guarantees a correct URL is computed, logged, and one line away — it does not mechanically prevent an agent from hand-writing one anyway. A stronger follow-up would validate that every `harbor-*` link in a posted comment appears in `trace-urls.tsv`, turning a wrong link into a CI failure instead of a silent blank page.

## Scope

`.github/skill-eval/` only — no skill, adapter, or eval-spec surface is affected.

### DIFF
--- a/.github/skill-eval/AGENTS.md
+++ b/.github/skill-eval/AGENTS.md
@@ -678,50 +678,63 @@ to pick up a trial, its directory must live under
 `/tmp/skill-eval/results/_viewer/<leg-slug>__<run_id>__<date>/` as a
 **real dir (not a symlink)**, flattened — no nested `<date>/` level.
 The `<leg-slug>` keeps concurrent legs from colliding on one viewer
-entry. **Copy** (don't move) from this leg's scoped results root:
+entry.
 
-```bash
-VIEWER_JOB="/tmp/skill-eval/results/_viewer/${LEG}__${GITHUB_RUN_ID}__<date>"
-mkdir -p "$VIEWER_JOB"
-cp -a "$RES/<date>/." "$VIEWER_JOB/"
-```
+**`run_leg.py` does this publish for you** (`publish_trace`), after
+every trial including timed-out ones. It copies (never moves) the
+date dir's *contents* into a pre-made job dir — the workflow's
+"Collect results" step runs *after* this agent and tars `$RES` for
+the artifact, so a `mv` would upload an artifact with no
+`result.json`. Copying keeps `$RES` intact for the collector (which
+excludes `agent/` from the public tarball) while the `_viewer` copy
+keeps `agent/` for the live Harbor Trace tab. You do not run `cp`
+yourself.
 
-`cp -a`, **not `mv`** — the workflow's "Collect results" step runs
-*after* this agent and scans + tars `$RES` for the artifact. A `mv`
-would leave `$RES` empty and the uploaded artifact would have no
-`result.json` or traces. Copying keeps `$RES` intact for the collector
-(which excludes `agent/` from the public tarball) while the `_viewer`
-copy keeps `agent/` for the live Harbor Trace tab.
+### Trace URLs — read them, never build them
 
-**Use the `mkdir -p` + `cp -a "$RES/<date>/."` (trailing `/.`) form
-above, not `cp -a "$RES/<date>" "$VIEWER_JOB"`** — the latter is not
-idempotent: on the *second* trial of a multi-step spec, `$VIEWER_JOB`
-already exists, so `cp -a <date> <existing-dir>` nests the trial as
-`$VIEWER_JOB/<date>/...` and Harbor only sees the first (top-level)
-trial. Copying the directory *contents* into a pre-made `$VIEWER_JOB`
-keeps every trial at the job's top level.
-
-Do this between trials so each new trial's traces are reachable
-via the SPA URL:
+`run_leg.py` writes one row per finished trial to
+`<results-root>/trace-urls.tsv`:
 
 ```
-https://harbor-${BREV_ENV_ID}.brevlab.com/jobs/<leg-slug>__<run_id>__<date>/tasks/<source>/<agent>/<provider>/<model>/<task>
+<include-task-name>\t<trial-dir>\t<url>
 ```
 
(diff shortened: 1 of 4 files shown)


sarathm-nvidia (approved):
LGTM

PR #1425, opened 2026-07-27
ci: drop the three container-source check workflows

zac-wang-nv:
Supersedes #1423.

## All three are non-functional gates

Each was disarmed by a *different* accident, which is why this went unnoticed — the checks are green, and the greenness means nothing.

| Workflow | Step 4 (the only step that guards anything) | Effect |
|---|---|---|
| `check-agent-container-source` | `!startsWith(github.ref_name, 'pull-request/')` | comparison never runs on a PR |
| `check-ui-container-source` | same | comparison never runs on a PR |
| `check-alert-ms-container-source` | `continue-on-error: true` | comparison runs, result discarded |

Agent and UI were required contexts that are vacuously green on every PR and cannot block anything. Their only live effect is turning `develop` red after a merge.

### alert-ms is failing right now and reporting success

It does run the comparison — and throws the answer away. Run [30295132615](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/30295132615) on `develop`:

```
[1/1] ghcr.io/nvidia-ai-blueprints/vss/vss-alert-ms:develop-latest
  manifest: UNREACHABLE (token endpoint returned 403: Forbidden)
  [FAIL] could not fetch the image manifest ...
Result: 1 failure(s) out of 1 unique ref(s).
```

Step conclusion: `success`. Job conclusion: `success`. **40 of the last 40 runs are green**; the comparison actually executed in 3 of the last 25, and failed every time it did.

## The guarantee already lives earlier, and is enforced

`build-dev-images.yml:191`, at push time, with working GHCR credentials:

```yaml
- name: Verify pushed labels satisfy the container-source gate
  env:
    GHCR_USERNAME: ${{ github.actor }}
    GHCR_TOKEN: ${{ secrets.GITHUB_TOKEN }}
  run: |
    python3 .github/scripts/ghcr_image_guard.py verify \
      --ref ... --expect-tree-sha ... --expect-source-path ... --expect-image-name ...
```

Same tree-SHA comparison, against the image being pushed, **blocking the push** instead of reporting after the merge. #1190 added that step and the `pull-request/` skip to these workflows in the same change — this finishes that migration.

This is also why the 403 never surfaced anywhere else: `ghcr_image_guard.py` passes `GHCR_USERNAME`/`GHCR_TOKEN` and **no** NGC key, so it never takes the broken code path. The three workflows deleted here are the only callers that pair `NGC_CLI_API_KEY` with a `ghcr.io` ref.

## Scope

Also removes `source_check_only_nonbuild.py` and its test: these three workflows were their only consumers.

**`check_container_tag_source.py` is retained** — it is a shared library imported by `release_set.py`, `compose_image_golden.py`, and `ghcr_image_guard.py`.

## Ruleset prerequisite — done

`Check vss-agent tag source` and `Check vss-agent-ui tag source` were required contexts on ruleset **Protect develop** (`15323753`). Merging this while they were still required would have left every PR perpetually pending on a status that can no longer report.

Both have now been removed. Verified — the required contexts on `Protect develop` are now:

```
DCO Sign-off
Check folder structure
Copyright Headers
Security Scan (detect-secrets)
SonarQube Scan (agent)
SonarQube Scan (ui)
SonarQube Scan (skills)
```

The repo's other eight rulesets (`Protect main`, `Protect release branches`, `Restrict pull-request branches`, …) were swept for lingering references to either context — none require them. **This is now safe to merge.**

## Verification

- `test_compose_image_golden.py` — 13 tests, OK
- `test_ghcr_image_guard.py` — 16 tests, OK
- `ghcr_image_guard`, `release_set`, `compose_image_golden` all still import cleanly
- No dangling references to the deleted files remain

Net: **−795 lines**.


akshayanv (approved):

kaushikc-nvidia (approved):

kaushikc-nvidia:
/ok to test [57a0127](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1425/commits/57a0127299b2cf6877f1b7af689ffc95fbdaa2ad)

hugoverjus (approved):

PR #1434, opened 2026-07-28
Enhance code coverage for Alert MS

danh-nguyen-nvidia:
## Checklist


services/alert/test/unit/handlers/test_event_loop_pipeline_mixin.py
```diff
@@ -0,0 +1,796 @@
...
+    async def test_a_system_prompt_alone_is_enough_to_continue(self, pipeline):
+        pipeline.prompt_manager.get_prompts_for_message.return_value = (None, "system")
+
+        assert await pipeline._prepare_message_context_async({}, "cam-1", {}, 0.0) == (
+            None,
+            "system",
+        )
+
+
+class TestResolveVideoUrlAsync:
```
dineshvenkat, line 392: Every message in this class comes from `dict(MESSAGE)`, which always includes `end` and `timestamp`. The code under test reads `message['end']` and `message['timestamp']` unguarded (at the video-URL call and its retry), so a message missing those keys raises an unhandled `KeyError` — a failure this service has actually hit. No test here feeds a message without `end`/`timestamp`, so that path is exercised only implicitly by the generic unexpected-error test and its intended outcome is never pinned.

services/alert/test/unit/handlers/test_event_loop_pipeline_mixin.py
```diff
@@ -0,0 +1,796 @@
...
+        await pipeline._publish_error_and_complete_async(
+            dict(MESSAGE), "user", "sys", "vlm_timeout", 0.0, {}
+        )
+
+        pipeline.vlm_enhanced_event_sink.publish_error_async.assert_awaited_once()
+
+
+class TestHandleVlmExceptionAsync:
+    @pytest.mark.asyncio
+    async def test_the_exception_is_classified_and_published(self, pipeline):
```
dineshvenkat, line 705: Several tests here mock the pipeline's own helper methods (`_apply_vlm_exception`, `_compute_fingerprint`, `_complete_event_after_publish`, `_publish_outcome_and_complete`, `_update_enrichment_with_mode`) and then assert only that those methods were called — this test, for instance, asserts `_apply_vlm_exception` and `_log_vlm_exception` were invoked. No test runs the full event path with the real helpers, so wrong arguments, a missing message mutation, incorrect failure classification, or broken ordering between these steps would still pass.

services/alert/test/unit/utils/test_event_utils.py
```diff
@@ -0,0 +1,243 @@
...
+        result = normalize_alert_message(
+            {
+                "sensor": {"id": "cam-1"},
+                "analyticsModule": {"id": "intrusion"},
+                "_normalized_added_fields": ["earlier"],
+            }
+        )
+        assert set(result["_normalized_added_fields"]) == {"earlier", "sensorId", "category"}
+
+    def test_object_only_message_raises_attribute_error(self):
```
dineshvenkat: This test asserts that the current behavior — raising `AttributeError` on an object-only message — is the expected outcome, and its docstring notes the behavior is broken/unfixed. Several other tests do the same (`test_prompt_manager.py::test_get_prompts_for_entity_is_broken`, `test_schema_util.py::test_none_raises_attribute_error`, `test_sink_elastic.py::test_a_blank_host_string_slips_past_the_guard`). Pinning known defects as passing tests counts a bug as "covered," and a later fix of the defect will make these tests fail as if it were a regression.

services/alert/test/unit/vlm/test_vlm_client_payload.py
```diff
@@ -0,0 +1,630 @@
...
+        assert kwargs["messages"][0]["content"][0]["type"] == "image_url"
+        assert "media_io_kwargs" not in kwargs.get("extra_body", {})
+
+    def test_system_prompt_is_forwarded(self, client):
+        client.analyze_image_url("http://host/f.png", "p", system_prompt="You are an analyst.")
+        messages = client.client.chat.completions.create.call_args.kwargs["messages"]
+        assert messages[0]["role"] == "system"
+
+
+class TestSyncAnalyzeVideoUrl:
```
dineshvenkat, line 462: The analyze tests here (and in `test_async_vlm_client.py`) verify payload construction and delegation against a mocked response object, then stop. No test exercises a malformed or unexpected VLM response — empty choices, missing content, missing verdict, wrong field type, or a truncated body — so the response-parsing failure path (a class of failure this service has hit) is not covered.

dineshvenkat (approved):

services/alert/test/unit/vlm/test_vlm_client_payload.py
danh-nguyen-nvidia: Fixed

services/alert/test/unit/utils/test_event_utils.py
danh-nguyen-nvidia: Fixed

services/alert/test/unit/handlers/test_event_loop_pipeline_mixin.py
danh-nguyen-nvidia: Fixed

services/alert/test/unit/handlers/test_event_loop_pipeline_mixin.py
danh-nguyen-nvidia: Fixed

danh-nguyen-nvidia:
/ok to test [022ea9f](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1434/commits/022ea9f76869a2faa3c53c5331edf13558145854)

PR #1442, opened 2026-07-28
feat(vss-cli): add summarize vss cli command

agmohare:
## What this adds

`vss summarize` summarizes a recorded video through LVS and persists the result
to unified memory. `run` posts to `/v1/summarize` at the endpoint `vss configure`
discovered, writes a record when it submits, and closes that record out when the
summary arrives. `status`, `get` and `list` answer from those records, so a
summary is recallable by job id or sensor id long after the process exits. Event
times are anchored to `--creation-time`, so memory holds absolute instants rather
than offsets into the clip.

Each `run` ends with a one-line completion marker on stdout — `job_id`, plus
either `summary` and `persist` or `status`, `record` and `error` — so a caller
learns the outcome without scraping logs. Exit codes separate the cases worth
handling differently: 0 success, 2 invalid input, 3 backend unreachable, 4 no
deployment, 5 handle not found, 6 summarized but not persisted, 7 timeout.

Deployment support:

- HAProxy publishes the `/lvs` route the CLI resolves through, and admits
  Elasticsearch document writes to the `vss-memory` index family only. Reads and
  search are unchanged, and no other index becomes writable.
- Helm mirrors both routes behind `cliRoutes`, off by default, and the
  `dev-profile-lvs` ingress allows 3600s so long summarizations run to
  completion instead of timing out at the edge.

## Testing

`ruff`, `mypy` and 774 hermetic tests at 89.6% coverage.

End to end on the LVS developer profile (RTX PRO 6000, Cosmos-Reason2-8B):

- A 210s VIOS clip summarized in 59s: exit 0, 7 chunks, 3 events, and
  `persist: {"status":"complete","index":"vss-memory","events":3}`. The record
  reads back through `status`, `get` and `list --sensor-id`, filed under the
  sensor id, with event times stored as `2025-01-01T00:00:00Z`-style instants.
- The edge admits the memory store's reads and writes, including a
  `vss-memory-*` override, and refuses writes to other indices along with
  deletes, `_settings`, `_update_by_query` and `_cluster/health`.
- Every failure exit exercised against the live stack: 2, 3, 4, 5, 6 with the
  summary intact when a write is refused, and 7 whose handle `status` can still
  read. The exit 3 and exit 7 jobs close as `failed` and `timeout` rather than
  staying `submitted`.


agmohare:
/ok to test [da6e700](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1442/commits/da6e7008e657768dc9cb3672ceecc70d5b10f904)

agmohare:
/ok to test [ff51d21](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1442/commits/ff51d212dcbd5c8a7f8b21909bf69594f320559a)

agmohare:
/ok to test [46e6b08](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1442/commits/46e6b08e6d69e2649ffc0c3fb0cf477d523f8169)

services/agent/packages/vss_cli/src/vss_cli/summarize.py
```diff
@@ -0,0 +1,668 @@
...
+
+
+def _post_summarize(args: argparse.Namespace, request: dict[str, Any]) -> dict[str, Any]:
+    url = args.backend_url.rstrip("/") + _SUMMARIZE_PATH
+    try:
+        response = httpx.post(url, json=request, timeout=float(args.request_timeout_seconds))
+    except httpx.TimeoutException as error:
+        raise SummarizeError(
+            f"summarization timed out after {args.request_timeout_seconds}s (resume via the job_id)",
+            EXIT_TIMEOUT,
```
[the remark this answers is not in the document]
zac-wang-nv: Still live at `1d45301aa`, re-anchored `:515 → :547`. The retry added this round changes how reliably the record gets written, not what it says.

1. `close("timeout", ...)` (summarize_group.py:547) writes `status: "timeout"`, which `TERMINAL_STATUSES` in `vss_core/memory/models.py:29` classes as **terminal** (`{"completed", "failed", "partial", "timeout"}` — re-read at this head). If LVS finishes server-side after the client-side timeout, memory permanently records a job that actually succeeded as timed out.
2. The record's `backend_ref` is `None`, and `MemoryService._maybe_reconcile` returns early on a falsy `backend_ref`. There is nothing to reconcile against, by construction. Both files are unchanged in the new base — re-checked, not assumed.
3. `exits.py:34` says *"The marker carries a resumable job_id; resume via status/get"*. The marker exists — I see it on stdout in the harness above — but `Result.job_id` still has **zero consumers** repo-wide, and "resume" resolves to reading back a terminal `timeout` record, not to recovering the run.

Net: exit 7 tells the caller *that* it timed out and lets them look that fact up; it cannot give back the hour of VLM time. The module docstring at summarize_group.py:31-34 still claims the record is "what makes the exit-7 'resume by job id' contract true rather than aspirational" — that sentence is the thing to fix, or the contract is. Not blocking.

services/agent/packages/vss_cli/src/vss_cli/summarize_group.py
```diff
@@ -0,0 +1,567 @@
...
+    it summarizes with, which is the one that has to appear in the payload.
+    RT-VLM is the fallback for a deployment recorded before ``vss configure``
+    probed ``lvs``, where the two are usually the same model anyway.
+    """
+    for name in ("lvs", "rt_vlm"):
+        service = deployment.services.get(name)
+        if service and service.models:
+            return service.models[0]
+    raise config_mod.ConfigError(
+        f"deployment at {deployment.base_url} reports no LVS or RT-VLM model, so --model cannot be defaulted. "
```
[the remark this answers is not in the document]
zac-wang-nv: Addressed at `46e6b08e6` and strengthened again this round — resolving, with one honest residual.

`_mark_terminal` (summarize_group.py:368) returns `bool` instead of swallowing, and `close()` (:495-520) turns that into `absent` / `closed` / `stale`. The `stale` case prints to stderr (:515) *and* travels in the emitted body, so an operator sees it and a harness can branch on it. This round adds a bounded retry (:400-411) before giving up, with a rationale I agree with: the terminal write follows a `submitted` write that already succeeded, so the store was reachable moments ago and a transient blip is the likely failure.

**The residual, stated plainly because Greptile's summary still names it:** a write that fails all three times *does* leave the earlier `submitted` record unchanged. That much of the original claim is still true. What changed is that it is no longer silent — the caller gets `record: "stale"` in the marker and a sentence on stderr — and it is now much less likely. I'm calling that addressed rather than open, but if you'd rather keep it open until the record is reconciled for real, that's a defensible read and Thread 1 is where that work would live.

**One unchanged nit:** the comment at summarize_group.py:480-482 still reads *"From here on every exit path closes the record out, so `status` never reports a job as perpetually submitted."* That is contradicted 35 lines below by the code's own stderr string (*"so `status` still reports it submitted"*, :517) and by `stale` existing. The behaviour is right; the comment asserts the old, stronger claim.

services/agent/packages/vss_cli/src/vss_cli/summarize_group.py
```diff
@@ -0,0 +1,594 @@
...
+                err=True,
+            )
+            return Result(
+                body={"job_id": job_id, "status": "timeout", "record": record},
+                exit=Exit.TIMEOUT,
+                job_id=job_id,
+            )
+        except httpx.HTTPError as error:
+            close("failed", str(error))
+            raise BackendUnreachableError("lvs", f"unreachable at {url}: {error}") from error
```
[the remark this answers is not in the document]
zac-wang-nv: Fixed at `d28e93646` — resolving. I verified this rather than reading the commit message, and the result is in the table above: the three failure modes that emitted **0 bytes** of stdout last round now each emit a marker, while the control (the timeout branch, which always emitted one) is unchanged at 92 bytes. Exit codes are identical to before the fix — 3 / 2 / 3 — so nothing downstream that branches on them moves.

The shape is the one the rest of the module already used: `failed()` (:522) closes the record, prints the diagnosis to stderr, and returns a `Result` carrying `{"job_id", "status": "failed", "record", "error"}`. The inversion I flagged last round is gone — the well-behaved path is no longer the less discoverable one.

Two follow-ons from this change are filed as separate comments rather than buried here: the new test that guards it does not actually fail when the retry is removed (finding 2), and the 4xx path now hand-copies the framework's error prefix (finding 3).

services/agent/packages/vss_cli/src/vss_cli/group.py
```diff
@@ -131,6 +134,7 @@ def _exit_for(exc: Exception) -> Exit | None:
     by_name = {
         "InvalidInputError": Exit.INVALID_INPUT,
         "IndexNotFoundError": Exit.NOT_FOUND,
+        "MemoryNotFoundError": Exit.NOT_FOUND,
```
zac-wang-nv, line 137: **P1** — This table needs one more row, or `list` crashes on every deployment that has not persisted yet. Carried from round 3 unchanged — `group.py` is byte-identical and `vss_core/memory/` is untouched by the new base, so this is neither fixed nor a rebase artefact.

`elasticsearch.ApiError` is **not** a subclass of `TransportError` in es-py 8.17 (`NotFoundError.__mro__` is `NotFoundError → ApiError → ApiError → Exception`). So a `search` against an index that does not exist escapes `ElasticsearchMemoryStore._search`'s `except (ESConnectionError, ESTransportError)` (`vss_core/memory/backends/elasticsearch.py:114`), reaches `_exit_for` unmapped, and `_guarded` re-raises it. `main()` only catches `ConfigError`/`Abort`/`ClickException`, so it lands as a traceback and exit 1.

Re-reproduced at `1d45301aa` against a stub ES answering `404 index_not_found_exception` on `_search`:

```
$ vss summarize list
EXIT=1
elasticsearch.NotFoundError: NotFoundError(404, 'index_not_found_exception', 'no such index [vss-memory]')

$ vss search list                       # inherits the same CommandGroup.list
EXIT=1                                  # same traceback

$ vss summarize status --job-id summarize-X
EXIT=5   vss: job_id not found: summarize-X          # correct
$ vss summarize get --job-id summarize-X
EXIT=5   vss: job_id not found: summarize-X          # correct
```

`status`/`get` are fine because `ElasticsearchMemoryStore.get` catches `ESNotFoundError` explicitly (elasticsearch.py:78). Only the `_search` path is uncovered, which is `list`.

This is also a regression in failure mode: at the merge-base `CommandGroup.list` raised `MemoryUnavailable("list")` with `exit_code = Exit.CONFIGURATION`, so the same situation exited 4 with a sentence. And it is precisely the case `_guarded`'s own docstring says it exists to prevent — *"Without this a missing index — the ordinary 'nothing ingested yet' case — exits 1 with an Elasticsearch traceback, which no harness can branch on."*

The write path already got this right: `memory.write_failures()` (memory.py:135-153) deliberately appends `ApiError` for exactly this reason. The read path has no equivalent.

Two ways out, and I'd suggest the second:
- add `"ApiError": Exit.BACKEND_UNREACHABLE` here (MRO-walking covers every 4xx/5xx subclass), or
- catch `ESNotFoundError` in `_search` and return `[]` — a `list` over a store with nothing in it is an empty list, not an error.

No test covers this: `test_read_verbs_fail_honestly_without_a_memory_index` (test_job_grammar.py:127) points `VSS_CONFIG_HOME` at an absent directory, so it exercises "no **deployment**", not "no **index**" — the name promises the case that is actually untested. Given how carefully the new failure paths are tested this round, this one is conspicuous.

services/agent/packages/vss_cli/tests/unit_test/cli/test_summarize.py
```diff
@@ -0,0 +1,975 @@
...
+    configured: config_mod.Deployment, monkeypatch: pytest.MonkeyPatch
+) -> None:
+    """A blink between the two writes must not strand the job at `submitted`.
+
+    The submitted write already proved the store reachable, so a terminal write
+    that fails once says little about the next attempt -- and the difference is
+    a record `status` would report as running forever.
+    """
+    _capture_post(monkeypatch, httpx.TimeoutException("slow"))
+    store = _FlakyStore(fail_on="timeout", times=summarize_group._TERMINAL_WRITE_ATTEMPTS - 1)
```
zac-wang-nv: **P2** — The retry is a good addition and the reasoning in the constant's comment is right. But this guard cannot fail if the retry is removed, so it does not protect the behaviour it is named for.

Both the setup and the assertion are derived from the constant under test — `times=_TERMINAL_WRITE_ATTEMPTS - 1` above, and `assert store.attempts == _TERMINAL_WRITE_ATTEMPTS` below. They move together, so the test is a tautology at any value of the constant.

I ran it rather than arguing it. Setting `_TERMINAL_WRITE_ATTEMPTS = 1` at this head — which disables retrying completely, deleting the feature this commit adds:

```
$ pytest -k loses_a_race
1 passed

$ pytest packages/vss_cli/tests/unit_test -q
116 passed
```

At `= 1`, `times` becomes `0`, so `_FlakyStore` never raises, one attempt is made, and `attempts == 1 == _TERMINAL_WRITE_ATTEMPTS` holds. Green suite, no retry.

Pinning the numbers fixes it:

```python
store = _FlakyStore(fail_on="timeout", times=2)
...
assert store.attempts == 3
```

or, if you'd rather not hardcode the tuning, keep the derived setup but assert the property instead: `assert store.attempts > 1, "a single failure must not be final"`. That one still fails at `= 1`.

Worth contrasting with the sibling test added in the same round: `test_a_backend_failure_puts_its_handle_on_stdout` is exactly right — I ran the whole new file against round 3's `summarize_group.py` and all five of its params fail there, along with this one. This is a narrow fix to one test, not a comment on the suite.

services/agent/packages/vss_cli/src/vss_cli/summarize_group.py
```diff
@@ -0,0 +1,622 @@
...
+        except httpx.HTTPError as error:
+            return failed(str(error), f"vss: lvs unreachable at {url}: {error}", Exit.BACKEND_UNREACHABLE)
+
+        if response.status_code >= 400:
+            detail = f"HTTP {response.status_code}"
+            if response.status_code >= 500:
+                return failed(detail, f"vss: lvs backend error {detail}", Exit.BACKEND_UNREACHABLE)
+            return failed(
+                detail,
+                f"[vss] invalid input: summarization rejected {detail}: {response.text[:500]}",
```
zac-wang-nv: **P2** — Small consequence of an otherwise good change. Now that this path returns instead of raising `InvalidInput`, it no longer gets the prefix from the framework — it hand-writes it, and nothing pins the two together.

`group.py:113-123` owns that string and documents it as a contract:

```python
class InvalidInput(click.ClickException):
    """A payload the action's input model rejected (exit 2).

    Carries the ``[vss] invalid input:`` prefix so a harness can distinguish a
    malformed call from a backend failure without parsing pydantic's output.
    """
    def format_message(self) -> str:
        return f"[vss] invalid input: {self.message}"
```

It is also the only line in this function that uses that prefix — its three neighbours emit `vss: lvs ...` — which is what makes it easy to miss. And no test anywhere in `packages/vss_cli/tests/` matches the string `invalid input`, so if `format_message` is ever reworded, every `raise`-based path follows it and this one silently does not. A harness branching on the prefix then sees two formats for the same class of error, with a green suite.

Cheapest fix that keeps the return-don't-raise shape: build the message from the exception class that owns it, e.g. `InvalidInput(f"summarization rejected {detail}: ...").format_message()`, so there is still exactly one definition of the prefix.

deploy/helm/developer-profiles/dev-profile-lvs/templates/vss-ingress.yaml
```diff
@@ -7,6 +7,24 @@
...
+
+  Off by default, as in dev-profile-search: enabling it publishes Elasticsearch
+  on the public origin with no per-path guard, because an Ingress annotation
+  cannot express one.
+
+  /lvs is bounded by the ingress-wide timeouts below, not by the CLI's own
+  --request-timeout-seconds, so a summarization that outlasts them arrives as a
+  504 rather than the CLI's timeout. The Docker edge scopes its longer timeout
+  to the LVS backend alone; the annotation here applies to every route, so it
+  stays as it is. Raise timeoutServer to lift the ceiling for all of them.
```
zac-wang-nv: **P2** — This is the right place for these routes and the comment is honest about the tradeoff — thank you for writing the ES exposure consequence down rather than leaving it implicit. One thing to reconsider before merge, and it is specifically the timeout paragraph.

**Correcting my own round-3 evidence first.** Last round I said a summarization that outlasts this ingress ends as "exit 3 with an empty stdout". That is no longer true — `d28e93646` fixed exactly that, and the caller now always gets a handle. The substrate divergence survives; the missing-handle half of my argument does not, and I'd rather retract it explicitly than quietly restate a weaker version.

You know the ceiling is ingress-wide; the helm-sync bot flagged the same thing and also left it (*"arguably drift, but I did not touch it since it predates this PR"*). That reasoning was correct **while nothing routed `/lvs` through this ingress**. This PR is what makes it operative: as of this head, 600s is the real cap on the feature the profile is named for.

The numbers do not line up in three directions at once:

| | value | scope |
|---|---|---|
| CLI `--request-timeout-seconds` default | **3600** | summarize_group.py:223 |
| docker `bk_lvs_strip` | **3600s** | haproxy.cfg.template:111, per-backend |
| helm `dev-profile-lvs` `timeout-server` | **600s** | vss-ingress.yaml:44, ingress-wide |
| helm `dev-profile-search` `timeout-server` | **3600s** | ingress.yaml:110, hardcoded |

The sibling profile — the one whose `cliRoutes` block this is modelled on — already sits at 3600s. `dev-profile-lvs` is the outlier, and it is the profile that runs long video summarization.

What it costs, run at this head with a stub returning the **exact** ingress failure (504, where last round I only ran 500):

```
EXIT=3
{"job_id":"summarize-01KZYP83…","status":"failed","record":"closed","error":"HTTP 504"}
stored document: status='failed'
```

So the same command, on the same video, past ten minutes: on helm it exits **3** and writes a **terminal `failed`** record for a job LVS is very likely still running; on docker it runs to completion, or past an hour exits **7** with `timeout`. `failed` is in `TERMINAL_STATUSES`, so `status` will report that job as failed forever. The handle is no longer missing — it now points at a confidently wrong answer, which is why I still think this is worth a decision.

Three options, in the order I'd pick them:
1. Scope the longer timeout to the LVS backend, the way docker does. Per-Service `haproxy.org/*` annotations are already used in-tree (`rtvi-cv/templates/service.yaml:26-29`), so if `timeout-server` is Service-scopable in this controller version that answers the "applies to every route" objection directly — worth a check rather than taking my word for it.
2. Default `timeoutServer` to `3600s` when `cliRoutes.enabled`, matching `dev-profile-search`.
3. If it stays at 600s, say so in the flag help — a caller passing `--request-timeout-seconds 3600` on helm is being told they have an hour when they have ten minutes.

Not blocking, since `cliRoutes` is off by default and this is one values override — but it should be a decision, not an inheritance.

services/agent/packages/vss_cli/src/vss_cli/summarize_group.py
```diff
@@ -0,0 +1,622 @@
...
+    )
+    model: str | None = Field(
+        None,
+        description="VLM to summarize with. Defaults to the model the deployment's RT-VLM reports serving.",
+    )
+    prompt: str | None = Field(None, description="VLM prompt.")
+    system_prompt: str | None = Field(None, description="VLM system prompt.")
+    chunk_duration: int | None = Field(None, ge=1, description="Chunk duration in seconds.")
+    chunk_overlap_duration: int | None = Field(None, ge=0, description="Chunk overlap duration in seconds.")
+    temperature: float | None = Field(None, ge=0.0, le=2.0, description="Sampling temperature.")
```
zac-wang-nv: **P2** — Carried from round 3, re-anchored `:151 → :160`; `vss_api_models.py` is unchanged in the new base, so the comparison still holds.

The sampling/chunking bounds don't match the LVS model they pass through to, in both directions. The in-tree server model is `SummarizationQuery` in services/video-summarization/src/vss_api_models.py:

| flag | this model | LVS `SummarizationQuery` | effect |
|---|---|---|---|
| `--temperature` | `ge=0.0, le=2.0` | `ge=0, le=1` (:658-662) | `--temperature 1.5` validates here, 422s there |
| `--top-k` | `ge=0`, no upper | `ge=1, le=1000` (:678-682) | `--top-k 0` and `--top-k 5000` both 422 |
| `--seed` | unbounded | `ge=1, le=2**32-1` (:687-690) | `--seed 0` 422s |
| `--max-tokens` | `ge=1`, no upper | `le=1000000` (:639-643) | large values 422 |
| `--chunk-duration` | `ge=1` | `ge=0, le=3600` (:696-701) | CLI **refuses** 0, LVS's documented "no chunking" |
| `--num-frames-per-chunk` | `ge=1` | `ge=0, le=120` (:756-760) | same shape |

Confirmed by construction:

```
CLI ACCEPTS {'temperature': 1.5}      -> payload: {..., 'temperature': 1.5}
CLI ACCEPTS {'top_k': 0}              -> payload: {..., 'top_k': 0}
CLI ACCEPTS {'seed': 0}               -> payload: {..., 'seed': 0}
CLI REJECTS {'chunk_duration': 0}
CLI REJECTS {'num_frames_per_chunk': 0}
```

Low stakes — the rejection is fast and lands as exit 2 before any VLM time — but the class docstring says these are "passed through to the VLM untouched and are named as the REST API names them", and the PR description says the flag surface is "derived from the request model". The names match; the ranges are a second, hand-maintained copy that already differs. Copying LVS's bounds verbatim would make the CLI's own validation the authoritative check it reads as.

Separately: `--num-frames-per-chunk` is the field LVS marks deprecated — `_alias_deprecated_num_frames_per_chunk` forwards it to `num_frames_per_second_or_fixed_frames_chunk`, which the CLI doesn't expose. Fine for now, worth a comment so it isn't mistaken for the current API.

services/agent/packages/vss_cli/src/vss_cli/group.py
```diff
@@ -359,13 +367,13 @@ def _list_command(self) -> click.Command:
         )
 
         def callback(**values: Any) -> None:
-            ctx = _context_from(values)
+            ctx = _memory_context(values)
             selected = {k: values[k] for k in ("since", "sensor_id", "status") if values.get(k)}
-            _emit(owner.list(selected, ctx), ctx)
+            _emit(_guarded(lambda: owner.list(selected, ctx)), ctx)
```
zac-wang-nv, line 409: **P2** — Carried from round 3; `group.py` and `_foundation/time.py` are both unchanged, so this reproduces as written.

Now that `list` reaches a real store, `--since` reaches `JobFilters.__post_init__` → `coerce_utc_instant` → `iso8601_to_datetime`, and the help text at group.py:364 promises something that isn't implemented:

```
$ vss summarize list --since 1h
EXIT=1
  File ".../vss_core/_foundation/time.py", line 39, in iso8601_to_datetime
    dt = datetime.fromisoformat(utc_timestamp_to_tz_timestamp(timestamp))
ValueError: Invalid isoformat string: '1h'
```

Same unmapped-exception mechanism as finding 1 (`ValueError` isn't in `_exit_for`), but a different fix: either parse durations, or drop "or duration" from the help so the flag only advertises what it accepts. The option itself predates this PR — it just had nothing to reach until now.

services/agent/packages/vss_cli/src/vss_cli/summarize_group.py
```diff
@@ -0,0 +1,622 @@
...
+    return memory.service.upsert(record), content
+
+
+class SummarizeGroup(CommandGroup):
+    """Summarize video and persist to memory."""
+
+    name: ClassVar[str] = "summarize"
+    summary: ClassVar[str] = "Summarize video and persist to memory"
+
+    Input: ClassVar[type[BaseModel] | None] = SummarizeInput
```
zac-wang-nv, line 480: **P2** — Carried from round 3, re-anchored `:429 → :444`.

A group that declares `Input` rather than `actions` can't declare `requires`, so it never gets the framework's uniform missing-backend diagnostic. `_run_command` (group.py:295) synthesizes `Action(name="run", summary=..., Input=self.Input)` and leaves `requires` at its `frozenset()` default, so `_require_services` returns immediately for `summarize` — the whole mechanism at group.py:158-181 is unreachable for any single-action group.

Mostly invisible, because `deployment.endpoint("lvs")` produces a good message on its own. The one case it shows: a deployment with neither `lvs` nor `rt_vlm` models fails first in `_default_model` with *"reports no LVS or RT-VLM model, so --model cannot be defaulted"*, which points at `--model` when the real problem is that LVS isn't deployed. Minor ordering wart too — with `--persist`, `self.memory(ctx)` opens the Elasticsearch connection before anything checks that `lvs` is routable at all.

Either let `_run_command` take a `requires` ClassVar for the no-actions case, or declare `requires={"lvs"}` via a one-entry `actions` tuple. Worth settling now — the next single-verb group inherits it.

services/agent/packages/vss_cli/src/vss_cli/summarize_group.py
```diff
@@ -0,0 +1,622 @@
...
+            ):
+                return "closed"
+            click.echo(
+                f"vss: could not record job {job_id} as {status} in unified memory, "
+                f"so `status` still reports it submitted",
+                err=True,
+            )
+            return "stale"
+
+        def failed(detail: str, diagnostic: str, code: Exit) -> Result:
```
zac-wang-nv, line 573: Flagging this positively, because it is the right fix and it is easy to under-notice next to the retry work in the same commit range.

The four backend-failure paths used to `raise` after `close()` had already written a terminal record, so the process spent the write and then exited without telling the caller the id it wrote under. This helper makes them return the same shape the timeout branch returns, and the docstring's reason for it — "a harness reads stdout, so raising would leave the handle for the record this just wrote nowhere it can be found" — is the correct framing.

I verified it instead of taking the diff's word for it. Same harness as last round, with the timeout branch as a control: `http500`/`http422`/`conn` went from **0 bytes** of stdout to a marker each, while the control stayed at 92 bytes. Exit codes 3 / 2 / 3 are unchanged, so nothing branching on them moves. Running the new test file against the previous head's module fails all five params of `test_a_backend_failure_puts_its_handle_on_stdout` — the guard is real.

Also worth noting the local import of `BackendUnreachableError` was removed along with the raises rather than left behind, and `time` was already imported at module scope for the ULID clock, so the retry's `time.sleep` didn't need a new import. No dead references left in the file.

deploy/docker/services/infra/haproxy/haproxy.cfg.template
```diff
@@ -177,7 +185,16 @@ frontend fe_http
...
+
+    # Unified memory persists a job as PUT /<index>/_doc/<job_id>, which the
+    # read-only method rule answers 405. Scoped to vss-* so the mdx-* search
+    # indices stay read-only. Two denies rather than `http-request allow`:
+    # allow also skips the backend's replace-path rules, so the request would
+    # reach Elasticsearch with the /elasticsearch prefix still on it.
+    acl es_memory_write method PUT
+    acl es_memory_doc_path path_reg ^/elasticsearch/vss-[^/]+/_doc/[^/]+
+    http-request deny status 405 if h_main p_elasticsearch !es_read_method !es_memory_doc_path
+    http-request deny status 405 if h_main p_elasticsearch !es_read_method es_memory_doc_path !es_memory_write
```
zac-wang-nv, line 203: Carried from round 3 — this file is byte-identical at this head, so the verification below stands unchanged.

This is the subtlest thing in the diff and it is exactly right — flagging that positively because a reviewer's instinct on "loosen a read-only ACL" is to be suspicious.

I ran the memory store against a logging stub through the same URL shape the ingress sees, and it issues precisely three requests:

```
GET  /elasticsearch/vss-memory/_doc/summarize-01ABC
PUT  /elasticsearch/vss-memory/_doc/summarize-01ABC?refresh=wait_for
POST /elasticsearch/vss-memory/_search
```

All three pass: `GET`/`POST` are already in `es_read_method`, and the `PUT` matches both `es_memory_doc_path` and `es_memory_write`, so neither deny fires. `DELETE` on the same path is still denied by the second rule, `PUT` to `mdx-*` by the first, and `es_admin_path`/`es_mutating_op` still fire because they are evaluated after. The two-denies-instead-of-`allow` reasoning is also correct — `http-request allow` would end evaluation before `bk_elasticsearch_strip`'s `replace-path`, and the request would arrive at Elasticsearch with the prefix still attached.

services/agent/packages/vss_cli/src/vss_cli/summarize_group.py
```diff
@@ -0,0 +1,622 @@
...
+        created_at = utc_now_iso()
+        input_data = _memory_input(inputs, options, request)
+        persist_error: str | None = None
+        if memory is not None:
+            # Write the job before doing the work. From here on every exit path
+            # closes the record out, so `status` never reports a job as
+            # perpetually submitted.
+            try:
+                memory.service.upsert(
+                    SummaryAdapter().submitted_record(job_id=job_id, created_at=created_at, input_data=input_data)
```
zac-wang-nv: **P2** — `vss_core.memory` already ships an adapter abstraction, and this PR — its first production consumer — reaches past it. I don't think the direct import is wrong, but the consequence should be a decision rather than a default.

What exists in `vss_core/memory/adapters.py`: a `MemoryAdapter` Protocol (`:40`), `register_adapter` (`:78`), `get_adapter` (`:85`), and both `SummaryAdapter` (`:202`) and `SearchAdapter` (`:259`) already decorated into the registry. What this PR does: imports the concrete class at ten sites (`:333, 342, 352, 360, 387, 393, 425, 428, 472, 485`) and never calls `get_adapter`.

I censused the repo rather than assume, from five independent angles:

- `get_adapter` has **exactly one caller repo-wide**, and it is a test — `test_memory.py:466`, inside `test_register_fake_future_adapter_without_common_code_changes`.
- **Nothing outside `vss_cli` reads or writes these records.** Zero hits for `nv.vss.memory` or `vss-memory` in any non-Python file — no UI, helm, notebook, eval or dashboard. Across the whole repo exactly **10 production files** touch the tier: three in `vss_cli` (`memory.py`, `group.py`, `summarize_group.py`) and seven memory-package internals. No other service, tool or harness.
- **No code path needs group→adapter dispatch.** `BackendReconciler.reconcile(record)` (`service.py:31`) takes a whole record and is group-agnostic; every `job.group` reference in the package is a query filter (`backends/elasticsearch.py:141`, `in_memory.py:30,54`) or validation (`models.py:60`), never adapter selection.

So the registry answers a question nothing asks: *"given a group string I didn't know at compile time, which adapter maps it?"* Your direct import answers the question that is actually asked — *"which adapter does this group use?"* — and answers it with a concrete type, which matters because the Protocol declares only `submitted/running/terminal_record`; `build_input`/`build_output` exist solely on the subclasses and have different signatures per group. `get_adapter()` could not have typed `:342` or `:360`.

**Not asking you to change `vss_core` here** — it's untouched by this PR and came in with #1583, so deleting the registry is out of scope. The ask is just that this gets tracked: after this PR, `register_adapter` / `get_adapter` / `clear_adapter_registry` / `_ADAPTER_REGISTRY` are provably dead code with a test asserting a capability nothing uses. Either a follow-up removes them, or a comment records why they're kept.

The one cheap thing worth doing *in* this PR: make the binding declarative rather than repeating the import ten times — a single `Adapter` classmethod or ClassVar on `SummarizeGroup`, with the call sites going through it. That's a local change, keeps the concrete type, and sets up Comment 12 without committing to it.

zac-wang-nv:
Two process notes, neither about the code.

**Three heads have now gone by without a sweep.** `1d45301aa` carries 13 check runs and every one is a bot: 11×SonarQube, Greptile, one skipped. `Test (vss-cli)`, `DCO Sign-off`, `Container Coordinates Golden`, `Check helm parity` and `Trigger Downstream Pipeline` have not run. `/ok to test` was posted for `46e6b08` at 07:06:36Z and never re-issued for `3a056a722`, `d28e93646`, or this head. The green 54-check board belongs to `46e6b08`, which is three heads and a rebase behind. Worth a `/ok to test 1d45301` before this is read as passing — especially now, because `Check helm parity` has never once evaluated the helm templates this PR adds. (It was red on `46e6b08` for an unrelated reason: a `$10k` API-budget 429 with `cost=$0.00`, meaning it never looked at the diff at all.)

I checked DCO by hand in the meantime: both commits carry `Signed-off-by: Arun G Mohare <arg@nvidia.com>` matching the author, so that one should pass when it runs. The local suite is green at 116 tests.

**The PR description is now stale in a way that matters.** It describes persistence as a stdin-JSON subprocess bridge and says `status`/`get`/`list` exit with `MemoryUnavailable`; the head persists in-process via `vss_core.memory` and wires all three verbs to the real store. It also does not mention the helm ingress work at all — which is the most operationally consequential part of the last two rounds, and the part an operator most needs to know is opt-in.

zac-wang-nv:
Design observation for the *next* group rather than a change request for this one. `summarize` is the second `CommandGroup`, which is the first moment the shared shape is visible — and two instances already disagree.

A job group needs three things: the **wire contract** (`SummarizeInput`), the **job/transport options** (`SummarizeOptions`), and the **memory mapping** (`SummaryAdapter`). Only the first is declarative. The other two are hand-assembled inside `run()`, and both existing groups hand-rolled the same line independently:

```python
search_group.py:388     tuning  = {k: v for k, v in ctx.extra.items() if k in SearchTuning.model_fields}
summarize_group.py:455  options = SummarizeOptions(**{k: v for k, v in ctx.extra.items() if k in SummarizeOptions.model_fields})
```

They have already diverged: `summarize` constructs the model and gets pydantic validation; `search` keeps a raw dict, gets none, and then hand-patches a negated flag (`search_group.py:390-391`). Even the concept's name diverged — `SearchTuning` vs `SummarizeOptions`. When two plugins independently reimplement the same framework job inconsistently, that's usually the framework's cue to own it.

There's a second signal pointing the same way. `CommandGroup.run()`'s contract says:

> *"Do the work. **Persistence and markers are the framework's job.**"* — `group.py:242`

But `group.py` contains zero writes; all three `upsert` calls live in `summarize_group.py` (`:403, :435, :484`). The base class declares a responsibility it doesn't discharge, and the first group to need it implemented it by hand. That mismatch is upstream of everything above — and it's also why the adapter registry has no caller, since the layer that would have called it was never built.

The shape that would follow the existing conventions — six ClassVars are already the declaration mechanism (`name`, `summary`, `Input`, `actions`, `extra_params`, `primitives`):

```python
class SummarizeGroup(CommandGroup):
    Input   = SummarizeInput      # today
    Options = SummarizeOptions    # framework builds extra_params AND injects it typed
    Adapter = SummaryAdapter      # framework resolves it; no string key to mismatch

    def run(self, action, inputs, options, ctx) -> Result: ...
```

Three practical notes if this is ever picked up:

- **It's a base-class change, not a plugin-contract change.** The entry-point contract is the structural `CommandGroupSpec` Protocol (`plugins.py:60`) — `api_version`, `name`, `summary`, `cli()`. `_ConfigureGroup` satisfies it without subclassing `CommandGroup` at all, so it would be unaffected.
- **Now is the cheapest moment.** `API_VERSION = 1` is enforced at load with a clear diagnostic (`plugins.py:146-149`), there are three first-party entry points, and zero third-party plugins in tree.
- **Keep the adapter reference lazy.** Every `vss_core` import in this file is currently function-local or `TYPE_CHECKING`, which is deliberate. I measured a module-level `from vss_core.memory import SummaryAdapter` at **69 ms** (it does *not* pull in the Elasticsearch client), against a 0.24 s baseline for `vss summarize --help`. A classmethod returning the class keeps the declaration and the laziness.

None of this blocks the PR. Recording it because the seam is easiest to set before a third group inherits it.

services/agent/packages/vss_cli/src/vss_cli/summarize_group.py
zac-wang-nv: Correcting my own comment above. I said no code path needs group→adapter dispatch — that is wrong, and `MemoryService.events()` is the counter-example I missed.

It queries by `sensor_id` with **no group filter** (`vss_core/memory/service.py:139`), so a shared index hands it summary *and* search records in one list. It then hardcodes `for key in ("events", "incidents", "results")` (`:144`) — which is the union of two adapters' `ext` keys: `events`/`incidents` are set by `SummaryAdapter.build_output` (`adapters.py:243-249`), `results` by `SearchAdapter.build_output` (`adapters.py:336`). That is adapter-owned knowledge open-coded in the service.

So the registry is not answering a question nobody asks. It is answering one `events()` already asks, and currently answers by hand. `KNOWN_GROUPS` (`models.py:29`) already declares `alert` and `vlm`, so when either ships with its own `ext` key, `events()` will silently return nothing for those records — no error, just missing data.

That changes the follow-up I suggested: it should be *let the adapter own its ext keys and have `events()` dispatch on `record.job.group`*, not *delete the registry*. Everything else in the comment stands — the census still shows nothing outside `vss_cli` reads or writes these records, and the direct `SummaryAdapter` import is still the right call for this PR, since the Protocol cannot type `build_input`/`build_output`.

Still nothing for this PR to change: `service.py` is not in the diff.

agmohare:
/ok to test [6c31541](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1442/commits/6c31541f00c17ce34065b9945657e5b652caeda9)

kaushikc-nvidia (dismissed):

services/agent/packages/vss_cli/src/vss_cli/group.py
agmohare: Fixed, and in two places rather than one, because the row alone would have mapped the common case to the wrong code.

vss_core/memory/backends/elasticsearch.py — _search now catches NotFoundError and returns []. A missing index means nothing has been ingested yet, which is an empty result, not an unreachable backend, so this is exit 0 rather than exit 3.

group.py — the table gains "ApiError": Exit.BACKEND_UNREACHABLE, which now catches the genuine refusals (a 403 from a locked-down cluster) that the index fix does not cover.

Pinned by test_elasticsearch_list_before_anything_is_ingested_is_empty in vss_core, and test_a_refused_read_is_a_typed_exit_not_a_traceback parameterized over list and get in vss_cli. Live against the lvs profile on RTXPRO6000BW: vss summarize list --memory-index vss-memory-nope → exit 0, [], no traceback.

services/agent/packages/vss_cli/tests/unit_test/cli/test_summarize.py
agmohare: addressed in [6c31541](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1442/commits/6c31541f00c17ce34065b9945657e5b652caeda9))

services/agent/packages/vss_cli/src/vss_cli/summarize_group.py
agmohare: **Fixed**: the message is now built through InvalidInput(f"summarization rejected {detail}: {response.text[:500]}").format_message(), so the prefix comes from the same place every other rejection gets it and rewording it in one spot moves both.

deploy/helm/developer-profiles/dev-profile-lvs/templates/vss-ingress.yaml
agmohare: Addressed in [6c31541](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1442/commits/6c31541f00c17ce34065b9945657e5b652caeda9)) . Matching the  CLI's default: timeoutClient and timeoutServer now default to 3600s instead of 600s, matching dev-profile-search and the Docker edge, so a long summarization is not cut off at the edge while the CLI is still waiting.

services/agent/packages/vss_cli/src/vss_cli/summarize_group.py
agmohare: addresed in [6c31541](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1442/commits/6c31541f00c17ce34065b9945657e5b652caeda9)) . Fixed in both directions for chunk_duration, chunk_overlap_duration, temperature, top_p, top_k, max_tokens, seed and num_frames_per_chunk, so the CLI no longer rejects values LVS accepts nor forwards values it refuses.

services/agent/packages/vss_cli/src/vss_cli/group.py
agmohare: addresed in [6c31541](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1442/commits/6c31541f00c17ce34065b9945657e5b652caeda9). Fixed at the boundary: --since takes a click callback that parses the value and raises BadParameter, so a duration is a clean exit 2 instead of a ValueError traceback, and the help no longer advertises one — it now reads "Only jobs at or after this ISO-8601 instant."

services/agent/packages/vss_cli/src/vss_cli/summarize_group.py
agmohare: addresed in [6c31541](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1442/commits/6c31541f00c17ce34065b9945657e5b652caeda9).  choosing the first option you offered: CommandGroup gains a requires ClassVar and _run_command passes it into the synthesized action (Action(name="run", …, requires=self.requires)),

services/agent/packages/vss_cli/src/vss_cli/summarize_group.py
agmohare: Ack

agmohare:
/ok to test [7842b72](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1442/commits/7842b72a6f3f556ee0f39dd4d110c221fa5d9038)

agmohare:
/ok to test [5e0a62c](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1442/commits/5e0a62c54501ec5a43c1539a3158808e715ef7f6)

agmohare:
/ok to test [e0fa84b](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1442/commits/e0fa84baeb247f68237b746cde004fb30d7c9c21)

agmohare:
/ok to test [d8d50f0](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1442/commits/d8d50f01a377f884e57ace0255d4e50c0a657c43)

deploy/helm/developer-profiles/dev-profile-lvs/templates/vss-ingress.yaml
```diff
@@ -19,15 +45,23 @@ metadata:
...
-    # pipeline and can take several minutes. Raise the HAProxy timeouts well
-    # above the 50s controller default so the ingress does not return a
-    # 504 Gateway Time-out while the summarization backend is still working.
-    haproxy.org/timeout-client: {{ $vi.timeoutClient | default "600s" | quote }}
-    haproxy.org/timeout-server: {{ $vi.timeoutServer | default "600s" | quote }}
-  {{- if $phoenixStackEnabled }}
+    # pipeline, which on a long video outlasts ten minutes. Raise the HAProxy
+    # timeouts well above the 50s controller default so the ingress does not
+    # return a 504 Gateway Time-out while the backend is still working.
+    haproxy.org/timeout-client: {{ $vi.timeoutClient | default "3600s" | quote }}
```
zac-wang-nv, line 51: **P2** — The reasoning in the rewritten comment is right, and all three facts it asserts check out at this head: the CLI's `request_timeout_seconds` does default to 3600 (summarize_group.py:240), the Docker edge does set `timeout server 3600s` on the LVS backend (haproxy.cfg.template:111), and `dev-profile-search` does use `3600s` (ingress.yaml:110). The problem is that the new default never fires.

`values-lvs.yaml:106-107` sets both keys explicitly to `"600s"`, and `default` only substitutes for an empty value. That file is the documented install path — the profile README passes `-f dev-profile-lvs/values-lvs.yaml` in all five `helm upgrade` invocations (lines 294, 304, 318, 474, 482) — and it is also the file that sets `vssIngress.enabled: true`, which is `false` in the chart's own `values.yaml:447`. So the only supported way to render this Ingress at all is through the file that pins 600s.

Rendered at this head (`d8d50f01a`, i.e. **after** the develop merge), both ways:

```
$ helm template vss ./dev-profile-lvs -f dev-profile-lvs/values-lvs.yaml \
    --set global.externalHost=example.com --show-only templates/vss-ingress.yaml
    haproxy.org/timeout-client: "600s"
    haproxy.org/timeout-server: "600s"

$ helm template vss ./dev-profile-lvs --set vssIngress.enabled=true \
    --set global.externalHost=example.com --show-only templates/vss-ingress.yaml
    haproxy.org/timeout-client: "3600s"
    haproxy.org/timeout-server: "3600s"
```

So the round-4 failure mode is unchanged as shipped: a summarization past ten minutes arrives as a 504, exits 3, and closes the record out as terminal `failed` for a job LVS is still running. The fix is to move the same change into `values-lvs.yaml:106-107`; the template default then becomes belt-and-braces rather than the thing doing the work. (`values-lvs.yaml` is not in this PR's diff, which is why this sits on the template.)

One update from your develop merge: that merge pulled in `a0a3be3e1` (#1678), which shifted this block from `:90-91` to `:106-107` but left the `600s` values untouched. At the previous head I predicted merging develop would not resolve this; the render above is that prediction measured rather than projected.

services/agent/packages/vss_cli/src/vss_cli/summarize_group.py
```diff
@@ -0,0 +1,666 @@
...
+            body["record"] = close("partial", str(error))
+            return Result(body=body, exit=Exit.PARTIAL, job_id=job_id)
+
+        body["persist"] = {
+            "status": "complete",
+            "index": memory.index,
+            "group": memory_mod.group_token(self.name),
+            "events": len(content["events"]),
+        }
+        return Result(body=body, exit=Exit.SUCCESS, job_id=job_id)
```
zac-wang-nv, line 684: **P2** — A direct consequence of the staleness fix: the marker's `record` key is now on every **failure** path and on neither **success** path, so the field a caller would switch on is the one field whose presence depends on the outcome.

The key is set at :569 (`failed`) and :589 (`timeout`) as inline dict entries, and at :625 (`partial`, always `"absent"`) and :652 (`partial`, `"closed"` or `"stale"`). It is not set here, nor on the `--no-persist` return at :621. Measured rather than read — running both success paths through the CLI runner:

```
persisted success   exit=0  keys=['job_id', 'persist', 'summary']   record=<<ABSENT>>
--no-persist        exit=0  keys=['job_id', 'summary']              record=<<ABSENT>>
```

That is exactly the shape the control experiment surfaced from the other side: with the fix reverted, `test_a_store_that_refuses_the_first_write_still_summarizes` fails with `KeyError: 'record'` rather than a wrong value. A marker-driven caller — the consumer Greptile's P1 was written for — has to reach for `.get("record")` and treat missing as success, which is the kind of implicit contract that is easy to get wrong on the one path that matters.

Both values are already known at these two returns and neither needs a new call: it is `"closed"` here (the `_record` upsert at :461 succeeded, or we would be in the `except` below) and `"absent"` at :621 (`memory is None`). Setting them makes `record` total, which is what lets `status`-reconciling callers key off one field instead of two. Worth a line in `test_summarize.py` pinning the success shape too — right now nothing asserts what the key is on exit 0.

services/agent/packages/vss_cli/src/vss_cli/summarize_group.py
```diff
@@ -0,0 +1,666 @@
...
+    )
+    model: str | None = Field(
+        None,
+        description="VLM to summarize with. Defaults to the model the deployment's RT-VLM reports serving.",
+    )
+    prompt: str | None = Field(None, description="VLM prompt.")
+    system_prompt: str | None = Field(None, description="VLM system prompt.")
+    # Bounds are copied from LVS's own SummarizationQuery
+    # (services/video-summarization/src/vss_api_models.py), so a value this
+    # model accepts is one the backend accepts. A second, looser set here would
```
zac-wang-nv: **P2** — This comment states the right contract and the numeric fields honour it, but it does not hold for the two list fields, and the test that enforces it cannot see them.

LVS constrains the *elements* of both lists, not the lists:

- `events` — each element `max_length=1024` (vss_api_models.py:915)
- `objects_of_interest` — each element `max_length=256` (vss_api_models.py:949)

`SummarizeInput` puts `max_length=1000` on the list itself (summarize_group.py:141 and :146). In pydantic that bounds the number of *items*, not the length of each one. The item cap is correct and does mirror LVS's own `max_length=1000`; the per-element cap has no counterpart here. Measured by constructing the model directly rather than by reading it:

```
events element of 2000 chars                ACCEPTED    (LVS caps 1024)
objects_of_interest element of 300 chars    ACCEPTED    (LVS caps 256)
objects_of_interest element of 5000 chars   ACCEPTED
scenario of 2000 chars                      REJECTED    ✓ mirrors LVS
1000 events                                 ACCEPTED    ✓ mirrors LVS
1001 events                                 REJECTED    ✓ mirrors LVS
```

So `scenario` — a plain `str` with the same 1024 cap — does mirror it, and the list length does too. It is specifically the constraint *inside* a list that was not carried across. The consequence is exactly the one this comment names: a 300-character `--object-of-interest` spends the round trip and comes back 422 from LVS, where the stated intent is exit 2 from the CLI.

The guard cannot catch this by construction: `_lvs_query_bounds()` collects only `ge`/`le` keywords out of `SummarizationQuery`, and `test_request_bounds_match_the_service_they_are_sent_to` is parametrized over 8 numeric fields. Strings and lists are outside its reach — see comment 9, where I flagged that test as genuinely non-vacuous, which it is, within that scope.

One annotation each keeps the "copied from LVS" property literally true:

```python
events: list[Annotated[str, Field(max_length=1024)]] = Field(max_length=1000, ...)
objects_of_interest: list[Annotated[str, Field(max_length=256)]] = Field(default=[], max_length=1000, ...)
```

LVS also applies `pattern=ANY_CHAR_PATTERN` to these, but that is `^(.|\n)*$` — it matches anything, so there is nothing to mirror there and I would not add it. Worth teaching `_lvs_query_bounds` to collect `max_length` alongside `ge`/`le` at the same time: it would have caught this one and will catch the next.

agmohare:
/ok to test  [bc38678](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1442/commits/bc386782ba260a060dd77be0f539b40f52f517f6)

agmohare:
/ok to test [bc38678](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1442/commits/bc386782ba260a060dd77be0f539b40f52f517f6)

zac-wang-nv (approved):
LGTM

PR #1356, opened 2026-07-22
Github based CI for Behavior Analytics and Video Analytics API

kartikayt-nvidia:
## Summary

Adds GitHub CI and container-build support for:

- `services/analytics/video-analytics-api`
- `services/analytics/behavior-analytics`

Both services now follow the existing VSS GitHub GHCR image-build and immutable release-set flow.

## Changes

- Added unit and integration CI jobs for both analytics services.
- Unit tests run only when the corresponding service directory changes.
- Integration tests run on every applicable CI pipeline.
- Enabled multi-architecture `linux/amd64` and `linux/arm64` image builds through the existing `Build Dev Images (GHCR)` workflow.
- Added both images to `deploy/docker/container-inventory.json`.
- Added container-source validation workflows for both images.
- Added service-scoped image change detection:
  - Video analytics API changes build only `vss-video-analytics-api`.
  - Behavior analytics changes build only `vss-behavior-analytics`.
  - Existing agent/UI/alert shared-build behavior remains unchanged.
- Added immutable image tag, digest, source path, and source-tree SHA information to the release set.

### Video analytics API

- Added CI-compatible Docker-host endpoint detection.
- Added configurable Elasticsearch and video analytics API endpoints.
- Updated warehouse tests to compare only against:
  `src/app/specification/openapi.json`
- Improved Docker Compose timeout and cleanup handling.

### Behavior analytics

- Fixed Kafka advertised-host configuration for host-network tests.
- Fixed Elasticsearch access across local and CI environments.
- Added Docker Compose status and log diagnostics.
- Improved cleanup behavior and timeout handling.

## Image flow

1. GitHub detects changes in an analytics service.
2. GitHub builds and pushes the corresponding multi-architecture image to GHCR.
3. The immutable image tag and digest are recorded in the release set.
4. The downstream `ci-vss-oss` pipeline evaluates that exact digest.
5. A successfully tested digest can be copied to NGC without rebuilding.

## Checklist
- [ ] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/HEAD/CONTRIBUTING.md).
- [ ] I have installed and run [pre-commit hooks](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/HEAD/CONTRIBUTING.md#1-enable-pre-commit-hooks) locally (`uv run pre-commit install` once, then hooks run on every `git commit`).
[...]

PR #1446, opened 2026-07-28
refactor(ci): derive develop-<sha> tags at every push tip; clean up PR tags

zac-wang-nv:
## Summary

Publishes an immutable `develop-<sha12>` tag for **every** GHCR image at each `develop` push tip, so a consumer derives the whole coordinate set from a commit SHA alone instead of fetching the release-set manifest. Adds cleanup of a closed PR's `pr-<N>-*` candidate images.

Carries forward the architecture findings surfaced while reviewing #1356. Those were out of scope there — #1356 extends the existing design; this changes it.

## Why the manifest exists today

The release set is the price of build avoidance, and the sparseness that matters is **across images at one commit**, not across commits.

Tags are already a pure function of (ref, commit) — `container_build_plan.candidate_coordinates` yields `develop-<sha12>`. But with build avoidance, only the images that were actually rebuilt at commit X get a `develop-<sha12>` tag. The rest are `reuse-pinned` — "not rebuilt, still lives at the older tag". So deriving coordinates from a SHA 404s for exactly those images, and something has to record where they really live. That something is the manifest.

**Sparse across commits is fine and stays that way.** Only the pushed tip is tagged; interior commits of a rebase-merge get nothing, and nothing resolves against them. What this PR changes is that at the tip, *every* GHCR image carries that tip's `develop-<sha12>` — complete across the 15 images, at one commit. That is what makes SHA-derivation total, and it is the only property the manifest was supplying for coordinate lookup.

## Retag — `advance_ghcr_alias.py`

**Retag every GHCR entry, not just `strategy == "build"`.** A tag set that skipped `mirror` / `reuse-pinned` entries would be incomplete at the commit, so a SHA-derived coordinate would 404 for those images. Selection is now on *evidence* — a `ghcr.io/` repository plus a resolved manifest digest — rather than on strategy. Side effect: fixes `develop-latest` going stale for re-mirrored images, which the old `build`-only filter skipped.

Coverage tracks what is *actually* published rather than a hardcoded list. **3 images today** — `vss-agent`, `vss-agent-ui`, `vss-alert-ms`: the managed set that the shared `VSS_CONTAINER_TAG` governs, and exactly the entries with `ghcr_build: true`. It grows on its own as the 5 staged `strategy: build` entries (`vss-behavior-analytics`, `vss-rt-embed`, `vss-rt-vlm`, `vss-video-analytics-api`, `vss-video-summarization`) are onboarded — no list to update. The `mirror` and `external-pin` entries live at `nvcr.io`, cannot be retagged, and stay pinned in git at this commit.

**Tree-SHA gate (`--verify-tree-sha`).** Refuses to publish a tag when a recorded `source_tree_sha` does not equal `git rev-parse <commit>:<source_path>`. This makes the retag *absolute* — it asserts "this image was built from this commit's source" against git and the registry, never against the previous commit's tags — and reuses the invariant `check_container_tag_source` already trusts. The label lives in the **config blob**, not a manifest annotation, so it survives round-trip through nvcr.io and the artifacts-promotion copy.

Fails **closed**: an incomplete tag set is visible and recoverable; a wrong one is neither. Entries with no `source_tree_sha` (`mirror` entries carry `upstream_digest`; their content is not repo-derived) pass unverified rather than failing.

**Why only the pushed tip.** The repo allows squash and rebase merges only — no merge commits — so `develop` is linear and every consumer resolves from the tip, never from an interior commit. Tagging the tip per push is sufficient and is what the workflow already does for `develop-latest`.

## PR tag cleanup — `cleanup_pr_tags.py` + `cleanup-pr-tags.yml`

Candidate tags are build scaffolding: once the PR closes they can never be rebuilt or referenced, so retaining them only grows the package. `develop-*` tags are the opposite and are kept forever.

**The safety rule that matters.** A GHCR *version* is a manifest digest and can carry several tags. Content-addressed reuse makes that routine — a PR whose tree matches develop's resolves to the same digest, so one version can be tagged `pr-1234-abc123abc123` *and* `develop-def456def456` *and* `tree-<sha>` at once. Deleting a version deletes every tag on it, so a version is deleted only when **every** tag on it belongs to the PR being cleaned. One foreign tag disqualifies the whole version; untagged versions are left alone.

Deleting a live develop image is unrecoverable without a rebuild; a retained tag costs storage. The conservative direction is the correct one, and cleanup is best-effort — the script swallows per-package HTTP errors and the job is `continue-on-error`.

Uses `pull_request`, not `pull_request_target`: this repo has never fired a single `pull_request_target` run, while `pull_request` fires normally (PRs are branched in-repo, so the token carries `packages: write`). `workflow_dispatch` is included for backfill, defaulting to dry-run.

## Also

Corrects `prepare_nightly_promotion`'s error message, which claimed to have verified a *"GitHub CI/downstream run"* while only asserting a successful `ci.yml` workflow run. `trigger-downstream-pipeline` is gated on `has-ghcr-build-entries` and is **skipped on develop** whenever every entry is `reuse-pinned` (the normal case — the PR already built those trees). Skipped jobs do not fail a workflow, so a green `ci.yml` can coexist with no downstream run at all. Verified skipped 3/3 on recent develop runs.

## Deliberately not in this PR

**Retiring `release-set.json` itself.** `VSS_RELEASE_SET_B64` is consumed by **8 files in `metromind/ci-vss-oss`** (`.gitlab/ci/00-foundations.yml`, `05-preflight.yml`, `ci/setup-eval-overrides.sh`, `ci/tools/vss_decision.py`, `promote_ghcr_release_set.py`, + tests). Removing the producer without a paired MR there breaks the GitLab release plane. This PR makes the manifest *redundant for coordinate lookup*; retiring it needs the two changes landed together.

Related follow-ups, each independently reviewable:
- Replace `wait-for-build-dev-images` (520 × 15s poll, `timeout-minutes: 140`) with a `workflow_run: completed` trigger — it exists only because Actions has no cross-workflow `needs:`.
- `nightly-promote-ghcr.yml` and `last-green-controller.yml` are on `develop` but not on `main`, which is the default branch and **391 commits behind**. `on: schedule` fires only from the default branch, so neither has ever run. Worth checking whether that lag is intentional — it silently disables every scheduled workflow in the repo.

## Test

26 new unit tests (14 retag incl. the tree-SHA gate, 12 cleanup incl. the shared-digest cases). All 15 existing guard suites pass. `ruff check` clean; workflow YAML validated.

### DIFF
--- a/.github/scripts/advance_ghcr_alias.py
+++ b/.github/scripts/advance_ghcr_alias.py
@@ -1,7 +1,41 @@
 #!/usr/bin/env python3
 # SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
-"""Advance a GHCR moving alias to each built digest in a release set."""
+"""Retag every GHCR image in a release set under one or more aliases.
+
+Two aliases are published post-merge on ``develop``:
+
+* ``develop-latest`` — the moving developer convenience alias.
+* ``develop-<sha12>`` — an **immutable per-commit** tag. Publishing this for
+  *every* GHCR image at every develop push tip is what lets a consumer derive a
+  complete, correct coordinate set from the commit SHA alone, with no
+  release-set manifest to fetch. That is the point of this script.
+
+Two behaviours differ from the original build-only alias mover:
+
+1. **Every GHCR entry is retagged, not just ``strategy == "build"``.** A tag set
+   that skipped ``mirror`` / ``reuse-pinned`` entries would be incomplete at the
+   commit, so a SHA-derived coordinate would 404 for those images. Entries are
+   selected on evidence — a ``ghcr.io/`` repository plus a resolved manifest
+   digest — rather than on strategy, so coverage tracks what is *actually*
+   published rather than a hardcoded list. That is 3 images today — vss-agent,
+   vss-agent-ui, vss-alert-ms, the managed set the shared ``VSS_CONTAINER_TAG``
+   governs — and grows on its own as ``ghcr_build`` is enabled for the staged
+   build entries. Non-GHCR pins (nvcr.io) cannot be retagged and are skipped;
+   they are pinned in git at this commit and stay derivable from the repo.
+
+2. **Tree-SHA gate.** ``--verify-tree-sha`` refuses to retag an image whose
+   recorded ``source_tree_sha`` does not equal ``git rev-parse <commit>:<source_path>``.
+   This makes the retag *absolute*: it asserts "this image was built from this
+   commit's source" against git and the registry, never against the previous
+   commit's tags. Entries with no ``source_tree_sha`` (``mirror`` entries carry
+   an ``upstream_digest`` instead — their content does not come from repo
+   source) have nothing repo-side to compare and pass the gate unverified.
+
+The gate fails **closed**: a mismatch raises rather than publishing a tag that
+would misrepresent provenance. An incomplete tag set is visible and recoverable;
+a wrong one is neither.
+"""
 from __future__ import annotations
 
 import argparse
@@ -15,6 +49,7 @@ from typing import Callable
 
 DIGEST_RE = re.compile(r"^sha256:[0-9a-f]{64}$")
 ALIAS_RE = re.compile(r"^[A-Za-z0-9_][A-Za-z0-9_.-]{0,127}$")
+TREE_RE = re.compile(r"^[0-9a-f]{40}$")
 Runner = Callable[[list[str]], str]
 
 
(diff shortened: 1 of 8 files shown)


.github/scripts/advance_ghcr_alias.py
```diff
@@ -53,10 +99,66 @@ def alias_plan(release_set: dict, alias: str) -> list[AliasUpdate]:
...
+        if not recorded or not source_path:
+            report.append(f"{name}: no source_tree_sha recorded; retag unverified")
+            continue
+        expected = tree_reader(repo_root, commit, str(source_path))
+        if expected is None:
+            mismatches.append(
+                f"{name}: {source_path!r} does not resolve to a tree at {commit}"
+            )
+            continue
+        if expected != str(recorded):
```
simings-nv, line 148: `verify_tree_shas` raises before the alias loop, and the workflow passes `--alias develop-latest` and `--alias develop-<short>` in one call. So one image with a stale `source_tree_sha` blocks both aliases for every image, not just that one. or I mis-read.

Worth skipping the mismatched entry and advancing the rest?

simings-nv: nvm, a second thought that this is unlikely to happen, so not a problem.

hugoverjus (approved):

hugoverjus:
LGTM

PR #1448, opened 2026-07-29
fix(ui): rename .dockerignore so Docker actually reads it

simings-nv:
## Description

`services/ui/.dockerignore` is read by nothing today.

`vss-agent-ui` builds with `context: "."` (`deploy/docker/container-inventory.json`), and Docker resolves `.dockerignore` from the **build context root**, or as `<dockerfile>.dockerignore` beside the Dockerfile under BuildKit. Neither exists, so every pattern in that file is inert. This renames it to the second form.

### What changes, measured by real builds at `432293977`

The build context under `services/ui` goes from **484 files to 363**. Of the 121 that stop entering, **89** are `*.test.ts` / `*.test.tsx` / `*.test.js` and **24** are markdown.

**This is context transfer and builder-layer cache, not image size.** Only one `COPY` in `services/ui/Dockerfile` reads the context (L15, builder stage); every runner-stage `COPY` is `--from=builder` and takes only build outputs, so these files never reached the published image.

The practical effect: today a `*.test.ts` edit changes the build context, so the builder `COPY` layer invalidates and the npm build reruns. After the rename it does not.

### Two things worth knowing before approving

**The patterns now evaluate at the repository root**, because the context is `.`, not at `services/ui`. So files matching them elsewhere in the tree also stop entering the context. Conversely five patterns written assuming `services/ui` was the context root (`Dockerfile`, `.idea`, `.vscode`, `.env*`, `.github/`) become no-ops for `services/ui`.

**Why this filename and not a repo-root `.dockerignore`.** `vss-agent-ui`'s content tag is `git rev-parse HEAD:services/ui` (`container_build_plan.py`), which is only sound because the sole context-reading `COPY` is `COPY services/ui/`. A root `.dockerignore` would break that invariant: editing it would change image content without moving `HEAD:services/ui`, so the content tag would not move and reuse could re-tag a stale manifest. Keeping the file inside `services/ui` preserves it.

It also matches existing house style: the 4 analytics services and `libs/analytics/spatialai-data-utils` already use `docker/Dockerfile.dockerignore`.

### How the precedence was verified

By real builds on two machines and three buildx client versions, because `.dockerignore` file selection is resolved **client-side by buildx** rather than by BuildKit:

| case | `drop.txt` | expected |
|---|---|---|
| no `.dockerignore` anywhere (control) | PRESENT | PRESENT |
| `services/ui/.dockerignore` (not ctx root) | **PRESENT** | PRESENT — the bug |
| `services/ui/Dockerfile.dockerignore` | **excluded** | excluded — the fix |
| `./.dockerignore` at context root (control) | excluded | excluded |

Controls included deliberately: without them, a build that copied nothing would also report `excluded`.

### Cost to land

One `linux/amd64` + `linux/arm64` rebuild of `vss-agent-ui`. The rename moves `git rev-parse HEAD:services/ui` even though the blob is byte-identical, so the content-tag reuse gate misses once.

## Checklist

- [x] Every commit on this PR is DCO sign-off'd
- [x] Pure rename, no content change (`R100`, blob unchanged)
- [x] No references to the old path anywhere in the tree
- [x] Verified by real builds, not from documentation


zac-wang-nv (approved):
LGTM

hugoverjus (approved):

PR #1449, opened 2026-07-29
feat(skill-eval): compute runner labels from the spec in plan_matrix

zac-wang-nv:
## What

Every matrix leg now carries `runs_on` — the runner label set implied by the spec's own `resources.platforms.<PLATFORM>` block:

```json
{"include": [{"name": "…", "slug": "…",
              "runs_on": ["self-hosted", "vss-eval", "gpu-l40s", "gpus-1"]}]}
```

**Nothing consumes it yet.** `skills-eval.yml` still pins the coordinator pool and `run_leg.py` still does fleet selection + flock. This is emit-only on purpose, so the mapping can be reviewed and diffed against today's placement before any GPU box is registered as a runner in its own right.

## Why here

The spec already declares its hardware, and that declaration is readable at **plan** time — no dataset, no `brev ls`, no box. `spec_platforms()` was already parsing the exact object and discarding everything but the keys:

```python
platforms = data.get("resources", {}).get("platforms", {})
return sorted(platforms) if isinstance(platforms, dict) else []
```

So the spec stays the source of truth for hardware; only the *resolution point* moves — from leg-time runtime matching to plan-time label computation, with Actions doing the placement.

## Mapping, with `pool_candidates` as the parity target

| spec | label | rationale |
|---|---|---|
| `platforms.L40S` | `gpu-l40s` | keys mirror the `PLATFORMS` tables in `adapters/*/generate.py` |
| `platforms.ANY` | *(none)* | GPU-independent |
| `gpu_count: N` | `gpus-N` | a **demand**. A box advertises every count it can satisfy — a 2-GPU box carries `gpus-1` *and* `gpus-2` — which is how "over-provisioned boxes remain valid" (#929) survives a static label set |
| `gpu_count: 0` | *(neither)* | `pool_candidates` guards its type filter with `if required_count > 0 and required_type`, so a zero-GPU spec already accepts any RUNNING box. Emitting a GPU label here would be **more** restrictive than today |
| absent | `gpus-1` | matches `int(metadata.get("gpu_count", 1) or 0)`. 15 of the 50 platform entries rely on this default |

## Over the real tree

50 legs collapse to four label sets:

```
18  self-hosted vss-eval gpu-l40s gpus-1
14  self-hosted vss-eval gpu-rtxpro6000bw gpus-2
11  self-hosted vss-eval gpu-rtxpro6000bw gpus-1
 7  self-hosted vss-eval
```

Those 7 are the zero-GPU legs — the `ANY` specs plus `detection-tracking-3d/routing`. Under labels they stop competing for GPU boxes entirely, which the runtime selector cannot express. That's a throughput win available the moment anything consumes the field.

## Two findings from the survey

**No spec declares `gpu_type`** — 0 of 50. `_rtx4090_supports()` routes on "spec metadata declares `gpu_type`", so that path is dead code, consistent with `RTX4090_TESTS` and `RTX4090_ALL_TESTS` both being empty frozensets. Nothing to preserve when the label model lands.

**Unknown platform keys now warn.** A typo'd platform would otherwise emit a label no box carries, and the job would queue until GitHub cancels it at 24 h. It still emits a deterministic slug, but says so on stderr.

## Incidental

`spec_platforms` is refactored onto a new `spec_platform_config`, which also hardens the malformed-spec path: a non-dict document or `resources` block previously raised `AttributeError` straight past the `except (OSError, ValueError)`. There is such a file on disk — the legacy list-shaped `evals.json`. Tests stub the new helper.

Also fixes two pre-existing `ruff` findings in the touched files (unused `targets` local, unused `sys` import). Both confirmed present on `develop`.

## Tests

- `RunsOnLabels` — each mapping rule, plus both `pool_candidates` coercion quirks (absent → 1, present-but-falsy → 0).
- `RealSpecCorpus` — reads the real tree on purpose: every spec on disk yields lowercase-safe labels, base labels first, at most one `gpu-*` and one `gpus-*`, and **never a GPU-type label without a count** (the zero-GPU parity rule).
- `BuildMatrix` — every leg carries `runs_on`; `missing_adapter` legs get base labels only (they commit an adapter, run no trial).

33 passed, `ruff` clean.

Note: `test_plan_matrix.py` is not in `ci.yml`'s hand-listed pytest set, so these run locally only — same gap as `test_run_leg.py`.

## Sequencing

This is step 1 of moving placement to Actions. It is independently reviewable and mergeable: no runner is re-registered, no workflow changes, no behaviour changes. Step 2 would be registering one GPU box with these labels and flipping `runs-on: ${{ matrix.runs_on }}` for that one platform.

### DIFF
--- a/.github/skill-eval/plan_matrix.py
+++ b/.github/skill-eval/plan_matrix.py
@@ -24,6 +24,14 @@ A skill whose adapter is missing collapses to a single `missing_adapter`
 leg (that leg's agent commits the one adapter to the PR branch), so N specs
 of an adapterless skill don't race to commit it N times.
 
+Each leg also carries `runs_on`: the runner label set implied by the
+spec's own `resources.platforms.<PLATFORM>` block (see runs_on_labels).
+This resolves the spec -> hardware mapping at PLAN time, where today
+run_leg.py re-derives it at LEG time from `brev ls` under a flock.
+Nothing consumes `runs_on` yet — it is emitted so the mapping can be
+reviewed against current placement before the GPU boxes are registered
+as runners in their own right.
+
 Env:
     PR_BASE        base branch, e.g. develop (diffed as FETCH_HEAD...HEAD)
     MANUAL_SKILLS_FILTER  workflow_dispatch sweep: a skill-dir name or `*`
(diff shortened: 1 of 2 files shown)


.github/skill-eval/plan_matrix.py
```diff
@@ -67,6 +75,94 @@
...
+    `if required_count > 0 and required_type`, so a zero-GPU spec ignores
+    the declared platform and "accepts any RUNNING box". 7 of the 50
+    platform entries are zero-GPU today (the ANY specs, and
+    detection-tracking-3d/routing on RTXPRO6000BW) — under labels they
+    stop competing for GPU boxes at all.
+    """
+    labels = list(BASE_LABELS)
+    count = _gpu_count(config) if config is not None else DEFAULT_GPU_COUNT
+    if count <= 0:
+        return labels
```
simings-nv, line 158: I see returning BASE_LABELS doesn't make this CPU only. BASE_LABELS includes vss-eval and is documented above as the label carried by GPU boxes. Would this cause GPU runners got matched?

.github/skill-eval/plan_matrix.py
```diff
@@ -67,6 +75,94 @@
...
+DEFAULT_GPU_COUNT = 1
+
+
+def _platform_label(platform: str) -> str | None:
+    """GPU-type label for a `resources.platforms` key."""
+    if platform in PLATFORM_LABELS:
+        return PLATFORM_LABELS[platform]
+    # Unknown key: still emit something deterministic, but say so — a
+    # typo'd platform would otherwise produce a label no box carries and
+    # the job would queue until GitHub cancels it at 24 h.
```
simings-nv, line 118: In case to ask, we still allows to emit labels gpu-${slug} even unknown?
What if someone put a typo and leave the job queued untiled timeout. 
Why not just fail fast with actionable error?

For the warning, do we want to log them in case we want to know how often it is.

.github/skill-eval/plan_matrix.py
```diff
@@ -67,6 +75,94 @@
...
+
+def _gpu_count(config: dict) -> int:
+    """Declared GPU demand, matching run_leg's coercion exactly."""
+    raw = config.get("gpu_count", DEFAULT_GPU_COUNT)
+    try:
+        return max(0, int(raw))
+    except (TypeError, ValueError):
+        # Explicit null / "" / garbage -> 0, same as run_leg's `or 0`.
+        return 0
+
```
simings-nv, line 136: These two in-tree specs already omit gpu_count, [Here](skills/vss-deploy-detection-tracking-2d/evals/deploy-evals.json:5-12)
- L40S with only modes: ["standalone"].
- similar for RTXPRO6000BW with only modes: ["standalone"].
    
The spec and adapter may disagree about GPU demand. This spec omits gpu_count, so the planner correctly applies its documented default and emits gpus-1. However, the detection-tracking-2d adapter defaults the same platform to `gpu_count = 2`. Because runs_on is evaluated before the adapter runs, GitHub can assign this leg to a one-GPU runner. If the workload truly needs two GPUs, can we declare gpu_count: 2 explicitly in both detection-tracking-2d specs and add a test preventing the spec and generated `task.toml` requirements from diverging?

.github/skill-eval/plan_matrix.py
```diff
@@ -67,6 +75,94 @@
...
+# and contributes no `gpu-*` label. Keys mirror the PLATFORMS tables in
+# .github/skill-eval/adapters/*/generate.py.
+PLATFORM_LABELS: dict[str, str | None] = {
+    "H100": "gpu-h100",
+    "L40S": "gpu-l40s",
+    "RTXPRO6000BW": "gpu-rtxpro6000bw",
+    "DGX-SPARK": "gpu-dgx-spark",
+    "IGX-THOR": "gpu-igx-thor",
+    "ANY": None,
+}
```
simings-nv, line 102: Could PLATFORM_LABELS become a shared platform catalog consumed by planning, validation, and runner registration? It currently mirrors adapter-local PLATFORMS tables, so copying it into provisioning would create another drift point

simings-nv (approved):
LGTM.
My comments are non-blocking for this PR, but before matrix.runs_on is activated, it would be good to clarify the placement semantics, fail fast on unknown platforms, align GPU counts, and establish a reusable source for platform labels.

hugoverjus (approved):

hugoverjus:
LGTM

PR #1453, opened 2026-07-29
fix(warehouse): configure RT-CV Kafka broker

hugoverjus:
## Summary
- configure RT-CV workers to consume camera provisioning events from the deployed Kafka broker
- apply the endpoint consistently to warehouse 2D, 3D, and MV3DT profiles
- prevent SDRC from falling back to unreachable `localhost:9092`

## Context
Downstream warehouse evaluation showed RT-CV starting successfully but remaining at zero active sources. SDRC had Kafka enabled without `WDM_KFK_BOOTSTRAP_URL`, so the worker defaulted to `localhost:9092` while the configurator published camera events to `kafka:29092`.

This upstreams the temporary compatibility patch from `ci-vss-oss` MR !313.

## Test plan
- [x] validated all three warehouse RT-CV template sections contain both the Kafka enable flag and `kafka:29092` broker
- [x] `git diff --check`


zac-wang-nv (approved):
LGTM

PR #1454, opened 2026-07-29
ci: dispatch OSRB review to protected GitLab pipeline

hugoverjus:
## Summary
- add a trusted default-branch dispatcher for completed License Diff runs
- keep private compliance systems and credentials outside the public repository
- create and complete a public `OSRB Review` check while the protected review runs
- pass the source License Diff run ID and canonical Actions URL to the protected reviewer
- link the public check to the License Diff evidence instead of the follow-up dispatcher run
- allow the existing downstream helper to pass generic variables without logging their values
- explain the required developer action directly in the License Diff comment and OSRB Review check
- add a public developer guide at `.github/OSRB_REVIEW.md`
- rebase the implementation onto the latest `main`

## Public data boundary
The public workflow exposes only the PR identity, source commit, public License Diff run, and final review result. It does not publish private approval evidence or service details.

## Merge order
1. Merge and activate the protected reviewer.
2. Protect and update its standing trigger ref, then validate one end-to-end test.
3. Merge this PR into `main` after approval and green CI.
4. Sync this dispatcher and developer guide from `main` into `develop`.

PR #1528, which adds the focused License Diff overview, is already merged into `develop`.

## Test plan
- [x] rebase onto current `main` with no conflicts
- [x] `python3 .github/scripts/test_osrb_dispatch.py` — 6 tests
- [x] `ruff check .github/scripts/osrb_check.py .github/scripts/test_osrb_dispatch.py .github/scripts/trigger-downstream-pipeline.sh`
- [x] parse both edited workflows as YAML
- [x] confirm the trigger receives `DOWNSTREAM_PROJECT_PATH` and passes the resolved numeric project ID to the poller
- [x] verify all review threads are resolved
- [ ] repository CI on the latest commit
- [ ] protected reviewer approval and merge
- [ ] protected trigger setup and end-to-end handoff

The `workflow_run` dispatcher executes from the default branch, so the handoff starts applying to subsequent PRs only after this workflow lands.

### DIFF
--- a/.github/scripts/test_osrb_dispatch.py
+++ b/.github/scripts/test_osrb_dispatch.py
@@ -13,6 +13,7 @@ from pathlib import Path
 from unittest import mock
 
 DIRECTORY = Path(__file__).parent
+WORKFLOW = DIRECTORY.parent / "workflows" / "osrb-review.yml"
 
 
 def load_python(name: str, path: Path):
@@ -50,6 +51,15 @@ class DispatchTests(unittest.TestCase):
 
     def test_check_external_id_is_private_pipeline_scoped(self) -> None:
         self.assertEqual(check.EXTERNAL_PREFIX, "gitlab-osrb:")
+
+    def test_dispatch_passes_canonical_license_diff_run_url(self) -> None:
+        workflow = WORKFLOW.read_text()
+        self.assertIn(
+            '"GITHUB_LICENSE_RUN_URL":"${{ github.event.workflow_run.html_url }}"',
+            workflow,
+        )
+        self.assertIn('LICENSE_RUN_URL: ${{ github.event.workflow_run.html_url }}', workflow)
+        self.assertIn('--run-url "$LICENSE_RUN_URL"', workflow)
 
 
 if __name__ == "__main__":
--- a/.github/scripts/test_osrb_dispatch.py
+++ b/.github/scripts/test_osrb_dispatch.py
@@ -1,0 +1,56 @@
+#!/usr/bin/env python3
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""Focused tests for the public-to-private OSRB dispatch boundary."""
+
+from __future__ import annotations
+
+import importlib.util
+import os
+import unittest
+from importlib.machinery import SourceFileLoader
+from pathlib import Path
+from unittest import mock
+
+DIRECTORY = Path(__file__).parent
+
+
+def load_python(name: str, path: Path):
+    loader = SourceFileLoader(name, str(path))
+    spec = importlib.util.spec_from_loader(name, loader)
+    assert spec
+    module = importlib.util.module_from_spec(spec)
+    loader.exec_module(module)
+    return module
+
+
+trigger = load_python("trigger_downstream", DIRECTORY / "trigger-downstream-pipeline.sh")
+check = load_python("osrb_check", DIRECTORY / "osrb_check.py")
+
+
+class DispatchTests(unittest.TestCase):
+    def test_extra_variables_are_string_map(self) -> None:
+        with mock.patch.dict(
+            os.environ,
+            {"DOWNSTREAM_EXTRA_VARIABLES_JSON": '{"OSRB_REVIEW":"true","PR":"42"}'},
+            clear=False,
+        ):
+            self.assertEqual(
+                trigger.configured_extra_variables(),
+                {"OSRB_REVIEW": "true", "PR": "42"},
+            )
+
+    def test_extra_variables_reject_non_string_values(self) -> None:
+        with mock.patch.dict(
+            os.environ,
+            {"DOWNSTREAM_EXTRA_VARIABLES_JSON": '{"PR":42}'},
+            clear=False,
+        ), self.assertRaises(SystemExit):
+            trigger.configured_extra_variables()
+
+    def test_check_external_id_is_private_pipeline_scoped(self) -> None:
+        self.assertEqual(check.EXTERNAL_PREFIX, "gitlab-osrb:")
+
+
+if __name__ == "__main__":
+    unittest.main()
--- a/.github/scripts/test_osrb_dispatch.py
+++ b/.github/scripts/test_osrb_dispatch.py
@@ -74,6 +74,11 @@ class DispatchTests(unittest.TestCase):
         self.assertIn('LICENSE_RUN_URL: ${{ github.event.workflow_run.html_url }}', workflow)
         self.assertIn('--run-url "$LICENSE_RUN_URL"', workflow)
 
+    def test_dispatch_does_not_choose_the_private_pipeline_code_ref(self) -> None:
+        workflow = WORKFLOW.read_text()
+        for variable in ("OSRB_CODE_REF", "OSRB_ALLOW_UNREVIEWED_CODE"):
+            self.assertNotIn(variable, workflow)
+
     def test_github_output_explains_developer_actions(self) -> None:
         guide = DEVELOPER_GUIDE.read_text()
         license_workflow = LICENSE_WORKFLOW.read_text()
--- a/.github/scripts/test_osrb_dispatch.py
+++ b/.github/scripts/test_osrb_dispatch.py
@@ -7,6 +7,7 @@ from __future__ import annotations
 
 import importlib.util
 import os
+import re
 import unittest
 from importlib.machinery import SourceFileLoader
 from pathlib import Path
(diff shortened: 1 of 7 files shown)


.github/workflows/osrb-review.yml
```diff
@@ -0,0 +1,145 @@
...
+        if: steps.pr.outputs.skip != 'true' && steps.trigger.outcome == 'success'
+        id: poll
+        continue-on-error: true
+        env:
+          DOWNSTREAM_CI_URL: ${{ secrets.DOWNSTREAM_CI_URL }}
+          DOWNSTREAM_CI_TOKEN: ${{ secrets.DOWNSTREAM_CI_TOKEN }}
+          DOWNSTREAM_PIPELINE_ID: ${{ steps.trigger.outputs.pipeline_id }}
+          DOWNSTREAM_PROJECT_ID: ${{ steps.trigger.outputs.project_id }}
+          POLL_INTERVAL_SECONDS: "30"
+          MAX_POLL_DURATION_SECONDS: "2400"
```
simings-nv: Don't you need DOWNSTREAM_PROJECT_PATH? Similar to trigger-downstream-pipeline.sh?

.github/workflows/osrb-review.yml
```diff
@@ -0,0 +1,145 @@
...
+        run: python3 .github/scripts/poll-downstream-pipeline.py
```
simings-nv, line 133: I checked poll-downstream-pipeline.py and it seems DOWNSTREAM_PROJECT_PATH is needed there?

hugoverjus: The poller needs the numeric `DOWNSTREAM_PROJECT_ID`, not the path. The trigger helper writes both `pipeline_id` and `project_id` to its step outputs, and this workflow passes them as `DOWNSTREAM_PIPELINE_ID` and `DOWNSTREAM_PROJECT_ID` to the poll step. No second path lookup is needed.

.github/workflows/osrb-review.yml
hugoverjus: `DOWNSTREAM_PROJECT_PATH` is already supplied to the trigger step from the existing repository secret (line 95 in the rebased workflow). The helper uses it to resolve the numeric GitLab project ID before creating the pipeline.

zac-wang-nv (approved):

.github/scripts/test_osrb_dispatch.py
zac-wang-nv, line 1: where is this script run?

.github/workflows/osrb-review.yml
```diff
@@ -0,0 +1,149 @@
...
+          GITHUB_TOKEN: ${{ github.token }}
+        run: |
+          python3 .github/scripts/osrb_check.py skip \
+            --repo "$REPO" \
+            --sha "$HEAD_SHA" \
+            --run-url "$LICENSE_RUN_URL" \
+            --external-id "$EXTERNAL_ID" \
+            --summary "PR targets ${{ steps.pr.outputs.base }}; OSRB was already enforced on the develop-targeted PR."
+
+      - name: Start OSRB check
```
zac-wang-nv: This step is what creates the `OSRB Review` check run on the commit, and it sits *after* `Resolve source PR`, gated on `steps.pr.outputs.skip != 'true'`. That ordering turns any earlier failure into a silent fail-open.

If `Resolve source PR` fails — closed or deleted PR, a GitHub API blip (no retry), or `gh` not present on the `[self-hosted, downstream-pipeline]` runner — `set -euo pipefail` aborts it and no check run is ever created. Then:

1. Nothing exists on `HEAD_SHA`.
2. `workflow_run` jobs don't appear in the PR's checks list, so there is no visible signal here — you'd have to know to open the Actions tab.
3. `Complete OSRB check` still runs anyway: its condition is `always() && steps.pr.outputs.skip != 'true'`, and with the step failed `skip` is the empty string, so `'' != 'true'` holds. It then raises at `osrb_check.py:117` — `no 'OSRB Review' check exists for <sha>` — as an uncaught traceback, so the only trace left reads like a script bug rather than "the compliance gate did not run".

If `OSRB Review` is wired into branch protection, the PR sits in *"Expected — waiting for status"* with no run to click. If it isn't wired in yet, the gate silently doesn't happen and nothing says so.

The four values `start` needs — `REPO`, `HEAD_SHA`, `LICENSE_RUN_URL`, `EXTERNAL_ID` — are all job-level `env` (L42-44 plus L36-40); none come from `Resolve source PR`. So this step can move above it and drop its `if:`. `start` is already idempotent — `find_check` then PATCH-or-POST at `osrb_check.py:100-104` — so hoisting it can't double-create. On the failure path the check would then already exist, and `Complete OSRB check` would mark it `failure`, which is visible on the PR, instead of raising.

One knock-on if you make that change: the release-branch path at L74 would then find a check already open, so it wants `complete --success` rather than `skip`.

.github/scripts/test_osrb_dispatch.py
hugoverjus: Nowhere, which was the real answer. Fixed: added an `OSRB Dispatch Wiring` job to `ci.yml` that runs it on every PR.

Worth having because this workflow runs on `workflow_run` from the default branch, so a PR never executes its own copy of it. These tests are the only pre-merge signal on the dispatch wiring. They already caught a missing `DOWNSTREAM_PROJECT_PATH` on the poll step, and there is now a test asserting the tests are wired in, so this cannot regress silently.

.github/workflows/osrb-review.yml
hugoverjus: You're right, and the traceback detail makes it worse than it looks. Fixed as you suggested: `Start OSRB check` is now the first step after checkout, unconditional.

Traced the failure path to confirm it behaves: `Resolve source PR` aborts, steps 3-5 are skipped, `Complete OSRB check` still runs on `always()`, the check now exists so it PATCHes to `failure` instead of raising, and the final assertion reddens the job. Visible on the PR either way.

One thing you flagged turns out to be already handled: `skip` calls `start()` before `complete()` (`osrb_check.py:150-159`), and `start` is find-then-PATCH-or-POST, so the release-branch path works unchanged against an already-open check. No switch to `complete --success` needed.

Also split the completion summary, because a gate that could not run says nothing about the dependencies and shouldn't read like a finding. That path now says the changes were not reviewed and not to merge on the strength of the check.

Added a test pinning `Start OSRB check` as the second step with no `if:`, so the ordering can't drift back.

.github/workflows/osrb-review.yml
hugoverjus: You were right and my earlier answer was wrong. `poll-downstream-pipeline.py` calls `require_env("DOWNSTREAM_PROJECT_PATH")` before it ever looks at `DOWNSTREAM_PROJECT_ID`, so the poll step would have died on a missing variable on every run — the project ID I was passing does not spare it.

Fixed by hoisting `DOWNSTREAM_CI_URL`, `DOWNSTREAM_CI_TOKEN` and `DOWNSTREAM_PROJECT_PATH` to job-level `env` so the trigger and the poller both get them.

Added `test_workflow_supplies_every_variable_the_helpers_require`, which reads the `require_env` calls out of both helpers and asserts the workflow provides each one. Next time a helper grows a requirement, that fails in CI rather than at dispatch time.

hugoverjus:
Pushed `ab5ffeb` with the review fixes, plus two more holes of the same kind found while tracing every branch of the job.

**From review**
- The dispatch tests ran nowhere. Added an `OSRB Dispatch Wiring` job to `ci.yml`. This workflow runs via `workflow_run` off the default branch, so a PR never executes its own copy of it — the tests are the only pre-merge signal on it.
- `Start OSRB check` sat after `Resolve source PR` and was gated on its output, so a failure there left no check on the commit at all. It is now the first step after checkout, unconditional. `workflow_run` jobs are not listed on the PR, so a missing check reads as "no gate", not "gate failed".
- The poller really did need `DOWNSTREAM_PROJECT_PATH`. Moved the downstream variables to job-level `env`.

**Found while checking the rest**
- The release-branch path completed the check itself, but the completion step was gated on `skip != 'true'`. An API error there left the check `in_progress` with nothing left to close it. The completion step is now unconditional and exits early only when the release path already finished.
- `complete()` raised when no check existed, which is exactly when a PR most needs one. It now publishes the conclusion instead.
- A failed dispatch was worded like a licensing finding. A dispatch that never ran and a review that found something are different things and now read differently.

Every path was traced against the step matrix: normal pass, review not approved, dispatch failure, PR resolution failure, release branch, release-branch marking failure, and a failed check creation. All seven end with a resolved check, and only the first two say anything about the dependencies themselves.

One residual: if `actions/checkout` itself fails, no check appears, since the scripts come from that checkout. Closing that would mean duplicating the API call in inline shell before the checkout, which trades a narrow failure for a broader one.

New tests: check-creation ordering, one asserting the workflow supplies every `require_env` the helpers ask for, one that every path resolves the check, and one that `complete()` publishes without a prior `start`.

PR #1469, opened 2026-07-29
feat(cli): discover command groups from vss.commands entry points

zac-wang-nv:
## What

External packages can add top-level `vss` command groups by declaring an entry point — no change to this repo required.

```toml
[project.entry-points."vss.commands"]
acme = "acme_vss:GROUP"

[project.entry-points."vss.command_summaries"]
acme = "Acme tide tables"
```

The plugin object needs three attributes and one method:

```python
class GROUP:
    api_version = 1
    name = "acme"
    summary = "Acme tide tables"
    def cli(self) -> click.Command: ...
```

The first-party `search` group registers through those same two stanzas, so the published contract is the one the shipped CLI depends on.

## Why Click and not Typer

Laziness, not ergonomics, decided this. argparse builds its parser tree eagerly, so plugin discovery would import every installed plugin on every invocation — including `vss --help` — and one broken plugin would stop the CLI from booting. Click exposes `Group.list_commands` / `Group.get_command` for exactly this deferred resolution.

Typer sits *on* Click, so you drop to the Click primitive for this anyway. It also has no shared-option-group mechanism (the wrong shape for the planned universal job model, where every group takes `--job-id`/`--emit`/`--timeout`), and it pulls `rich`, whose decorated output fights the requirement that the final stdout line be a bounded JSON completion marker.

This takes `nvidia-vss-cli` from zero third-party dependencies to one. Click 8.5 drops `colorama` and is heading to zero runtime deps of its own.

## Two entry-point groups

`vss.commands` names the object to import. `vss.command_summaries` carries the one-line help as **data** — entry-point values are opaque strings until `.load()` parses them as `module:attr`, so `vss --help` renders a full command table while importing nothing. Summaries are per-command rather than per-distribution (`dist.metadata["Summary"]`), which matters as soon as one wheel ships two groups.

## Failure isolation

A group that fails to load becomes a `BrokenCommand`: `vss --help` still lists it, other groups still run, and invoking it reports the load error with the offending distribution named. `VSS_DISABLE_PLUGINS=name` hides it entirely (pytest's `-p no:name`).

`BrokenCommand` deliberately swallows arbitrary argv. Without that, Click's parser rejects the plugin's own flags first and the user sees `No such option '--sensor'` instead of the real reason — found while testing against a real broken package.

The ~25-line loader is copied rather than depended on: `click-plugins` is end-of-life (`Development Status :: 7 - Inactive`, last substantive release 2019) and its published build still imports `pkg_resources`.

## Test removed

`test_search_skill_contract.py` asserted on ~30 literal prose fragments of `vss-search-archive/SKILL.md`:

```python
assert "never starts a `kubectl port-forward`" in skill_text
```

That couples a Python unit test to Markdown wording and fails on any editorial change. The skill evals exercise those commands for real.

## Verification

Built and pip-installed a separate `acme-vss` distribution and confirmed end to end:

| Check | Result |
|---|---|
| Third-party group appears in `vss --help` with its declared summary | ✅ commas & spaces intact |
| `vss acme tides --sensor cam-west-77` | ✅ runs |
| Entry-point name beats callback name (`acme_group` → `acme`) | ✅ (Click 8.2 strips `_group` suffixes) |
| `VSS_DISABLE_PLUGINS=acme` | ✅ hidden |
| `acme_vss` imported during `vss --help` | ✅ **not** imported |
| `vss_cli.search` imported during `vss --help` | ✅ **not** imported |
| Plugin broken at import → `vss --help`, `vss search` | ✅ both still work |
| Broken group invoked with its own flags | ✅ reports the load error, names the dist |

`vss search --help` renders correctly even with `vss_core` uninstalled — the search runtime is genuinely never touched.

735 tests pass (126 `vss_cli` + 610 `vss_core`, includes 14 new). ruff, ruff format, and mypy clean.

> [!NOTE]
> Pre-commit's ruff hooks could not execute locally — they invoke `uv`, which cannot resolve the workspace because `nvdataset` needs the NGC index. ruff check, ruff format, and mypy were run directly against the same config instead. Worth a second look in CI.

## Not in this PR

The extension unit is a **job type**: a plugin should implement only the work, with `run`/`status`/`get`/`list` derived by the framework, since the SDD makes those pure reads against the memory index. `vss_core` has no memory module yet, so this PR ships the discovery half. `CommandGroupSpec.cli()` returns a `click.Command` today; the action/Pydantic-`Input` layer that also generates MCP tool schemas lands with the memory tier.

`search.py` keeps its argparse internals behind a passthrough — it owns 53 documented options that the skill and its evals quote verbatim, and restating that surface in the same change as the plugin system would put two large risks in one PR.


zac-wang-nv:
### Added a second commit: `fix(deps)` — relocated artifactory index + regenerated lock

Two problems surfaced while testing a lockfile-based install, both now fixed in `ec42b5813`.

**1. The `ngc-data-platform` index URL is dead.** It pointed at `urm.nvidia.com`, which returns 403 anonymously *and* with credentials. Since the index is `explicit = true` and `nvidia-vss[eval]` requires `nvdataset` from it, **every** `uv lock` / `uv sync` against this workspace failed to resolve — even for packages unrelated to eval:

```
Because nvdataset was not found in the package registry and nvidia-vss:eval
depends on nvdataset>=0.77 ... your workspace's requirements are unsatisfiable.
```

Repointed at `artifactory.pdx.nvidia.com/.../sw-ngc-data-platform-pypi-local`, which serves the index anonymously. This is a repo-wide fix independent of this PR — happy to split it out if you'd rather land it on its own.

**2. `uv.lock` was stale w.r.t. this PR.** The first commit added `click>=8.1` to `nvidia-vss-cli` without refreshing the lock, so `uv sync --frozen` produced a CLI that couldn't import its own framework. Regenerating fixes it:

```toml
[[package]]
name = "nvidia-vss-cli"
dependencies = [
    { name = "click" },
    { name = "nvidia-vss-core" },
]
```

### ⚠️ Dependency churn worth a look before merge

Full re-resolution moved more than `nvdataset`, because the lock had drifted from what the indexes now offer:

| Package | Change | |
|---|---|---|
| `argparse` | 1.4.0 → removed | |
| `cryptography` | 46.0.7 → 49.0.0 | |
| `langchain-litellm` | 0.6.1 → **0.5.1** | downgrade |
| `multi-storage-client` | 0.45.1 → **1.0.1** | major bump |
| `nvdataset` | 0.77.0 → 0.93.0 | expected |
| `questionary` | → 2.1.1 added | |

None are directly constrained by any workspace `pyproject.toml` — all transitive. The litellm downgrade and the multi-storage-client major bump touch the **agent runtime**, not the CLI, so someone who owns that path should sanity-check them. If they're unwelcome, the alternative is landing the index fix alone and regenerating the lock in a separate change with pins.

### Verified

- License gate: `OK: all 234 runtime Python dep(s) carry a permissive license` (click is BSD-3-Clause)
- `uv sync --frozen --package nvidia-vss-cli` → `+ click==8.1.8`, `vss --help` works
- 126 `vss_cli` unit tests pass
- All pre-commit hooks pass (they previously couldn't run at all, because they shell out to `uv`, which hit the dead index)

zac-wang-nv:
### Correction to the comment above — the litellm downgrade is fixed, not accepted

My earlier comment reported `langchain-litellm 0.6.1 -> 0.5.1` as unavoidable churn. That was wrong, and the amended commit (`3648b523`) fixes it properly.

**Root cause.** Nothing pins `langchain-litellm` — `nvidia-nat-langchain` requires it unversioned. Left free, the resolver picks **0.5.1**, because alone in the 0.6.x/0.7.x range it declares *no* `cryptography` dependency and therefore relaxes the solution. It isn't a conflict and 0.6.1 isn't yanked; the solver just had no reason to prefer it.

Adding a floor to the existing security constraint block resolves cleanly to **0.6.6** — an upgrade over the 0.6.1 that was there before:

```toml
"langchain-litellm>=0.6.1",     # Floor only. Unconstrained, the solver picks 0.5.1
                                # (it alone drops the cryptography dep), silently
                                # reverting the 0.6.x line
```

`0.7.0` stays out of reach because it needs `langchain-core>=1.4.7` against the existing `langchain-core>=1.3.3,<1.4` CVE pin — correctly blocked by a security constraint, not by accident.

**No security pin was touched.** `cryptography` keeps its original unbounded `>=46.0.7` and stays at **46.0.7**, because 0.6.6 caps it below 47 naturally. I briefly tried capping cryptography to chase this and reverted it — worth stating explicitly since narrowing a CVE constraint would have been the wrong fix.

**Revised lock movement — no downgrades:**

| Package | Change | |
|---|---|---|
| `argparse` | 1.4.0 → removed | stdlib backport |
| `langchain-litellm` | 0.6.1 → **0.6.6** | upgrade |
| `multi-storage-client` | 0.45.1 → 0.51.0 | minor |
| `nvdataset` | 0.77.0 → 0.92.0 | newer index |
| `questionary` | → 2.1.1 | added |
| `cryptography` | **unchanged** (46.0.7) | |

Down from 6 changed packages including a downgrade and a major bump, to 5 with neither. Lock diff is 56 lines instead of 143.

Verified: license gate passes, `uv sync --frozen --package nvidia-vss-cli` yields a working `vss`, 126 unit tests pass, `uv lock` is a no-op.

zac-wang-nv:
### Rebased onto develop (post-#1363); dependency changes dropped

**#1363 supersedes everything I'd added here on the dependency side.** Rebased onto `98a1249a6` and reduced this PR back to the plugin work plus a 5-line lock hunk.

**What I dropped, and why:**

- **The `langchain-litellm` floor** — unnecessary. #1363 raises `langchain-core` to `>=1.4.9,<1.5` and `cryptography` to `>=48.0.1`, which lands **langchain-litellm 0.7.0** on develop. That's newer than the 0.6.6 I was reaching for, and my `>=0.6.1` floor would now actively conflict (0.6.x caps `cryptography<47`).
- **The full `uv lock` regeneration** — it dragged `nvdataset` 0.89→0.93 and `multi-storage-client` 0.50→1.0.1, neither related to this PR.

**What's left** is one hunk:

```diff
 [[package]]
 name = "nvidia-vss-cli"
 dependencies = [
+    { name = "click" },
     { name = "nvidia-vss-core" },
 ]
```

click 8.4.2 already resolves in this workspace (#1363 pulled it in), so only the edge was missing — no version selection changes.

### Still outstanding, and not in this PR

The `ngc-data-platform` index still points at `urm.nvidia.com`, which returns **403** for anonymous *and* authenticated requests. The working URL is:

```
https://artifactory.pdx.nvidia.com/artifactory/api/pypi/sw-ngc-data-platform-pypi-local/simple
```

Because that index is `explicit = true` and `nvdataset` resolves only from it, **any** `uv lock` against this workspace currently fails outright — which is also why the ruff/mypy pre-commit hooks can't run (they shell out to `uv`). I kept the fix out of this PR because with `index-strategy = "unsafe-best-match"` it pulls newer `nvdataset`/`multi-storage-client`, and that belongs in a dependency PR rather than a CLI one. Happy to raise it separately.

### Verified after rebase

- `uv sync --frozen --package nvidia-vss-cli` → installs click 8.4.2, `vss --help` works
- 126 `vss_cli` unit tests pass
- All pre-commit hooks pass

PR #1473, opened 2026-07-30
feat(skills): base-profile Kubernetes operate via VSS_PUBLIC_URL

pratik-bhatt-nv:
## Summary
- Extend `deployment_resolution.md` and `skills/README.md` with a **base** Ingress contract (`vss.<ip>.nip.io`, `/vst`, public RT-VLM at `/v1` — not `/vlm/v1`).
- Wire `vss-ask-video` to `VSS_PUBLIC_URL` (legacy alias `VSS_ENDPOINT`) for VIOS + direct VLM on Kubernetes; gate Docker `docker inspect` / host-port discovery off the K8s path.
- Align `vss-generate-video-report` Mode A with the same resolution and prefer `VSS_PUBLIC_URL` for browser clip rewrites (Compose `VSS_PUBLIC_HOST`/`PORT` remains the Docker fallback).

Operators can use the [docs Agent Skills walkthrough](https://docs.nvidia.com/vss/latest/quickstart.html#agent-skills-walkthrough) prompts after `export VSS_PUBLIC_URL=…` against a Helm base deploy. `vss-manage-video-io-storage` already consumed this contract; deploy-profile Helm publish and Mode B remain out of scope.

## Test plan
- [ ] Confirm base Helm Ingress exposes `/vst` and `/v1` (`vssIngress.vlm.enabled`)
- [ ] `export VSS_PUBLIC_URL=<base ingress origin>`
- [ ] Upload / list sensors / snapshot via `vss-manage-video-io-storage`
- [ ] Ask a visual question via `vss-ask-video` (no `/vlm/v1`, no port-forward)
- [ ] Generate Mode A report via `vss-generate-video-report`
- [ ] Spot-check Docker path still works when `VSS_PUBLIC_URL` is unset (`HOST_IP` probes)
- [ ] Playbook compliance / skills-check on PR


pratik-bhatt-nv:
Fixed in 3b572ca36.

The rewrite now reduces the VIOS URL to a path (stripping repeated `http://` schemes, which VIOS `/url` sometimes emits) and restores the `/vst` prefix for bare `/storage/...` paths, matching the compat mapping Docker HAProxy already applies. Paths already under `/vst` are left alone so they don't get double-prefixed. The skill also now tells the agent to verify the link with a GET before rendering it — GET rather than HEAD, since VST lazy-renders clips and 404s on HEAD until the first GET materializes the file.

skills/vss-generate-video-report/SKILL.md
```diff
@@ -72,6 +74,31 @@ Never route reports through VSS-agent `POST /generate`.
...
+When operating against a deployed VSS stack (especially **base** on Helm), resolve
+public endpoints once. Follow
+[`../vss-build-vision-agent/references/deployment_resolution.md`](../vss-build-vision-agent/references/deployment_resolution.md):
+
+```bash
+if [ -n "${VSS_PUBLIC_URL:-}" ]; then
+  DEPLOYMENT_KIND="kubernetes"
+  VSS_PUBLIC_URL="${VSS_PUBLIC_URL%/}"
+  VSS_VIOS_URL="${VSS_PUBLIC_URL}/vst"
+  VST_API_BASE="${VSS_VIOS_URL}/api/v1"
```
pratik-bhatt-nv, line 92: Harbor root cause (non-blocking; separate follow-up): `base_profile_report` failed at step 3 after VST reachability, clip materialization, and local CR3 validation passed. Its timeline probes returned `MethodNotAllowed` or `null`, then the agent constructed a clip timestamp from `2025-01-01T00:00:00.000Z` without receiving a valid `{startTime,endTime}` range. This endpoint-resolution change does not alter the existing Mode A timeline flow, so the failure is not introduced by this PR. If evaluator compliance is needed later, make the VST path always resolve the main stream, GET `/storage/<streamId>/timelines`, and derive the clip window from a returned range before calling `/url`.

skills/vss-build-vision-agent/references/deployment_resolution.md
```diff
@@ -15,30 +15,37 @@ A Kubernetes deployment must publish one public Ingress origin, including the
...
 
 | Operate-skill name | Helm / runtime equivalents |
 |---|---|
-| `VSS_PUBLIC_URL` | `global.externalHost`, main Ingress host (`vss-search.<ip>.nip.io`), `AGENT_BASE_URL`, `VSS_AGENT_EXTERNAL_URL`, and deploy-minted `VST_EXTERNAL_URL` when it equals that origin |
+| `VSS_PUBLIC_URL` | `global.externalHost`, main Ingress host, `AGENT_BASE_URL`, `VSS_AGENT_EXTERNAL_URL`, and deploy-minted `VST_EXTERNAL_URL` when it equals that origin |
 | `VSS_VIOS_URL` | `${VSS_PUBLIC_URL}/vst` |
 | `VST_API_BASE` | `${VSS_VIOS_URL}/api/v1` — all VIOS `curl` targets |
 | `AGENT_URL` | `${VSS_PUBLIC_URL}` for Kubernetes operate skills |
-| `VSS_STREAMER_URL` | Separate streamer Ingress host (`streamer.<ip>.nip.io`); **not** under `/vst` |
+| `VLM_ENDPOINT` | `${VSS_PUBLIC_URL}/v1` when the Ingress exposes RT-VLM (base Helm `vssIngress.vlm.enabled`) — OpenAI-compatible root ending in `/v1` |
```
pratik-bhatt-nv, line 41: Scope check (non-blocking): the base chart defines `vssIngress.vlm.enabled: true` and its Ingress template routes `/v1` to RT-VLM, so this public-endpoint contract matches the deployed chart. The remaining Harbor failures do not exercise this base Helm route: two are verifier crashes without `reward.txt`; the others are Docker build/runtime behaviors (wrong foundation/profiles, writes below `deploy/docker`, or missing VLM request fields). They should be tracked separately rather than expanded into this base-Helm operate-skills PR.

skills/vss-generate-video-report/SKILL.md
```diff
@@ -154,14 +195,23 @@ Guardrails (required):
...
-deploy layer exports the browser-facing host:port as `$VSS_PUBLIC_HOST` /
-`$VSS_PUBLIC_PORT` (and scheme as `$VSS_PUBLIC_HTTP_PROTOCOL`) in every
-profile `.env` — Brev or bare-metal — so the report-link rewrite is:
+Only create `BROWSER_CLIP_URL` for URLs shown in the rendered report.
+
+**Kubernetes** — rewrite to the public Ingress origin from `VSS_PUBLIC_URL`:
+
+```bash
+: "${VSS_PUBLIC_URL:?Set VSS_PUBLIC_URL before rewriting clip URLs on Kubernetes}"
+BROWSER_CLIP_URL=$(echo "$RAW_URL" | sed -E "s|^https?://[^/]+|${VSS_PUBLIC_URL%/}|")
```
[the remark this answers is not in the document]
zac-wang-nv: Confirmed fixed in `3b572ca36` — re-verified against the charts at `5102836`. The finding was real and slightly broader than stated: base Ingress routes only `/vst` to `vss-vios-ingress` with a `/` catch-all to `vss-agent-ui` (`dev-profile-base/templates/vss-ingress.yaml`), so the authority-only rewrite put every Mode A clip link on the UI. Docker survived the same `sed` only because `haproxy.cfg.template` ships two compat backends the Ingress has no equivalent of — `bk_vst_storage_compat` (`^/storage/(.*)` → `/vst/storage/\1`) and `bk_vst_prefixed_compat` (`^/[^/]+:[0-9]+/vst/(.*)` → `/vst/\1`, which absorbs the double-`http://` VIOS defect).

The replacement handles both: it reduces to a path, strips repeated schemes, and leaves `/vst/...` alone so a `VST_EXTERNAL_URL`-configured deploy does not get `/vst/vst/...`. Using GET rather than HEAD for the probe is the right call too. Two small residual notes on the new block are in a separate inline comment.

skills/vss-generate-video-report/SKILL.md
zac-wang-nv: Went through the Harbor results independently. **Your triage checks out on every factual claim** — the two verifier crashes are real and are exactly `profile_in_2_rt_cv_person_detection` and `profile_an_1_stored_video_summarization_api`, both `RewardFileNotFoundError` with the agent side having completed successfully; the rest are Docker build/runtime behaviours (`FOUNDATION=alerts` instead of `search`, writes under `deploy/docker/`, missing `mm_processor_kwargs`/`media_io_kwargs`, `vss-rtvi-cv` exiting 255). And `base_profile_report` genuinely is not yours: no line in this diff touches the timeline flow (`git diff <base>..HEAD -- skills/vss-generate-video-report/SKILL.md | grep timeline` is empty), and none of these evals exercise the base Helm route at all.

Three things to add on top of that.

**1. The green rollup hides all of it.** All 11 eval jobs report `pass` in `gh pr checks`, but 9 of the 10 eval comments carry a `### Failing checks` section and 5 had a step fail hard enough to skip the rest. Anyone merging off the status rollup would reasonably believe the evals passed. Worth stating in the PR description which eval failures are known-and-tracked, because the checkmarks will not say so.

**2. The flip side of "not introduced by this PR" is "not covered by this PR's CI."** `base_profile_report` stopped at step-3, so steps 4–8 — *"Generate video report"*, which is the Mode A path carrying the `3b572ca36` clip-URL rewrite — never executed. Combined with every eval being a Docker deploy while this PR changes the Kubernetes path, nothing in CI touched the code this PR actually changes. That is not an objection to merging, but it does mean the base Helm path rests entirely on manual validation, which loops back to the unchecked test-plan boxes.

**3. A plausible root cause for the timeline probes, and it is a one-line fix in this file.** The endpoint is documented correctly — `vss-manage-video-io-storage/references/api-reference.md:134` gives `GET /vst/api/v1/storage/<streamId>/timelines`. But in Step 1 of this skill, item 2 is bare prose ("Fetch `/storage/<streamId>/timelines` for the recorded range…") with no command and no `${VST_API_BASE}`, while item 3 immediately below it got an explicit `${VST_API_BASE}/storage/file/<streamId>/url?...` curl **in this PR**. An agent left to guess the timelines URL is a good explanation for probes coming back `MethodNotAllowed`. Since this PR is already normalising VIOS calls onto `${VST_API_BASE}`, giving item 2 the same treatment is in scope and cheap:

```bash
curl -s --max-time 5 "${VST_API_BASE}/storage/<streamId>/timelines" | jq -r '.[] | "\(.startTime) \(.endTime)"'
```

That is also exactly the "derive the clip window from a returned range before calling `/url`" you proposed — it just lands here rather than in a follow-up.

**Separately:** one eval check looks self-inconsistent and will keep failing regardless of skill changes — `profile_in_1_streaming_dense_captions` check 8, where the judge found the proposal set `RTVI_VLM_KAFKA_ENABLED`/`RTVI_VLM_KAFKA_TOPIC` correctly but *"ruled the check's claim about the default topic name factually incorrect per the trajectory evidence."* That reads like an eval-spec bug rather than an agent failure; worth filing against the eval, not the skill.

Credit where due: the `VST_EXTERNAL_URL`-configured case — where VIOS mints clip URLs that already contain `/vst` — is now handled by the `/vst/*` arm. Worth an explicit test-plan row so it does not regress.

skills/vss-generate-video-report/SKILL.md
```diff
@@ -154,14 +195,42 @@ Guardrails (required):
...
+esac
+BROWSER_CLIP_URL="${VSS_PUBLIC_URL%/}${CLIP_PATH}"
+```
+
+Verify the result before putting it in the report — it must begin with
+`${VSS_PUBLIC_URL}/vst/`. Probe with GET, not HEAD: VST lazy-renders clips and
+returns 404 to HEAD until a GET materializes the file:
+
+```bash
+curl -sf -o /dev/null -w '%{http_code}\n' "${BROWSER_CLIP_URL}"
```
zac-wang-nv: *Non-blocking hardening — the logic is correct as written, this is about failure modes at the edges.* The rewrite is right: reducing to a path, stripping repeated schemes, and leaving `/vst/...` untouched covers both shapes VIOS mints. Two small things on the probe:

1. **No `--max-time`.** Every other `curl` in this skill carries `--max-time 5` (lines 119, 122, 125, 148, 162–163, 377). An agent following this step against an unreachable Ingress hangs on the default timeout instead of failing fast.
2. **The stated invariant is not enforced.** The prose says the result "must begin with `${VSS_PUBLIC_URL}/vst/`", but the `case` at 216–219 has no `*)` arm — an unexpected shape falls through, keeps its original path, and `BROWSER_CLIP_URL` is built anyway. A bare relative path (`storage/temp_files/x.mp4`, no leading slash) is the concrete case: `s|^[^/]*||` strips `storage`, leaving `/temp_files/...`, which matches neither arm and silently produces a wrong link.

Folding the invariant into the probe covers both:

```bash
case "${BROWSER_CLIP_URL}" in
  "${VSS_PUBLIC_URL%/}"/vst/*) ;;
  *) echo "Refusing to render clip link off the VIOS route: ${BROWSER_CLIP_URL}" >&2; exit 1 ;;
esac
curl -sf --max-time 5 -o /dev/null -w '%{http_code}\n' "${BROWSER_CLIP_URL}"
```

skills/vss-ask-video/SKILL.md
```diff
@@ -86,14 +86,45 @@ already serving. At runtime it needs:
...
+  VSS_PUBLIC_URL="${VSS_ENDPOINT}"
+fi
+
+if [ -n "${VSS_PUBLIC_URL:-}" ]; then
+  DEPLOYMENT_KIND="kubernetes"
+  VSS_PUBLIC_URL="${VSS_PUBLIC_URL%/}"
+  VSS_VIOS_URL="${VSS_PUBLIC_URL}/vst"
+  VST_API_BASE="${VSS_VIOS_URL}/api/v1"
+  # Stock base Helm exposes RT-VLM at /v1 (not /vlm/v1).
+  : "${VLM_ENDPOINT:=${VSS_PUBLIC_URL}/v1}"
```
zac-wang-nv: This unconditional default pre-empts the verified discovery block further down. Step 2 (lines 250–257) is guarded by `if [ -z "${VLM_ENDPOINT:-}" ] && [ -n "${VSS_PUBLIC_URL:-}" ]` — with `VLM_ENDPOINT` already assigned here, that whole block is dead on the Kubernetes path, and it is the copy that actually confirms the endpoint with `curl -sf .../models` before adopting it.

Two consequences on K8s:

1. `/v1` is adopted without ever being probed. On a profile whose Ingress does not expose it (search — see the comment on `deployment_resolution.md:60`) the skill proceeds with an endpoint that resolves to the UI.
2. The VIA-E-114-04 ladder at line 342 (*"Discovery above always runs first … each confirmed live with `/v1/models`"*) is no longer reachable on K8s. "No endpoint resolves" can never be true, so the user is never offered the four options. The run instead dies at line 328 with `Could not resolve a VLM model id for …; set VLM_MODEL explicitly` — which points at the wrong variable, since the real problem is that the endpoint is not a VLM.

Simplest fix is to delete this line and let Step 2 do the work — it already handles `VSS_PUBLIC_URL`, probes, and back-fills `VLM_MODEL` from `.data[0].id`. If you want the value available early, gate it on the same probe.

Same shape in `vss-generate-video-report/SKILL.md`: line 90 sets it, and the block at 311–315 that would set it is consequently dead. Harmless there (no probe in either copy), but it reads as live logic.

skills/vss-build-vision-agent/references/deployment_resolution.md
```diff
@@ -49,9 +56,42 @@ VSS_PUBLIC_URL="${VSS_PUBLIC_URL%/}"
 AGENT_URL="${AGENT_URL:-${VSS_PUBLIC_URL}}"
 VSS_VIOS_URL="${VSS_VIOS_URL:-${VSS_PUBLIC_URL}/vst}"
 VST_API_BASE="${VST_API_BASE:-${VSS_VIOS_URL}/api/v1}"
+# When Ingress exposes RT-VLM at /v1 (base profile default):
+VLM_ENDPOINT="${VLM_ENDPOINT:-${VSS_PUBLIC_URL}/v1}"
```
zac-wang-nv: The comment says "base profile default" but the assignment is unconditional, and this is the *shared* Consumer-derived block — the one the search consumers read. `dev-profile-search/templates/ingress.yaml` has no `/v1` path (`/api/chat`, `/api`, `/chat`, `/websocket`, `/static`, `/docs`, `/openapi.json`, `/redoc`, `/generate`, `/vst`, `/phoenix`, `/video-analytics-api`, `/`), and RT-VLM there is deliberately internal — `rtvi-internal-ingress.yaml` on a separate host. So on search this hands out a `VLM_ENDPOINT` that lands on the UI catch-all. Same at line 156 in the Kubernetes operate block.

Worth noting the table at line 41 in this same file already gets it right — *"`${VSS_PUBLIC_URL}/v1` **when the Ingress exposes RT-VLM** (base Helm `vssIngress.vlm.enabled`)"*. It is only the code blocks that drop the condition.

Since `/v1` is now conditional per profile, a bare reachability probe is not enough to tell a VLM from the catch-all. Recommend making the shape check the documented one:

```bash
curl -sf --max-time 5 "${VLM_ENDPOINT}/models" | jq -e '.data[0].id' >/dev/null \
  || { echo "No OpenAI-compatible VLM at ${VLM_ENDPOINT}"; VLM_ENDPOINT=; }
```

Then either carry the line-41 condition down into the code blocks, or leave `VLM_ENDPOINT` unset here and let each skill probe-then-adopt (which is what the `vss-ask-video/SKILL.md:107` comment is asking for anyway).

skills/vss-build-vision-agent/references/deployment_resolution.md
```diff
@@ -49,9 +56,42 @@ VSS_PUBLIC_URL="${VSS_PUBLIC_URL%/}"
...
+| Direct VLM (ask / report Mode A) | `${VSS_PUBLIC_URL}/v1` → `GET …/models`, `POST …/chat/completions` |
+| Phoenix (optional) | `${VSS_PUBLIC_URL}/phoenix` |
+
+Base operate skills for the quickstart walkthrough:
+
+- `vss-manage-video-io-storage` — VIOS via `${VST_API_BASE}`
+- `vss-ask-video` — VIOS clip URL + direct VLM at `${VLM_ENDPOINT}` (**not** Agent `/generate`)
+- `vss-generate-video-report` Mode A — same VIOS + VLM path (**not** Agent `/generate`)
+
+Do **not** use `${VSS_PUBLIC_URL}/vlm/v1`. Stock base Helm and Docker HAProxy expose
```
zac-wang-nv: First half is right and worth keeping — `/vlm/v1` is absent from both, and grep confirms no stale references survive anywhere in the tree. Second half over-reaches: Docker HAProxy does not expose RT-VLM at `/v1` either. `haproxy.cfg.template` has no `/v1` ACL and no VLM backend at all (`grep -nE 'v1|vlm|VLM'` returns nothing), so `${public_origin}/v1/models` on Compose falls through `use_backend bk_vss_ui if h_main` to the UI.

The skills themselves are consistent — the Docker path resolves the VLM from host ports `:30082` / `:8018`, never the public origin. It is only this sentence that would send someone the wrong way. Suggest:

> Stock base Helm exposes RT-VLM under **`/v1`**; neither base Helm nor Docker HAProxy serves `/vlm`. On Docker Compose the VLM is not published on the public origin at all — use the host port (`:30082` NIM / `:8018` RT-VLM).

*(skills-review-bot independently flagged this same line — it is one of the two of its seven "high" findings that hold up.)*

zac-wang-nv:
The Kubernetes-vs-Docker resolution block now exists in four places and has already drifted: [`vss-ask-video/SKILL.md:95`](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/510283604b0cb7b150b524eec4e0451b60ffd5bd/skills/vss-ask-video/SKILL.md#L95) (has the `VSS_ENDPOINT` alias and the `VLM_ENDPOINT` default), [`vss-generate-video-report/SKILL.md:83`](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/510283604b0cb7b150b524eec4e0451b60ffd5bd/skills/vss-generate-video-report/SKILL.md#L83) (default, no alias), [`vss-manage-video-io-storage/SKILL.md:61`](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/510283604b0cb7b150b524eec4e0451b60ffd5bd/skills/vss-manage-video-io-storage/SKILL.md#L61) (error-exit branch, no VLM), and the canonical copy at [`deployment_resolution.md:54`](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/510283604b0cb7b150b524eec4e0451b60ffd5bd/skills/vss-build-vision-agent/references/deployment_resolution.md#L54). The two `VLM_ENDPOINT` comments above are one bug replicated across copies, which is the argument for not having four.

Both edited skills are also over the STR-002 target and this PR pushes them further (`vss-ask-video` 590 lines, `vss-generate-video-report` 631 — the compliance bot flagged both). Since `deployment_resolution.md` is already linked from both, folding the block into it and leaving a two-line "resolve endpoints per the contract, then use `${VST_API_BASE}` / `${VLM_ENDPOINT}`" pointer would fix the size warning and the drift in one move. Not blocking — worth doing before a third skill copies the block.

zac-wang-nv:
Independent pass over the skills-review-bot's 7 "high" findings, since all are single-lens (agreement 1/6) and only some survive.

**Hold up (2):**
- `deployment_resolution.md:88` — Docker HAProxy does not expose RT-VLM at `/v1`. Inline comment on that line.
- `vss-generate-video-report/SKILL.md:228` — clip probe has no `--max-time`. Inline comment on that line.

**Do not hold up (5):**
- `deployment_resolution.md:41` — "phantom Helm key". Agreed with your scope-check reply; verified the same way (`values.yaml:156-159` and `values-base.yaml:74-76` both set `vlm.enabled: true`, and `vss-ingress.yaml:4` reads it via `$vlmEnabled := default false (index $vlmIngress "enabled")` to gate `/v1`).
- `vss-generate-video-report/SKILL.md:215` — *"sed path-strip silent no-op on already-absolute path"*. The no-op is the intended behaviour: `s|^[^/]*||` matching zero characters on `/storage/...` is exactly what should happen. (There is a real adjacent edge — a path with **no** leading slash — but that is the missing `*)` arm noted inline at line 228, not this `sed`.)
- `vss-ask-video/SKILL.md:102`, `vss-ask-video/SKILL.md:364`, and `vss-generate-video-report/SKILL.md:154` all sit outside this PR's changed lines.

zac-wang-nv (approved):
LGTM, but we need to unify the endpoint for VLM to be `/v1/vlm/models/` /v1/vlm/chat/completion` in a followup PR

PR #1480, opened 2026-07-30
fix(ci): fail fast when the companion GHCR build has already failed

simings-nv:
## Description

**This is not about test failures.** The bridge already stops on the first blocking downstream failure and already ignores `allow_failure: true` jobs — that part works. This is one layer earlier: **you cannot run tests at all if the images never got built, and nothing noticed the build had died.**

`Wait for Build Dev Images` could not tell *"the build hasn't finished"* from *"the build failed"*, so a failed companion build was invisible and the job polled out its whole budget waiting for an artifact that was never coming.

### Measured

Four runs between 2026-07-23 and 07-29 did this. 520 polls at 15s, ~132 minutes each, roughly **9 hours of delayed verdicts**, each ending on:

```
GHCR build run for f2da3e71421a is not ready; retrying in 15s (519/520)
RuntimeError: no successful GHCR build run found for f2da3e71421a
```

The information was available on the first poll.

### Cause

Two filters, either of which alone would have caused it:

- the workflow-runs query asked for `status=success`
- `select_release_set_run` additionally required `conclusion == "success"`

So a failed run and a nonexistent one both came back as `None`, and the caller could only conclude "keep waiting."

### Change

`select_release_set_run` now returns the newest run matching `head_sha` + `head_branch` regardless of conclusion, and the caller distinguishes three cases:

| Companion run | Behaviour |
|---|---|
| `completed` + `success` | proceed |
| `completed` + `failure`/`cancelled`/`timed_out`/`startup_failure`/`stale`/`neutral`/`skipped` | **raise now**, naming run id, conclusion and URL |
| anything else | keep polling |

A conclusion is only interpreted once `status == "completed"`. **A GitHub re-run reuses the run id and resets status while the old conclusion may still read stale** — without that gate this would abort somebody's retry. `action_required` also keeps polling, since approval can resume the same run. An unrecognised conclusion keeps polling too, so an unfamiliar value can cost time but never cause a false failure.

The query is now scoped by `branch` and pages at 100, so a busy SHA cannot evict the relevant run.

### Impact, stated honestly

**This is a latency fix and does not change pass rate.** Those four runs still fail — roughly 130 minutes sooner. The frequency is four occurrences in one seven-day sample, so treat the ~9 hours as indicative rather than a rate.

One operational note: fail-fast removes the accidental 130-minute window in which someone could manually re-run the build and rescue the still-polling job. Recovery becomes re-run Build Dev Images, then re-run the wait job.

### Tests

24 pass. New coverage in `BuildWaitTest`, including the two that guard the change itself:

- `test_query_does_not_filter_on_success` — asserts the request contains `head_sha=` and `branch=` and **not** `status=`, so the fix cannot be silently undone
- `test_in_flight_github_rerun_is_not_aborted` — same run id, `run_attempt: 2`, `in_progress` with a stale `failure` conclusion, must not raise
- every terminal non-success conclusion fails immediately, asserting `sleep` was never called
- still-in-progress at the final attempt times out without downloading an artifact
- an older failed execution does not abort a newer running one

## Checklist

- [x] Every commit is DCO sign-off'd
- [x] `test_update_pr_ghcr_candidates.py` already runs in `ci.yml`, so the new behaviour is gated
- [x] No other caller of `select_release_set_run`; `prepare_nightly_promotion.py` imports only `download_release_set_artifact` and is unaffected
- [x] Nothing in either repo depends on the old error string

### DIFF
--- a/.github/scripts/test_update_pr_ghcr_candidates.py
+++ b/.github/scripts/test_update_pr_ghcr_candidates.py
@@ -24,27 +24,49 @@ class CandidateCommentTest(unittest.TestCase):
         self.assertIsNone(module.pr_number("develop"))
         self.assertIsNone(module.pr_number("pull-request/not-a-number"))
 
-    def test_select_release_set_run_requires_exact_ref_sha_and_success(self):
+    def test_select_release_set_run_requires_exact_ref_and_sha(self):
+        """Exact match only; conclusion is the caller's business, not the selector's."""
         runs = [
-            {
-                "id": 1,
-                "head_sha": "a" * 40,
-                "head_branch": "pull-request/1190",
-                "conclusion": "failure",
-            },
-            {
-                "id": 2,
-                "head_sha": "a" * 40,
-                "head_branch": "pull-request/1190",
-                "conclusion": "success",
-            },
+            {"id": 1, "head_sha": "b" * 40, "head_branch": "pull-request/1190",
+             "conclusion": "success", "created_at": "2026-07-01T00:00:00Z"},
+            {"id": 2, "head_sha": "a" * 40, "head_branch": "develop",
+             "conclusion": "success", "created_at": "2026-07-01T00:00:00Z"},
+        ]
+        self.assertIsNone(
+            module.select_release_set_run(runs, "a" * 40, "pull-request/1190")
+        )
+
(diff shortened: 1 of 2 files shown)


zac-wang-nv (approved):
LGTm

PR #1485, opened 2026-07-30
fix(ci): retag develop-<sha> at every develop tip, including no-change commits

zac-wang-nv:
## Summary

`develop-<sha12>` is missing at any commit where no first-party image changed — which is the **common** case, not a rare one. `tree-<sha>` content reuse (#1385) means an unchanged source tree never rebuilds, so `changes.count == 0` happens routinely.

Three of the last six develop commits are in this state. Reported twice today against `develop-b982b2680ed3` and `develop-ac8e5175076f`, both giving `manifest unknown`.

## Two causes, both introduced by #1446

**1. The guard.** The post-merge steps were gated on `changes.count > 0`:

```yaml
if: github.ref_name == 'develop' && fromJSON(needs.changes.outputs.count) > 0
```

That was correct when the step only moved `develop-latest` for freshly built images — no build, nothing to move. #1446 widened selection to *evidence* but left the guard, so the retag skipped exactly when it was most needed. Dropped from all three steps (Buildx setup, GHCR login, retag) — they must stay in agreement or the retag runs without a registry login.

**2. Even ungated it would have failed.** On a no-change commit every release-set entry looks like this:

```json
{ "name": "vss-agent",
  "image": "ghcr.io/nvidia-ai-blueprints/vss/vss-agent",
  "tag": "develop-latest",
  "digest": null,
  "strategy": "reuse-pinned" }
```

All 15 entries, digest `null` — nothing was built to resolve one. `_retaggable` required a digest, so `alias_plan` would have found nothing and raised `"release set has no retaggable GHCR entries"`. A silent skip would have become a hard failure.

Selection now needs a `ghcr.io/` repository plus *any* resolvable source reference. `imagetools create` accepts a tag as readily as a digest, so a null-digest entry retags from its recorded tag. Semantically correct: nothing changed at this commit, so `develop-latest`'s content **is** this commit's content. `advance()` reports the resolved digest for those rather than asserting parity, since the release set gave it nothing to compare against.

## Scope

Built images are unaffected — the build step already publishes `develop-<sha12>` directly, so the retag only ever mattered for the reuse-pinned and mirror set. Non-GHCR pins (`nvcr.io`) stay excluded and remain derivable from git at the commit. The tree-SHA gate is unchanged and still fails closed.

## Test

Three new cases: an all-reuse-pinned release set retags from `develop-latest`; a tag-sourced retag with no digest reports instead of asserting; a non-GHCR entry stays excluded when it also lacks a digest. 16 tests pass; `ruff` and workflow YAML clean.

## Follow-up

This unblocks the nightly mirror in ci-vss-oss. Two things still outstanding there:

- The tag job derives a **7-char** short SHA (`rev-parse --short`); GitHub publishes 12 (`commit_sha[:12]`). `40-promote-release.yml` already uses `--short=12` — the nightly tag job needs the same. Git's default width is also unstable, so it's the wrong primitive for a tag contract regardless.
- [ci-vss-oss!315](https://gitlab-master.nvidia.com/metromind/ci-vss-oss/-/merge_requests/315) is over-built now that #1447 put all five images on the shared coordinate; it should shrink to setting `VSS_CONTAINER_TAG` plus an existence probe.

### DIFF
--- a/.github/scripts/advance_ghcr_alias.py
+++ b/.github/scripts/advance_ghcr_alias.py
@@ -16,12 +16,12 @@ Two behaviours differ from the original build-only alias mover:
 1. **Every GHCR entry is retagged, not just ``strategy == "build"``.** A tag set
    that skipped ``mirror`` / ``reuse-pinned`` entries would be incomplete at the
    commit, so a SHA-derived coordinate would 404 for those images. Entries are
-   selected on evidence — a ``ghcr.io/`` repository plus a resolved manifest
-   digest — rather than on strategy, so coverage tracks what is *actually*
-   published rather than a hardcoded list. That is 3 images today — vss-agent,
-   vss-agent-ui, vss-alert-ms, the managed set the shared ``VSS_CONTAINER_TAG``
-   governs — and grows on its own as ``ghcr_build`` is enabled for the staged
-   build entries. Non-GHCR pins (nvcr.io) cannot be retagged and are skipped;
+   selected on evidence — a ``ghcr.io/`` repository plus a resolvable source
+   reference — rather than on strategy, so coverage tracks what is *actually*
+   published rather than a hardcoded list. That is 5 images today (vss-agent,
+   vss-agent-ui, vss-alert-ms, vss-video-analytics-api, vss-behavior-analytics)
+   and grows on its own as ``ghcr_build`` is enabled for the staged build
+   entries. Non-GHCR pins (nvcr.io) cannot be retagged and are skipped;
    they are pinned in git at this commit and stay derivable from the repo.
 
 2. **Tree-SHA gate.** ``--verify-tree-sha`` refuses to retag an image whose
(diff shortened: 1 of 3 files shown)


.github/scripts/advance_ghcr_alias.py
```diff
@@ -90,10 +93,15 @@ def alias_plan(release_set: dict, alias: str) -> list[AliasUpdate]:
         name = str(image.get("name") or "")
         if not tag:
             raise ValueError(f"{name}: incomplete immutable coordinate")
+        # Prefer the digest when the build resolved one; fall back to the tag,
+        # which imagetools resolves at create time. On a no-change commit the
+        # recorded tag is develop-latest, and since nothing changed at this
+        # commit its content IS this commit's content.
+        source = f"{repository}:{tag}@{digest}" if DIGEST_RE.fullmatch(digest) else f"{repository}:{tag}"
```
simings-nv: Consider reuse the existing content-addressed tree-<source_tree_sha> tags, not develop-latest?

Worried this uses the develop-latest which is mutable.

simings-nv: If the expected tree tag is missing or has mismatched provenance, the workflow should rebuild or fail closed. An existing `develop-<sha>` should no-op only when its digest matches and fail otherwise. This preserves build avoidance for unchanged images while keeping the tree and digest gates effective, and it also removes the `develop-validated` race.

simings-nv (approved):
LGTM

zac-wang-nv:
merged-via-#1486

PR #1497, opened 2026-07-31
Make normalize_alert_message null-safe for partial analytics payloads

danh-nguyen-nvidia:
The guard admitted a message when any one of sensor / analyticsModule / object was present, then dereferenced all three unconditionally. An nv.Incident payload carrying flat sensorId and analyticsModule but no nested sensor raised "AttributeError: 'NoneType' object has no attribute 'get'", reproducible with test/protobuf/test_data/sample_incident.json. The mirror case (sensor without analyticsModule) failed the same way.

The blast radius is the whole batch, not the one message: the call site in enhance_alert_with_vlm.py normalizes inside a list comprehension whose exception unwinds to the handler wrapping process_batch_vlm, so a single partial payload dropped every other message in the batch before VLM processing, leaving only an "Error processing batch" log line.

Read each id behind its own guard, and admit the message when any id was found rather than only sensor_id or category, so an object-only payload yields objectIds instead of returning early. Existing flat fields are still never overwritten.

test_object_only_message_is_normalized was already pinned xfail(strict) against this defect, so drop the marker now that it passes and add regression tests for both partial shapes.

Refs #1344

## Checklist


services/alert/src/utils/schema_util.py
```diff
@@ -162,21 +175,25 @@ def convert_behavior_to_protobuf_behavior(behavior: dict) -> nvSchemaBehavior:
...
+    protobuf_behavior.place.CopyFrom(place_to_nv_place(behavior.get("place")))
 
     # Sensor, Analytics Module, Object, and Event
-    protobuf_behavior.sensor.id = behavior.get("sensor", {}).get("id", "")
-    protobuf_behavior.analyticsModule.id = behavior.get("analyticsModule", {}).get("id", "")
+    sensor = _as_dict(behavior.get("sensor"))
+    am = _as_dict(behavior.get("analyticsModule"))
+    obj = _as_dict(behavior.get("object"))
+
+    protobuf_behavior.sensor.id = sensor.get("id", "")
```
dineshvenkat: only nested sensor.id is read here, so a payload with just a flat sensorId publishes an empty sensor id

services/alert/test/unit/api/test_alert_service.py
```diff
@@ -163,11 +163,21 @@ async def test_key_is_empty_when_nothing_identifies_the_alert(self, service):
...
+        the alert was well-formed. The converter now coerces the block, so the
+        alert is published with the flat fields it did carry; one that carries
+        nothing identifying is dropped later by the ``sensorId``/``timestamp``/
+        ``end`` guard, the same way a sensor-less protobuf alert already is.
+        """
+        _body, status = await service.submit_nvschema_alert(
+            {"sensorId": "cam-9", "sensor": None}
+        )
+        assert status == 202
+        service.kafka_producer.produce.assert_called_once()
```
dineshvenkat: the test never decodes what was published, so it does not actually check the flat field survived

services/alert/test/unit/utils/test_event_utils.py
```diff
@@ -139,15 +139,28 @@ def test_tracker_merges_with_pre_existing_entries(self):
...
+            {
+                "sensorId": "HWY_20_AND_LOCUST__WBA",
+                "category": "collision",
+                "analyticsModule": {"id": "Collision Detection Module"},
+            }
+        )
+
+        assert result["sensorId"] == "HWY_20_AND_LOCUST__WBA"
+        assert result["category"] == "collision"
+        assert result["notification_type"] == "alert"
```
dineshvenkat: this is an incident payload, and the test expects it to come out labelled an alert

services/alert/src/utils/schema_util.py
```diff
@@ -188,7 +205,7 @@ def convert_behavior_to_protobuf_behavior(behavior: dict) -> nvSchemaBehavior:
 
     # Embeddings
     for embedding in behavior.get("embeddings", []):
-        protobuf_behavior.embeddings.add().vector.extend(embedding.get("vector", []))
+        protobuf_behavior.embeddings.add().vector.extend(_as_dict(embedding).get("vector", []))
```
dineshvenkat, line 218: (Note: original finding at line 213, snapped to nearest changed line) map_geo_location calls .get on the locations block without checking it is a dict, so a mistyped value raises

dineshvenkat:
Thanks for this — the null-guard in commit 1 is right, and commit 3 is the strongest part of the PR. Round-tripping a real `Behavior` through `SerializeToString` and the actual decoder, and reading `SUBMESSAGE_FIELDS` off the descriptor instead of listing them, means a submessage added to the proto later gets covered without anyone remembering to extend the test. Two of the three retargeted tests are clean — both had `xfail(strict)` dropped with byte-identical bodies, which the strict marker forced.

Two things sit outside the diff, so I couldn't leave them as line comments.

**1. A malformed message still takes down its whole batch.**

Commit 1's message documents this, and the fix doesn't close it — `normalize_alert_message` still isn't total. The tracker merge dereferences a payload-controlled field:

```python
existing = updated.get('_normalized_added_fields') or []
updated['_normalized_added_fields'] = list({*existing, *added})
```

On the redisStream path that value is client-controlled JSON, so it is reachable. Run against this branch:

```
{'sensor': {'id': 'cam-1'}, '_normalized_added_fields': 5}     -> TypeError: 'int' object is not iterable
{'sensor': {'id': 'cam-1'}, '_normalized_added_fields': [{}]}  -> TypeError: unhashable type: 'dict'
```

And in a three-message batch where only the middle one is bad, all three are lost. `process_batch_vlm` normalizes inside a list comprehension whose only handler is the batch-level `except`, so the two valid messages never reach VLM and one `"Error processing batch"` line is all that remains. The same function already does per-message catch-and-skip twice for JSON decode, a few lines above.

I don't think this has to land in this PR — the crash you set out to fix is genuinely gone, and this blast radius predates you. But it would be good to have it tracked rather than dropped.

**2. This PR masks the redisStream incident misroute rather than fixing it.**

`source_redis_stream.py` emits `kind='anomaly'`; the scheduler maps anything that isn't `'incident'` to `Behavior`, and normalization then runs on it. So genuine incidents on that path get `notification_type='alert'` stamped. Kafka sets `kind` correctly, which is why this only shows up on redis.

Before this PR that combination crashed — destructive, but loud. After it, the incident survives and is treated as an alert the rest of the way: `notification_type` selects the alert index and the alert fingerprint function, stamps `type: mdx-vlm-alerts`, and routes through `convert_behavior_to_protobuf_behavior`, which reads none of `sensorId`, `category`, `objectIds` or `isAnomaly`. The incident never reaches the incident index, so VSS alert verification won't find it.

Two details that make it hard to notice: `notification_type` is set outside the `if added:` block, so when the payload already carries flat `sensorId`/`category` — as incidents do — nothing is recorded in `_normalized_added_fields` and the relabel leaves no trace. And `strip_normalization_fields` has no caller outside tests, so nothing undoes it downstream.

Not asking you to fix this here either. What I'd like before merge is a follow-up bug filed and cross-referenced from 6535815, so the masking is a known open item rather than something the next person has to rediscover. I've also left a note on `test_analytics_module_without_sensor_is_normalized`, since that test now pins "incident becomes an alert" as expected behaviour and whoever fixes the misroute will need to change it.

Everything above was verified by running it against this branch, except the routing consequences in section 2, which I traced by reading. Happy to be corrected on any of it.

services/alert/src/utils/schema_util.py
danh-nguyen-nvidia: Good catch — fixed in b99dbd5b3.

`sensor.id` now falls back to the flat `sensorId` when the nested block does not supply one, and the nested id still wins when both are present. This matches the Kafka key derivation in `AlertSubmissionService`, which already accepts the flat form as an identifier — the converter was the odd one out.

Verified end to end: `{"sensorId": "cam-9", "sensor": null}` now decodes with `sensorId` set and clears the `sensorId`/`timestamp`/`end` guard, instead of being accepted with a 202 and then silently discarded.

services/alert/test/unit/api/test_alert_service.py
danh-nguyen-nvidia: Fixed in b99dbd5b3. The test now decodes the published bytes and asserts `sensor.id == "cam-9"`.

You are right that this was the root of it — had the test decoded the payload originally, it would have caught the dropped identifier above rather than passing on `produce()` having been called.

services/alert/test/unit/utils/test_event_utils.py
danh-nguyen-nvidia: Agreed — fixed in b99dbd5b3. Dropped the `notification_type == "alert"` assertion so the test no longer turns the mislabelling into an expectation.

The payload is kept because it is the reported repro shape, but the docstring now records that it is incident-shaped, that it only reaches normalization when a source mislabels its batch kind, and that the label is a separate defect left deliberately unasserted.

(`test_sensor_without_analytics_module_is_normalized` still asserts the label, since that payload carries a nested `sensor` and is a genuine alert.)

services/alert/src/utils/schema_util.py
danh-nguyen-nvidia: Fixed in b99dbd5b3. `map_geo_location` now coerces the block through `_as_dict`, and the coordinate entries with it, so a null entry inside `coordinates` is tolerated too.

This one was an inconsistency in my own change — the earlier pass covered `sensor`, `analyticsModule`, `object` and `place` but skipped the location blocks, so the same malformed shape raised in one place and was tolerated in another. Added parametrized coverage for `locations`/`smoothLocations` as null, string, list and int.

services/alert/src/utils/schema_util.py
```diff
@@ -162,21 +175,30 @@ def convert_behavior_to_protobuf_behavior(behavior: dict) -> nvSchemaBehavior:
...
+    sensor = _as_dict(behavior.get("sensor"))
+    am = _as_dict(behavior.get("analyticsModule"))
+    obj = _as_dict(behavior.get("object"))
+
+    # ``Behavior`` has no flat ``sensorId`` field, so a client that carries the
+    # id there rather than under ``sensor`` would publish an empty sensor and be
+    # dropped downstream for a missing ``sensorId``. The Kafka key derivation in
+    # ``AlertSubmissionService`` already accepts the flat form as an identifier;
+    # honour the same fallback here so the two agree.
+    protobuf_behavior.sensor.id = sensor.get("id") or behavior.get("sensorId") or ""
```
dineshvenkat: this reads behavior["sensorId"] straight into a protobuf string field — a non-string value like {"sensorId": 7} returned 202 before this commit and now raises TypeError, so the endpoint answers 500

services/alert/src/utils/schema_util.py
```diff
@@ -188,17 +210,18 @@ def convert_behavior_to_protobuf_behavior(behavior: dict) -> nvSchemaBehavior:
...
     return protobuf_behavior
 
 def map_geo_location(proto_geo_location, geo_location_dict):
     """Maps geo-location data from dict to protobuf."""
+    geo_location_dict = _as_dict(geo_location_dict)
     if geo_location_dict:
         proto_geo_location.type = geo_location_dict.get("type", "")
         for coord in geo_location_dict.get("coordinates", []):
-            point = GeoLocation.Point(point=coord.get("point", []))
+            point = GeoLocation.Point(point=_as_dict(coord).get("point", []))
```
dineshvenkat: coercing each coordinate entry turns a bare [lon,lat] array — the standard GeoJSON shape — into an empty point, so a client sending [[1.0,2.0],[3.0,4.0]] now gets a 202 with its trajectory published as empty geometry, where before it raised

services/alert/src/utils/schema_util.py
danh-nguyen-nvidia: Confirmed and fixed in b3e2fdac7 — that was a regression I introduced, not something the original change had.

Reproduced it against the previous commit: `{"sensorId": 7}` gave `sensor.id=""` and a 202 before, `TypeError` and a 500 after.

The fallback now takes the flat value only when it is a string. Every other type is ignored again, so the submission is accepted and then dropped by the `sensorId` guard exactly as it was before. Added parametrized coverage for `7`, `null`, a list and a dict.

services/alert/src/utils/schema_util.py
danh-nguyen-nvidia: Agreed, and fixed in b3e2fdac7 by removing the coercion.

Coercing the entries was an over-reach on my part — your original comment asked for the `locations` block to be type-checked, and I extended it to the coordinates as well without thinking through what defaulting one would mean. Verified your case: `[[1.0,2.0],[3.0,4.0]]` raised before and came back as `coords=[[], []]` under a 202 after, so the trajectory was published with its geometry silently dropped.

The entries are uncoerced again, and there is now a comment and a test pinning the reason so it does not get reintroduced.

Separately: should the bare `[lon, lat]` form be accepted outright? The mapper only understands `{"point": [...]}` today. That looked like a question about the input contract rather than something to settle by defaulting, so I left it out of this commit — happy to pick it up if you want it supported.

dineshvenkat (approved):

danh-nguyen-nvidia:
/ok to test [4bd227b](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1497/commits/4bd227ba931e74bfc7e357898bc98142df1f135e)

PR #1507, opened 2026-07-31
fix(ci): gate downstream on source and deploy changes, not on GHCR builds

zac-wang-nv:
## The problem

Downstream acceptance is gated on `has-ghcr-build-entries` — *"did any GHCR image get rebuilt this run"*:

```yaml
if: needs.wait-for-build-dev-images.outputs.has-ghcr-build-entries == 'true'
```

That's a poor proxy twice over:

- **Build avoidance means a real source change can rebuild nothing.** Content-addressed reuse re-tags instead of building, so `false` doesn't mean "nothing changed".
- **Deploy-only changes never rebuild anything** — yet compose, profiles and `containers.env` edits are exactly what acceptance exists to catch.

Net effect: config and deploy edits get **no downstream coverage at all**, along with every change to the 13 images not published to GHCR. Reported on #1504.

## The new gate

```
(source changed AND (ghcr_build OR trigger_downstream_from_source))
OR deploy/ changed
```

Scoped on what *changed*, not what got built.

**`trigger_downstream_from_source`** is a new optional inventory flag. An image not published to GHCR can still want downstream coverage when its source changes — mirrored and externally pinned components are deployed by the same profiles and break the same evals. This lets a module opt in without being GHCR-built.

| | |
|---|---|
| Covered automatically (`ghcr_build`) | `vss-agent`, `vss-agent-ui`, `vss-alert-ms`, `vss-video-analytics-api`, `vss-behavior-analytics` |
| Can opt in per-image | the other 13 — `vss-vios-nvstreamer`, `vss-rt-cv`, `vss-video-summarization`, … |

## Fails open

An unresolvable diff runs downstream rather than skipping it, matching the posture used elsewhere in this workflow. A spurious pipeline costs runner time; a spurious skip costs coverage nobody notices is missing.

**That's why `fetch-depth: 0` is needed** on the `wait-for-build-dev-images` checkout — it was shallow, so the diff would always fail, the gate would always fail open, and downstream would run for every push. Safe, but it defeats the gate.

## Not removed

`has_ghcr_build_entries` is still emitted. It answers whether images were *built*, which is a different question from whether downstream should *run*, and any other consumer can keep using it.

## Test

10 tests: GHCR source change, opted-in non-GHCR source change, unflagged source change skipping, deploy-only change running with no source change, unrelated change skipping, unresolvable diff running, and a `source_path` prefix not matching loosely — `services/agent` must not catch `services/agent-extras`.

## Follow-up, not in scope here

With the decision now driven by changed paths rather than the release-set artifact, `wait-for-build-dev-images` itself (520 × 15s poll, `timeout-minutes: 140`) exists mainly to fetch that artifact. Worth retiring separately once the release-set dependency is fully unwound.


kaushikc-nvidia (dismissed):

kaushikc-nvidia:
/ok to test [20a9e58](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1507/commits/20a9e587c426ea49c79a648c50f6fe92a86f9757)

PR #1513, opened 2026-07-31
fix(lvs): forward scenario/events to CA-RAG and retry empty aggregation results

david-duu:
## Summary

Two independent defects in the CA-RAG aggregation path, both surfacing as empty or degraded summaries.

**1. `scenario` / `events` were dropped before reaching the aggregation LLM**

- They were only forwarded to the CA-RAG summarization params when `enable_vlm_structured_output=False`
- `enable_vlm_structured_output` defaults to `True` and the reported payload never overrides it, so the aggregation LLM received no context about the scenario or the event types being detected
- Confirmed deterministically from LVS debug logs for the same request:
  - without fix: `params={'time_overlap_threshold': 0.1, 'max_events_per_batch': 50, 'kafka_enabled': 'false', 'enable_llm_merging': 'false', 'caption_source': 'sse', 'uuid': ...}`
  - with fix: `params={..., 'caption_source': 'sse', 'scenario': 'warehouse', 'events': ['forklift', 'crash'], 'uuid': ...}`
- Fix moves the forwarding outside the `enable_vlm_structured_output` guard

**2. A single bad aggregation sample surfaced as a fully empty response**

Reproduction on matching hardware (H100 80GB, `cosmos3-nano-reasoner` modelopt-fp8 NIM served locally on GPU) showed roughly 5-10% of requests returning HTTP 200 with `total_events=0`, `events=[]` and `video_summary=""`. Across 40 calls the failures followed no endpoint — an empty result landed on `/summarize` as well as `/v1/summarize` — so this is **not** an endpoint parity defect. Both routes are the same handler behind stacked FastAPI decorators. The aggregation LLM (`nvidia-nemotron-nano-9b-v2` at temperature 0.2) intermittently samples a response that yields nothing parseable.

Fix adds an empty-result guard that re-runs the CA-RAG aggregation call when the result carries neither events nor a video summary:

- Aggregation reads already-persisted captions and runs an LLM over them, so it holds no state of its own and re-running it is safe
- A result with no events but a non-empty `video_summary` is a legitimate "nothing notable happened" answer and is **not** retried
- Non-JSON free-form text is treated as a usable summary as long as it is not blank
- Error responses are returned immediately without retry, preserving the existing `DependencyError` classification
- When every attempt comes back empty the last response is returned unchanged, so a genuinely empty aggregation is not converted into a new error path
- Applies to both the file path (`summarization`) and the livestream path (`summarization_online`)
- Retry budget defaults to 2, overridable via `LVS_AGGREGATION_EMPTY_RETRIES`; `0` restores the previous single-call behaviour

Note on scoping: fix 1 is a genuine defect but is not the cause of the empty output — it was verified to survive that change. Fix 2 targets the empty-output symptom directly. The bug is best re-scoped to "intermittent empty summarization output with CR3 FP8, both endpoints" rather than endpoint parity.

## Test plan

- [x] `pytest tests/unit/` — 26 passed (4 scenario/events forwarding + 22 aggregation guard)
- [x] Empty-result detection covers: `None`, blank, whitespace-only, no-events-and-no-summary JSON, whitespace-only summary, dict results
- [x] Non-empty detection covers: summary-only, events-only, free-form text, non-dict JSON
- [x] Retry behaviour covers: no retry on first good result, retry until non-empty, bounded retries returning the last response, retries disabled, error responses not retried, metadata taken from the successful attempt
- [x] `LVS_AGGREGATION_EMPTY_RETRIES` parsing covers unset, valid, invalid, and negative values
- [ ] Re-run the 40-call sample (10 per endpoint per build) on `viking-prod-349-au` to confirm the empty-output rate drops
- [ ] Confirm no regression in summarization quality for existing warehouse/retail/security scenarios

### DIFF
--- a/services/video-summarization/CLAUDE.md
+++ b/services/video-summarization/CLAUDE.md
@@ -230,6 +230,7 @@ VIA uses extensive environment variable configuration. Key patterns:
 - Database backends selected via `LVS_DATABASE_BACKEND`
 - `KAFKA_ENABLED`: Enables Kafka integration for livestream and file paths
 - `LVS_CAPTION_SOURCE`: Controls caption source for file-path Kafka aggregation (`sse` default, or `db` for Elasticsearch)
+- `LVS_AGGREGATION_EMPTY_RETRIES`: Extra attempts at a CA-RAG aggregation call that returns neither events nor a video summary (default `2`, `0` disables)
 
 ### Logging
 
--- a/services/video-summarization/src/via_stream_handler.py
+++ b/services/video-summarization/src/via_stream_handler.py
@@ -3728,6 +3728,17 @@ This is very important and you must follow this strictly.
             if "params" not in ca_rag_config["functions"]["summarization"]:
                 ca_rag_config["functions"]["summarization"]["params"] = {}
 
+            # scenario and events are always forwarded so the CA-RAG aggregation
+            # LLM has context regardless of enable_vlm_structured_output.
+            if req_info.scenario:
+                ca_rag_config["functions"]["summarization"]["params"][
+                    "scenario"
+                ] = req_info.scenario
+            if req_info.events:
+                ca_rag_config["functions"]["summarization"]["params"][
+                    "events"
+                ] = req_info.events
+
             if not req_info.enable_vlm_structured_output:
                 if req_info.schema:
                     ca_rag_config["functions"]["summarization"]["params"][
@@ -3737,14 +3748,6 @@ This is very important and you must follow this strictly.
                     ca_rag_config["functions"]["summarization"]["params"][
                         "batch_response_method"
                     ] = req_info.batch_response_method
-                if req_info.scenario:
-                    ca_rag_config["functions"]["summarization"]["params"][
-                        "scenario"
-                    ] = req_info.scenario
-                if req_info.events:
-                    ca_rag_config["functions"]["summarization"]["params"][
-                        "events"
-                    ] = req_info.events
                 if req_info.auto_generate_prompt is not None:
                     ca_rag_config["functions"]["summarization"]["params"][
                         "auto_generate_prompt"
--- a/services/video-summarization/src/via_stream_handler.py
+++ b/services/video-summarization/src/via_stream_handler.py
@@ -89,6 +89,12 @@ def _safe_collection_name(stream_id) -> str:
 
 MAX_MILVUS_STRING_LEN = 65535
 
+# Extra attempts at a CA-RAG aggregation call when it returns neither events nor
+# a video summary. The aggregation LLM intermittently samples an unparseable
+# response, which surfaces to the caller as HTTP 200 with total_events=0 and
+# video_summary="". Override with LVS_AGGREGATION_EMPTY_RETRIES; 0 disables.
+DEFAULT_AGGREGATION_EMPTY_RETRIES = 2
+
 
 class RequestInfo:
     """Store information for a request"""
@@ -478,6 +484,8 @@ class ViaStreamHandler:
                 "(sse=use SSE captions for aggregation, db=retrieve from Elastic DB)",
                 self._caption_source,
             )
+
+        self._aggregation_empty_retries = self._read_aggregation_empty_retries()
 
         if not args.disable_ca_rag:
             try:
@@ -2020,7 +2028,9 @@ class ViaStreamHandler:
                 sub_state,
             )
 
-            agg_response = ctx_mgr.call({"summarization_online": sub_state})
+            agg_response = self._call_aggregation_with_empty_guard(
+                ctx_mgr, "summarization_online", sub_state, req_info.source_id
+            )
 
             if agg_response.get("error"):
                 logger.error(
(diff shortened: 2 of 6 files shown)


pshekdar-nv:
`<p><b>SUMMARY</b><br>
Symptom confirmed on matching hardware, but it is NOT an endpoint parity issue, and PR 1513 does not eliminate it.</p>

<p><b>ENVIRONMENT</b> (matched to the bug report)</p>
<ul>
  <li>Host: viking-prod-349-au, H100 80GB</li>
  <li>RTVI VLM: <code>nvcr.io/nvstaging/vss-core/vss-rt-vlm:3.2.1-26.07.2</code></li>
  <li>Model served locally on GPU: <code>MODEL_PATH=ngc:nim/nvidia/cosmos3-nano-reasoner:modelopt-fp8-final_format_fix</code>, <code>VLM_MODEL_TO_USE=cosmos-reason3</code></li>
  <li><code>VIA_VLM_ENDPOINT</code> and <code>VIA_VLM_OPENAI_MODEL_DEPLOYMENT_NAME</code> cleared, so nothing routes to an external openai-compat endpoint. <code>/models</code> confirms <code>nim_nvidia_cosmos3-nano-reasoner_modelopt-fp8-final_format_fix</code>.</li>
  <li>Payload: the one from the bug description (warehouse scenario, events forklift+crash, 1min.mp4, chunk_duration 30, offset 0-30).</li>
</ul>

<p><b>RESULTS</b> (40 calls total, 10 per endpoint per build)<br>
"Empty" = HTTP 200 with <code>total_events=0</code>, <code>events=[]</code>, and <code>video_summary=""</code> — i.e. the exact reported symptom.</p>

<table>
  <thead>
    <tr>
      <th>Build</th>
      <th>Endpoint</th>
      <th>Empty</th>
      <th>Total</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td>develop (no fix)</td>
      <td>/summarize</td>
      <td>0/10</td>
      <td rowspan="2">1/20</td>
    </tr>
    <tr>
      <td>develop (no fix)</td>
      <td>/v1/summarize</td>
      <td>1/10</td>
    </tr>
    <tr>
      <td>bug/6542210 (fix)</td>
      <td>/summarize</td>
      <td>1/10</td>
      <td rowspan="2">2/20</td>
    </tr>
    <tr>
      <td>bug/6542210 (fix)</td>
      <td>/v1/summarize</td>
      <td>1/10</td>
    </tr>
  </tbody>
</table>

<p><b>FINDING 1</b> — the symptom is real and reproducible, but only against the actual CR3 FP8 NIM on GPU.<br>
An earlier attempt on a workstation whose <code>.env</code> pointed at an external openai-compat endpoint never produced it in 20+ calls, which is why matching the reported model and hardware mattered.</p>

<p><b>FINDING 2</b> — it is NOT endpoint-specific.<br>
With the fix applied, an empty result landed on <code>/summarize</code>, not just <code>/v1/summarize</code>. Both routes are the same handler behind stacked FastAPI decorators, and across 40 calls the failures follow no endpoint. The original framing ("/summarize works, /v1/summarize does not") looks like a single-observation artifact. Recommend dropping the parity framing.</p>

<p><b>FINDING 3</b> — PR 1513 does not resolve this symptom.<br>
2 empty in 20 with the fix vs 1 in 20 without. At this sample size those rates are statistically indistinguishable, so this is not evidence the fix made anything worse — only that the empty-output failure mode clearly survives it.</p>

<p><b>ON PR 1513</b> — it is still worth keeping on its own merits. I verified deterministically from the LVS debug logs that it changes the CA-RAG config for the same request:</p>
<ul>
  <li>without fix: <code>params={'time_overlap_threshold': 0.1, 'max_events_per_batch': 50, 'kafka_enabled': 'false', 'enable_llm_merging': 'false', 'caption_source': 'sse', 'uuid': ...}</code></li>
  <li>with fix: <code>params={..., 'caption_source': 'sse', 'scenario': 'warehouse', 'events': ['forklift', 'crash'], 'uuid': ...}</code></li>
</ul>
<p>Because <code>enable_vlm_structured_output</code> defaults to True and the request never overrides it, <code>scenario</code> and <code>events</code> were previously dropped before reaching the aggregation LLM. That is a genuine defect the PR corrects — it just is not the cause of the empty output. This supports the analysis in comment #2: intermittent LLM/aggregation variability rather than a code-side endpoint defect.</p>

<p><b>RECOMMENDATION</b><br>
Re-scope this bug to "intermittent empty summarization output with CR3 FP8 (approx 5–10% of requests, both endpoints)" and investigate the CA-RAG aggregation step rather than endpoint parity.</p>

<p>Candidate directions:</p>
<ul>
  <li>The aggregation LLM (<code>nvidia/nvidia-nemotron-nano-9b-v2</code> at temperature 0.2) occasionally returning nothing parseable.</li>
  <li>Adding an empty-result guard or retry in the aggregation path so a single bad sample does not surface as a fully empty response.</li>
</ul>

<p>Happy to run a larger sample or capture the aggregation request/response on a failing run if that would help.</p>`

david-duu:
Fix is this ENV should be set to false. It should be automatically set in compose.yaml

https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/develop/services/video-summarization/docker/deploy/compose.yaml

LVS_DROP_EMPTY_EVENT_FIELDS=False

bug/6542210 without flag: 21/120 empty
With LVS_DROP_EMPTY_EVENT_FIELDS=false: 0/120 empty
Errors: 0
Empty-guard retries: 0

Root cause:
The VLM still produced 18 blank event types and 3 malformed JSON captions, but CA-RAG recovered all 21 using the stored event list. Sample recovered outputs correctly inferred forklift.

nathankw (approved):

PR #1516, opened 2026-08-02
feat(skill-eval): record what the GPUs actually do during each eval leg

simings-nv:
## Not a duplicate of the DCGM work (#1678)

Both changes are about GPU visibility, so the first question is whether one replaces the other. They don't: they instrument different planes and share no files.

| | [#1678](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1678) (DCGM) | this PR |
|---|---|---|
| Plane | product **deployment** | **CI**, the `vss-eval-*` boxes |
| Touches | `deploy/docker`, `deploy/helm`, `skills/vss-deploy-profile` | `.github/skill-eval` |
| Answers | what is a running stack doing right now | did this leg use the GPUs its spec declared |
| Shape | live exporter, operator-facing | per-leg artifact, joined to run/spec/step |

`gpu_trace.py` exists because the eval boxes have no DCGM exporter, and #1678 as scoped does not give them one. Closing this PR would not deliver the eval-fleet numbers by another route.

If DCGM does reach the eval fleet later and can be joined to job identity, that is the better mechanism and this module should be retired rather than extended. That condition is written into the module docstring so the decision is not re-litigated from memory.

---

## What this changes

Every skill-eval leg now records what the GPUs on its box actually did for the duration of each Harbor invocation, and ships it in the results artifact the workflow already uploads.

## Why

Nothing in the pipeline could answer **"this spec declares `gpu_count: 2`, did it use them?"**

There is no retrospective source for it either: `nvidia-smi` accounting mode is disabled on the eval boxes, and they carry no metrics exporter. Sampling during the run is the only way to get the number.

The comparison that matters is `declared gpu_count` versus GPUs that did any work, so the declaration travels in the sidecar next to the measurement. Publishing a utilisation curve on its own would argue for the one conclusion that is hardest to reverse.

## Design

| Decision | Why |
|---|---|
| Hook in `run_leg.py` | The only point where GPU identity and job identity coexist: it holds the per-box `flock` and knows run/spec/step. Shared by `skills-eval.yml` and `skills-eval-daily.yml`, so one hook covers both. |
| Rides the existing artifact | No new service, no new network path, no credential on the eval box. Survives `_reset_docker_runtime` wiping the box between trials. |
| Default on, non-blocking | Opt-in telemetry that nobody enables collects nothing. Every path swallows its own exceptions and the context manager always yields. |
| Hard-bounded remote lifetime | The sampler runs under `timeout -k`, so it dies on its own even if this process is SIGKILLed and never calls stop. |
| Allowlist, not redaction | Seven `--query-gpu` fields, structurally validated per row against an anchored regex. Redaction fails open; an allowlist fails closed. |

Also archives the tarball when a GPU trace exists but `result.json` does not: a trial that hits the agent ceiling dies before Harbor writes one, and that is precisely the trace worth reading.

### A design alternative that was ruled out with evidence

Writing into `/logs/artifacts` would let harbor collect the trace with no fetch path at all. It does not work: `envs/brev_env.py:235` wipes that directory during harbor startup — *after* `run_leg.py` has already started the sampler — so writes continue into an unlinked inode; and the runs that matter most, `rc=124` and cancellation, never reach `_download_artifacts` at all.

## What review found

Three adversarial rounds. The product code held up better than the suite.

**Round 1** — three blockers, all reproduced:
- `mkdir && nohup ... &` backgrounds the whole AND-list, so `$!` was a forked subshell rather than `timeout`, killing it left the sampler running, and the subshell inherited stdout so the start call blocked until the *sampler* exited. Measured 3.03s under bash, 3.04s under sh, **0.01s under zsh** — invisible on a laptop, fatal over SSH.
- The recorded PID was remote text interpolated unquoted into a shell command.
- `ln.count(",") >= 6` was an allowlist in name only.

**Round 2** — four more, two of them living *inside* round 1's repair:
- A predictable `/tmp` path is a file-clobbering primitive; shell `>` follows symlinks. Reproduced by truncating a victim file.
- `MAX_FETCH_BYTES` was applied after `communicate()` had already buffered everything: a 32 MiB emit grew RSS by 125 MB against an 8 MiB "cap".
- `gpu_name` took every non-PID line of the start output and published it in the sidecar.
- PID syntax is not PID identity.

**Round 3** — 29 mutations passed all 37 tests, including two that need a single character:
- Deleting `DIR_RE`'s trailing anchor turns a remote reply into arbitrary command execution, because the value is interpolated into `rm -rf {dir}`. `DIR_RE` appeared **zero times** in the test file, while `PID_RE` — guarding an identical interpolation — had two dedicated tests.
- Deleting `ROW_RE`'s trailing anchor puts a private key in an uploaded artifact. Every existing negative case failed on the *prefix*, so a well-formed row carrying a garbage *suffix* was never tried.
- `if not rows:` → `if rows is None:` publishes "declared 2 GPUs, 0 samples over 70 minutes" as fact.
- Deleting `--query-gpu=` entirely, appending `gpu_uuid,gpu_serial`, adding `-i 0`, or changing `--format` all passed, because the `nvidia-smi` fake ignored argv.
- Removing `-k` removed the SIGKILL escalation, because the `timeout` fake parsed the duration off argv and threw it away.

## Product fixes since the first draft

- **SIGTERM now unwinds.** Python converts only SIGINT into an exception, so SIGTERM skipped every `finally` — and `skills-eval.yml` sets `cancel-in-progress: true`, so every push to a PR SIGTERMs up to eight in-flight legs at once. Each leaked a sampler for up to 139 minutes onto a box handed straight to the next leg, and lost the partial trace. Measured before and after.
- **`start_cmd` is single-attempt.** It is not idempotent: a blind retry starts a second sampler whose pid is never recorded, so nothing can kill it.
- **`nvidia-smi` writes via `-f`** rather than a shell redirect. A redirected stdout is block-buffered and SIGTERM does not flush it, so minutes of samples were discarded at exactly the moment cleanup ran. This also puts the nonce directory in argv, which makes the identity check exact rather than "is this some `nvidia-smi`" on a box that runs them constantly.
- **The kill is decoupled from the fetch guard.** A reply where `PID=` validated and `DIR=` did not left a running sampler nothing would kill, logged as a benign non-start.
- **Orphaned `/tmp` directories are reaped.** Nothing else on the box does.

## Tests

51 tests, no external dependencies, stable across repeated runs.

The fakes now implement their contracts rather than the caller's assumptions: `nvidia-smi` builds rows **from the parsed field list** and honours `--format`, `-i`, `-f` and `-l`, so a drift between the command and `CSV_HEADER` changes the output instead of being silent; `timeout` runs a real timer, escalates to SIGKILL, returns 124, and refuses a missing `-k`.

Thirteen mutations confirmed to fail the suite, including all six of the round-3 findings above.

```
cd .github/skill-eval && PYTHONPATH=. python3 -m unittest discover -s tests -p 'test_gpu_trace.py'
```

Note: `test_run_leg.py` has 3 pre-existing failures in `_rtx4090_supports`, unrelated to this change and present on `develop`.

## Kill switch

`EVAL_GPU_TRACE=0`.

### DIFF
--- a/.github/skill-eval/gpu_trace.py
+++ b/.github/skill-eval/gpu_trace.py
@@ -401,8 +401,12 @@ def _finish(
         # but it is the difference between "some gputrace sampler" and "any
         # process on the box". Both samplers are bounded by their own timeout,
         # so the weaker match costs at most a slightly early stop.
+        # `-ww` for the reason spelled out at the fetch-path guard below, and
+        # this site is worse off: `vss-gputrace` occurs in the sampler's argv
+        # only inside `-f "$d/trace.csv"`, which is the LAST thing on a 204-char
+        # command line, so any width limit at all discards it.
         _remote(instance,
-                f"ps -o args= -p {pid} 2>/dev/null | grep -qF vss-gputrace "
+                f"ps -ww -o args= -p {pid} 2>/dev/null | grep -qF vss-gputrace "
                 f"&& kill {pid} 2>/dev/null; true", attempts=1)
         print(f"[gpu-trace] tried to stop sampler {pid} on {instance} "
               f"but have no trace path", flush=True)
@@ -423,7 +427,25 @@ def _finish(
     # -F, not a bare pattern: DIR_RE permits `.`, which is a BRE wildcard, so
     # `grep -q /tmp/vss-gputrace.ABCD1234` also matches `/tmp/vss-gputraceXABCD1234`.
     # A fixed-string match makes the nonce mean what it looks like it means.
-    kill = (f"ps -o args= -p {pid} 2>/dev/null | grep -qF -- {remote_dir} "
+    #
+    # `-ww`, and it is load-bearing rather than tidy. `ps` truncates `args` to
+    # the display width, and the sampler's command line is 204 characters with
+    # the nonce starting at column 169. Any width under that returns a prefix
+    # that CANNOT contain the nonce, so `grep` reports no match, `&&` skips the
+    # kill, and the sampler runs to its 2h20m hard stop on a box that has already
+    # been handed to the next leg -- a guard that is present, correct-looking and
+    # inert. Two real ways to land under 169 columns, both measured on
+    # procps-ng 3.3.17:
+    #   * COLUMNS is set in the environment (COLUMNS=80 and =100 both discard it);
+    #   * stdout is a pty whose window size was never set, which defaults to
+    #     exactly 80x24 -- and `brev exec`/`ssh` allocate precisely that.
+    # Note it is NOT simply "no tty": with no tty anywhere and COLUMNS unset,
+    # procps prints unlimited width, which is why this can look green locally.
+    # `-ww` removes the limit unconditionally and is immune to all of them.
+    # A single `-w` is NOT enough and is the tempting half-fix: it widens to 132
+    # columns, still short of 169, so the guard stays inert while looking
+    # repaired. Measured at COLUMNS=80: bare=80 chars, -w=132, -ww=207.
+    kill = (f"ps -ww -o args= -p {pid} 2>/dev/null | grep -qF -- {remote_dir} "
             f"&& kill {pid} 2>/dev/null; ") if pid else ""
     # `head -c` bounds the remote side; _read_capped bounds ours. Both are
     # needed: the remote cap cannot bound brev's own stdout.
--- a/.github/skill-eval/gpu_trace.py
+++ b/.github/skill-eval/gpu_trace.py
@@ -164,7 +164,8 @@ def _read_capped(proc: subprocess.Popen, deadline_sec: float) -> str:
     return "".join(chunks)
 
 
-def _remote(instance: str, command: str, timeout: int = EXEC_TIMEOUT_SEC) -> str | None:
+def _remote(instance: str, command: str, timeout: int = EXEC_TIMEOUT_SEC,
+            attempts: int = 2) -> str | None:
     """Run `command` on the box. Returns stdout, or None on any failure.
 
     Tries `brev exec` first (managed instances), then `ssh <alias>` (registered
@@ -180,13 +181,13 @@ def _remote(instance: str, command: str, timeout: int = EXEC_TIMEOUT_SEC) -> str
     an ssh that forked would survive; `envs/brev_env.py::_kill_proc_group`
     already fixes this exact problem the same way for harbor.
     """
-    attempts = (
+    candidates = (
         ["brev", "exec", instance, command],
         ["ssh", "-o", "BatchMode=yes", "-o", "ConnectTimeout=15",
          "-o", "StrictHostKeyChecking=no", instance.lower(), command],
-    )
+    )[:max(1, attempts)]
     deadline = time.monotonic() + timeout
-    for cmd in attempts:
+    for cmd in candidates:
         remaining = deadline - time.monotonic()
         if remaining <= 1:
             break
@@ -300,15 +301,33 @@ def trace(
     # `set -C` (noclobber) is belt-and-braces inside that fresh directory: it
     # refuses to write through an existing name at all.
     start_cmd = (
+        # Reap orphans first. Nothing else on the box does: _reset_docker_runtime
+        # touches only containers and volumes, the stray-agent reaper matches only
+        # claude, and the /logs wipe does not reach /tmp. Every abrupt-death path
+        # leaves a directory behind, and these boxes stay up for weeks. 180 min is
+        # safely past the hard stop, so this can never hit a live sampler.
+        f"find /tmp -maxdepth 1 -name 'vss-gputrace.*' -mmin +180 "
+        f"-exec rm -rf {{}} + 2>/dev/null; "
         f"set -C; d=$(mktemp -d /tmp/vss-gputrace.XXXXXXXX) || exit 0; "
         f"echo DIR=$d; "
+        # `-f FILE` rather than `> FILE`. A redirected stdout is block-buffered
+        # and SIGTERM does not flush it, so up to a full buffer of samples is
+        # discarded at exactly the moment cleanup runs -- on a 2-GPU box at 10s
+        # that is minutes of data, and a short leg would return nothing at all,
+        # logged indistinguishably from an unreachable box. Letting nvidia-smi
+        # own the file also puts the nonce directory into its argv, which is
+        # what makes the identity check below exact rather than a guess.
         f"nohup timeout -k 10 {hard_stop} nvidia-smi "
         f"--query-gpu={QUERY_FIELDS} --format=csv,noheader,nounits "
-        f"-l {INTERVAL_SEC} >\"$d/trace.csv\" 2>/dev/null </dev/null & "
+        f"-l {INTERVAL_SEC} -f \"$d/trace.csv\" >/dev/null 2>&1 </dev/null & "
         f"echo PID=$!"
     )
     try:
-        out = _remote(instance, start_cmd)
+        # Single attempt. `start_cmd` is NOT idempotent: a second run creates a
+        # second mktemp directory and a second sampler whose pid we never see,
+        # so nothing can kill it and it lives to the 140-minute hard stop. The
+        # fetch below is safely re-runnable and keeps both attempts.
+        out = _remote(instance, start_cmd, attempts=1)
         # Only two prefixes are read, and both are validated. There is
         # deliberately no free-form field: an earlier revision assigned every
         # other line to `gpu_name` and published it in the sidecar, so anything
@@ -361,6 +380,14 @@ def _finish(
     # `timeout` forwards SIGTERM to nvidia-smi, so killing its pid is enough.
     # The `rm` matters: `_reset_docker_runtime` does not clear /tmp, and an
     # unbounded pile of traces on a box that has been up six weeks is litter.
+    # The kill and the fetch were coupled to one guard, but only the fetch needs
+    # the directory. A reply where PID= validated and DIR= did not left a running
+    # sampler that nothing would ever kill, reported as a benign non-start.
+    if pid and not remote_dir:
+        _remote(instance, f"kill {pid} 2>/dev/null; true", attempts=1)
+        print(f"[gpu-trace] killed sampler {pid} on {instance} but have no trace path",
+              flush=True)
+        return
     if not remote_dir or not remote_csv:
         print(f"[gpu-trace] no remote trace directory for {instance}", flush=True)
         return
@@ -369,7 +396,12 @@ def _finish(
     # reuses it, and `kill <number>` then hits an unrelated process on a shared
     # box. Verified: handing back the pid of an unrelated `sleep 60` terminated
     # it. So the pid must still LOOK like our sampler before it is signalled.
-    kill = (f"ps -o args= -p {pid} 2>/dev/null | grep -q nvidia-smi "
+    # Match this invocation's nonce directory, not merely "some nvidia-smi".
+    # These boxes run nvidia-smi constantly -- deploy scripts, health checks,
+    # containers -- so a recycled pid that happens to be any nvidia-smi would be
+    # killed, possibly one belonging to another leg. The nonce is in argv now
+    # that the sampler uses `-f`, which makes this exact.
+    kill = (f"ps -o args= -p {pid} 2>/dev/null | grep -q {remote_dir} "
             f"&& kill {pid} 2>/dev/null; ") if pid else ""
     # `head -c` bounds the remote side; _read_capped bounds ours. Both are
     # needed: the remote cap cannot bound brev's own stdout.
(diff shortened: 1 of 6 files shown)


simings-nv:
## Addressed in 3890d1db6

### The 4/5 blocker: PID-only cleanup (`gpu_trace.py`)

Confirmed and fixed. `_finish`'s `pid and not remote_dir` branch ran a bare `kill <pid>`. The code eight lines below it already documented why that is wrong, with a reproduction — so this was an inconsistency, not a disagreement. The branch was *added* in the previous round to fix a different leak (a reply where `PID=` validated and `DIR=` did not left a sampler nothing would kill), and the nonce guard did not travel with it.

It cannot use the nonce, because **the nonce is the directory that failed to validate**. So it now matches the `vss-gputrace` prefix the sampler carries in its own argv via `-f "$d/trace.csv"`:

```
ps -o args= -p <pid> | grep -qF vss-gputrace && kill <pid>
```

Weaker than the exact nonce — it cannot tell this leg's sampler from another leg's — but both are bounded by their own `timeout`, so the worst case is a slightly early stop instead of signalling an arbitrary process.

### The missing coverage you flagged on the test file

Also fixed: `test_a_valid_pid_with_a_rejected_directory_still_checks_identity` drives that exact branch and asserts both that an identity check exists and that it **gates** the kill via `&&`. Mutation-checked — restoring the bare `kill <pid>` fails it.

### Two things found alongside it

- **`grep -q` → `grep -qF`** on the main guard. `DIR_RE` permits `.`, a BRE wildcard, so `grep -q /tmp/vss-gputrace.ABCD1234` also matched `/tmp/vss-gputraceXABCD1234`.
- **`_token` chain collision.** `chain` sat inside the 100-char truncation while `step` was deliberately placed outside it. `_token(r, "x"*100, p, 1, "chainA") == _token(r, "x"*100, p, 1, "chainB")` — the same silent overwrite chain support was added to prevent. Both discriminators are now outside the truncation.

Tests 51 → 54, each new one mutation-checked before being trusted. One pre-existing assertion pinned the literal string `grep -q `, so tightening to `grep -qF` broke it; it now asserts the property rather than the spelling.

.github/skill-eval/gpu_trace.py
```diff
@@ -0,0 +1,478 @@
...
+    # it. So the pid must still LOOK like our sampler before it is signalled.
+    # Match this invocation's nonce directory, not merely "some nvidia-smi".
+    # These boxes run nvidia-smi constantly -- deploy scripts, health checks,
+    # containers -- so a recycled pid that happens to be any nvidia-smi would be
+    # killed, possibly one belonging to another leg. The nonce is in argv now
+    # that the sampler uses `-f`, which makes this exact.
+    # -F, not a bare pattern: DIR_RE permits `.`, which is a BRE wildcard, so
+    # `grep -q /tmp/vss-gputrace.ABCD1234` also matches `/tmp/vss-gputraceXABCD1234`.
+    # A fixed-string match makes the nonce mean what it looks like it means.
+    kill = (f"ps -o args= -p {pid} 2>/dev/null | grep -qF -- {remote_dir} "
```
zac-wang-nv, line 426: **The identity guard never matches, so the `kill` behind it never runs.** The sampler is not stopped by cleanup at all -- it survives until the remote `timeout -k 10 8400` hard stop, 2h20m later, on a box that has already been released to the next leg.

### Mechanism

`ps -o args=` truncates its output to the terminal width, and with no tty -- which is the case under `brev exec` -- that falls back to **80 columns**. The sampler's command line is 204 characters and `remote_dir` starts at **column 168**:

```
timeout -k 10 8400 nvidia-smi --query-gpu=timestamp,index,utilization.gpu,utiliz   <- all ps returns
                                       ation.memory,...,-l 10 -f /tmp/vss-gputrace.AbCdEfGh/trace.csv
                                                                 ^^^^^^^^^^^^^^^^ what grep needs, discarded
```

So `grep -qF` searches 80 characters that cannot contain the nonce, reports failure, and `&&` skips the kill. Captured from inside this PR's own test harness:

```
ps stdout: 'python3 .../timeout -k 10 8400 nvidia-smi --query-gpu=timestamp'
nonce:     /tmp/vss-gputrace.fqIOPUeq
nonce in ps stdout: False
guard says: NOMATCH
```

The pid itself is fine: issuing a bare `kill <same pid>` at that moment took the sampler down within 1s. The guard is the only thing preventing it.

### This PR's own test catches it, and is currently failing

`test_the_sampler_is_actually_stopped` fails on the **liveness** assertion, not a string check:

```
AssertionError: 2 != 0 : sampler survived the cleanup
```

I polled the survivors: alive at t+1s, +2s, +3s, +5s, gone by +8s -- which is `SMI_RUN_SEC`, the fake sampler's own expiry. They are not being killed, they are timing out. 3/3 runs, same shape.

The reason this has looked green is that `ps` keys on terminal width, so the outcome depends on the window of whoever runs the suite:

```
COLUMNS=400                        -> 1 passed
COLUMNS unset (CI / pipe default)  -> 1 failed
COLUMNS=80                         -> 1 failed
```

### Fix

`-ww` removes the width limit:

```diff
-    kill = (f"ps -o args= -p {pid} 2>/dev/null | grep -qF -- {remote_dir} "
+    kill = (f"ps -ww -o args= -p {pid} 2>/dev/null | grep -qF -- {remote_dir} "
```

I applied exactly that and re-ran with `COLUMNS` unset: `assertEqual(alive_after, 0)` passes, the sampler dies. Two tests then fail on a different assertion because they pin the literal string `ps -o args= -p` -- that is the same over-specification that previously broke when `grep -q ` was tightened to `grep -qF`, and it wants the same treatment: assert that an identity check happens, not the exact flags.

Worth a line in the mutation list too, since this failure mode is invisible to every text-shaped assertion: a guard that is present, correct-looking, and inert.

zac-wang-nv (commented):
Two findings on the sampler cleanup path in `gpu_trace.py`. Same root cause and the same one-token fix, but at two call sites, so I've flagged them separately.

The short version: `ps -o args=` truncates at 80 columns when there is no tty, the nonce it greps for sits at column ~168, so both identity guards always fail and neither `kill` ever fires. This PR's own `test_the_sampler_is_actually_stopped` catches it and is failing; it passes only in a wide terminal.

.github/skill-eval/gpu_trace.py
```diff
@@ -0,0 +1,478 @@
...
+        # signals whatever inherited it. Verified there with an unrelated
+        # `sleep 60`; nothing about this branch makes that safer.
+        # The literal `vss-gputrace` prefix is in the sampler's argv (it owns
+        # the file via `-f "$d/trace.csv"`), so match on that: weaker than the
+        # exact nonce -- it cannot tell this leg's sampler from another's --
+        # but it is the difference between "some gputrace sampler" and "any
+        # process on the box". Both samplers are bounded by their own timeout,
+        # so the weaker match costs at most a slightly early stop.
+        _remote(instance,
+                f"ps -o args= -p {pid} 2>/dev/null | grep -qF vss-gputrace "
```
zac-wang-nv, line 405: Same defect as the guard at line 426, and this branch is worse off.

`vss-gputrace` appears in the sampler's argv only inside `-f "$d/trace.csv"`, which is at the very end of the command line -- so it is past the same 80-column cut, and this guard is inert too.

The consequence is specific to this branch: it is the path taken when `PID=` validated but `DIR=` did not, and the comment above it is right that it exists so a sampler is not left running with nothing to kill it. As written it does not kill it either, then returns having logged `tried to stop sampler ... but have no trace path`, which reads as though the attempt was made.

Same one-token fix (`ps -ww`). Flagging separately from line 426 so both call sites get it.

.github/skill-eval/gpu_trace.py
simings-nv: Thanks Zac, good catch. Confirmed at both sites and fixed with `ps -ww` in e4e36e01.

One correction on the mechanism, and it makes the bug slightly worse than described. It is not the absence of a tty. With no tty anywhere and `COLUMNS` unset, procps-ng 3.3.17 prints unlimited width and the guard matches, which is why the test passed for me. The real trigger is any width under 169, where the nonce starts:

```
COLUMNS unset (no tty)  -> ps len 207   MATCH     OK
COLUMNS=80              -> ps len 80    NOMATCH   2 != 0 : sampler survived the cleanup
COLUMNS=100             -> ps len 100   NOMATCH   2 != 0
COLUMNS=400             -> ps len 207   MATCH     OK
```

The one that reaches production is a pty whose window size was never set, which defaults to exactly 80x24, and that is what `brev exec` and `ssh` allocate. Verified by running `ps` under `pty.spawn` with a default winsize: len 80, nonce absent. So the guard was inert on the normal remote path, not only in CI.

Worth flagging that `-w` alone is not enough, since it is the cheaper looking repair. At `COLUMNS=80`: bare 80 chars, `-w` 132, `-ww` 207. 132 is still short of column 169, so it stays inert while looking fixed.

On the tests, you are right that the string assertions were the wrong instrument. Both `ps -o args= -p` pins now assert the property instead. That is the second time an over specified string there rejected a strictly stronger version of what it was asserting, after `grep -q ` vs `grep -qF`, so I treated it as a pattern rather than a one off.

The more useful fix is that `Stub` now pins `COLUMNS=80`. The suite's result was a function of whoever's terminal ran it, which is how a guard this broken stayed green. Pinning makes the existing assertions deterministic and hostile at the same time, matching what a fresh pty hands the remote command.

Added `test_the_guard_still_fires_at_every_plausible_terminal_width`, which asserts the sampler is dead at 80/100/160/400/unset. Behaviour rather than text, for exactly the reason you gave. Three mutations confirmed to fail it: dropping `-ww` from either site, and the `-w` half fix. Added to the mutation list.

Also fixed a flake I hit on the first run: the sidecar test forbade the substring `"10."`, which also matches its own unix timestamps (`"finished_at": 1786557910.4098206`). Roughly one run in fifty. It now matches a dotted quad, which is what it meant.

Rebased onto `develop` to clear the conflict. 55 tests, green at `COLUMNS` unset, 80 and 400.

.github/skill-eval/gpu_trace.py
simings-nv: Fixed here too in e4e36e01, same `ps -ww`. You were right that this site is worse off: `vss-gputrace` sits inside `-f "$d/trace.csv"` at the very end of the command line, so any width limit at all discards it.

Details and the measurements are in the reply on the other thread, including a correction to the mechanism. Short version is that it is not the missing tty, it is a default sized pty at 80 columns, which is what `brev exec` and `ssh` give us.

Also took your point about the misleading log line. That branch no longer reports a stop attempt it did not make.

simings-nv: Correction to my last line: I did not touch the log string. What changed is that the kill now actually fires, so `tried to stop sampler ...` is true where before it was not. The wording is accurate now by behaviour, not by edit. Sorry for the sloppy phrasing.

If you would still rather it named the outcome instead of the attempt, say so and I will make it report whether the signal landed.

.github/skill-eval/gpu_trace.py
```diff
@@ -0,0 +1,500 @@
...
+    #   * stdout is a pty whose window size was never set, which defaults to
+    #     exactly 80x24 -- and `brev exec`/`ssh` allocate precisely that.
+    # Note it is NOT simply "no tty": with no tty anywhere and COLUMNS unset,
+    # procps prints unlimited width, which is why this can look green locally.
+    # `-ww` removes the limit unconditionally and is immune to all of them.
+    # A single `-w` is NOT enough and is the tempting half-fix: it widens to 132
+    # columns, still short of 169, so the guard stays inert while looking
+    # repaired. Measured at COLUMNS=80: bare=80 chars, -w=132, -ww=207.
+    kill = (f"ps -ww -o args= -p {pid} 2>/dev/null | grep -qF -- {remote_dir} "
+            f"&& kill {pid} 2>/dev/null; ") if pid else ""
```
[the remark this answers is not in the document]
simings-nv: The race is real in principle. Two measurements say the current form is the safer of the options available here.

`pid_max` on these boxes is 4194304, so reuse inside a single shell pipeline needs four million allocations between the `ps` read and the `kill`.

The alternative that removes the pid from the equation entirely, `pkill -f <nonce>`, is worse. The cleanup command carries the nonce in its own argv, so it matches itself:

```
1481484  SAMPLER          timeout -k 10 8400 nvidia-smi ... -f /tmp/vss-gputrace.PROBE001/trace.csv
1481553  NOT-THE-SAMPLER  bash -c ... /tmp/vss-gputrace.PROBE001 ...
```

Under `brev exec` that second line is the shell running the cleanup, so `pkill -f` would terminate its own command shell. That trades an implausible race for a reproducible bug.

The failure is also asymmetric. A missed kill self-resolves, because the sampler runs under `timeout -k` and dies on its own at the hard stop. A wrong kill does not self-resolve. The current form fails in the direction that costs least, so keeping it.

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32384848714)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32385978070)

PR #1524, opened 2026-08-03
Fix (RTVI-CV): make DeepStream runtime paths writable after STORAGE_UID drop

prisrivastav-nv:
### Issue:

**1. Tracker ReID (alerts `2d_cv` / Smart Cities)** 

On the alerts profile in verification mode (MODE=2d_cv), vss-rtvi-cv failed during NvDCF tracker startup with:
`NvDsInferCudaEngineGetFromTltModel: Failed to open TLT encoded model file …/deepstream-9.1/samples/models/Tracker/resnet50_market1501.etlt`

```
Server running at port: 9010
gstnvtracker: Loading low-level lib at /opt/nvidia/deepstream/deepstream/lib/libnvds_nvmultiobjecttracker.so
~~ CLOG[src/modules/FrameTransform/trtModule.cpp, loadTRTEngine() @line 305]: Engine file does not exist: /opt/nvidia/deepstream/deepstream/samples/models/Tracker/resnet50_market1501.etlt_b100_gpu0_fp16.engine
[NvMultiObjectTracker] Load engine failed. Create engine again.
uri:/api/v1/stream/add
method:GET
NvDsInferCudaEngineGetFromTltModel: Failed to open TLT encoded model file /opt/nvidia/deepstream/deepstream-9.1/samples/models/Tracker/resnet50_market1501.etlt
free(): double free detected in tcache 2
```

**2. HTTP file upload / `/stream/add` (HTTP ingest)**
Camera/file ingest via `camera_url: http://…` failed with `STREAM_ADD_FAIL, Failed to download HTTP/HTTPS file`. RTSP ingest and embedding still worked; only the CV path that downloads into DeepStream’s streams dir failed.

### Root cause
RT-CV starts as root (Option A / VIA-2430), then drops to `STORAGE_UID:STORAGE_GID` (default `1001:1001`) before `metropolis_perception_app`.
Several image-owned DeepStream paths stay owned by other UIDs (e.g. `triton-server` / 1000) with mode `0755`, so UID 1001 cannot write there. `HOME` also remains `/root` after `setpriv`, so GStreamer/dconf cache writes fail.
- **ReID:** etlt was staged with a plain `cp` into one Tracker path and never followed the model-tree ownership contract used for `/opt/storage`.
- **HTTP download:** plugin uses hardcoded `HTTP_DOWNLOAD_DIR=…/samples/streams/`; that directory was unwritable by 1001.
`STORAGE_UID=1000` is not a fix — it would break host bind-mount ownership for `/opt/storage` (and is only a coincidence with this image’s `triton-server` UID).

### Fix
**Docker Compose and Helm** (`ds-start.sh` both copies):
1. **ReID:** `setup-tracker-reid.sh` installs `resnet50_market1501.etlt` with explicit permissions into every DeepStream Tracker dir the runtime may resolve; owns those dirs as `STORAGE_UID`; caches/reuses the ReID TensorRT engine under `/opt/engines`.
2. **Class-wide (before privilege drop):** `prepare_runtime_user_environment`:
   - Redirects `HOME` / `XDG_CACHE_HOME` / `GST_REGISTRY` to a writable runtime home (`/tmp/rtvi-cv-home`, overridable via `RTVI_CV_RUNTIME_HOME`)
   - Non-recursive prepare + chown of `samples/streams`, `samples/models/Tracker`, and `DS_APP_DIR` under both `deepstream/` and `deepstream-*/`
### Scope
- **ReID path:** alerts `2d_cv`, Smart Cities (`rtdetr-gdino`)
- **Streams / HOME prepare:** all RT-CV families (search, alerts, warehouse) that drop to `STORAGE_UID`
- **Not a UID change:** default remains `1001`

### Verified
- ReID: etlt staged readable; engine built/written as UID 1001; tracker init succeeds
- Streams: after owning `samples/streams` as 1001, `/stream/add` with HTTP URL returns `STREAM_ADD_SUCCESS` and the file lands owned by 1001

## Checklist


kaushikc-nvidia:
/ok to test [4e1793c](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1524/commits/4e1793cd3f70f04f982a39bb93fca05e29991b48)

deploy/docker/services/rtvi/rtvi-cv/setup-tracker-reid.sh
```diff
@@ -0,0 +1,192 @@
...
+
+install_etlt_to_tracker_dirs() {
+    local src="$1"
+    local tdir dest
+
+    collect_tracker_dirs
+    for tdir in "${TRACKER_DIRS[@]}"; do
+        dest="${tdir}/${REID_BASENAME}"
+        install -d -m 0755 "$tdir"
+        if [[ -e "$dest" && ! -L "$dest" ]]; then
```
zac-wang-nv: **TODO 2 (small — the one I'd suggest folding into this PR):** this early-out skips the `install -m 0644` that is the point of the change.

To be clear about severity — I do **not** think this blocks the fix. Per [`troubleshooting.md:67`](skills/vss-deploy-detection-tracking-2d/references/troubleshooting.md#L67) the etlt ships deeper in the sources tree and is *absent* from `Tracker/`, so on a fresh container `-e "$dest"` is false, the branch isn't taken, and the permission fix applies. This is hardening for the paths where the file is already there:

- A container restarted in place after having run the old inline `cp` — the destination was populated by code that set no explicit mode, so whatever `cp` produced (source mode masked by umask) is preserved rather than normalised to `0644`.
- Any future DeepStream image that *does* ship an etlt at that path with a restrictive mode — the early-out would silently keep it, and the failure would look identical to the one this PR fixes.

It's also a behaviour change from the code being replaced: the old `cp` overwrote unconditionally, so a stale or wrong-mode etlt at the destination was always corrected; now it's preserved.

`install` is idempotent and cheap, so the simplest version is to make it unconditional and drop the branch (keeping `--symlink` handling if you want it):

```bash
install -m 0644 "$src" "$dest"
echo "TRACKER_REID: INSTALLED  $src  ->  $dest"
```

Same line in the Helm copy — the two files are byte-identical, so both need it.

deploy/docker/services/rtvi/rtvi-cv/ds-start.sh
```diff
@@ -93,10 +93,75 @@ ensure_models_from_manifest() {
...
+        return 0
+    fi
+
+    chown -R "${uid}:${gid}" "${runtime_home}"
+
+    # Non-recursive: streams accumulates downloaded videos; only the dir must
+    # be writable so fopen(..., "wb") for HTTP_DOWNLOAD_DIR succeeds.
+    append_deepstream_sample_dirs streams runtime_dirs
+    append_deepstream_sample_dirs models/Tracker runtime_dirs
+    runtime_dirs+=("${DS_APP_DIR}")
```
zac-wang-nv, line 151: **TODO 3 (follow-up, pre-existing — explicitly *not* a regression from this PR):** `/opt/engines` isn't in this enumeration, and it's the path the rest of `ds-start.sh` most clearly expects the *dropped* process to write.

160 lines below, `start_rtdetr_gdino()` does:

```bash
ENGINES_DIR="/opt/engines"
mkdir -p "${ENGINES_DIR}/gdino" "${ENGINES_DIR}/rtdetr-its"     # root:root, default umask
...
sed -i "...s|^model-engine-file=.*|model-engine-file=${ENGINES_DIR}/rtdetr-its/model_epoch_035.fp16.onnx_b${NUM_SENSORS}_gpu0_fp16.engine|..."
```

`/opt/engines` is a host bind mount in both affected profiles (`$VSS_APPS_DIR/engines/:/opt/engines/` — [`dev-profile-alerts/compose.yml:93`](deploy/docker/developer-profiles/dev-profile-alerts/compose.yml#L93), [`smartcities/compose.yml:84`](deploy/docker/industry-profiles/smartcities/compose.yml#L84)), so the intent is plainly "persist the RT-DETR engine across restarts". But the dirs are made by root and nothing chowns them, so post-`setpriv` the app is UID 1001 writing into a `root:root 0755` tree. Either way it goes, the persisted-engine design doesn't hold:

- If nvinfer serializes to the configured `model-engine-file` path → the write fails and the engine rebuilds every start.
- If nvinfer instead serializes next to the ONNX (`/opt/storage/rtdetr-its/`, which `download-models.sh` *does* chown to `STORAGE_UID`) → the write succeeds but lands outside the bind mount, so the sed'd `model-engine-file` never matches next start and it rebuilds anyway.

The GDINO branch is unaffected because `trtexec` builds `model_gdino_trt.plan` as root pre-drop — which is exactly why the asymmetry is easy to miss.

Both `ENGINES_DIR` and the privilege drop predate this PR, so nothing here regresses. Flagging it because it's the same class of problem as this commit's title, and the function's own comment already names "Tracker engine dir" as a target — the detector engine dir is its sibling. One line if you want it now:

```bash
[[ -n "${ENGINES_DIR:-}" ]] && runtime_dirs+=("${ENGINES_DIR}" "${ENGINES_DIR}/gdino" "${ENGINES_DIR}/rtdetr-its")
```

The `:-` guard matters: `ENGINES_DIR` is only assigned on the `rtdetr-gdino` path and this function runs for all three model families under `set -u`.

If it was left out deliberately — e.g. you already know nvinfer writes next to the ONNX — a one-line comment saying so would save the next reader the trip.

deploy/docker/services/rtvi/rtvi-cv/setup-tracker-reid.sh
```diff
@@ -0,0 +1,192 @@
...
+# Used by ds-start.sh for rtdetr-gdino profiles (alerts, smartcities). Derived
+# from skills/vss-deploy-detection-tracking-2d/scripts/setup_tracker_reid.sh with
+# multi-path etlt install and explicit permissions for the privilege-dropped app
+# user (STORAGE_UID/STORAGE_GID).
+
+set -euo pipefail
+
+REID_BASENAME="resnet50_market1501.etlt"
+DEST_DIR="/opt/nvidia/deepstream/deepstream/samples/models/Tracker"
+DEST="$DEST_DIR/$REID_BASENAME"
```
zac-wang-nv, line 18: **TODO 4 (nits, lowest priority — batch these whenever TODO 1 is picked up):** dead code carried over from the reference script.

- `DEST` is never read. `shellcheck --severity=warning` flags exactly this and nothing else in the file (`SC2034`); the multi-dir rewrite replaced it with the per-`$tdir` `dest` local.
- `--symlink` (L26) and `--wait` (L28) are never passed by either call site.
- `-h|--help` (L30) runs `sed -n '1,25p' "$0"`, but lines 1–25 here are the header comment, `set -euo pipefail`, the variable block and a dangling `while`/`case` — it prints code, not usage. The reference used `sed -n '18,59p'` against a real Usage/exit-code block that this file's header dropped while keeping the `sed`.
- This PR now ships two implementations of "find every DeepStream `Tracker/` dir": `collect_tracker_dirs` here (L44) and `append_deepstream_sample_dirs models/Tracker` in `ds-start.sh` — four copies once you count Docker + Helm. The newer one guards against both `readlink -f` results being empty (with a comment explaining why); `collect_tracker_dirs` compares them directly, so two empty results read as "same dir" and the versioned trees get skipped. Benign today (`DEST_DIR` is always `TRACKER_DIRS[0]` and resolves through the `deepstream` symlink anyway), but the two probably shouldn't disagree inside one PR.

If TODO 1 goes the "drop the cache machinery" route, most of this disappears with it.

zac-wang-nv (commented):
Reviewed against `fed78dfc9`. **The crash fix is sound and I'm not blocking on any of the below** — the primary goal (unblock the `NvDsInferCudaEngineGetFromTltModel: Failed to open TLT encoded model file` failure on alerts / smartcities) is met, and I verified the wiring reaches every affected deployment path. Everything I raise here is a follow-up **TODO**.

Why I'm confident the crash itself is fixed: the in-tree troubleshooting note ([`troubleshooting.md:67`](skills/vss-deploy-detection-tracking-2d/references/troubleshooting.md#L67)) documents the root cause as *"the tracker config references `/opt/nvidia/deepstream/deepstream/samples/models/Tracker/resnet50_market1501.etlt` but the etlt ships deeper in the perception-app sources tree"*. So on a fresh container the destination is empty, the new helper's `install -m 0644` runs, and the tracker can open it as UID 1001. The second commit then covers the adjacent write paths (`samples/streams`, `Tracker/`, app cwd) and moves `HOME`/`GST_REGISTRY` off `/root` before `setpriv`.

What I checked, for the record:
- Both new script copies are byte-identical (`sha256 d9bc9b1c…`), as is the new 65-line `ds-start.sh` block (`sha256 ad194c59…`). `Check helm parity` is green.
- Compose: rendered the merged config with `docker compose config` on an alerts-shaped `extends` override — `/opt/scripts/setup-tracker-reid.sh` survives the child's own `volumes:` list. Helm: both ConfigMaps that embed `files/ds-start.sh` also embed the new script, mounted at `/startup-script` with `defaultMode: 493`, no `items:` filter. `statefulset-standalone-mv3dt.yaml` runs a different script and is correctly untouched.
- `chown` needs root: `runAsUser: 0` on all four rtvi-cv statefulsets, no `readOnlyRootFilesystem`, so `/tmp/rtvi-cv-home` is writable.
- Ran it rather than read it: `bash -n` clean on all four scripts; `shellcheck --severity=warning` clean except one `SC2034` (TODO 4); and I extracted `append_deepstream_sample_dirs` and executed it against a fake DeepStream tree — correct on the normal single-version layout (dedupes through the `deepstream` symlink) and on a two-version layout (emits both).

Of the four TODOs, **TODO 2 is the only one I'd suggest folding into this PR** — two lines, in a file you're already touching. The rest are fine as follow-ups or a tracking issue.

---

## TODO 1 — engine-cache half of the helper never populates on K8s or on `down`/`up` (follow-up)

`setup-tracker-reid.sh` is invoked exactly once, pre-launch, at [`ds-start.sh:278`](deploy/docker/services/rtvi/rtvi-cv/ds-start.sh#L278) (docker) and [`ds-start.sh:255`](deploy/helm/services/rtvi/charts/rtvi-cv/files/ds-start.sh#L255) (helm) — before the app is exec'd. The tracker builds the ReID engine ~90–120 s *after the first frame flows*, into the `Tracker/` dir, and nothing runs the script again afterwards.

Credit where due: chowning `Tracker/` to `STORAGE_UID` pre-launch (second commit) is what lets the dropped process write its engine there at all, so the `docker compose restart` case now genuinely populates the cache on the next start. Still open:

- `docker compose down && up` → fresh container, engine lost before it's ever cached.
- Any Kubernetes container restart → kubelet makes a new container, layer always fresh, cache never populates.
- Helm doubly so: `engineCache.enabled` defaults to `false` ([`statefulset.yaml:18`](deploy/helm/services/rtvi/charts/rtvi-cv/templates/statefulset.yaml#L18)), so `/opt/engines` isn't even mounted.

The reference implementation runs it **twice**, and the docs say so explicitly — [`troubleshooting.md:119`](skills/vss-deploy-detection-tracking-2d/references/troubleshooting.md#L119): *"`setup_tracker_reid.sh` runs twice: once before launch … and once after launch (moves a freshly-built engine into the cache so the next deploy is fast)"*, with the post-launch call at [`run_app_and_wait.sh:296`](skills/vss-deploy-detection-tracking-2d/scripts/run_app_and_wait.sh#L296) using `--wait 180`. This PR carries the `--wait` polling block over but no call site passes `--wait`, so that block is currently unreachable.

Not a crash — the cost is that every fresh alerts/smartcities container keeps paying the ~90–120 s rebuild. Two clean ways out whenever this is picked up: **(1)** drop the cache machinery from the deploy copy (helper shrinks to ~60 lines, no dead flags), or **(2)** add the post-launch call the reference does. Worth saying which is intended even if the work lands later.

---

## TODO 5 — worth one real alerts deploy before or shortly after merge (verification, no code change)

CI is in good shape — you're vetted, and the head carries 43 check-runs with the gate sweep green (`Check helm parity`, `Copyright Headers`, `DCO Sign-off`, `License Diff (OSRB CSV)`, `Detect changed first-party images`, the full `CI` matrix). Just noting precisely what that green does *not* cover, so nobody reads it as "the scripts were exercised":

- **0 images were built.** `Build ${{ matrix.name }} → GHCR`, `Publish … multiarch manifest` and the matrix job are all `skipped` — correct, since these scripts are bind-mounted (Compose) and ConfigMap-mounted (Helm) rather than baked in. But it means nothing in the sweep ran the container.
- **No job parses or lints this bash.** The only `bash -n` in `.github/workflows/ci.yml` is at lines 242–243 against `$EMBED_INSTALLER` / `$VLM_INSTALLER`; there is no `shellcheck` step anywhere in `.github/workflows/`.

I covered the cheap half locally (`bash -n`, `shellcheck`, and executing the new discovery function). What that can't cover is `chown`/`install` against a real image under `set -euo pipefail`, where a non-zero exit crash-loops the container. Since this changes container startup for alerts and smartcities, one real deploy is the missing signal.

deploy/docker/services/rtvi/rtvi-cv/setup-tracker-reid.sh
zac-wang-nv: Following up on this one, because there's a sharper case than the permissions angle above — and it needs no unusual mode bits, just an interrupted write.

The test is `-e`, which is **exists** (any type, any size). The non-empty test is `-s`. So a **0-byte or truncated etlt takes this early-out and is never repaired**. Truth table, run against real files:

| path | `-e` | `-s` | `-L` | `-f` | branch taken? |
|---|---|---|---|---|---|
| `empty.etlt` (0 bytes) | yes | no | no | yes | **YES → skips install** |
| `full.etlt` | yes | yes | no | yes | YES → skips install |
| `link_ok.etlt` | yes | yes | yes | yes | no → install runs |
| `link_broken.etlt` | no | no | yes | no | no → install runs |
| `missing.etlt` | no | no | no | no | no → install runs |

Row 1 is the problem: a partial NGC pull, a copy that died mid-write, or a full disk leaves a zero-length file at `$dest`. The script then prints `TRACKER_REID: ALREADY_PRESENT`, reports success, and leaves the tracker to load a corrupt model — and it will keep doing that on every subsequent start, because the condition never becomes false on its own. The operator-visible symptom would look like the failure this PR fixes, but with a log line claiming the staging step succeeded.

Two things worth keeping as they are:

- The `! -L` half is deliberate and right. A symlink at that path was planted by this script's own `--symlink` / engine-cache path, so it should be re-pointed rather than treated as already-staged. Note `-e` follows symlinks while `-L` doesn't, so a *broken* symlink is `-e` false / `-L` true and correctly falls through to be replaced (row 4).
- The unconditional `chown` on L86 is fine — this is only about the file's mode and contents, not the directory's ownership.

Same fix as above, and it happens to cover both cases at once, since `install` overwrites whatever is there regardless of size or mode:

```bash
install -m 0644 "$src" "$dest"
echo "TRACKER_REID: INSTALLED  $src  ->  $dest"
```

If you'd rather keep an early-out to avoid the rewrite, `-s` instead of `-e` would at least catch the truncated case — but given `install` is idempotent and this runs once per container start, dropping the branch is simpler than getting the predicate exactly right.

(Still not blocking — same TODO status as the parent comment.)

zac-wang-nv (approved):
LGTM

PR #1530, opened 2026-08-03
feat(skill-eval): time each leg phase so a silent gap is readable

simings-nv:
## Why

A skill-eval leg reaches the Actions log as one undated block. `skills_eval_agent` runs `run_leg.py` through a Bash tool that surfaces captured output only when the tool returns, so the wrapper's existing log lines all arrive together with no clock on them.

From run `30733390364`:

```
detection-tracking-3d · deploy       130.3 of 132 min = one silent gap
stored_video_summarization_harbor    103.5 of 108 min = one silent gap
```

Both gaps begin right after `[tool] Bash :: python3 .github/skill-eval/run_leg.py ...`. From outside there is no way to tell a lock wait from a slow image pull from the agent working.

That gap is blocking a decision. We are weighing whether to pre-download and mount model weights on the runners, and nobody can say whether download time is even a material share of a leg, because we emit nothing while it runs.

## What

No behaviour change. Two additions to `run_leg.py`, and no workflow change.

**A heartbeat** prints elapsed leg time and the current phase every `HEARTBEAT_SEC` (default 60, env-overridable). Because captured output is replayed in order, every existing log line can now be placed to within one heartbeat even though the block still arrives at once. That is why the elapsed time is carried in the text rather than left to the log's own line timestamps, which do not exist for captured output. The first tick fires immediately rather than after a full interval, so a leg that dies inside the first minute still emits a timing line.

**Per-phase wall time** written to `phase-timings.json` under the results root. The workflow's existing "Collect results for workflow artifact" step already uploads that directory. Phases are the lock wait and each Harbor invocation, which is the split the queue-time question turns on.

```json
{"run_id": "...", "slug": "...", "total_s": 8192.0,
 "phases": [{"phase": "lock-wait", "start_s": 0.0, "end_s": 7412.0, "seconds": 7412.0},
            {"phase": "harbor:step-1", "start_s": 7412.0, "end_s": 8192.0, "seconds": 780.0}]}
```

The lock wait sets the heartbeat's **label** as well as its recorded interval. It calls `record_phase` directly rather than going through `phase()`, because its recorded name depends on how it ends (`lock-wait`, `lock-wait-timeout`, `lock-wait-failed`). An earlier revision therefore left the label at `startup` for the whole wait: the longest gap a leg can have, 16 minutes in the run above and bounded only by `lock_timeout_sec`, was the one phase every tick named wrongly.

## Layout

The instrumentation lives in its own module, `leg_timing.py`, rather than inline. It is self-contained, carries its own state, and `run_leg.py` is long enough that a reader looking for the lock or the Harbor command should not have to scroll past it. That file grows by 98 lines instead of 278.

Callers read the label through `leg_timing.current_phase()` rather than importing the global, which would bind once at import and never see a change.

The tests split on the same seam. `test_leg_timing.py` covers the module on its own; `test_run_leg.py` keeps the cases that drive the contract *through* `main()` (the lock-wait label, SIGTERM unwinding, a heartbeat that cannot start), because those are about the wiring rather than the module.

## Cancellation is the normal exit

`skills-eval.yml` sets `cancel-in-progress`, so every push SIGTERMs the in-flight legs. Three things follow, and without them the legs that cancel after the longest waits are exactly the ones that record nothing:

- **SIGTERM is converted into an unwinding exit.** Python turns only SIGINT into an exception; SIGTERM keeps `SIG_DFL` and kills the interpreter without running any `finally`, so the timing write never happened at all.
- **The lock wait is recorded on `BaseException`**, not `Exception`, so a `KeyboardInterrupt` mid-wait still leaves its interval in the artifact.
- **The artifact is written to a sibling and renamed.** This write runs in `main()`'s `finally`, which is where a second signal lands, and a partial write leaves truncated JSON that a reader cannot distinguish from a leg that recorded nothing.

## Safety

Timing must never change a verdict, so every write is best effort:

- `leg_log` swallows its own exceptions; a `print` can raise `BrokenPipeError`, and one of these calls sits in `main()`'s `finally`
- the heartbeat is started inside a guard, so a runner at its thread limit loses visibility rather than the leg
- the heartbeat thread is a daemon and cannot hold exit
- an unwritable results root logs and continues, leaving no partial sibling
- phases are recorded even when the body raises, since a leg that dies mid-Harbor is exactly the one worth timing
- `HEARTBEAT_SEC` is clamped to `[1, 3600]` and rejects non-finite values: `float()` accepts `"inf"` and `"1e999"`, and `max(1.0, inf)` is `inf`, which reaches `Thread.join(timeout=...)` and raises `OverflowError`
- the phase label is restored *before* the end line is logged, and the stop event is checked before every tick, so no line can claim a phase that has already closed

No new dependency; `threading` is stdlib.

## Scope

This times the leg. It deliberately **does not sample the network**, which an earlier revision of this PR did.

The workload runs on a remote Brev box via `envs.brev_env:BrevEnvironment` while this wrapper runs on the coordinator, so any counter read here describes the wrong machine: a box downloading model weights for ten minutes shows up as a quiet link. Measuring the transfer needs remote sampling of the kind `gpu_trace.py` does over `brev exec`, and is a follow-up once these timings show whether that time is where we think it is.

Separating an NGC pull from a weight download *inside* a Harbor invocation needs instrumentation in the deploy path, and is also a follow-up.

## Tests

```
python3 .github/skill-eval/tests/test_run_leg.py
```

69 tests in the file, 15 covering this change, each mutation-checked. Reverting the immediate first tick, the lock-wait label, either label restore, the end-line ordering, the `BaseException` catch, the guarded heartbeat start, the SIGTERM handler, or the atomic rename each fails a dedicated test.

CI runs `test_run_leg.py` by explicit class rather than by file, so this PR adds `test_leg_timing.py` and the two remaining new classes to that list in `ci.yml`. Without that none of these 15 tests execute in CI, including `InstrumentationNeverChangesTheVerdict`, which pins the one invariant the feature rests on. Verified against the step's exact invocation: 79 passed.

The SIGTERM case runs in a subprocess on purpose: delivered in-process, a removed handler would take the whole test runner down instead of failing one test.

`test_a_dead_heartbeat_thread_does_not_fail_the_leg` previously asserted the opposite of its own name, inside a class asserting the opposite invariant. It now pins the behaviour the name describes.

Three failures in `test_run_leg.py` predate this change (RTX 4090 pool capability tables) and are unrelated. Verified by running the suite on `develop`: same 3 failures.

### DIFF
--- a/.github/skill-eval/leg_timing.py
+++ b/.github/skill-eval/leg_timing.py
@@ -1,0 +1,224 @@
+#!/usr/bin/env python3
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""Elapsed-time heartbeat and per-phase wall time for one skill-eval leg.
+
+A leg's log lines carry no clock. `skills_eval_agent` runs `run_leg.py` through
+a Bash tool that surfaces captured output only once the tool returns, so a
+132-minute leg reaches the Actions log as a single undated block: run
+30733390364 spent 130.3 of 132 minutes inside one gap, and 103.5 of 108 in
+another leg, with no way to tell a lock wait from a slow deploy.
+
+Two cheap answers, both here. A heartbeat stamps elapsed time and the current
+phase into the same stream, so every neighbouring line can be placed to within
+HEARTBEAT_SEC even though the whole block arrives at once. Each phase's wall
+time is written next to the results, where the workflow's existing collect step
+already uploads it, so the split can be aggregated across legs instead of read
+by eye.
+
+TIMING MUST NEVER CHANGE A VERDICT. Every write here is best effort and a
+failure to record is logged and swallowed. `leg_log` swallows its own
+exceptions because one of these calls sits in `run_leg.main`'s finally, where
+raising would replace Harbor's exit code with a logging failure.
+
+WHAT THIS DELIBERATELY DOES NOT DO is sample the network. Harbor runs the
+workload on a remote Brev box via `envs.brev_env:BrevEnvironment` while the
+wrapper runs on the coordinator, so any counter read here describes the wrong
+machine: a box downloading model weights for ten minutes would show up as a
+quiet link. Measuring the transfer needs remote sampling of the kind
+`gpu_trace.py` does, and is a follow-up once these timings show whether that
+time is where we think it is.
+
+STATE IS MODULE-LEVEL and single-writer by design. `run_leg.main` is the only
+writer; the heartbeat thread only reads `_CURRENT_PHASE`, and it is joined
+before `_PHASES` is serialised. Read the label through `current_phase()` rather
+than importing the global, which would copy it once and never see a change.
+"""
+
+from __future__ import annotations
+
+import contextlib
+import json
+import math
+import os
+import threading
+import time
+from pathlib import Path
+
+
+MAX_HEARTBEAT_SEC = 3600.0
+
+
+def _heartbeat_interval() -> float:
+    """Seconds between heartbeats, clamped so a typo cannot spin or break a leg."""
+    raw = os.environ.get("RUN_LEG_HEARTBEAT_SEC", "60")
+    try:
+        interval = float(raw)
+    except ValueError:
+        print(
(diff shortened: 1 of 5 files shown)


zac-wang-nv (approved):
LGTM

PR #1538, opened 2026-08-04
feat(skills): alerts-profile Kubernetes operate via VSS_PUBLIC_URL

pratik-bhatt-nv:
## Summary
- Wire `vss-manage-alerts`, `vss-query-analytics`, and report Mode B/C to the shared `VSS_PUBLIC_URL` contract for alerts Ingress (`/alert-bridge`, `/va-mcp`, `/vst`).
- Force public endpoints when `VSS_PUBLIC_URL` is set (ignore leftover Docker `:9080` / `:9901` / `:30888`); gate `docker ps` and ES `:9200` CV-verdict probes to Docker only.
- Document alerts public routes in `deployment_resolution.md` / `skills/README.md`; leave Compose host-port operate paths unchanged.

## Test plan
- [ ] With `VSS_PUBLIC_URL` unset and `HOST_IP` set: Alert Bridge / VIOS / VA-MCP still resolve to `:9080` / `:30888` / `:9901`; mode detection via `docker ps` still works
- [ ] With `VSS_PUBLIC_URL` set: `$AB` / `$VST` / `VA_MCP_URL` force public prefixes; no `docker ps` / port-forward / Service DNS
- [ ] K8s: `GET $AB/health`, create/list/delete realtime rule, `GET .../realtime/incidents`
- [ ] K8s: VA-MCP two-step initialize + tools/call via `${VSS_PUBLIC_URL}/va-mcp/mcp`
- [ ] K8s: NvStreamer upload still requires separate `VSS_STREAMER_URL` (not derived from `VSS_PUBLIC_URL`)
- [ ] Optional `ALERTS_MODE` skips mode ask; without it, skill asks real-time vs verification


pratik-bhatt-nv:
Follow-up fix pushed for Greptile P1:

- VA-MCP readiness now probes `${VA_MCP_URL}/health` (Ingress rewrite → `/health`) in `vss-query-analytics` and report Mode B
- Guard empty `SESSION_ID` before tools/call
- `always-on.md` / `on-demand-verification.md` use `$AB` (no hardcoded `:9080`); always-on enable path names `ALWAYS_ON_RULES_CONFIG` / `realtime-config.yaml`

skills/vss-query-analytics/SKILL.md
```diff
@@ -55,20 +58,46 @@ Queries incidents, alerts, and metrics stored in Elasticsearch via MCP JSON-RPC
...
 
-1. Probe the VA-MCP endpoint:
+1. Probe VA-MCP liveness via `/health` (Ingress rewrites
+   `${VSS_PUBLIC_URL}/va-mcp/health` → `/health` on the Service). Do **not**
+   use `GET /mcp` or the service root as the readiness check — those are not
+   reliable health routes.
    ```bash
-   curl -sf --max-time 5 "http://${HOST_IP}:9901/mcp" >/dev/null 2>&1 || \
-     curl -sf --max-time 5 "http://${HOST_IP}:9901/" >/dev/null
+   curl -sf --max-time 5 "${VA_MCP_URL%/}/health" >/dev/null
```
zac-wang-nv, line 105: **P1 — `/health` probe contradicts this skill's own eval check (reproduced)**

This is correct on the merits, but it contradicts this skill's own eval check — and that is no longer a prediction. **The `query_analytics` leg ran on `7ab1462a6` and failed on exactly this:**

> **step-2** — `The trajectory probed VA-MCP liveness with curl -sf --max-time 5 http://localhost:9901/mcp (or a GET / fallback) before issuing any tools/call`: *"The agent probed VA-MCP liveness at /health, not at /mcp or / as required by the check."*

Step-2 scored `0.67 (2/3)`, and steps 3, 4, 5 and 6 were all **skipped** on `prior-step fail`. So one stale check costs the entire spec. The spec sets no `verifier` and therefore routes to the default generic judge — the judge was not lenient.

`evals/query_analytics.json` is byte-identical to merge-base and has never been touched by this PR — `sha256 4224f941…` at all five of `18b0302bc`, `bf3b4c613`, `e31618a50`, `7ab1462a6`, `f8f15ee61`.

| | probe | matches check 26? |
|---|---|---|
| merge-base | `"http://${HOST_IP}:9901/mcp"` `\|\|` `".../"` | ✓ |
| first commit `e31618a5` | `"${VA_MCP_MCP}"` `\|\|` `"${VA_MCP_URL%/}/"` | ✓ |
| head `f8f15ee61` | `"${VA_MCP_URL%/}/health"` | ✗ **(reproduced)** |

The skill also now says explicitly, three lines up: *"Do **not** use `GET /mcp` or the service root as the readiness check."* An agent following the skill exactly **cannot** satisfy check 26.

To be clear on provenance: the check is pre-existing and untouched by this PR, but the contradiction is introduced here — `7ab1462a6` moved the probe away from a check it left frozen. The fix is to update check 26 in this same PR; the probe change and the check that grades it have to move together. Worth doing before merge, since the leg will keep failing at every head otherwise.

skills/vss-query-analytics/SKILL.md
```diff
@@ -91,14 +120,15 @@ This skill reads from the Elasticsearch/VA-MCP stack brought up by the VSS **ale
 
 ```bash
 # Step 1: initialize — get session ID from response HEADER
-SESSION_ID=$(curl -si -X POST http://${HOST_IP:-localhost}:9901/mcp \
+SESSION_ID=$(curl -si -X POST "${VA_MCP_MCP}" \
```
zac-wang-nv, line 143: **P1 — `${VA_MCP_MCP}` is resolved in a different fenced block → empty expansion**

`${VA_MCP_MCP}` is set in a *different* fenced block (`Endpoint resolution`, L63-79). If fenced blocks don't share a shell this expands to empty and the command becomes `curl -si -X POST ""` → `curl: (3) URL rejected`. Same at `:131`.

Three things in this PR's own tree say they don't share a shell:

- `vss-generate-video-report/SKILL.md:687` states it verbatim: *"Each fenced block runs as its own shell, so `$MCP` / `$SID` / `$CT` / `$AC` from Step 1 do NOT carry over — re-establish them."*
- `vss-ask-video/SKILL.md:265-267` re-derives the whole `VSS_PUBLIC_URL` resolution inside the second block that needs it — the merged base/LVS convention.
- This PR guards `${DEPLOYMENT_KIND:-docker}` in **all 8** blocks that read it, which is only necessary if state doesn't carry.

At merge-base this was `http://${HOST_IP:-localhost}:9901/mcp` — self-contained, with a `localhost` default — so on Docker this is a regression on the skill's most-used path, not just a K8s gap.

The `[ -n "$SESSION_ID" ] || { …; exit 1; }` guard at `:128` is a real improvement — it converts a silent bad request into a hard stop. It doesn't fix the resolution though, and its own message interpolates `${VA_MCP_URL}`, which is empty in exactly the case that triggers it. Suggest `VA_MCP_MCP="${VA_MCP_MCP:-http://${HOST_IP:-localhost}:9901/mcp}"` at the top of the block, or re-deriving the 4-line resolution.

Note the fail-open direction too: `${DEPLOYMENT_KIND:-docker}` defaults to **docker**, so if resolution doesn't carry on a K8s deployment the guards fall through to `docker ps` and the ES `:9200` probe — precisely what this PR set out to prevent.

skills/vss-generate-video-report/SKILL.md
```diff
@@ -662,12 +665,13 @@ Use for "generate an SOP compliance report" over a sensor + time range. Data com
 - Confirm the SOP tools are present (once). The four `get_sop_*` tools are added by the SOP patch and are **not** in the base `/vss-query-analytics` tool set, so call the VA-MCP endpoint directly (two-step MCP JSON-RPC: `initialize` → `tools/list`):
 
 ```bash
-MCP="http://${HOST_IP}:9901/mcp"
+# Resolve VA_MCP_URL first (Endpoint resolution above). Force public path on K8s.
+MCP="${VA_MCP_URL%/}/mcp"
```
zac-wang-nv, line 678: **P1 — `VA_MCP_URL` not re-derived here; `MCP` becomes `/mcp`**

It's the clearest case of the cross-block family, because the sentence at L687 (for the Step 2 twin) already tells the reader that variables from an earlier fenced block do **not** carry over. `VA_MCP_URL` is set ~570 lines earlier at L96/L101, so `${VA_MCP_URL%/}` expands to empty and `MCP` becomes `/mcp`.

The comment "Resolve VA_MCP_URL first (Endpoint resolution above)" is an instruction to a human reader; it isn't executable. The Mode B probe at `:133` **does** defend itself — `"${VA_MCP_URL:-http://${HOST_IP}:9901}/health"` — so the file is internally inconsistent: the probe the follow-up touched is guarded, the two call sites it didn't touch are not. Applying the `:133` pattern here and at `:688` would close it.

Worth noting no eval currently reaches this code: `base_profile_report` step-5 (*"Verify VA-MCP incident analytics is ready"*) was **skipped** because step-4 failed on unrelated VLM content. This path is unexercised, not passing.

skills/vss-manage-alerts/references/alert-subscriptions.md
```diff
@@ -103,7 +107,7 @@ Resolve the user's sensor name to three values needed for the payload: `sensor_i
 **2a. Fetch the sensor list:**
 
 ```bash
-curl -s "http://${HOST_IP}:30888/vst/api/v1/sensor/list" | jq .
+curl -s "$VST/vst/api/v1/sensor/list" | jq .
```
zac-wang-nv: **P1 — `$VST`/`$AB` resolved in a different *file*; 7 bare sites, zero guards**

Same cross-block issue, one level worse: `$VST` and `$AB` are set in a **different file** (`../SKILL.md:74-78`). This is loaded as a Workflow D sub-playbook, so there's no guarantee the parent's resolution block ran in the same shell. Seven blocks depend on it bare — `$VST` at `:110`, `:139`; `$AB` at `:188`, `:272`, `:277`, `:350`, `:382` — where merge-base had working literal URLs at every one.

The follow-up made this sharper rather than moot. `: "${AB:?…}"` is now applied in `verification.md`, `on-demand-verification.md` and `always-on.md` — **three of the four operate playbooks**. Grepping `:?` in `alert-subscriptions.md` at head returns nothing: it is the only one left with zero guards, and it has the most bare references. Adding the same one-liner for both `AB` and `VST` at the top of this file would finish the pattern the follow-up started.

pratik-bhatt-nv:
Addressed Zac’s P1s in a signed follow-up:

1. **`query_analytics` eval** — step-2 checks now expect `GET …/health` (not `/mcp` or `/`), matching the skill.
2. **Cross-block resolution** — `vss-query-analytics` probe + two-step MCP blocks re-derive `VA_MCP_URL` / `VA_MCP_MCP` in-shell.
3. **Report Mode C** — both SOP MCP blocks re-derive `VA_MCP_URL` (no bare `/mcp`).
4. **`alert-subscriptions.md`** — Setup now resolves + guards `$AB` / `$VST` locally.

skills/vss-query-analytics/SKILL.md
pratik-bhatt-nv: Addressed in [c9e3cd8](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/commit/c9e3cd8df292420188d833ce41a652800c86e007): the eval now requires the `/health` readiness probe, matching the skill. The latest `query_analytics` Skills Eval also passes the endpoint-verification step.

skills/vss-query-analytics/SKILL.md
pratik-bhatt-nv: Addressed in [c9e3cd8](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/commit/c9e3cd8df292420188d833ce41a652800c86e007): the two-step MCP block now re-derives `VSS_PUBLIC_URL`, `VA_MCP_URL`, and `VA_MCP_MCP` before initializing or calling VA-MCP, so it does not rely on state from another fenced block.

skills/vss-generate-video-report/SKILL.md
pratik-bhatt-nv: Addressed in [c9e3cd8](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/commit/c9e3cd8df292420188d833ce41a652800c86e007): both Mode C fenced blocks now independently resolve `VA_MCP_URL` and construct `MCP` before their MCP requests.

skills/vss-manage-alerts/references/alert-subscriptions.md
pratik-bhatt-nv: Addressed in [c9e3cd8](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/commit/c9e3cd8df292420188d833ce41a652800c86e007) and [a99f6cd](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/commit/a99f6cdda82b9a3d6d1a75ce4f0ebd721e6c97f5): this playbook now resolves and guards `AB` and `VST` locally, then derives and uses `VST_API_BASE`. It no longer depends on the parent skill shell.

zac-wang-nv (approved):
LGTM

PR #1554, opened 2026-08-05
Fix 9 BLOCKER SonarQube issues in services/vios

gous-75:
## Summary

Fixes all 9 **BLOCKER**-severity SonarQube issues on
[`TEGRASW_VST_VST_vms_shim`](https://sonar.nvidia.com/project/issues?id=TEGRASW_VST_VST_vms_shim&branch=develop) (`develop`).

Analysis runs against the GitLab mirror `L4TMM/vms_shim`; the same tree lives here under `services/vios/`.

Opened as a **draft**: these are machine-generated changes to error-handling and null-guard logic and have **not been compiled** — see Validation below.

## Changes by module

| Module | File | Rule | Fix |
|---|---|---|---|
| `media` | `media_pipelines/gstdemux.cpp` | `cpp:S999` | backward `goto` → `while (true) { … continue; }` |
| `media` | `media_utils/mm_utils.cpp` | `cpp:S999` | `retry:`/`exit:` labels → `do { … } while (…)` |
| `media` | `local_stream/clip_reader_producer.cpp` | `cpp:S912` ×3 | hoist calls out of `&&`/`\|\|` right-hand operands |
| `rtsp_server` | `RtspLoadBalancer.h` | `cpp:S3584` ×2 | **memory leak** — `new` → `std::make_unique` + `.release()` |
| `utilities` | `utilities/utils.cpp` | `cpp:S999` | `ping_retry:` → `do { … } while (retry-- > 0)` |
| `webrtc_streamer` | `PeerConnectionManager.cpp` | `cpp:S999` | `retryForRelayCandidate:` → `while (true)` + `break` |

One squashed commit per module; each commit body lists the rule, line and SonarQube issue key.

## Semantics worth reviewing

- **`RtspLoadBalancer.h`** — the real bug: `new RtspServer(...)` leaked whenever `isError()` was true, since the pointer was neither stored nor deleted. `make_unique` frees on that path; `.release()` into `m_servers` preserves the existing raw-pointer ownership model.
- **`clip_reader_producer.cpp`** — the `caps_str &&` short-circuit is a deliberate **null guard**. The calls are hoisted inside an `if (caps_str) { … }` block so the guard is preserved; behaviour is unchanged.
- **`mm_utils.cpp`** — retry count is unchanged (3 attempts). This also removes a latent bug: previously, if `gst_discoverer_new()` failed, `goto exit` fell into the retry check and jumped back to `retry:` with a **NULL** discoverer. The loop is now skipped entirely and `g_object_unref` only runs on success.
- **`PeerConnectionManager.cpp`** — the original `goto` jumped backward over `Json::Value local_iceCandidates;`, destroying and reconstructing it each retry. It is now hoisted out of the loop with an explicit `local_iceCandidates = Json::Value();` reset to preserve that.
- **`utils.cpp`** — attempt count preserved (1 initial + `retry`).

## Validation

- [x] Semantic review of every hunk
- [x] Brace balance verified; no orphaned labels (remaining `goto`s in `gstdemux.cpp` / `PeerConnectionManager.cpp` are pre-existing at other sites, with their labels intact)
- [ ] **Not compiled** — no local CUDA/GStreamer/DeepStream toolchain. Please let CI build before this leaves draft.

Two files carry large reindentation churn from removing the `goto` labels (`gstdemux.cpp` ~190 lines, `mm_utils.cpp` ~330). `git diff -w` reduces these to 5 and 14 real lines respectively — review with `-w` for signal.

> Generated by the SonarQube AI Fix Agent (backend: `claude`). Please review each change before merging.

### DIFF
--- a/services/vios/src/framework/media/media_pipelines/gstdemux.cpp
+++ b/services/vios/src/framework/media/media_pipelines/gstdemux.cpp
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2020-2021 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-FileCopyrightText: Copyright (c) 2020-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  * SPDX-License-Identifier: Apache-2.0
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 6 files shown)


gous-75:
Superseded by #1659, which combined this and the other nine SonarQube remediation PRs into a single change so the union was reviewed and built together rather than ten times in isolation.

#1659 is merged to `develop` (commit `5d99fbe8a`, 300 files). Every file changed on this branch is present there — verified by comparing this branch's changed-file set against the merged commit before closing.

Closing as superseded, not rejected.

PR #1564, opened 2026-08-05
Fix alerts eval Kafka container name

hugoverjus:
## Summary
- align the dense-captioning alerts eval with the deployed `kafka` container name
- update final incident-consumer guidance and examples
- retain legacy and standalone Kafka container detection as fallbacks

## Test plan
- [x] Validate `alerts_profile_api.json`
- [x] Run dense-captioning skill compliance check
- [x] Run `git diff --check`


zac-wang-nv (approved):
LGTM

PR #1565, opened 2026-08-05
Fix SonarQube HIGH: Memory & ownership

gous-75:
## Memory & ownership

Manual memory management, C-style allocation, and special member functions -> RAII / smart pointers / rule-of-zero.

Automated SonarQube fixes (HIGH impact severity), one squashed commit per module.

**Rules:** `cpp:S5025`, `cpp:S1231`, `cpp:S3490`, `cpp:S3624`, `cpp:S5019`, `cpp:S1699`, `cpp:S5506`
**Results:** {'fixed': 311, 'failed': 170}
**Commits:** 8

Every module was compiled via `cd services/vios && ./build.sh package module=streamprocessing,sensor` before its commit; a failing module is reverted and its issues marked failed.

> Review **commit-by-commit** - each commit is one module. Use `git diff -w`: some rules cause reindentation that dwarfs the real change.

> Generated by the SonarQube fix agent. Draft until reviewed.

### DIFF
--- a/services/vios/src/framework/media/media_pipelines/VideoQueue.h
+++ b/services/vios/src/framework/media/media_pipelines/VideoQueue.h
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2022 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-FileCopyrightText: Copyright (c) 2022-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  * SPDX-License-Identifier: Apache-2.0
  *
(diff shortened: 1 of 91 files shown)


gous-75:
Superseded. Regenerating with a new harness that builds after every batch of fixes and commits only when green.

This branch does not compile as a whole: several fixes converted a member to `std::unique_ptr` in the .cpp while leaving the declaration in the paired .h as a raw pointer. Per-module verification could not catch that, because each module was verified in isolation and nothing ever built the union.

Rather than repair 311 changes produced without scope resolution — where the ones that compile may still be incomplete in ways the compiler cannot see — these are being regenerated.

PR #1574, opened 2026-08-06
Fix/bva svc name resolution

prisrivastav-nv:
### **1. Video Analytics API service name swap in Alerts & Search combined build**

The combined Alerts + Search build eval failed: the resolved `COMPOSE_PROFILES` contained `vss-video-analytics-api-alerts` instead of the expected `vss-video-analytics-api-fusion` (26/27 keys matched). The agent **swapped** one service name to the wrong Foundation's variant — mixing search and alerts names.

### Root cause
Which key survives a singleton convergence is **foundation-relative** knowledge the framework delegates to owner contracts. Three of the five aliased owners (`rt-cv`, `vios`, `elk`) carry that local statement; the analytics-API owner (`alerts.md`/`search.md`) was the only one missing it, so the agent took the alerts key at face value. The swap is a foundation-membership error, not a `container_name` collision.

### Approach
Fix the swap at the layer that owns foundation-relative key choice (the owner contract), stated **agnostically** ("use the selected Foundation's key," no primary/secondary, no cross-references). Add an independent, foundation-agnostic graph-validation net for the distinct duplicate case. No changes to `composition.md` (avoids global/owner conflation).

### Fix
- Owner contracts: added the missing per-Foundation key anchor to the analytics-API and behavior-analytics owners, phrased Foundation-agnostically.
- Validator: `resolved.yml` now fails if two services share a `container_name` (deploy-safety net) + unit tests.
- Eval: reworded the combined query to be capability-driven and search-forward, reducing Foundation-selection ambiguity.

---

### **2. Follow-on eval hardening across build-vision profiles**

Later runs surfaced failures unrelated to the swap. Triaging each cleanly separated two classes: genuine agent-guidance gaps and over-strict/ambiguous eval checks — fixed at the owning layer, never by papering over the other.

- **Sharper Foundation deltas.** Owner contracts now make prune-vs-keep and key-vs-bundle choices unambiguous — exposed read surfaces (analytics-API, ingress) prune by default, converged configs mount verbatim, and initializer keys stay on the selected Foundation while only the bundle varies. Net: fewer mixed-Foundation and under/over-pruning drifts.
- **Grounded capability wiring.** Filled concrete gaps the agent couldn't infer — direct VLM serving endpoint, explicit Kafka output wiring, resolvable verifier-config paths, and deploy-time `vss configure`. Net: proposals and builds that actually resolve and deploy.
- **Trustworthier checks.** Corrected eval specs that penalized valid builds — realistic artifact allowances, provisioning wording matched to the true caller-driven model, and removal of a forward-reference that made the judge conflate build-time docs with runtime. Net: failures now signal real defects, not check artifacts.

---

## Checklist


skills/vss-build-vision-agent/references/services/alerts.md
```diff
@@ -8,6 +8,15 @@
 | Video-analytics MCP | `vss-va-mcp` |
 | Alerts analytics API | `vss-video-analytics-api-alerts` |
 
+`vss-video-analytics-api-alerts` is the **same** `vss-video-analytics-api` container
```
zac-wang-nv, line 11: can we modify the source rather than document the diff here?

PR #1575, opened 2026-08-06
Fix 12 HIGH SonarQube issues: memory ownership (RAII conversions)

gous-75:
Fixes 12 HIGH-severity SonarQube issues in `services/vios` by replacing manual
memory management with RAII ownership.

| Rule | Issues | What changed |
|---|---|---|
| `cpp:S5025` | 8 | Manual `new`/`delete` -> `std::unique_ptr` |
| `cpp:S1231` | 2 | `malloc` -> typed RAII allocation |
| `cpp:S3490` | 2 | Manual resource management -> RAII |

**12 files, +70 / -78.** Sources only.

### Why these are safe

Each conversion changes *where ownership is expressed*, not *when the object
dies*. A member that was `new`ed in the constructor and `delete`d in the
destructor becomes a `unique_ptr` initialised in the same place and released at
the same point — the destruction order and lifetime are unchanged. Two
singletons (`CompositeNotifier`, `WebhookNotifier`) use the `PrivateTag` idiom
so `make_unique` can reach a private constructor without widening the API.

The bug class these fixes prevent is a leak or double-free on an error path
where an early `return` skipped the manual `delete`.

### Verification

- `./build.sh package module=sensor,streamprocessing` — **0 errors**
- Every fix was compiled individually before being committed, and the branch
  was rebuilt as a whole after assembly, so no commit here depends on an
  uncommitted change in another.

### Note on scope

Each fix updates the `.cpp`, its paired header, and every use site together.
A partial conversion — converting the definition but leaving the declaration a
raw pointer — does not compile, so the per-commit build is what guarantees
completeness here.

Signed-off-by: gous <gmulla@nvidia.com>

### DIFF
--- a/services/vios/src/framework/media/overlays/ll_overlay.cpp
+++ b/services/vios/src/framework/media/overlays/ll_overlay.cpp
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2024-2025 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-FileCopyrightText: Copyright (c) 2024-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  * SPDX-License-Identifier: Apache-2.0
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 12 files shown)


gous-75:
Closing: the memory-ownership group is being completed on a separate machine, which will raise a single PR covering all ~479 issues in the group. These 12 fixes are a subset of that work and would otherwise conflict with it.

No other group PR depends on this branch — verified none of them contain these commits.

PR #1576, opened 2026-08-06
Enhanced Spatial AI Data Utils integration

kartikayt-nvidia:
- Updated version suffix pattern to include tree SHA and commit SHA for better traceability.
- Introduced a new function to determine the downstream build type based on the relevance of changes.
- Improved validation in `spatialai_publish_variables` to ensure correct SHA matching.
- Modified CI workflow to handle the new tree SHA and updated suffix format for Spatial AI Data Utils.

## Checklist

### DIFF
--- a/.github/scripts/prepare_downstream_release_set.py
+++ b/.github/scripts/prepare_downstream_release_set.py
@@ -12,7 +12,12 @@ import re
 import sys
 from pathlib import Path
 
-from detect_changed_images import changed_paths, commit_exists, resolve_diff_base
+from detect_changed_images import (
+    changed_paths,
+    commit_exists,
+    paths_changed_under,
+    resolve_diff_base,
+)
 from release_set import load_inventory, validate_release_set
 from update_pr_ghcr_candidates import GitHubApi, download_release_set
 
@@ -22,8 +27,10 @@ DEPLOY_PREFIX = "deploy/"
 # downstream coverage when its source changes -- mirrored and externally pinned
 # components are deployed by the same profiles and break the same evals.
 TRIGGER_FLAG = "trigger_downstream_from_source"
-
-
+SPATIALAI_DATA_UTILS_PATH = "libs/analytics/spatialai-data-utils"
+SPATIALAI_VERSION_SUFFIX_PATTERN = re.compile(
+    r"\.dev[1-9][0-9]*\+t([0-9a-f]{40})\.g([0-9a-f]{12})\.gh\.r[1-9][0-9]*"
+)
 PR_REF_PATTERN = re.compile(r"pull-request/(\d+)")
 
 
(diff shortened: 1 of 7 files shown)


hugoverjus (dismissed):

.github/scripts/prepare_downstream_release_set.py
zac-wang-nv, line 1: why do we need release-set artifact?

.github/scripts/prepare_downstream_release_set.py
```diff
@@ -134,6 +141,65 @@ def downstream_variables(release_set: dict) -> dict[str, str]:
     }
 
 
+def downstream_build_type(
```
zac-wang-nv: what is spatialai-reconcile build type?  

can you create another branch on vss-ci-oss to handle this pipeline?

.github/workflows/ci.yml
```diff
@@ -1510,22 +1497,17 @@ jobs:
...
+    # every downstream acceptance run where SDU did not change. Check only this
+    # job's direct dependencies, then apply the gate prepared by the wait job.
     if: >-
       !cancelled() &&
       !contains(needs.*.result, 'failure') &&
       !contains(needs.*.result, 'cancelled') &&
       !contains(needs.*.result, 'skipped') &&
-      (needs.prepare-source.outputs.trigger-from-src-change == 'true' ||
-       needs.wait-for-build-dev-images.outputs.trigger-from-new-image == 'true')
+      needs.wait-for-build-dev-images.outputs.run-downstream == 'true'
```
zac-wang-nv, line 1351: why do we combine them?

hugoverjus (approved):

.github/scripts/prepare_downstream_release_set.py
kartikayt-nvidia: spatialai-reconcile is no longer a BUILD_TYPE; normal downstream release-set handoff uses only ghcr-acceptance. SDU uses SPATIALAI_PIPELINE=true and targets the separate spatialai-publisher branch, which exists in the CI repository.

.github/workflows/ci.yml
kartikayt-nvidia: The old inline “source changed OR new image exists” expression was replaced by one explicit run-downstream decision from prepare_downstream_release_set.py. It runs downstream for relevant image-source or deploy/ changes, even if no image rebuild occurs, and avoids the old split logic drifting apart.

.github/scripts/prepare_downstream_release_set.py
kartikayt-nvidia: The artifact is the immutable handoff from Build Dev Images to the downstream job. It pins the exact source commit, image digests, release-set ID, and candidate tag that completed upstream validation. Passing it avoids rediscovering images downstream and ensures GitLab validates the same bytes GitHub produced.

PR #1579, opened 2026-08-06
Fix 83 HIGH SonarQube issues: void* API signatures (cpp:S5008)

gous-75:
Fixes 83 HIGH-severity SonarQube issues in `services/vios`.

Replacing `void *` in typedefs, members, parameters and return types with concrete types.

**Risk: HIGH.** These change function signatures, so the blast radius reaches every caller and, for exported symbols, the ABI. They compile cleanly — review each signature deliberately.

### Rules
- `cpp:S5008`

### How these were produced and verified

Issues were fixed in waves: each issue fixed independently in its own worktree,
the results merged into a single tree, and **that combined tree compiled before
anything was committed**. A fix that failed to build was isolated and dropped
rather than dragging the wave down with it, so every commit on this branch
compiles.

Each fix updates the file SonarQube flagged **and** its header and use sites
together — a partial conversion does not compile, so the per-wave build is what
guarantees completeness.

### Verification
`./build.sh package module=sensor,streamprocessing` — clean.

### Not included
- `cpp:S134`: Structural refactor (nesting depth / cognitive complexity)
- `cpp:S4423`: TLS/crypto behaviour change

Signed-off-by: gous <gmulla@nvidia.com>

### DIFF
--- a/services/vios/include/device_manager.h
+++ b/services/vios/include/device_manager.h
@@ -74,6 +74,7 @@ class ISensorControlInterface;
 class ISensorDiscoveryInterface;
 class SensorControl;
 
+typedef void (*destroyControlObject_t) (ISensorControlInterface*);
 typedef void (*destroyDiscoveryObject_t) (ISensorDiscoveryInterface*);
 
 typedef void (*cb_ptr_t)(const string, shared_ptr<struct SensorInfo>, bool);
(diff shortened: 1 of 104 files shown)


gous-75:
Superseded by #1659, which combined this and the other nine SonarQube remediation PRs into a single change so the union was reviewed and built together rather than ten times in isolation.

#1659 is merged to `develop` (commit `5d99fbe8a`, 300 files). Every file changed on this branch is present there — verified by comparing this branch's changed-file set against the merged commit before closing.

Closing as superseded, not rejected.

PR #1566, opened 2026-08-05
fix(agent): remediate current fixable CVEs

hugoverjus:
## Summary
- upgrade `cryptography` from 49.0.0 to 50.0.0 to fix CVE-2026-69247 / GHSA-g6cj-pr64-35w5
- install Debian's patched `liblzma5` 5.4.1-1+deb12u1 in the distroless runtime to fix CVE-2026-34743
- preserve patched package metadata so image scanners report the runtime version accurately

## Security scans
Grype 0.104.3 with the vulnerability database built 2026-08-05 scanned a fresh local image built from this branch (`--scope squashed`).

- baseline: 68 matches (2 critical, 10 high, 20 medium, 1 low, 33 negligible, 2 unknown)
- remediated: 66 matches (2 critical, 9 high, 19 medium, 1 low, 33 negligible, 2 unknown)
- removed: CVE-2026-34743 (`liblzma5`) and GHSA-g6cj-pr64-35w5 / CVE-2026-69247 (`cryptography`)
- added: none
- `pip-audit`: no known vulnerabilities in the exported agent runtime dependency closure

The 66 remaining Grype matches are inherited from the NVIDIA distroless Python 3.13 base and have no fix usable by this component today. Seven Python findings list only Python 3.15 alpha/beta/GA fixes, while the agent currently supports `>=3.13,<3.15`; the remaining Debian findings have no vendor-fixed version. These results are reported rather than suppressed.

## Test plan
- [x] built `services/agent/docker/Dockerfile` through `agent-runtime`
- [x] Grype scan of the fresh final image
- [x] pip-audit of the exported frozen runtime closure
- [x] runtime import/round-trip check for cryptography 50.0.0 and liblzma5 5.4.1-1+deb12u1
- [x] 2,971 agent unit tests
- [x] Ruff check and format check
- [x] mypy (131 source files)
- [x] no-patented-codecs image scan
- [x] `uv lock --check`

## Checklist
- [x] DCO sign-off included
## Supplied VSS 3.3 report reconciliation
Only `services/agent` findings were considered; UI/JavaScript and other-service findings are intentionally out of scope.

[...]

PR #1583, opened 2026-08-06
feat(vss_core): add nv.vss.memory/1.0 unified memory module

liamy-nv:
## Description
Land the contract and store/service adapters (in-memory + Elasticsearch) with unit tests so CLI persist/recall can stack on a smaller reviewable base.

## Checklist


liamy-nv:
/ok to test [ac1b0f1](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1583/commits/ac1b0f1be56211ac483e10ac5250495777c3666b)

services/agent/packages/vss_core/src/vss_core/memory/service.py
```diff
@@ -0,0 +1,228 @@
...
+        match: str | None = None,
+        limit: int = 50,
+    ) -> list[dict[str, Any]]:
+        """Extract event collections from persisted records for an asset.
+
+        Reads ``output.ext.events``, ``output.ext.incidents``, and
+        ``output.ext.results`` only — never pipeline Elasticsearch indices.
+        """
+        del window  # reserved for duration windows per design FR-5
+        records = self._store.query(MemoryQuery(sensor_id=asset_id, limit=_EVENTS_RECORD_SCAN_CAP))
```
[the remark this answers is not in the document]
liamy-nv: Acknowledged — intentional for v1. limit now applies only after event-level filters; _EVENTS_RECORD_SCAN_CAP bounds how many jobs we load per asset in one call. Proper fix is store-side pagination / filter pushdown (match, time range, anchor), tracked as follow-up. 10k jobs/asset is well above current expected usage.

services/agent/packages/vss_core/src/vss_core/memory/models.py
```diff
@@ -0,0 +1,183 @@
...
+PENDING_STATUSES: frozenset[str] = frozenset({"submitted", "running"})
+
+
+class MemoryGroupEnum(StrEnum):
+    """Enumerated job groups accepted by ``nv.vss.memory/1.0``."""
+
+    SUMMARY = "summary"
+    SEARCH = "search"
+    ALERT = "alert"
+    MEDIA = "media"
```
zac-wang-nv: **P2** — Can we name this `vios`?

Worth settling, and it's slightly tangled — three of the five group values already diverge from their command names: `summarize`→`summary`, `alerts`→`alert`, `vios`→`media`. The first two are grammatical (verb→noun, plural→singular); `vios`→`media` is the only *semantic* rename, which is presumably what makes it feel off.

It matters concretely because §3.1 specifies `job_id = <group>-<ULID>`, so a clip job produced by `vss vios clip run` mints `media-01J…`. (The doc muddles this itself — §5.2's example is `"summarize-01JZX8…"`, using the command name, not the group value.)

**Before renaming, though, the bigger question is whether this member should exist at all.** SDD v0.9 added **VIOS-8: *"VIOS operations don't write into memory"***, which contradicts three older sections that this PR implemented:
- §2 group notes: clip/snapshot are *"media-producing jobs … and **persists a common record**"*
- §3.1: *"Everything else is a point call (… vios, vlm): completes within the invocation, **writes one terminal record**"*
- §5.2: `media` is in the group enum at all

If VIOS-8 is the new decision, `media` is dead on arrival and should be dropped rather than renamed (along with `output.ext.kind` for clip/snapshot). If VIOS-8 was meant narrowly — say, sensor/stream CRUD doesn't persist but clip/snapshot jobs do — then the rename question stands and `vios` is the more honest name.

My read: `media` ages better *if* the group survives, since it describes what the record holds rather than which service produced it, and a future non-VIOS media source wouldn't need a second group. But that's a weak preference next to the VIOS-8 question, which decides whether there's anything to name.

services/agent/packages/vss_core/pyproject.toml
```diff
@@ -36,7 +36,7 @@ authors = [{ name = "NVIDIA Corporation" }]
 maintainers = [{ name = "NVIDIA Corporation" }]
 # The base install is intentionally light (NAT-free). Keep sorted!!!
 dependencies = [
-  "aiohttp~=3.14",
+  "aiohttp~=3.13",
```
zac-wang-nv: **P1** — Why do we downgrade from 3.14 to 3.13?

This reverts #1566 (`fix(agent): remediate current fixable CVEs`), which raised the floor specifically for CVE-2026-50269. The sibling manifest still says so: `services/agent/pyproject.toml:51` → `"aiohttp>=3.14.0",  # Fix CVE-2026-50269 (multipart header injection)`.

The change is in the feature commit itself (`fe70c04d9`), whose parent already had `~=3.14` — so it isn't merge fallout, and neither the commit message nor the PR description mentions a dependency change.

It also leaves the lockfile stale. On this head:

```
$ uv lock --check
The lockfile at `uv.lock` needs to be updated, but `--check` was provided.   # exit 1
```

Reverting only this one line makes it pass again (exit 0), so this line is the sole cause.

Two reasons it reached CI unflagged, both worth a follow-up beyond this PR: the jobs use `uv sync --frozen` (install from the existing lock, no manifest check) rather than `--locked`; and the pre-commit dependency gate is anchored to `^services/agent/(pyproject\.toml|uv\.lock)$` (`.pre-commit-config.yaml:74`), which doesn't match `packages/*/pyproject.toml` — so no nested workspace-package manifest is gated, even though those carry the library dependency floors.

Not P0: the workspace still resolves aiohttp **3.14.3**, because `services/agent`'s `>=3.14.0` governs and `--frozen` installs from the existing lock — so nothing ships vulnerable today. The exposure is the `vss_core` wheel's own advertised floor (it builds as a standalone distribution) plus the lock drift.

Suggest restoring `aiohttp~=3.14`. If the downgrade is deliberate, it needs a comment saying why the CVE floor is being relaxed, and a `uv lock` re-run.

services/agent/packages/vss_core/src/vss_core/memory/models.py
```diff
@@ -0,0 +1,183 @@
...
+    type: str = ""
+    info: dict[str, Any] = Field(default_factory=dict)
+
+
+class TimestampPoint(BaseModel):
+    """A single UTC ISO-8601 timestamp point."""
+
+    model_config = ConfigDict(extra="forbid")
+
+    timestamp: str
```
zac-wang-nv, line 80: **P2** — `str` is not a safe type for an ISO-8601 instant, and it is not just this field: every timestamp in the contract is a bare `str` with no validation — `job.created_at` and `job.updated_at` (`models.py:47-48`), this one, and the `since`/`until`/`start_time`/`end_time` filters on `MemoryQuery`, `JobFilters`, and `events()`. §5.1 specifies *"UTC ISO-8601; the record's temporal envelope"*, but the type enforces none of it:

```
JobInfo(created_at="banana", updated_at="not-a-timestamp-at-all")   # validates fine
TimestampPoint(timestamp="")                                        # validates fine
```

That matters more than usual here because the whole time-filtering path is **lexicographic string comparison** (`store.py:65`), which is only sound if every operand has an identical shape. Two ways that breaks with strings that are all legal ISO-8601 for the same instant:

```
'2026-07-22T12:00:00Z'        < '2026-07-22T12:00:00+00:00'   -> False   # 'Z'(0x5A) vs '+'(0x2B)
'2026-07-22T12:00:00.500000Z' < '2026-07-22T12:00:00Z'        -> True    # fractional sorts first
```

The second is reachable today, not hypothetical: `utc_now_iso()` (`adapters.py:35`) delegates to `datetime_to_iso8601`, which preserves fractional seconds, so real records carry microseconds while a caller-supplied bound is typically whole-second. Reproduced:

```
record created_at = 2026-07-22T12:00:00.500000Z
query(MemoryQuery(since="2026-07-22T12:00:00Z")) -> 0 hits      # record is AFTER the bound
```

Three consequences: filters silently return fewer rows rather than erroring; a garbage timestamp that passes pydantic will be *rejected by Elasticsearch* at index time once the field is date-mapped — so validation failure surfaces as a write error from the backend instead of a schema error at the boundary; and the two stores genuinely disagree. I confirmed that last one against a live Elasticsearch 9.4.4, same record, same query:

```
query                                                  ES  InMemory   agree?
----------------------------------------------------------------------------
since='2026-07-22T12:00:00Z' (record is …12:00:00.500000Z)  1     0   NO
until='2026-07-22T12:00:00Z'                                1     1   yes
```

ES maps `job.created_at` as a real `date` and gets it right in both directions; `InMemoryStore` drops the record on `since`. So the same `MemoryQuery` returns different results depending on which backend is configured — and the unit tests only exercise the one that's wrong.

Typing these as `AwareDatetime` (pydantic parses and normalizes, and `model_dump(mode="json")` still emits ISO-8601 on the wire, so §5.2's wire format is unchanged) fixes validation, ordering, and store divergence at one stroke. If the wire type must stay `str`, a validator calling the existing `iso8601_to_datetime` in `_foundation.time` gets most of it.

services/agent/packages/vss_core/src/vss_core/memory/elasticsearch.py
zac-wang-nv, line 1: **P2** — Same `backends/` question from the Elasticsearch side: `ElasticsearchMemoryStore` has its own top-level module while `InMemoryStore` sits inside `store.py` next to the protocol it implements. Both belong under `memory/backends/`, leaving `store.py` as the contract alone — the shape `vss_core/knowledge/` already uses (`base.py` + `factory.py` + `adapters/`).

Full rationale on the `store.py` comment in this review.

services/agent/packages/vss_core/src/vss_core/memory/store.py
```diff
@@ -0,0 +1,149 @@
...
+    if filters.group is not None and record.job.group != filters.group:
+        return False
+    if filters.status is not None and record.job.status != filters.status:
+        return False
+    if not _sensor_match(record, filters.sensor_id):
+        return False
+    return _time_in_range(record.job.created_at, filters.since, filters.until)
+
+
+class InMemoryStore:
```
zac-wang-nv, line 109: **P2** — `InMemoryStore` is a backend too, and it sits in a different place from the other one: `InMemoryStore` lives in `store.py` beside the `MemoryStore` protocol, while `ElasticsearchMemoryStore` gets its own top-level module. Worth moving both under `memory/backends/`.

This is not just symmetry — it's the convention the sibling package already uses. `vss_core/knowledge/` is exactly this shape:

```
knowledge/            memory/ (proposed)
  base.py       <->     store.py            # the contract only
  factory.py    <->     service.py          # build_memory_service, lazy backend import
  adapters/     <->     backends/           # concrete implementations
    es_caption.py         elasticsearch.py
    arango_graph.py       in_memory.py
    llama_index.py
```

and `vss_core/pyproject.toml:18-21` states the rationale in the package comment: *"factory.py resolves adapters via importlib, so heavy backend deps only load when a backend is selected"*. `memory/` already follows that rule (`build_memory_service` imports `elasticsearch` lazily) — a `backends/` package makes the rule visible in the tree instead of implied.

`backends/` is also the better name here specifically, even though `knowledge/` says `adapters/`: this module already has an `adapters.py` meaning something completely different (the group adapters that map payloads into records). Reusing `adapters/` for stores would collide with that.

Two things that make this cheap right now: nothing outside the module imports `vss_core.memory` yet, so there are no callers to update; and the only edits needed are the four paths in `__init__.py`'s `_LAZY_EXPORTS` plus the `TYPE_CHECKING` block. It gets meaningfully more expensive once the CLI PR and any harness plugins are importing these names.

zac-wang-nv:
**P2** — Conformance pass against the SDD (v0.9, 2026-08-10). Recording it here because the schema fidelity is genuinely high and the gaps are narrow — a note for the follow-up PR, not a request to expand this one.

**Matches the doc exactly**, verified rather than assumed: the §5.2 envelope and both `Literal` sets; *"legacy fields dropped (body.\*, origin, record_type, memory_id)"* — asserted precisely in `test_no_legacy_fields_in_serialization`; §5.3/UM-7 (`vss-memory`, ES `_id = job.job_id`, lifecycle updates one doc); UM-11 (`EmbeddingRef` by reference); §6.1's memory-first `status`/`get` with a single write-through poll and a never-polling `list`; §6.2's executors-are-not-a-registry; §2.1's three ext collections; and §8's no-process-env rule (grep for `os.environ`/`getenv` returns zero). The `events()` parameter list is a one-for-one match to §2.1's operation shape.

**One gap, and the doc moved after the code.** §5.2 v0.9 adds `"intent": "critic|introspection|video-qa"` to `input`, per the 2026-08-10 revision *"Add intent to the memory module"*. The PR head is 2026-08-09, so this isn't a miss — but it needs a coordinated bump, because `MemoryInput` is `extra="forbid"`: once any producer emits `intent`, `model_validate` raises rather than ignoring the field. Adding `intent: str | None = None` is a one-line, backward-compatible change and worth doing here if the doc is settled, since it costs a schema revision later.

**A related `extra="forbid"` edge:** §5.1 maps `input.sensors[]` to nvschema `nv.Sensor {id, type, description, location, coordinate, info}`, but `SensorInfo` carries only `{id, type, info}`. §5.2's own JSON also shows three, and `SearchAdapter.build_input` correctly folds unknown keys into `info` — so the adapters are fine. A raw nvschema-shaped sensor dict handed to `UnifiedMemoryRecord.model_validate` would hard-fail, though, which is worth a docstring note on the boundary.

**Doc-side drift worth a doc edit, not a code change:** §11 places the new code at `src/lib/memory/`, but §8 says *"sibling to search_core"* — `search_core` is at `vss_core/search_core/`, so `vss_core/memory/` satisfies the intent and the `lib.memory` path in §7.1/§8/§11 is stale.

**One genuine doc gap, surfaced by finding 4.** The record schema in §5.2 needs no change — the wire JSON and the pydantic models stay exactly as they are; `output.ext` remains an open dictionary. What's missing is a layer the SDD never mentions: §5.3 names the index (`vss-memory`, single writer) and what goes in it ("the full JSON record"), but says nothing about **how it is indexed**. That silence is what let the query code invent `.keyword` assumptions and a `multi_match` over object fields, with nothing to check them against.

This is doc work rather than PR work, so I am drafting it separately rather than asking for it here. Worth adding to §5.3: the index mapping is an explicit `_index_template` — `keyword` for the identity fields the filters use, `flattened` for the open subtrees (`output.ext`, `input.params`, `sensors[].info`) — provisioned by `vss configure memory` (§4.0 C5, UM-13) before the first write. And one line that follows from it and *is* a §5.2-level rule: **values inside the open dictionaries are keyword-only for query purposes** — anything needing numeric or date query semantics has to be promoted to a declared field. That constraint exists whether or not it's written down; better written down before groups start putting numbers in `params`.

Drafted as CM-1…CM-13 for §4.0, with a per-field mapping table and the evidence behind each: [SDD §4.0 configure-memory requirements](https://docs.google.com/spreadsheets/d/1_KtZ4ZDJ8iRvfLdEGRgyuUU1d_uYqI1VxOKo_6MlOGo/edit).

Correctly deferred, for the record: `vss memory lookup` (it's `lookup_by_slug` — markdown tier, which §5.3 explicitly excludes), the markdown notes tier itself, the alert/media/vlm adapters (§11 delivery items 5–7, with extensibility already proven by a registration test), and completion markers / exit codes / `--no-persist` (CLI layer).

services/agent/packages/vss_core/src/vss_core/memory/service.py
```diff
@@ -0,0 +1,228 @@
...
+    for i, event in enumerate(events):
+        for key in ("event_id", "id", "uuid"):
+            if str(event.get(key, "")) == anchor_event_id:
+                index = i
+                break
+        if index is not None:
+            break
+    if index is None:
+        return []
+    if direction == "before":
```
zac-wang-nv, line 208: **P1** — `direction="before"` and `"after"` are inverted, and `limit` then truncates from the wrong end.

`collected` is built by iterating `self._store.query(...)`, and both stores sort **newest-first** (`store.py:131` sorts `updated_at` `reverse=True`; `elasticsearch.py:174` sorts `desc`). So a lower list index means a *later* event — `events[:index]` is everything **after** the anchor in time, not before.

Reproduced on this head — three jobs at 10:00 / 11:00 / 12:00, one event each, anchored on the 11:00 event:

```
collected order (no anchor): ['evt-LATE', 'evt-ANCHOR', 'evt-EARLY']
direction="before" -> ['evt-LATE']     # 12:00 — chronologically AFTER the anchor
direction="after"  -> ['evt-EARLY']    # 10:00 — chronologically BEFORE the anchor
```

This is the exact question the feature exists to answer — SDD §2.1 motivates `memory events` with *"what happened just before that event?"*.

The `limit` slice at `service.py:142` compounds it: `collected[:limit]` takes from the head, so for `"before"` you get the events *furthest* from the anchor. With 10 events anchored on the oldest and `limit=3`:

```
before+limit=3 (anchor=e-00, the oldest) -> ['e-09', 'e-08', 'e-07']
```

Worth noting the deeper issue: `collected` is ordered by *job* recency and then by each record's `ext` list order, so it is never sorted by event time. Even with the branches swapped, "adjacent" would mean "adjacent in job-write order". Since §2.1 frames these as temporal follow-ups, this wants an explicit sort on the event timestamp before slicing.

`_adjacent_events` has no test coverage at all — one case per direction would have caught this.

services/agent/packages/vss_core/src/vss_core/memory/service.py
```diff
@@ -0,0 +1,228 @@
...
+                    collected.append(event)
+
+        if match:
+            needle = match.casefold()
+            collected = [item for item in collected if needle in str(item).casefold()]
+
+        if start_time or end_time:
+            collected = [item for item in collected if _event_in_window(item, start_time, end_time)]
+
+        if anchor_event_id:
```
zac-wang-nv, line 139: **P1** — When `anchor_event_id` matches no event, `_adjacent_events` returns `[]` (`service.py:206-207`), so `events()` returns an empty list — indistinguishable from "anchor found, no neighbours in that direction".

The SDD assigns that case its own exit code. §8: exit **5** = *"Handle not found (unknown `job_id`/`asset_id`/**`event_id`**)"*, whose harness policy is *"weak/conflicting memory hit → disambiguate"*. NFR-4 makes the same point generally: *"empty windows, weak handles, partial results distinguishable from success"*.

As written, a typo'd or stale `--anchor-event-id` produces exit 0 with `[]`, and the harness concludes "nothing happened before that event" rather than "that handle is unknown, disambiguate". The two call for opposite agent behaviour.

The no-records case already gets this right (`MemoryNotFoundError` at `service.py:117`), so this is just the anchor case needing the same treatment — raising `MemoryNotFoundError(f"anchor_event_id not found: {anchor_event_id!r}")` would make it consistent and map cleanly onto exit 5 when the CLI lands.

Greptile's thread 2 gestured at this ("an existing anchor to return an empty result") but read it as a symptom of the record cap; it is independent of the cap and survives the `_EVENTS_RECORD_SCAN_CAP` fix.

services/agent/packages/vss_core/src/vss_core/memory/elasticsearch.py
```diff
@@ -0,0 +1,179 @@
...
+            filters.append({"range": {"job.created_at": range_body}})
+        if text:
+            must.append(
+                {
+                    "multi_match": {
+                        "query": text,
+                        "fields": [
+                            "input.query",
+                            "output.answer",
+                            "output.ext.events",
```
zac-wang-nv, line 158: **P1** — These three fields are listed in the `multi_match` but are not searchable as written, so `MemoryQuery(text=...)` never matches anything inside `output.ext`.

Under dynamic mapping ES maps `output.ext.events` as an **object**, and only its leaves get concrete field names — `output.ext.events.description`, `output.ext.events.id`, and so on. A `multi_match` naming the object itself resolves to no concrete field, matches nothing, and raises no error.

Verified against a real Elasticsearch 9.4.4 (the version `deploy/` pins), indexing one record through `ElasticsearchMemoryStore` with one distinctive term per collection:

```
field list                              yellow   pallet   ramp
--------------------------------------------------------------
current (object names)                       0        0      0
output.ext.{events,results,incidents}.*      1        1      1
```

`InMemoryStore._matches_query` (`store.py:88-92`) stringifies those same three keys and does match, so the two implementations of one contract disagree — and because the ES tests use a fake that ignores the query body (finding 8), nothing catches it.

**Suggest deleting the three ext entries rather than fixing them.** They have never worked, `input.query` and `output.answer` cover the user-facing text, and §2.1's `--match` is unaffected either way: `events()` never sends `text` to the store, it queries by `sensor_id` and filters ext in Python at `service.py:132-134`. Nothing in the SDD requires `vss memory query --text` to reach ext.

Two notes if you do delete them. `store.py:88-92` should lose the matching ext haystacks at the same time, or the two stores still disagree, just in the opposite direction. And avoid the tempting wildcard — `output.ext.*` expands to numeric siblings like `output.ext.result_count` and errors (`failed to create query: For input string: "yellow"`) unless `lenient` is set, which then swallows real type errors.

Making ext genuinely searchable is a mapping question rather than a query one, and it is being specified separately as SDD requirements CM-1…CM-13 (`vss configure memory`) — out of scope here.

services/agent/packages/vss_core/tests/unit_test/memory/test_elasticsearch.py
```diff
@@ -0,0 +1,119 @@
...
+        return {"result": "created"}
+
+    def get(self, *, index: str, id: str) -> dict[str, Any]:
+        from elasticsearch import NotFoundError as ESNotFoundError
+
+        if id not in self.docs:
+            raise ESNotFoundError("not found", {}, {"_id": id})
+        return {"_id": id, "_source": self.docs[id]}
+
+    def search(self, *, index: str, body: dict[str, Any]) -> dict[str, Any]:
```
zac-wang-nv, line 66: **P2** — This fake ignores `body` entirely and returns every stored doc, so nothing in this file constrains the generated query. `test_build_search_body_filters` only asserts `body["size"] == 5` and `"bool" in body["query"]`, which holds for almost any query.

Confirmed by control experiment — I stripped **all** filters and the `multi_match` out of `_build_search_body` so it emitted a bare `match_all`:

```json
{"size": 5, "sort": [{"job.updated_at": {"order": "desc"}}], "query": {"bool": {"must": [{"match_all": {}}]}}}
```

All 3 tests still passed. So `len(store.query(MemoryQuery(group="summary"))) == 1` at line 100 is satisfied by the fake's unconditional return, not by group filtering.

Since `ElasticsearchMemoryStore` is the only production backend (`build_memory_service` selects it whenever `--es-endpoint` is set), this is the surface most worth pinning. Asserting on the built body directly — the `term` filters, the `range` on `job.created_at`, the `multi_match` field list — would cost a few lines and guard the part most likely to drift.

services/agent/packages/vss_core/src/vss_core/memory/service.py
```diff
@@ -0,0 +1,228 @@
...
+        window: str | None = None,
+        match: str | None = None,
+        limit: int = 50,
+    ) -> list[dict[str, Any]]:
+        """Extract event collections from persisted records for an asset.
+
+        Reads ``output.ext.events``, ``output.ext.incidents``, and
+        ``output.ext.results`` only — never pipeline Elasticsearch indices.
+        """
+        del window  # reserved for duration windows per design FR-5
```
zac-wang-nv, line 114: **P2** — `window` is accepted and silently discarded, so `--window 30s` on the anchor shape returns whatever the hardcoded ±5-event slice gives (`service.py:213-215`) with no signal that the bound was ignored. SDD §2.1 lists it as part of that shape: `--anchor-event-id <id> --direction {before,after,around} --window <dur>`. Same NFR-4 concern as finding 3 — a silently-dropped filter is not distinguishable from an applied one.

Either implement it or reject it at the boundary (exit 2) until it exists; accepting-and-ignoring is the one option that misleads the caller.

Minor, while you're here: the citation `per design FR-5` points at a requirement ID that no longer exists. FR-4/FR-5 were reinterpreted away at doc revision 0.5 (*"`events` dissolved as a separate group … FR-4/FR-5 reinterpreted"*); v0.9's requirement IDs are C\*, R-C\*, UM-\*, NFR-\*, VIOS-\*, VLM-\*. §2.1 is the stable anchor to cite.

services/agent/packages/vss_core/src/vss_core/memory/service.py
zac-wang-nv, line 115: **P2** — `start_time`/`end_time` are never pushed down to the index. Only `sensor_id` is, so this fetches every record for the asset up to the cap and filters temporally in Python at `service.py:137`. This is the ES query the call actually issues:

```json
{"size": 10000,
 "sort": [{"job.updated_at": {"order": "desc"}}],
 "query": {"bool": {"must": [{"match_all": {}}],
                    "filter": [{"term": {"input.sensors.id.keyword": "cam-1"}}]}}}
```

One `term` filter, no `range`. Every matched record is transferred whole and validated through pydantic — including the full `output.ext` payload the filter is about to discard. A 200-event summary record serializes to ~22 KB, so the cap implies roughly **225 MB in a single response**; `size: 10000` also sits exactly on Elasticsearch's default `index.max_result_window`, so there is no headroom.

**The obvious fix is wrong, which is worth writing down before someone tries it.** `MemoryQuery.since/until` maps to a range on `job.created_at` (`elasticsearch.py:149`) — when the *job ran*, not the period the events describe. A summary executed at 14:00 over 10:00–11:00 footage must match `events(start_time="10:00", end_time="11:00")`, but a `created_at` pushdown would drop it:

```
job.created_at   : 2026-07-22T14:00:00Z     <- since/until filters on this
input.window     : 2026-07-22T10:00:00Z -> 2026-07-22T11:00:00Z
event timestamp  : 2026-07-22T10:30:00Z     <- what start_time/end_time mean
```

The sound pruning key is `input.window` — §5.1 calls it *"the record's temporal envelope"* — via an overlap test (`window.end >= start_time AND window.start <= end_time`). A record whose window doesn't overlap the request cannot contain matching events. That needs a new field on `MemoryQuery` and a matching clause in `_build_search_body`; neither exists today, which is presumably why it wasn't done.

You already flagged the general shape of this in the Greptile thread (*"proper fix is store-side pagination / filter pushdown (match, time range, anchor)"*), so this isn't re-raising it — just pinning down which field is the correct one and why `since`/`until` isn't, so the follow-up doesn't start from the wrong end. Worth a code comment on the cap in the meantime.

services/agent/packages/vss_core/src/vss_core/memory/service.py
```diff
@@ -0,0 +1,228 @@
...
+    if es_endpoint:
+        from .elasticsearch import DEFAULT_MEMORY_INDEX
+        from .elasticsearch import ElasticsearchMemoryStore
+
+        es_store = ElasticsearchMemoryStore(
+            endpoint=es_endpoint,
+            index=memory_index or DEFAULT_MEMORY_INDEX,
+        )
+        return MemoryService(es_store, reconciler=reconciler)
+    raise ConfigurationError("memory service requires --es-endpoint (or an injected store); process env is not read")
```
zac-wang-nv, line 179: **P2** — Worth checking against SDD v0.8 (2026-08-06), which made persistence optional: §1.1 says *"The job persistence store is optional, if not there, the cli just return the result without persisting the job to the database."*

Here an unconfigured store is a hard `ConfigurationError` — exit 4 (*"fix config"*) per §8. That's the right call if the CLI is expected to catch it and fall back, but nothing in this module offers that fallback, and the no-arg `MemoryService()` path quietly defaults to `InMemoryStore()` instead — a *third* behaviour (writes that vanish at process exit). Three different answers to "no storage configured" is a lot of surface for the CLI PR to reconcile.

The same question applies to reachability: `ElasticsearchMemoryStore.upsert` raises `BackendUnreachableError` on ES failure, which maps to exit 3 (*"retry / report infra failure"*). §7.1 specifies the opposite for the persist stage — *"callback exceptions ⇒ exit 6 with `persisted: false` in the marker (retrieval succeeded — the harness must not re-run the job to retry persistence)"*. Exit 3 tells the harness to retry the whole job; exit 6 tells it to keep the answer. Worth deciding here, in the library, rather than papering over it at the CLI layer.

Not blocking — the CLI that consumes this doesn't exist yet, so this is a "settle it before wiring" flag rather than a defect.

services/agent/packages/vss_core/src/vss_core/memory/service.py
```diff
@@ -0,0 +1,228 @@
...
+
+
+def _event_in_window(event: dict[str, Any], start: str | None, end: str | None) -> bool:
+    stamp = event.get("timestamp") or event.get("start_time") or event.get("start") or event.get("ts")
```
zac-wang-nv, line 183: **P2** — When an event carries none of these four keys, `_event_in_window` returns `True`, so `start_time`/`end_time` silently pass it through.

Nothing normalizes or requires a timestamp on the way in: `SummaryAdapter.build_output` stores whatever dicts the caller hands it, and this PR's own fixtures (`{"id": "evt-1", "description": "forklift"}`) have no timestamp at all. So for the shapes the module currently produces, the window filter is a no-op — and §2.1's other headline use case, *"what happened between 00:01 and 00:03?"*, returns everything.

Related: `SearchAdapter.build_input` drops a window entirely unless **both** `start` and `end` are present (`adapters.py:277`), so a half-open range vanishes rather than being resolved or rejected — §3.3 has adapters resolving the effective range, and VIOS-4 says an omitted bound *"defaults to the covering segment, never to an invented timestamp"*.

Requiring a timestamp at the adapter boundary would fix the whole cluster at one seam, and pairs naturally with the typed-timestamp change in finding 7.

services/agent/packages/vss_core/src/vss_core/memory/models.py
```diff
@@ -0,0 +1,183 @@
...
+
+    model_config = ConfigDict(extra="forbid", populate_by_name=True)
+
+    schema_: Literal["nv.vss.memory/1.0"] = Field(default=SCHEMA_ID, alias="schema")
+    job: JobInfo
+    input: MemoryInput = Field(default_factory=MemoryInput)
+    output: MemoryOutput = Field(default_factory=MemoryOutput)
+    error: MemoryError | None = None
+
+    @model_validator(mode="after")
```
zac-wang-nv, line 165: **P2** — This validator's body is `return self` — its own comment explains it deliberately doesn't validate anything. It costs a pydantic hook on every record construction and reads as an unfinished check.

There is a real decision hiding behind it, and it connects to the group-naming question: §3.1 specifies `job_id = <group>-<ULID>`, but §5.2's example is `"summarize-01JZX8…"` while the `group` literal is `summary`. Either settle that and enforce it here, or drop the hook; leaving a no-op validator named after the rule suggests the rule is enforced when it isn't.

Three more in the same vein, flagged together since the repo guidelines call for nothing speculative:
- `models.py:28` `MemoryGroupEnum` — defined and listed in `__all__`, referenced nowhere in the tree (the `MemoryGroup` `Literal` is what's actually used).
- `service.py:219` `_PersistCallback` — assigned solely so the `Callable` import isn't unused. It looks like a placeholder for §7.1's `callback(record) → PersistResult` contract; if that's the intent, it belongs in the PR that implements the persist pipeline, typed properly.
- `store.py:42` / `adapters.py:40` `@runtime_checkable` — nothing calls `isinstance`/`issubclass` on `MemoryStore` or `MemoryAdapter` anywhere, so the decorator is inert. Worth dropping rather than keeping, because it advertises a guarantee it can't deliver: `isinstance` against a Protocol matches on **method names only**, so a class with the right four names and entirely wrong signatures passes.

  ```
  class Bogus:
      def upsert(self): return "not a record"
      def get(self): return 42
      def query(self): ...
      def list_jobs(self): ...

  isinstance(Bogus(), MemoryStore)  ->  True
  ```

  The static side is the part that works, and it does: adding a required parameter to `InMemoryStore.get` or `ElasticsearchMemoryStore.get` makes mypy fail at `service.py:56` and `service.py:178` respectively. Those two binding sites are the *only* thing enforcing conformance, since neither store inherits `MemoryStore`. A store never passed through a `MemoryStore`-typed slot would drift unnoticed, and parameter *names* aren't enforced either — renaming `get(job_id=…)` to `get(job_identifier=…)` passes mypy clean while breaking every keyword caller. If definition-site enforcement is wanted, this costs nothing at runtime:

  ```python
  if TYPE_CHECKING:
      def _conforms(x: ElasticsearchMemoryStore) -> MemoryStore:
          return x
  ```

services/agent/packages/vss_core/src/vss_core/memory/__init__.py
```diff
@@ -0,0 +1,82 @@
...
+    from .service import MemoryNotFoundError
+    from .service import MemoryService
+    from .service import build_memory_service
+    from .store import InMemoryStore
+    from .store import JobFilters
+    from .store import MemoryQuery
+    from .store import MemoryStore
+
+
+def __getattr__(name: str) -> Any:
```
zac-wang-nv, line 75: **NOTE** — Nice pattern, and it discharges NFR-2 (*"all groups run from the base `nvidia-vss` install"*) concretely: the module-level `__getattr__` keeps `import vss_core.memory` free of elasticsearch/NAT/torch, the `TYPE_CHECKING` block keeps type checkers and IDEs whole, and `test_bare_memory_import_is_nat_free` proves it in a **subprocess** rather than trusting `sys.modules` in the already-polluted test process. That last detail is what makes the guard real — it would catch a future import creeping into `models.py` or `store.py`.

Also worth calling out: `test_events_match_scans_beyond_former_record_cap` is a genuine regression test for the Greptile record-cap thread — it builds 250 decoy records and asserts the single match on the *oldest* one survives, which fails against the old `limit * 4` cap. Good follow-through on review feedback.

services/agent/packages/vss_core/src/vss_core/memory/models.py
liamy-nv: Removed media since VIOS-8 states "VIOS operations don’t write into memory" - can pick it up later if it is changed.

services/agent/packages/vss_core/pyproject.toml
liamy-nv: Not sure why this reverted, just fixed

services/agent/packages/vss_core/src/vss_core/memory/models.py
liamy-nv: Timestamps are now real UTC instants end-to-end

services/agent/packages/vss_core/src/vss_core/memory/store.py
liamy-nv: Moved both stores under memory/backends/ matching the knowledge layout

services/agent/packages/vss_core/src/vss_core/memory/service.py
liamy-nv: ack - left a comment on _EVENTS_RECORD_SCAN_CAP / events() noting time stays in-process for now, since/until must not be used (that's job.created_at), and the follow-up pushdown is input.window overlap

services/agent/packages/vss_core/src/vss_core/memory/service.py
liamy-nv: Ack — not changing behavior in this PR; contract for the CLI follow-up: optional persist = skip build_memory_service/upsert (ConfigError only when durable memory was requested without an endpoint); bare MemoryService()→InMemoryStore is test-only; persist-stage ES failures map to exit 6/persisted:false, not exit 3.

services/agent/packages/vss_core/src/vss_core/memory/service.py
liamy-nv: Fixed: SummaryAdapter.build_output now requires a parseable event timestamp; events() time filters fail closed on missing/unparseable stamps; SearchAdapter.build_input rejects half-open windows instead of silently dropping them.

services/agent/packages/vss_core/src/vss_core/memory/models.py
liamy-nv: Removed the no-op _validate_job_id_prefix hook

liamy-nv: Dropped MemoryGroupEnum, the _PersistCallback placeholder, and @runtime_checkable on MemoryStore/MemoryAdapter — none were used, and isinstance against those Protocols wasn’t a real guarantee.

services/agent/packages/vss_core/src/vss_core/memory/service.py
liamy-nv: Fixed: sort by event timestamp first; before/after are chronological; limit keeps neighbours closest to the anchor.

services/agent/packages/vss_core/src/vss_core/memory/service.py
liamy-nv: Fixed: Raises MemoryNotFoundError (exit 5 when CLI lands), not empty [].

services/agent/packages/vss_core/src/vss_core/memory/backends/elasticsearch.py
```diff
@@ -0,0 +1,179 @@
...
+            filters.append({"range": {"job.created_at": range_body}})
+        if text:
+            must.append(
+                {
+                    "multi_match": {
+                        "query": text,
+                        "fields": [
+                            "input.query",
+                            "output.answer",
+                            "output.ext.events",
```
liamy-nv: Fixed: Dropped dead output.ext.* fields from ES and matching ext haystacks from InMemoryStore so both stores only search input.query / output.answer.

services/agent/packages/vss_core/tests/unit_test/memory/test_elasticsearch.py
liamy-nv: Fixed: test_build_search_body_filters now asserts the full filter list, range, and multi_match fields.

services/agent/packages/vss_core/src/vss_core/memory/service.py
liamy-nv: Fixed: window=... raises ValueError until §2.1 duration support exists (no silent ignore).

liamy-nv:
/ok to test [1a92cfa](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1583/commits/1a92cfabd9c95d23aa82c3b4670af2d0cd766126)

services/agent/packages/vss_core/pyproject.toml
zac-wang-nv: Confirmed fixed — and worth knowing *why* it mattered, since "not sure why this reverted" undersells it: that one character was failing four required checks.

`Lint (Python)`, `Test (pytest)`, `Test (vss-cli)` and `Type Check (mypy)` all failed on `bd9f13bba` with `dns error ... artifactory.pdx.nvidia.com`. `uv run` implicitly re-locks when `pyproject.toml` drifts from `uv.lock`, and re-resolution queries every configured index — including the internal `ngc-data-platform` one, which GitHub-hosted runners can't reach. Develop never re-resolves, which is why it stayed green.

Control, offline: with `~=3.13`, `uv lock --check` exits 1; with `~=3.14` it exits 0 in 80 ms.

Confirmed in the pipeline now that your `/ok to test 1a92cfa` sweep has run: `Test (pytest)`, `Test (vss-cli)` and `Type Check (mypy)` are all green. `Lint (Python)` is still red, but for an unrelated reason — three ruff errors, noted in the general comment.

services/agent/packages/vss_core/src/vss_core/memory/models.py
zac-wang-nv: Resolved, thanks — and the removal is clean. Repo-wide census shows no remaining reference to the `media` group outside the new negative test, so nothing is orphaned. `test_media_group_rejected_per_vios8` pins the decision to VIOS-8, which is the right way to record it.

services/agent/packages/vss_core/src/vss_core/memory/models.py
zac-wang-nv: Fixed at the right seam, and checked rather than assumed. `IsoInstant` carries `PlainSerializer(..., when_used="json")`, so `model_dump_memory()` still emits ISO-8601 and §5.2's wire format is unchanged; the ES store re-serializes `since`/`until` through `datetime_to_iso8601` before building the `range`, so no `datetime` leaks into the query body; and `_time_in_range` now compares aware instants.

That last one closes the store divergence I reported — the `since='…T12:00:00Z'` vs record `…T12:00:00.500000Z` case that returned 1 hit on Elasticsearch and 0 in-memory now agrees. `coerce_utc_instant` rejecting naive datetimes at the boundary is a good addition I hadn't asked for.

services/agent/packages/vss_core/src/vss_core/memory/store.py
zac-wang-nv: Done, and the layout matches `vss_core/knowledge/` exactly — `store.py` is the contract alone. The `_LAZY_EXPORTS` and `TYPE_CHECKING` re-points are correct, and `test_bare_memory_import_is_nat_free` still passes, so the lazy-import boundary survived the move.

services/agent/packages/vss_core/src/vss_core/memory/backends/elasticsearch.py
zac-wang-nv: Closed by the same commit — `ElasticsearchMemoryStore` now sits under `backends/` alongside `InMemoryStore`.

services/agent/packages/vss_core/src/vss_core/memory/models.py
zac-wang-nv: All four gone — the no-op validator, `MemoryGroupEnum`, `_PersistCallback`, and `@runtime_checkable` on both protocols. Agreed on your reasoning for the last one; the static binding sites were what actually enforced conformance.

services/agent/packages/vss_core/src/vss_core/memory/service.py
zac-wang-nv: That comment is exactly what was needed — it records *which* field is the correct pruning key and why `since`/`until` isn't, so the follow-up doesn't start from the wrong end. No further ask here.

services/agent/packages/vss_core/src/vss_core/memory/service.py
zac-wang-nv: Good — that contract is the part that needed settling, and stating it here means the CLI PR inherits a decision instead of re-litigating it. Agreed on not changing behaviour in this PR.

services/agent/packages/vss_core/src/vss_core/memory/service.py
zac-wang-nv: Fixed, and the fix went past what I reported. I flagged in passing that `collected` was ordered by *job* recency, so "adjacent" would have meant adjacent in job-write order even with the branches swapped — `_sort_events_by_time` now sorts by event timestamp before slicing, which is the real fix. `before` also correctly returns the neighbours *closest* to the anchor rather than the furthest, which was the `limit` half of the finding.

Verified against your new guard tests: I checked out the previous head's `service.py` underneath them and both `test_events_adjacent_directions_are_chronological` and `test_events_before_limit_keeps_closest_neighbours` fail there and pass here, so they're guarding the behaviour rather than restating it.

Two consequences of the new sort worth deciding explicitly — see my comment on `service.py:176`: the unanchored path now returns oldest-first (the comment above it says newest-first), and `limit=0` now means unbounded there.

services/agent/packages/vss_core/src/vss_core/memory/service.py
zac-wang-nv: Fixed — `MemoryNotFoundError(f"anchor_event_id not found: {anchor_event_id!r}")` is exactly the shape that maps onto §8's exit 5, and `test_events_unknown_anchor_raises` fails against the previous head. Thanks for taking the whole finding rather than just the empty-list symptom.

services/agent/packages/vss_core/src/vss_core/memory/backends/elasticsearch.py
zac-wang-nv: Fixed the right way — deleted rather than patched, and importantly you did **both** halves: the three `output.ext.*` entries are gone from the `multi_match` *and* the matching haystacks are gone from `InMemoryStore._matches_query`. That second half is what keeps the two backends agreeing; dropping only the ES side would have left them disagreeing in the opposite direction.

One residual on this same field list, which I'd raised separately and is easier to see now that it's down to two entries: `input.intent` isn't in it. The field was added to the envelope in `1307c0743`, so free-text search still can't reach it. Now that `test_build_search_body_filters` pins the list by exact equality, adding it is a deliberate two-line change — worth folding in here.

services/agent/packages/vss_core/src/vss_core/memory/service.py
zac-wang-nv: Fixed — rejecting at the boundary is the right one of the two options; accepting-and-ignoring was the only genuinely misleading choice. The message naming §2.1 is a good touch. `test_events_window_rejected_until_implemented` fails against the previous head, so it's guarded.

services/agent/packages/vss_core/tests/unit_test/memory/test_elasticsearch.py
zac-wang-nv: Agreed, this one's closed — and you undersold it. `test_build_search_body_filters` now asserts `must`, the entire `filter` list and `sort` by exact equality, which is exactly the surface I was asking to pin, and asserting the whole list rather than indexing into it also avoids the positional brittleness I was about to raise.

The decisive check: the control experiment I ran against the old version — stripping every filter and the `multi_match` down to a bare `match_all` — passed all three tests then and would fail against this. `_FakeES.search` still ignores `body`, but with the built body fully constrained that's a cosmetic gap rather than a hole.

services/agent/packages/vss_core/src/vss_core/memory/service.py
zac-wang-nv: The direction is right and rejecting half-open windows is a good catch, but the guard covers one of the three collections the filter applies to, and the follow-up commit gave the same rows a second, opposite behaviour. Details and repros in my comment on `service.py:176` — short version: untimed rows are silently dropped when a time bound is present, and preferentially surfaced when it isn't.

services/agent/packages/vss_core/src/vss_core/memory/service.py
```diff
@@ -0,0 +1,312 @@
...
+        collected = _sort_events_by_time(collected)
+        if anchor_event_id:
+            return _adjacent_events(
+                collected,
+                anchor_event_id,
+                direction=direction or "around",
+                limit=max(limit, 0),
+            )
+        # Newest-first for unanchored recalls (matches prior store ordering intent).
+        return collected[-max(limit, 0) :] if limit else collected
```
zac-wang-nv: **P1** — Untimed event rows now have two opposite behaviours, and neither is signalled. One root cause: `events()` reads `events`, `incidents` and `results` (`service.py:146`), but only `SummaryAdapter.build_output`'s events pass through `_require_event_timestamp` (`adapters.py:233`). `incidents` come through the open `ext` passthrough and `results` are written by `SearchAdapter.build_output` (`adapters.py:317`) with no timestamp requirement at all.

**With a time bound they're silently dropped** (`_event_in_window` fails closed). Through the public adapter API, window covering all of 2026:

```
no time filter : ['evt-1', 'inc-1', 'res-1', 'res-2']
windowed       : ['evt-1']            # 3 of 4 dropped
```

**Without one they're preferred.** `_sort_events_by_time` sorts unparseable stamps last via `datetime.max`, and this line returns the tail — so untimed rows land exactly in the limit window. Ten real timestamped events plus one untimed search result:

```
limit=3 : ['e-08', 'e-09', 'res-UNTIMED']
limit=1 : ['res-UNTIMED']             # all ten real events evicted
```

So `vss memory events cam-1 --limit 1` returns the one row with no time information and hides every actual event. Both directions break the NFR-4 rule about partial results being distinguishable from success.

Cleanest fix is to finish the seam — require a timestamp in `SearchAdapter.build_output` the same way you did for `SummaryAdapter`, and decide what `incidents` must carry. Sorting untimed rows to `datetime.min` instead would at least stop them displacing real events, but it's a mitigation rather than a decision.

**Two smaller things on this same line.**

`return collected[-max(limit, 0):] if limit else collected` has two edge cases that used to return `[]`:

```
limit=0  -> everything   (falsy limit takes the else branch)
limit=-5 -> everything   (max(-5,0)==0, and collected[-0:] == collected[0:])
anchored, limit=0 -> []  (the two paths now disagree)
```

And the comment above it — `# Newest-first for unanchored recalls (matches prior store ordering intent)` — says the opposite of what the code does: `collected` is oldest→newest and `[-limit:]` preserves that, so it selects the newest N and returns them **oldest-first**, where the previous behaviour returned newest-first. The selection is defensible; worth either fixing the comment or reversing the slice, and stating which order `events()` guarantees.

Worth a test for each: nothing currently exercises `incidents`/`results` under a time bound, or an unanchored `limit`.

services/agent/packages/vss_core/src/vss_core/memory/adapters.py
```diff
@@ -0,0 +1,393 @@
...
+        video_id: str | None,
+        media_ref: dict[str, Any] | None,
+        params: dict[str, Any] | None,
+        window: TimeWindow | None = None,
+    ) -> MemoryInput:
+        sensors: list[SensorInfo] = []
+        if video_id:
+            info = dict(media_ref or {})
+            sensors.append(SensorInfo(id=str(video_id), type=str(info.get("source") or "video"), info=info))
+        return MemoryInput(query=prompt, sensors=sensors, window=window, params=dict(params or {}))
```
zac-wang-nv: **P2** — `input.intent` validates and round-trips, but nothing in the module can populate it. `SummaryAdapter.build_input` (`adapters.py:221`) and `SearchAdapter.build_input` (`adapters.py:293`) both construct `MemoryInput(query=…, sensors=…, window=…, params=…)` without it, so only direct `MemoryInput(...)` construction can set the field.

The schema half of SDD v0.9 is done; reachability isn't. Either thread it through both adapters or say explicitly that the CLI PR owns that wiring.

zac-wang-nv (commented):
Re-reviewed at `1a92cfabd`. **All 14 findings from round 1 are now closed** — four P1s and ten P2s, across three heads in six hours. Thanks; this was a genuinely thorough response, and in two places the fix went past what I reported: the event-time sort behind the adjacency fix, and dropping the `output.ext` haystacks from `InMemoryStore` as well as from the `multi_match`. The ES query-body assertions are also stronger than your note suggested.

I verified the P1 fixes rather than taking the commit subject: I checked out the previous head's `service.py`, `backends/elasticsearch.py` and `backends/in_memory.py` underneath your four new guard tests, and all four fail there and pass here — so the tests guard the behaviour rather than restate it. Suite is 28 passing, up from 17 at round 1.

**Two P1s keep this at REQUEST CHANGES, and both post-date round 1.**

**1. `Lint (Python)` is red** — thanks for running the sweep. `ruff check` exits 1 with three errors, reproduced locally:

```
backends/elasticsearch.py:10:1  I001  [*] Import block is un-sorted or un-formatted
service.py:46:74                RUF003     Comment contains ambiguous `–` (EN DASH)
tests/…/test_memory.py:5:1      I001  [*] Import block is un-sorted or un-formatted
```

`ruff check --fix` clears the two I001s; RUF003 is a one-character edit — the en-dash is in `A summary run at 14:00 over 10:00–11:00` at `service.py:46`, inside the `_EVENTS_RECORD_SCAN_CAP` comment. The comment itself is right and worth keeping. Two of the three predate this commit (the rename reordered the imports), and they went unseen because no sweep ran on `dfd95c487`.

**2. Untimed rows in `output.ext` now have two opposite behaviours** — silently dropped when a time bound is present, and preferentially surfaced when it isn't. `--limit 1` over ten timestamped events plus one untimed search result returns only the untimed row. Repros on the `service.py` comment, along with two smaller items on the same line (`limit=0`/negative now means unbounded on the unanchored path, and the ordering comment says the opposite of what the slice does).

Also still open and small: `input.intent` is inert — neither adapter sets it, and it isn't in the `multi_match` list. Noted on its thread and in a new comment.

**The rest of CI is good news:** `Test (pytest)`, `Test (vss-cli)` and `Type Check (mypy)` are green, which confirms the aiohttp fix in the pipeline — 42 success / 8 skipped / 1 failure across 51 checks.

Note for the record — my round-1 summary said CI was green. That was wrong: it counted a partially-reported check board. The head was red at 51 checks with 4 failures, and the aiohttp pin was the cause.

zac-wang-nv:
**P2 / design** — raising this as a direction to settle in the SDD rather than a change to make in this PR.

Nesting events inside a summarization record's `output.ext.events[]` makes event-time search structurally impossible. `MemoryQuery.since`/`until` range on `job.created_at` — when the job *ran* — while the events' own instants live in an open `dict[str, Any]` subtree with no mapping. That's why `events()` fetches up to 10,000 full records and filters them in Python, and it's the root cause of the untimed-row problem on the `service.py` comment.

Promoting events to first-class records in the same index — same `job_id`, distinguished by a record kind — would make a time-range query a real ES `range` and per-event anchoring a lookup rather than a scan.

**I checked whether the schema already has a discriminator for this. It doesn't.** The four near-misses and why none fits:

| Field | What it is |
|---|---|
| `job.group` | `Literal["summary","search","alert","vlm"]` — which subsystem produced it, wrong axis |
| `job.operation` | `Literal["run"]` — **single-valued today** |
| `SensorInfo.type` | the sensor's type, nested in `input.sensors[]` |
| `EmbeddingRef.kind` | the embedding's kind, nested in `output.Embedding[]` |

`job.operation` is the cheapest home — already in the envelope, already indexed, and currently admits exactly one value, so widening it to `Literal["run","event"]` introduces no conflict. It reads slightly off semantically ("operation" = what the job did), so the alternative is a real top-level `kind` at a higher cost — see (1) below.

**Two constraints worth naming before anyone starts:**

1. **`record_type` was deliberately dropped, and its absence is pinned by a test.** §5.2 lists *"legacy fields dropped (body.\*, origin, record_type, memory_id)"* and `test_no_legacy_fields_in_serialization` asserts it never appears in the serialized blob. So a record kind isn't a code change — it reverses a documented decision and needs a schema revision plus that test updated.
2. **`_id` equals `job.job_id`, so same-`job_id` siblings would clobber each other.** `elasticsearch.py:68` calls `index(id=record.job.job_id)`, and the module docstring says lifecycle transitions upsert the same document. Event records sharing the parent's `job_id` would overwrite the parent summary and each other, so this needs a compound `_id` (`{job_id}#event#{n}`, or the event's own id).

Worth doing, in my view — it dissolves the record-cap/pushdown finding and the untimed-row finding at the root instead of patching both. And it lands squarely in the §5.3 gap I raised in round 1: the doc names the index and what goes in it but says nothing about *how* it's mapped, which is what let the query layer invent assumptions. The `vss configure memory` requirements (CM-1…CM-13) are where the index template for a mapped event-time field would live.

No action needed in this PR — flagging it so the CLI follow-up doesn't build on the nested shape.

services/agent/packages/vss_core/src/vss_core/memory/service.py
liamy-nv: Untimed rows — SearchAdapter.build_output and summary ext.incidents now require timestamps (same seam as events). Untimed rows sort to datetime.min so they can’t steal newest-limit. Unanchored path returns newest-first; limit<=0 → [] on both paths.

services/agent/packages/vss_core/src/vss_core/memory/adapters.py
liamy-nv: Intent - Threaded through SummaryAdapter.build_input and SearchAdapter.build_input.

liamy-nv:
/ok to test [06d5bbf](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1583/commits/06d5bbf69119540a7083fc0bae671b017610696f)

zac-wang-nv (dismissed):

zac-wang-nv:
@liamy-nv  can you check the CI/Lint issue?

liamy-nv:
/ok to test [431c45c](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1583/commits/431c45c2c080f2824105aead18d3ca0b0d5e7841)

zac-wang-nv (approved):
LGTM

PR #1584, opened 2026-08-06
ci(skills-eval): pin daily runs to nightly commit

sarathm-nvidia:
## Summary

Updates the scheduled Skills Eval Daily workflow to evaluate a stable commit associated with the UTC-dated nightly tag.

## Behavior

For scheduled runs:

1. Derive `nightly-YYYYMMDD` using the current UTC date.
2. Resolve the tag to its underlying commit SHA.
3. If the exact tag is unavailable, resolve the current `develop` commit SHA.
4. Pass the resolved SHA as the ref for all downstream jobs, ensuring matrix planning and every evaluation leg use the same commit.

Manual workflow dispatches continue using the branch or tag selected in the GitHub UI.

Resolving to a SHA ensures the entire run remains pinned to one commit even if a lightweight nightly tag is moved later.

## Implementation

- Adds a modular `resolve-skills-eval-ref.sh` resolver.
- Handles lightweight and annotated tags, including peeling annotated tags to their commit SHA.
- Fails on authentication or transport errors rather than silently falling back.
- Records the selected tag/branch and resolved SHA in the workflow logs.
- Loads the resolver from the same default-branch commit as the scheduled workflow.

## Validation

- Nightly annotated tag resolves to its underlying commit SHA.
- Missing exact nightly tag falls back to the `develop` commit SHA.
- A neighboring nightly tag is not selected.
- Bash syntax and workflow ref-routing checks pass.

## Follow-up recommendation

The nightly tag publisher should create annotated tags and protect the `nightly-*` namespace against updates or deletion. Annotated tags improve traceability; protection provides immutability.

### DIFF
--- a/.github/scripts/resolve-skills-eval-ref.sh
+++ b/.github/scripts/resolve-skills-eval-ref.sh
@@ -1,0 +1,88 @@
+#!/usr/bin/env bash
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+# Resolve the source ref for the daily skills evaluation.
+#
+# Usage: resolve-skills-eval-ref.sh [YYYYMMDD]
+#
+# Environment overrides keep the policy easy to adjust without coupling it to
+# the workflow:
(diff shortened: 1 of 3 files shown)


zac-wang-nv (approved):
LGTM

PR #1592, opened 2026-08-07
hermes integration fixes

nvleeryan:
## Description

- Harness profiles for OpenClaw and Hermes — Refactors deploy nemoclaw notebooks to sandbox-first CLI templates driven by AGENT_HARNESS_PROFILES, and enables Hermes webhook support.
- Configurable AGENT_RUNTIME from the notebook — Lets you pick openclaw or hermes in the notebook (or via env), with derived CLI/workspace/UI settings from the harness profile.
- Drop hardcoded nemoclaw / nemohermes branches — Replaces scattered runtime conditionals with profile-derived commands and labels so both harnesses share one path.
- Restart gateway after skill install — Restarts the agent gateway after uploading skills (with recover fallback) so new skills load without relying on a later optional step.
- Auto-detect sandbox name — Orchestrator notebook detects the NemoClaw sandbox name from the live sandbox list instead of requiring manual config.
- Fix vision-agent skill frontmatter — Quotes the multiline YAML description so unquoted vision: no longer breaks frontmatter parsing.

## Checklist


deploy/docker/scripts/deploy_nemoclaw.ipynb
```diff
@@ -208,23 +211,100 @@
...
+        "    \"mcp_status_cmd\": \"{cli} {sandbox} mcp status {server}\",\n",
+        "    \"mcp_add_cmd\": \"{cli} {sandbox} mcp add {server} --url {url}\",\n",
+        "    \"mcp_remove_cmd\": \"{cli} {sandbox} mcp remove {server}\",\n",
+        "    \"config_set_cmd\": \"{cli} {sandbox} config set --key {key} --value {value} --config-accept-new-path\",\n",
+        "    \"gateway_restart_cmd\": \"{cli} {sandbox} gateway restart\",\n",
+        "    \"recover_cmd\": \"{cli} {sandbox} recover\",\n",
+        "    \"gateway_token_cmd\": \"{cli} {sandbox} gateway-token --quiet\",\n",
+        "    \"dashboard_url_cmd\": \"{cli} {sandbox} dashboard-url --quiet\",\n",
+        "    \"connect_cmd\": \"{cli} {sandbox} connect\",\n",
+        "}\n",
```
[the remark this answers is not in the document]
nvleeryan: Not a bug — this matches the current NemoClaw CLI grammar (v0.0.80+).

Upstream switched to a sandbox-name-first form: {cli} <sandbox> <subcommand> (e.g. nemoclaw demo skill install …, nemohermes demo connect). Global commands stay unprefixed (nemoclaw list, nemohermes onboard); sandbox-scoped commands put the name first.

OpenClaw (nemoclaw) and Hermes (nemohermes) share that same grammar — only the binary differs. Confirmed by both CLIs’ help: “sandbox commands start with a sandbox name.”

The older nemoclaw sandbox <cmd> <name> form is obsolete. _AGENT_SANDBOX_CMDS / AGENT_HARNESS_PROFILES intentionally encode the new shared templates.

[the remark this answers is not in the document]
nvleeryan: tested both nemoclaw and nemohermes and both worked and were able to deploy base profile from the nemoclaw agent E2E

deploy/docker/scripts/deploy_vss_orchestrator.ipynb
```diff
@@ -1013,7 +1041,7 @@
         "exit\n",
         "\"\"\".strip() + \"\\n\"\n",
         "\n",
-        "CONNECT_CLI = \"nemohermes\" if AGENT_RUNTIME == \"hermes\" else \"nemoclaw\"\n",
+        "CONNECT_CLI = \"nemoclaw\"\n",
         "print(f\"Running reachability probe through: {CONNECT_CLI} {NEMOCLAW_SANDBOX_NAME} connect\")\n",
         "result = subprocess.run(\n",
         "    [CONNECT_CLI, NEMOCLAW_SANDBOX_NAME, \"connect\"],\n",
```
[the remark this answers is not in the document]
nvleeryan: nemohermes is not a separate CLI; it's an alias entry point of the same npm package with the Hermes agent preselected, and connect reads the agent from the sandbox registry rather than from the binary you typed.  

They're the same implementation. Both ~/.local/bin/* wrappers exec into node_modules/nemoclaw/bin/, and the Hermes entry point is three lines:

nemohermes.js Lines 1-9
`// NemoHermes — alias for NemoClaw with the Hermes agent pre-selected.
process.env.NEMOCLAW_AGENT = "hermes";
process.env.NEMOCLAW_INVOKED_AS = "nemohermes";
module.exports = require("../dist/nemoclaw");`

Same dist/nemoclaw bundle, so every subcommand and its grammar are identical. Since both bins ship in the one nemoclaw package, nemoclaw is always on PATH whenever nemohermes is.

connect resolves the harness from the registry. The probe calls agentRuntime.getSessionAgent(sandboxName), which is registry-first (registry.getSandbox(name) → getRegisteredAgent(sb)), and Hermes-specific work keys off the stored field, sb.agent !== "hermes" — never the invoked binary. So nemoclaw demo connect against a Hermes sandbox drives the Hermes gateway.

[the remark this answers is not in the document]
nvleeryan: yes

nvleeryan:
/ok to test https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1592/commits/f419dd1bcddea71d09d34d296d0350e3c442234b

zac-wang-nv (commented):
there seems to be some noise from the notebook, can you clean those temporary/local fields(id, etc) ?

nvleeryan:
@zac-wang-nv - i guess by noise you meant what I fixed [here](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1592/changes/06938c9869d6270880c6a11359258cdb26ec94c6)

deploy/docker/scripts/deploy_nemoclaw.ipynb
```diff
@@ -828,22 +919,37 @@
       "metadata": {},
       "outputs": [],
       "source": [
-        "!nemoclaw sandbox mcp {NEMOCLAW_SANDBOX_NAME} status vss_orchestrator >/dev/null 2>&1\n",
-        "if _exit_code == 0:\n",
-        "    print(\"MCP server 'vss_orchestrator' already registered.\", flush=True)\n",
+        "if not ORCHESTRATOR_ENABLE_HTTPS:\n",
```
zac-wang-nv, line 922: **P2** — The gate itself is right: with HTTPS off the agent still reaches the locally deployed orchestrator MCP that `deploy_vss_orchestrator.ipynb` configures and starts, so skipping the `mcp add` here is correct. The gap is documentation — at the one place a reader actually hits the skip, the notebook says only that the cell "does nothing", not where the MCP then comes from.

Suggest adding a short note right after **3.5 Register the VSS Orchestrator MCP** along the lines of:

> With `ORCHESTRATOR_ENABLE_HTTPS = False` (the default) this section is a no-op: the agent uses the locally deployed orchestrator MCP served by `deploy_vss_orchestrator.ipynb` (sections 3.3–4), so no registration is needed here. Set `ORCHESTRATOR_ENABLE_HTTPS = True` in **both** notebooks only if you want to register an HTTPS MCP endpoint with the sandbox instead.

1.3 already carries half of this — *"The server itself is configured and started in `deploy_vss_orchestrator.ipynb`"* (line 171) — but that is ~750 lines earlier, and the 3.5 cell text at line 910 stops at *"running this cell does nothing in that case"*, which reads like something is missing rather than handled elsewhere.

While you are in there, three prose spots still describe the pre-`a4c2a9eab` behaviour and now contradict the default path:
- `deploy_nemoclaw.ipynb:10` (cell 0) — *"register or confirm the VSS Orchestrator MCP path"* — and `:26` (overview) — *"register or confirm VSS Orchestrator MCP access"*
- `deploy_vss_orchestrator.ipynb:87` — *"that notebook registers the MCP URL with the sandbox"* (stated unconditionally)
- `deploy_vss_orchestrator.ipynb:9` — lists *"the `vss_orchestrator` MCP is registered"* as an unconditional precondition

`deploy/docker/scripts/nemoclaw/README.md:62` is in the same category — it documents registering over plain http, which is now the case that needs no registration at all. That file is already covered by finding 2.

zac-wang-nv:
**P2** — The CLI-grammar migration stops at the notebooks; `deploy/docker/scripts/nemoclaw/README.md` still documents the old form throughout.

Per your rebuttal on the Greptile thread, `nemoclaw sandbox <cmd> <name>` is obsolete as of v0.0.80+ — the version this PR pins in `NEMOCLAW_INSTALL_REF`. That README is the repo's stated reference for the flow (it has a "Why canonical" section), and all nine of its commands are in the obsolete grammar:

- `nemoclaw sandbox policy add "$SB"` (line 45) → `nemoclaw "$SB" policy-add`
- `nemoclaw sandbox skill install "$SB" "$skill"` (49) → `nemoclaw "$SB" skill install "$skill"`
- `nemoclaw sandbox upload "$SB" …` (55, 58) → `nemoclaw "$SB" upload …`
- `nemoclaw sandbox mcp "$SB" add …` (62) → `nemoclaw "$SB" mcp add …`
- `nemoclaw sandbox config set "$SB" …` (68) → `nemoclaw "$SB" config set …`
- `nemoclaw sandbox gateway token "$SB"` (73) → `nemoclaw "$SB" gateway-token`
- plus the prose at lines 11–12 and 17

`assets/vss_nemoclaw_policy.yaml:1` already uses the new grammar, so the README is now the only place left on the old one. Updating it here keeps the migration in one commit; if you would rather split it out, that is reasonable too — but it should not sit stale.

skills/vss-build-vision-agent/SKILL.md
```diff
@@ -1,7 +1,16 @@
 ---
 name: vss-build-vision-agent
-description: >-
-  Add agent-ready vision capabilities — dense captioning, detection, search, alerting, summarization — to an agent or application through a customizable, self-contained vision stack built on the NVIDIA VSS Blueprint. Use this skill when a developer or agent wants to give their app vision: pick capabilities via guided intake ("build a vision agent", "add vision capabilities") or describe them in natural language ("create a profile for streaming dense captioning", "add agentic search to my base deployment"). Route, compose, configure, and deploy stock base, alerts, LVS, or search developer profiles and lean custom combinations expressed as delta overlays using one current developer profile as the Foundation.
+description: >
```
zac-wang-nv: **P2** — I think this file can be dropped from the PR entirely; the bug it fixes is already fixed on `develop`.

The branch state you fixed had `description:` as a plain scalar containing `app vision: pick`, which is genuinely unparseable — so the fix was correct when you made it. But #1586 ("fix(skills): fix build vision agent frontmatter") landed independently and is now the merge-base version: `git rev-parse 635efd682:skills/vss-build-vision-agent/SKILL.md` is blob `dab6dc325`, already using `>-`.

I parsed both versions with `yaml.safe_load`: merge-base and head both succeed, same keys, same `name`. The only net difference is the chomping indicator — `>-` (merge-base) strips the trailing newline, `>` (head) keeps one — so the head's description is the same 713 characters plus a trailing `\n`, identical after `.strip()`, with the rest being a line re-wrap.

The reason it is worth dropping rather than leaving: touching this file is what expands the eval matrix to eight `vss-build-vision-agent` Harbor legs on this PR — the ones that burned ~40 minutes apiece at the previous head and failed on pool-box SSH and full disks. A no-functional-change re-wrap is a lot of GPU time for nothing.

deploy/docker/scripts/deploy_vss_orchestrator.ipynb
```diff
@@ -955,7 +937,8 @@
         "\n",
         "<span style=\"color:red\"><strong>Important:</strong> make sure VSS is already deployed on the host before running this step.</span>\n",
         "\n",
-        "Run the next cell only after the agent has finished deploying VSS in **section 5**. It runs `nemoclaw` / `nemohermes <sandbox> connect` and feeds a probe script into that real sandbox session. Do **not** replace this with `docker exec`; that can use a different path and produce misleading results.\n",
+        "Run the next cell only after the agent has finished deploying VSS in **section 5**. It auto-detects the default sandbox (and prints its onboarded harness) from `nemoclaw list --json` (override via `NEMOCLAW_SANDBOX_NAME`), then runs `nemoclaw <sandbox> connect` and feeds a probe script into that real sandbox session.\n",
```
zac-wang-nv: **P2** — The parenthetical is stale. `f419dd1bc` removed the `AGENT_RUNTIME` lookup and the `print("AGENT_RUNTIME:", ...)` line from this cell; I re-checked at this head and the cell's only identity print is `print("Sandbox:", NEMOCLAW_SANDBOX_NAME)` — `AGENT_RUNTIME` no longer appears in it at all. Nothing reports the onboarded harness, so suggest dropping "(and prints its onboarded harness)".

Same cell, unrelated and very minor: `_nemoclaw_list_payload()` handles `FileNotFoundError` and `CalledProcessError` but not `json.JSONDecodeError`, so a CLI that emits a banner before its JSON would surface a raw decode traceback rather than the "set `NEMOCLAW_SANDBOX_NAME` to override" hint the other two paths give.

deploy/docker/scripts/deploy_nemoclaw.ipynb
```diff
@@ -208,23 +211,100 @@
...
+        "        \"onboard_args\": \"--non-interactive\",\n",
+        "        \"retry_onboard_on_fail\": True,\n",
+        "        \"workspace_remote_dir\": \"/sandbox\",\n",
+        "        \"ui_mode\": \"dashboard_url\",  # plain URL; gateway-token is the :8642 API bearer\n",
+        "        \"verify_kind\": \"sandbox_docs\",\n",
+        "        \"verify_cmds\": (\n",
+        "            (\"{label} top-level docs\", \"ls -1 {workspace_remote_dir}/*.md 2>/dev/null || true\"),\n",
+        "            (\n",
+        "                \"VSS Orchestrator MCP host alias\",\n",
+        "                \"curl -s -o /dev/null --max-time 5 http://host.openshell.internal:9988/ && echo host alias reachable || true\",\n",
```
zac-wang-nv: **P2** — The host alias and MCP port are hardcoded here while every neighbouring field in the profile table is derived, so overriding `HOST_INTERNAL_ALIAS` or `VSS_ORCHESTRATOR_MCP_PORT` leaves this Hermes verification probing the wrong address and reporting "reachable" about a host the deployment is not using.

The literal came over verbatim from the old inline branch, so this is not a regression — but the values are only defined below the table, which is likely why it was left. A `{host_alias}` / `{mcp_port}` placeholder filled at `.format()` time alongside `{workspace_remote_dir}` would close it.

nvleeryan:
/ok to test https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1592/commits/872e9dd32bcbf7d1300d0ee6bc409916bd117ea9

deploy/docker/scripts/deploy_nemoclaw.ipynb
```diff
@@ -208,23 +211,100 @@
...
+        "        \"onboard_args\": \"--non-interactive\",\n",
+        "        \"retry_onboard_on_fail\": True,\n",
+        "        \"workspace_remote_dir\": \"/sandbox\",\n",
+        "        \"ui_mode\": \"dashboard_url\",  # plain URL; gateway-token is the :8642 API bearer\n",
+        "        \"verify_kind\": \"sandbox_docs\",\n",
+        "        \"verify_cmds\": (\n",
+        "            (\"{label} top-level docs\", \"ls -1 {workspace_remote_dir}/*.md 2>/dev/null || true\"),\n",
+        "            (\n",
+        "                \"VSS Orchestrator MCP host alias\",\n",
+        "                \"curl -s -o /dev/null --max-time 5 http://{host_alias}:{mcp_port}/ && echo host alias reachable || true\",\n",
```
zac-wang-nv: **P2** — The alias and port are derived now, which closes the part I raised. The scheme is the one value still hardcoded, and it is derived two lines of context away: `MCP_SCHEME = "https" if ORCHESTRATOR_ENABLE_HTTPS else "http"`, and `ORCHESTRATOR_MCP_URL` is built from it.

So with `ORCHESTRATOR_ENABLE_HTTPS = True` the orchestrator MCP listens on `https://{alias}:{port}`, while this Hermes verification still probes `http://`. curl gets a TLS listener speaking plaintext, exits non-zero, and the `|| true` swallows it — the operator simply never sees the `host alias reachable` line and has no indication the check was asking the wrong question.

Impact is genuinely small: this is section 3.8 diagnostics, not the deploy path, and it only misreports in the non-default HTTPS configuration. Adding `scheme=MCP_SCHEME` to the `_verify_fmt` dict and using `{scheme}://{host_alias}:{mcp_port}/` would finish the job in the same shape as the fix you just made.

nvleeryan:
/ok to test https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1592/commits/36398a392fb57f5646ff853abdba0b54b029cee5

zac-wang-nv (approved):
LGTM

PR #1602, opened 2026-08-07
Support float violation incident configs and cover 3d incidents

yunHe95:
- Upgrade the distroless base image to 3.13-v4.0.9
- Accept fractional seconds for the violation incident threshold and expiration window; only the FOV object count stays an integer. The int-only rule rejected the values the docs already recommended.
- Lower the default expiration window to 0.5s so it sits below the 1s threshold; at window >= threshold one tolerated gap spans the whole threshold, so two isolated detections raise a full-duration incident.
- Enable the three incident types in the warehouse-3d deploy profile and integration fixture, with explicit threshold/expiration values
- Compare mdx-incidents for warehouse_3d. Object ids after the primary come from a set, so their order varies between processes; compare the remainder as a multiset rather than a list.

## Checklist


kartikayt-nvidia (approved):

zac-wang-nv (approved):
OSRB good

PR #1615, opened 2026-08-07
DEBUG: probe the downstream gate expression inputs (do not merge)

zac-wang-nv:
**Temporary diagnostic — do not merge.**

`trigger-downstream-pipeline` has skipped on every PR since #1567 (2026-08-05), including runs where the release-set step prints `Downstream gate: run`. 8/8 such runs skipped; before #1567 the job ran.

Ruled out so far:
- **Not the dependency graph.** `wait-for-build-dev-images` never skips — 10/10 `success`. Its `always()` rescue works.
- **Not a failed sibling.** All 12 direct needs of the trigger job are `success`.
- **Not an empty output.** GitHub logs `Set output 'run-downstream'`.

Since `success()` should therefore be true, the remaining suspect is the value comparison `needs.wait-for-build-dev-images.outputs.run-downstream == 'true'`.

This job mirrors the trigger job's `needs` exactly but runs under `always()`, and prints the literal strings the expression engine sees — the raw value, the equality result, `success()`, and `toJSON` of the wait job outputs and the full `needs` context. `containers.env` is touched so the gate resolves to `run`.

Fix PR: #1614 (works regardless of mechanism, but its description asserts a cause I could not defend and needs correcting once this lands).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1417,6 +1417,20 @@ jobs:
               "${{ needs.spatialai-data-utils-test.outputs.package-version-suffix }}" \
             --release-set-output downstream-release-set.json
 
+      # TEMPORARY probe -- do not merge. Shows the literal bytes the script wrote,
+      # which is what the gate comparison actually reads.
+      - name: Probe release-set step outputs
+        if: always()
+        env:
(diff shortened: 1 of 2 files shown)


zac-wang-nv:
closed

PR #1623, opened 2026-08-10
Enable GHCR builds for SDRC, VSS configurator and VSS RT Config Adaptor services

munjalp6:
```markdown
## Summary

Adds GHCR build pipeline support for:

- `sdr-mw-l`
- `vss-configurator`
- `vss-rt-config-adaptor`

This follows the existing GHCR image flow used by other services.

## Changes

- Added the three services to the GHCR container inventory as GitHub-built images.
- Added GHCR build metadata, Dockerfile paths, build contexts, platforms, and Compose image names for the new images.
- Extended image change detection so `vss-configurator` rebuilds when either:
  - `services/configurators/vss-configurator` changes
  - `libs/analytics/spatialai-data-utils` changes
- Added multi-source content hashing for images built from more than one source path.
- Removed `spatialai-data-utils` from `vss-configurator` `pyproject.toml` and `uv.lock`.
- Updated the `vss-configurator` Dockerfile to install `spatialai-data-utils` from the current monorepo source during image build, avoiding stale lockfile churn when SDU changes trigger an image rebuild.
- Updated Compose and Helm image defaults to use the managed GHCR `develop-latest` channel.
- Updated Compose golden data and Helm/Compose channel policy checks for the new image coordinates.
- Added downstream release-set and promotion trigger support for the new images.
- Updated GHCR alias, release-set, provenance, and image-guard helpers to handle the new image set.
- Added SonarQube project wiring and coverage generation for the three services.
- Added CI unit-test jobs for SDRC, VSS configurator, and VSS RT config adaptor.
- Removed stale Spatial AI Data Utils jobs from the main CI workflow.

## Validation

Ran local validation for workflow/helper logic and service unit tests:

- `sdr-mw-l`: 77 passed
- `vss-configurator`: 149 passed
- `vss-rt-config-adaptor`: 23 passed

Also ran script/unit validation for:

- container build planning
- changed image detection
- release-set closure
- GHCR alias advancement
- GHCR image guard
- Compose image golden checks
- Helm release-channel policy
- downstream release-set preparation
- promotion trigger preparation
- YAML parsing
- `git diff --check`

## Notes

The actual GHCR image push and NGC/NVCR promotion were not run locally because they require GitHub Actions runner context and registry credentials.
```


mkothari-commits (dismissed):
_This is Hinton, Mugdha's AI Agent Assistant_

✅ **OSRB review CLEAN → approving.** CI/CD infrastructure PR — GHCR build enablement for 3 services (sdr-mw-l, vss-configurator, vss-rt-config-adaptor) + internal SDU library decoupling. No external OSS dep changes. All 11 required OSRB CI checks green.

Ran the [`pr-osrb-review` v1.7 skill](https://gitlab-master.nvidia.com/malabdallat/hinton-ai-agent-vss-team/-/tree/main/.claude/skills/tpm/pr-osrb-review) against PR #1623 HEAD `79b4e33e`.

**Gate coverage:**

| Gate | Status | How this PR clears it |
|---|---|---|
| Copyright/SPDX headers | ✅ | CI green |
| Missing 3rd-party in license file | ✅ | No 3rd-party manifest changes |
| Licence-terms change | ✅ | 2 CSV rows only: internal `spatialai-data-utils` removal + tzdata Apache-2.0 same-licence notation change |
| Non-permissive vs approved set | ✅ | No external OSS changes |
| Required CI OSRB gates | ✅ | All 11 required checks green |

**License Diff bot verdict:** *"No changes require OSRB re-engagement."* 2 CSV rows:

- `spatialai-data-utils` REMOVED from `services/configurators/vss-configurator/uv.lock` — **this is an internal NVIDIA library** (`libs/analytics/spatialai-data-utils` in the same monorepo). Being decoupled from the configurator's uv-managed lock and installed separately in the Dockerfile via `uv pip install`. Not 3rd-party OSS.
- `tzdata` 2026.3 → 2025.3, 2026.2 (Apache-2.0 unchanged)

**Bernd's 3 re-engagement triggers:** none fire — no new OSS deps, no licence changes, no external use changes.

**Positive OSRB signals:**
- Well-formed PR body with per-service test evidence (sdr-mw-l 77 passed, vss-configurator 149 passed, vss-rt-config-adaptor 23 passed)
- All 6 commits DCO-signed
- Real correctness fix: typo `nvcr.io/nvidian/distroless/python` → `nvcr.io/nvidia/distroless/python` in Dockerfile
- Architecture improvement: SDU source changes no longer force configurator uv.lock churn; image-change detection rebuilds on either source
- SonarQube coverage extended to the new services

LGTM.

services/configurators/vss-configurator/docker/Dockerfile
```diff
@@ -25,24 +25,28 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
...
 # External CI should build with:
 # docker build -f services/configurators/vss-configurator/docker/Dockerfile .
 COPY --from=ghcr.io/astral-sh/uv:0.11.24 /uv /usr/local/bin/uv
-COPY libs/analytics/spatialai-data-utils /libs/analytics/spatialai-data-utils
 COPY services/configurators/vss-configurator/pyproject.toml \
     services/configurators/vss-configurator/uv.lock \
     ./
 RUN uv sync --frozen --no-dev
 
+# Install Spatial AI Data Utils from the current monorepo source outside the
```
zac-wang-nv: **P1** — This is the one thing I'd want settled before merge, because a lot of the PR's build machinery follows from it.

`spatialai-data-utils` is **already a versioned, publishable distribution**: `libs/analytics/spatialai-data-utils/release/setup.py` sets `_BASE_VERSION = "2.0.2"`, `pyproject.toml` declares `dynamic = ["version"]` with a full `[project]` block, license-files, and its own pinned dependency set, and `VERSION_SUFFIX` is already wired for dev builds. So consuming it as a versioned library is not a refactor — the packaging exists.

**It was also already a declared, locked dependency.** At the merge-base, `vss-configurator/pyproject.toml` had:

```toml
dependencies = [ ..., "spatialai-data-utils" ]

[tool.uv.sources]
spatialai-data-utils = { path = "../../../libs/analytics/spatialai-data-utils/release" }
```

`ed5a191ce` removed the declaration and the source mapping and deleted **952 lines of `uv.lock`**, replacing it with `uv pip install /libs/analytics/spatialai-data-utils/release` here. The Dockerfile comment gives the reason: *"outside the configurator lock so SDU source/dependency changes do not require lock churn."*

I'd push back on that specifically: lock churn is the mechanism working, not overhead. It is what makes the image reproducible and what makes the rebuild trigger fall out for free.

**What the change cost.** Once the dependency is undeclared, `vss-configurator`'s `source_path` no longer determines its content — so the PR had to add a way to notice SDU changed. That is the entire multi-source apparatus: `source_paths` in the inventory, `source_tree_hash` (`container_build_plan.py:125`), and `entry_source_paths` threaded through `advance_ghcr_alias.py`, `check_container_tag_source.py`, `container_build_plan.py`, `release_set.py`, `ghcr_image_guard.py`, and `prepare_downstream_release_set.py`. All of it is new — `entry_source_paths` is **absent on `develop`**.

And it serves exactly one image: **`source_paths` appears on 1 of the 18 inventory entries.** A repo-wide build-system feature across six scripts, for one consumer, whose need for it was created by this same PR.

**Suggested direction:** keep SDU declared — either `spatialai-data-utils==2.0.2` resolved from a registry, or the existing `[tool.uv.sources]` path entry retained so the lock keeps capturing it. Then `vss-configurator` is single-source again and the pre-existing single-path code handles it unchanged.

Two things make the unwind cheap: `source_tree_hash` deliberately returns the exact git tree SHA in the single-path case (`container_build_plan.py:142-143`), so the other 10 single-source images keep their existing `tree-<sha>` tags and nothing rebuilds; and this also closes finding 13, since SDU's pinned transitive deps go back into `uv.lock` instead of resolving at docker build time.

If there is a reason the lock genuinely cannot hold SDU — a release-cadence constraint I can't see from the repo — then say so in the Dockerfile comment and I'll withdraw this. But it should be a stated tradeoff rather than a silent one, because the cost is six scripts of build machinery.

deploy/docker/container-inventory.json
```diff
@@ -182,10 +196,15 @@
     },
     {
       "name": "sdr-mw-l",
-      "strategy": "external-pin",
-      "platforms": ["linux/amd64"],
+      "strategy": "build",
```
zac-wang-nv, line 197: **P1** — Flipping these three images to `strategy: build` makes GitHub the producer instead of mirroring the NGC/Jenkins-built image. The amend from `81860ff47` to `619cf0a26` removed the three `NVIDIA-Software-License-Agreement.pdf` files and their Dockerfile `COPY` lines, so the GHCR-built images will ship without an artifact the images they replace carried.

Re-verified at the current head `5876c1645`: `81860ff47` carried `services/sdrc/`, `services/configurators/vss-configurator/`, and `services/configurators/vss-rt-config-adaptor/` copies of the PDF; the current head carries none of the three, and the five that remain in-tree are exactly the five already on `develop` — so this PR touches no SLA path at all in its diff, which is why nothing flags it. Worth adding that `License Diff (OSRB)` is **green with 0 changes** on this head, so whatever prompted the removal, it was not that gate rejecting the file.

Two independent reasons to think the PDF belongs in the image:

1. **Every sibling GHCR-managed image ships it**, and has on `develop` since well before this PR:
   - `services/agent/docker/Dockerfile:43,167` → `/vss-agent/`
   - `services/ui/Dockerfile:76`
   - `services/alert/Dockerfile:85` → `/app/`
   - `services/video-summarization/docker/Dockerfile:246` → `/opt/nvidia/via/`

   `sdr-mw-l`, `vss-configurator`, and `vss-rt-config-adaptor` would be the only managed images without it.

2. **The NGC builds of these three images do carry it.** The plan doc removed in the same amend said so at line 179: "The Jenkins pipelines inject `NVIDIA-Software-License-Agreement.pdf` before…". Since these GHCR images are meant to be promoted into the same NVCR coordinates, dropping it is a regression in the distributed artifact, not a no-op.

Note that `services/configurators/vss-configurator/README.md:315` only lists two legal requirements (Apache-2.0 SPDX headers, third-party license text in the image) and does not mention the SLA — so the README and the code do now agree. That is the one argument for the current state, and it may be the reason for the backout.

Could you say which it is? If the PDF was dropped because a gate rejected the binary (OSRB, License Diff, a size or file-type check), that is worth recording — `services/alert/.gitignore:4` carries an explicit `!NVIDIA-Software-License-Agreement.pdf` un-ignore rule, which suggests the other services hit exactly that friction and solved it with an allowlist entry rather than by dropping the file. The three new services have no `.gitignore` at all, so the same fix would be three one-line files.

If it was dropped deliberately after a legal review, then the four sibling Dockerfiles and the `services/agent`/`services/ui`/`services/alert` READMEs are the ones that are wrong, and that is worth its own issue rather than a silent divergence here.

.github/scripts/prepare_promotion_trigger.py
```diff
@@ -53,6 +56,14 @@ def promotion_variables(
         raise ValueError(
             "behavior analytics artifacts-promotion config path is required"
         )
+    if "sdr-mw-l" in built_names and not sdr_mw_l_config:
```
zac-wang-nv: **P1** — These three config paths arrive from `nightly-promote-ghcr.yml:66-68` as `vars.SDR_MW_L_ARTIFACTS_PROMOTION_CONFIG_PATH`, `vars.VSS_CONFIGURATOR_ARTIFACTS_PROMOTION_CONFIG_PATH`, and `vars.VSS_RT_CONFIG_ADAPTOR_ARTIFACTS_PROMOTION_CONFIG_PATH`. An unset GitHub repo variable renders to the empty string, so the first nightly whose release set contains any of the three images dies at "prepare promotion trigger."

The code itself is right — failing loud beats promoting to the wrong NVCR path. This is a **sequencing** requirement: the three repo variables need to exist before or with the merge, otherwise the failure surfaces overnight and reads like a pipeline regression rather than missing config. Could you confirm they are set, or say which are still pending?

**Second, a naming split that is worth more attention than a nit — one half of it is a cross-repo contract.**

There are two distinct namespaces here, and the four pre-existing images follow a consistent rule: **the GitHub repo variable carries the `VSS_` prefix, and the downstream variable handed to GitLab strips it.**

| image | repo var (`vars.*`) | downstream var (to GitLab) |
|---|---|---|
| agent/UI | `VSS_AGENT_UI_…` | `AGENT_UI_…` |
| alert | `VSS_ALERT_…` | `ALERT_…` |
| video analytics API | `VSS_VIDEO_ANALYTICS_API_…` | `VIDEO_ANALYTICS_API_…` |
| behavior analytics | `VSS_BEHAVIOR_ANALYTICS_…` | `BEHAVIOR_ANALYTICS_…` |
| **sdr-mw-l** | **`SDR_MW_L_…`** ← no `VSS_` | `SDR_MW_L_…` ✅ |
| **vss-configurator** | `VSS_CONFIGURATOR_…` ✅ | **`VSS_CONFIGURATOR_…`** ← keeps `VSS_` |
| **vss-rt-config-adaptor** | `VSS_RT_CONFIG_ADAPTOR_…` ✅ | **`VSS_RT_CONFIG_ADAPTOR_…`** ← keeps `VSS_` |

All three new images break the rule, but on *opposite sides*, so they can't all be right.

- `sdr-mw-l` breaks the **repo-var** side. That one really is cosmetic — free to rename now, annoying once the variable exists.
- The two configurators break the **downstream** side (`prepare_promotion_trigger.py:85,89`), and that is not cosmetic. The downstream variable name is the contract the GitLab promotion pipeline reads. If that side expects the stripped convention the other four follow, these two names won't be picked up and artifact promotion for them silently does nothing — which is a much quieter failure than the `raise ValueError` above.

I could not verify this from here: the only references to these names anywhere in this repo are this script, its tests, and the workflow, and there is no `ci-vss-oss` checkout to check the consuming side against. **Could you confirm which spelling the GitLab promotion job actually reads for these two?**

Note the tests don't settle it either — `test_prepare_promotion_trigger.py:73,77` assert the same strings the code emits, so they pass either way. They pin the implementation, not the contract.

**Third, and the reason the other two keep happening: this is a map being carried as seven parallel scalars.**

The fixture values in `test_prepare_promotion_trigger.py:40-52` show what these actually are:

```
configs/vss-3.2.0/vss-core-agent.yml
configs/vss-3.2.0/vss-core-vlm-verifier.yml
configs/vss-3.2.0/vss-core-behavior-analytics.yml
configs/vss-3.2.0/vss-core-sdr-mw-l.yml
configs/vss-3.2.0/vss-core-configurator.yml
configs/vss-3.2.0/vss-core-rt-config-adaptor.yml
```

One directory, one leaf per image. It is an `image name → config path` mapping, and it is currently threaded through **six in-repo sites plus an invisible seventh**. Adding one image means touching all of them: a `workflow_dispatch` input (`nightly-promote-ghcr.yml:28-36`), an `env:` entry (`:66-68`), a CLI flag (`:107-109`), a function kwarg (`prepare_promotion_trigger.py:23-25`), a validation branch (`:59-66`), an emit branch (`:82-91`), an argparse flag (`:102-104`) — and a GitHub repo variable that never appears in a diff at all.

Consider putting the mapping in `container-inventory.json`, which `github-first.md` already calls "the single source of truth for what gets built":

```json
{ "name": "vss-configurator", "ghcr_build": true,
  "promotion_config": "configs/vss-3.2.0/vss-core-configurator.yml" }
```

`promotion_variables` then becomes one loop over the release set instead of seven hand-written pairs of `if` blocks, and several problems close at once:

- **Finding 3 stops being a 2am failure.** An unset repo variable renders to `""` and dies overnight; a missing inventory key is visible in the PR and catchable by the existing `release_set.py closure` check at review time.
- **The naming split above disappears**, because the downstream key is derived by one rule rather than hand-spelled seven times.
- **The `vss-3.2.0` in every path becomes one reviewable change** at the next release instead of seven edits in repo settings, where a missed one is invisible until it fires.
- **Onboarding cost drops from seven code sites to one inventory line**, which is what `github-first.md` promises onboarding costs.

The honest counterargument: repo variables can be changed without a merge, which matters if these paths move on the GitLab side's cadence rather than this repo's. But they are not secrets — they are paths, they already embed a version string, and they therefore churn with releases anyway. I don't think that buys enough to justify seven plumbing points and an overnight failure mode.

I'd take this as a follow-up rather than a merge blocker, since the code as written is correct. But it is the same shape as findings 17 and 18: a declaration that belongs in the inventory living somewhere else, with bespoke machinery built to compensate. Worth deciding once as a pattern rather than three times.

.github/scripts/detect_changed_images.py
```diff
@@ -45,19 +46,27 @@
...
+# Analytics/configurator images have independent tag variables and build only when
+# their own source inputs change.
 SHARED_TAG_IMAGE_NAMES = frozenset({"vss-agent", "vss-agent-ui", "vss-alert-ms"})
 
 # Behavior analytics compiles architecture-sensitive native dependencies and
 # must not build arm64 through QEMU on an amd64 runner. The build workflow
 # expands it into one job per platform, then combines the native results into
 # the same multiarch candidate expected by the release-set flow.
-NATIVE_PLATFORM_IMAGE_NAMES = frozenset({"vss-behavior-analytics"})
+NATIVE_PLATFORM_IMAGE_NAMES = frozenset(
```
zac-wang-nv: **P2** — The native-runner routing itself is right, and I'd keep it: `ubuntu-24.04-arm` is a real hosted label, the combine job at `build-dev-images.yml:281` recomposes the multiarch manifest properly, and building these on real hardware beats arm64-under-QEMU. Three things about *how* the set is declared, though.

**a. It belongs in `container-inventory.json`, not in the script.** `github-first.md` calls the inventory "the single source of truth for what gets built", and this is a per-image build property sitting outside it. The inventory already carries optional per-entry flags with exactly this shape — `ghcr_build` (11 of 18 entries), `trigger_downstream_from_source` (1), `lfs_include` (1), `source_paths` (1) — so a 4-of-18 boolean is idiomatic, not a new pattern.

The entries reaching `split_build_matrices` *are* the raw inventory dicts, so the change is small:

```json
{ "name": "vss-configurator", "ghcr_build": true, "native_platform_build": true, ... }
```

```python
# Images that compile architecture-sensitive native code are built on real
# arch-matched runners instead of arm64-under-QEMU. Declared per entry in
# container-inventory.json via "native_platform_build": true.
def is_native_platform_build(entry: dict) -> bool:
    return entry.get("native_platform_build") is True
```

with the two comprehensions in `split_build_matrices` switching to `if not is_native_platform_build(entry)` / `if is_native_platform_build(entry)`.

I applied this in a scratch tree and verified it is **behaviour-preserving**: the derived partition over the real inventory is **identical** to the current hardcoded set (native = the same 4, standard = the other 5, 9 = 4 + 5, no overlap), the `native_platform_matrix` still expands to the same 8 jobs on the same runners, and the flag does **not** leak into the job matrix because `matrix_entry()` enumerates keys explicitly. Full suite **21/21 green**. Cost is 4 inventory lines, ~8 script lines, and 2 fixture lines.

**b. The comment's stated rationale doesn't match the membership.** It reads "Behavior analytics compiles architecture-sensitive native dependencies and must not build arm64 through QEMU" — still singular, and not a criterion that selects these four. `vss-agent` is **excluded** despite `services/agent/docker/Dockerfile:16-17` saying *"ninja-build, libcairo2-dev, pkg-config are needed for building pycairo on ARM64 (no pre-built wheel available, so it compiles from source)"* — an ARM64-specific source compile, i.e. the exact described condition, going through QEMU. `vss-alert-ms` is also excluded with `gcc` + `libssl-dev` + `librdkafka-dev` + `build-essential` (`services/alert/Dockerfile:31-42`), a stronger native profile than the two configurators that were added. If the real criterion is empirical ("arm64 QEMU build exceeds N minutes"), say that — it's a legitimate reason and it tells the next onboarder how to decide.

**c. Membership is unguarded — I proved this by mutation.** `test_native_images_use_arch_specific_runners:329` asserts against a hand-built fixture containing only `vss-behavior-analytics`; none of the three new names appears in any native-matrix test. I removed all three from the frozenset and ran the suite: **21/21 still passed.** A regression that silently drops one back onto QEMU would surface as a slow or timing-out arm64 build, not a red assertion. Deriving from the inventory fixes this structurally — the fixtures then have to declare intent — but it's worth one assertion over the real `container-inventory.json` so the test tracks reality rather than a fixture.

**Also worth a follow-up (not this PR):** `.github/github-first.md` documents five onboarding steps and never mentions this split; `detect_changed_images` appears once in the "How it fits together" diagram as `→ build matrix`, singular. Someone onboarding by the book silently gets QEMU arm64 with no hint a native path exists. Step 1 is where the reader is already looking at `platforms`, so it's the natural home.

.github/workflows/ci.yml
```diff
@@ -54,6 +54,10 @@ jobs:
       build-video-analytics-api: ${{ steps.analytics-changes.outputs.video }}
       current-behavior-analytics-image: ${{ steps.current-images.outputs.behavior }}
       current-video-analytics-api-image: ${{ steps.current-images.outputs.video }}
+      spatialai-data-utils-changed: ${{ steps.changed-paths.outputs.spatialai-data-utils-changed }}
```
zac-wang-nv: **P2 — small leftover from the SDU job removal: this output now has no consumers.**

`6cdd7e9ab` correctly deleted `spatialai-data-utils-test` and `spatialai-data-utils-publish`, which turned the golden check green. But `spatialai-data-utils-test` was the only reader of this output — it gated on `if: needs.prepare-source.outputs.spatialai-data-utils-changed == 'true'` — so with that job gone, the output at `:57` and its write at `:140` are dead.

A repo-wide grep at this head returns only the definition and the write, and no reader:

```
$ git grep -n "spatialai-data-utils-changed" 6cdd7e9ab
ci.yml:57:   spatialai-data-utils-changed: ${{ steps.changed-paths.outputs.spatialai-data-utils-changed }}
ci.yml:140:      f"spatialai-data-utils-changed={str(spatialai_changed).lower()}\n"
```

For comparison, `develop`'s `ci.yml` has **no** `spatialai` reference at all — `git show origin/develop:.github/workflows/ci.yml | grep spatialai` is empty. So dropping these two lines converges with the base rather than diverging from it.

**One thing to keep, though** — don't delete the computation. `spatialai_changed` itself is still live at `:128`:

```python
vss_configurator_changed = (
    ci_contract_changed
    or paths_changed_under(paths, "services/configurators/vss-configurator")
    or spatialai_changed
)
```

That is the "SDU source changed ⇒ rebuild the configurator" coupling this PR exists to add, and it's correct. Only the *exported job output* is orphaned. Removing `:57` and the `output.write(...)` at `:140` is sufficient; `:113–115` and `:128` should stay.

Entirely cosmetic — dead YAML costs nothing at runtime and the guard test only asserts on the job name, so this is not blocking. Flagging it so it doesn't get cargo-culted forward.

services/configurators/vss-configurator/docker/Dockerfile
```diff
@@ -25,24 +25,28 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
...
+# configurator lock so SDU source/dependency changes do not require lock churn.
+COPY libs/analytics/spatialai-data-utils /libs/analytics/spatialai-data-utils
+RUN uv pip install --python .venv/bin/python /libs/analytics/spatialai-data-utils/release
```
zac-wang-nv: **P2 — the SonarQube fix works, but it took 23 packages out from under a lockfile. Worth a conscious yes.**

`7edd18c03` fixes the numpy/cp313 failure the right way — moving the SDU install into the Docker builder, which has `build-essential` and CPython headers. Both `Build vss-configurator` arches and `SonarQube Scan (vss-configurator)` are green. No argument with the direction.

The side effect is that the second pass is unlocked:

```dockerfile
RUN uv sync --frozen --no-dev                       # 27 packages, locked
RUN uv pip install --python .venv/bin/python \
    /libs/analytics/spatialai-data-utils/release    # resolved fresh at build time
```

I diffed the two locks to see what that actually costs. `uv.lock` went **70 → 27** packages; of the 43 that left, **20 are still pinned exactly** by `spatialai-data-utils/release/pyproject.toml` (including `numpy==1.26.4` — I checked, since that was the whole point). The remaining **23 are transitive and now float at `docker build` time**:

```
attrs, botocore, contourpy, cycler, fonttools, jmespath, joblib,
jsonschema-specifications, kiwisolver, narwhals, portalocker, pyparsing,
python-dateutil, pywin32, pyyaml, referencing, rpds-py, s3transfer, six,
termcolor, threadpoolctl, tzdata, yacs
```

So two builds of this same commit can now produce different `vss-configurator` images.

The part I'd actually flag: `services/configurators/vss-configurator/3rdParty_Licenses.md` still has **70 entries** — matching the old lock — and declares exact versions for the floating ones (`## attrs:26.1.0`, `## pyyaml:6.0.3`, `## six:1.17.0`, `## rpds-py:2026.6.3`, …). `7edd18c03` doesn't touch that file. Nothing is wrong *today*, and `License Diff (OSRB CSV)` is green — but `01825ec39` on `develop` ("align license artifacts with lock and build") deliberately tied that manifest to the lock, and this quietly unties 23 of them, with no gate that would notice the drift.

Cheapest fix that keeps your stated goal (SDU changes shouldn't churn the configurator lock): give SDU its own `uv.lock` and `uv pip install --require-hashes`/`--frozen` against it, so the second pass is pinned too. If you'd rather accept the drift, a line in the Dockerfile comment saying the SDU closure is intentionally unpinned would save the next person this diff.

.github/workflows/build-dev-images.yml
```diff
@@ -265,10 +265,10 @@ jobs:
...
           } >> "$GITHUB_STEP_SUMMARY"
 
-  # Behavior analytics compiles native dependencies (notably OpenCV). Build
-  # each architecture on its matching GitHub-hosted runner so arm64 never runs
-  # through QEMU, then combine the platform images without rebuilding.
-  build-native-behavior-platforms:
+  # Some images compile architecture-sensitive native dependencies or inherit
+  # base images that should not run through QEMU. Build each architecture on
+  # its matching GitHub-hosted runner, then combine the platform images.
+  build-native-platforms:
```
zac-wang-nv, line 271: **P2** — Not a blocker: CI is green, and I want to be clear up front that **nothing here needs fixing for this PR to merge.** Flagging it because the path to green is worth a look.

This rename is the load change: the job that used to build one image (behavior-analytics) on native runners now builds every native image. On `619cf0a26` that produced repeated push failures which rendered as:

```
ERROR: failed to push ghcr.io/nvidia-ai-blueprints/vss/sdr-mw-l:pr-1623-619cf0a26493-amd64:
  denied: permission_denied: Error from intermediary with HTTP status code 403 "Forbidden"
  { "message": "You have exceeded a secondary rate limit. Please wait a few minutes..." }
```

**Please don't let anyone chase GHCR package permissions on the strength of that message.** buildx renders it as `denied: permission_denied` / `403 Forbidden`, which reads exactly like a missing package or a token without `packages: write`. It is neither — the body is a GitHub *secondary rate limit*. Every layer built fine; only the push was refused. Three independent controls confirm it:

1. `vss-configurator` — equally new, same workflow, same token — pushed successfully on **both** arches while `sdr-mw-l` failed. A permissions or package-creation problem cannot produce that split.
2. `vss-rt-config-adaptor` succeeded on arm64 and failed on amd64 in the same run. Same reasoning.
3. **The decisive one:** the rebase to `5876c1645` changed no PR content whatsoever (identical added/removed line multiset), and CI went green. Red→green on an unchanged tree isolates the cause to the environment.

The reason I'm still raising it at all: **the cost of reaching green was high, and this PR makes that cost structural.** `Build Dev Images (GHCR)` needed **four attempts** at `619cf0a26` — three consecutive failures on an identical tree. That is not a rare flake.

- The managed set goes 6 → 9 images, and every one is `["linux/amd64","linux/arm64"]` — so the native-platform matrix goes 12 → 18 jobs pushing multi-hundred-MB images to GHCR simultaneously.
- There is no `max-parallel` on any of the three build matrices in this file (`:117`, `:277`, `:414`), and `fail-fast: false` means all 18 keep going.
- There is no retry around the push, so a single 429-behind-a-403 loses the whole job after a full build. `sdr-mw-l` burned ~3.5 min of pyinstaller work before being refused at the last step.

Two options, either fine, and both are follow-up-able rather than merge-blocking:
1. `max-parallel: 6` (or similar) on the `native-platform-matrix` strategy — throttles the burst without slowing the common case, since most PRs rebuild one or two images.
2. Retry just the push step (`nick-fields/retry`, or `docker/build-push-action` behind a retry wrapper) so a transient 403 costs seconds instead of a rebuild.

Worth noting this run is the *worst* case on purpose: `container_build_plan.py` joining `BUILD_CONTRACT_PATHS` (see comment 9) meant all 9 images rebuilt at once. That is the same burst a nightly or any future build-contract edit will produce, so it is not a one-off.

In fairness to the other reading: `5876c1645` passed on the first attempt, so it's entirely possible this never bites again. I'd note only that it's one sample, and it ran ~05:11–05:40 UTC versus the failing run's ~14:00 UTC — off-peak timing is a plausible confounder, so I wouldn't treat first-try-green as proof the burst is safe. Your call whether it's worth pre-empting.

.github/scripts/check_container_tag_source.py
```diff
@@ -88,6 +95,21 @@ def compose_names(self) -> tuple[str, ...]:
         image_name="vss-behavior-analytics",
         source_path=Path("services/analytics/behavior-analytics"),
     ),
+    "sdr-mw-l": ImageConfig(
+        image_name="sdr-mw-l", source_path=Path("services/sdrc")
+    ),
+    "vss-configurator": ImageConfig(
+        image_name="vss-configurator",
+        source_path=Path("services/configurators/vss-configurator"),
+        source_paths=(
```
zac-wang-nv: **P2** — The same two paths are declared in `deploy/docker/container-inventory.json:132`. Those two declarations now sit on opposite sides of the build/verify boundary:

- **Build side** stamps the label from the inventory — `build-dev-images.yml:139` runs `container_build_plan.py source-hash --source-paths '${{ matrix.source_paths }}'`, and `matrix.source_paths` comes from `detect_changed_images.matrix_entry()` → `entry_source_paths(entry)` → the JSON.
- **Verify side** recomputes it here — `source_content_hash()` hashes `config.all_source_paths()`.

Add a third shared library to the inventory and miss this file (or the reverse) and the two sides hash different path sets. `source_tree_hash` digests `path\0tree\n` pairs, so a divergence in the *path list alone* changes the result even when every tree is identical. The failure lands after the image is built, tagged and pushed: every `check-vss-configurator-container-source` run fails with a mismatch that reads like a stale image rather than config drift.

Cheap fix — a test that closes the loop (there is no `test_check_container_tag_source.py` today, so this would be a new file):

```python
from check_container_tag_source import IMAGE_CONFIGS
from release_set import entry_source_paths, inventory_by_name, load_inventory

def test_image_configs_match_the_inventory_source_paths():
    entries = inventory_by_name(load_inventory(REPO_ROOT))
    for name, config in IMAGE_CONFIGS.items():
        assert [p.as_posix() for p in config.all_source_paths()] == \
            entry_source_paths(entries[name]), f"{name} drifted from the inventory"
```

Deriving `IMAGE_CONFIGS` from `load_inventory()` at import time removes the duplication outright, but that is a bigger change than this PR needs.

.github/scripts/check_container_tag_source.py
```diff
@@ -715,6 +750,22 @@ def read_image_manifest_labels(
 
 
 def discover_compose_files(repo_root: Path) -> list[Path]:
+    tracked = run_git(
```
zac-wang-nv, line 731: **P2** — `discover_compose_files` used to glob `deploy/docker/**/*.yml|yaml` off the filesystem; it now enumerates git-tracked files and only falls back to globbing when `ls-files` fails or returns nothing. Net effect: an untracked compose file under `deploy/docker` is no longer checked by the container-source gate.

That is probably the behavior you want — untracked files are not part of the tree the gate reasons about. But it narrows the coverage of a *gate*, it is not mentioned in the commit message or PR body, and it does not follow from "enable GHCR builds for three services."

If it is load-bearing (say, CI leaves generated compose files in the workspace and they were producing false failures), a one-line comment saying so keeps the next reader from reverting it. If it is opportunistic, it would read better as its own PR.

.github/scripts/check_container_tag_source.py
```diff
@@ -389,14 +422,16 @@ def rescue_ignored_only(
     confirms its ``source_path`` tree matches ``build_tree_sha`` (so we know we
     found the exact source the image was built from) before diffing to HEAD.
     """
+    if len(config.all_source_paths()) != 1:
```
zac-wang-nv: **P2** — `rescue_ignored_only` is what lets a service mark test-only files as "not built into the image" via `.ignore_source_code_check`, so a docs or test edit does not fail the container-source gate. Any image with more than one source path now skips that path entirely, silently.

Inert today: `services/agent/.ignore_source_code_check` is the only such file in the repo and agent is single-source. The trap is later — the first person to add `services/configurators/vss-configurator/.ignore_source_code_check` will watch it be ignored with no diagnostic. A `print()` on this bail-out ("multi-source image; ignore-file rescue not supported") turns a silent no-op into a five-second diagnosis.

deploy/helm/services/bp-configurator/values.yaml
```diff
@@ -15,8 +15,8 @@ useReleaseNamePrefix: null
 replicas: 1
 
 image:
-  repository: ""
-  tag: ""
+  repository: ghcr.io/nvidia-ai-blueprints/vss/vss-configurator
```
zac-wang-nv, line 18: **P2** — Worth a conscious yes rather than a silent one: `repository`/`tag` were `""`, and `deployment.yaml` wrapped them in `required "bp-configurator.image.repository is required"`. That `required` is gone now (replaced by `include "bp-configurator.image"`), and the values carry real defaults.

A parent chart that forgets to set the image used to get a loud template error; it now silently renders the GHCR developer channel. That is the intended new default for this repo and it matches how the agent/UI/alert charts already behave, so I read this as a deliberate trade — flagging only so it is deliberate.

deploy/helm/industry-profiles/warehouse-operations/warehouse-2d-app/templates/warehouse-extra-services.yaml
```diff
@@ -6,6 +6,13 @@
 {{- $ds := $warehouse.dsConfigurator | default dict }}
 {{- if ($ds.enabled | default false) }}
 {{- $image := $ds.image | default dict }}
+{{- $global := .Values.global | default dict }}
+{{- $prefix := index $global "container_prefix" | default "" }}
+{{- $repository := required "warehouse.dsConfigurator.image.repository is required" $image.repository }}
+{{- if $prefix }}
+{{- $repository = printf "%s/vss-rt-config-adaptor" (trimSuffix "/" $prefix) }}
+{{- end }}
+{{- $tag := index $global "container_tag" | default (required "warehouse.dsConfigurator.image.tag is required" $image.tag) }}
```
zac-wang-nv: **P2** — Two small things:

**a. `default` evaluates both of its arguments.** Helm computes `required "…" $image.tag` before `default` ever looks at `global.container_tag`, so setting the global override does **not** let a caller omit `dsConfigurator.image.tag` — it still errors out. If the intent is "global wins, per-chart value optional":

```gotemplate
{{- $tag := index $global "container_tag" | default $image.tag }}
{{- if not $tag }}{{ fail "warehouse.dsConfigurator.image.tag is required" }}{{ end }}
```

**b. In the 2d chart this block never renders.** `warehouse-2d-app/values.yaml` has no `warehouse.dsConfigurator` key at all, so `{{- if ($ds.enabled | default false) }}` is always false. The compose golden agrees — `warehouse-2d-app.yml` carries `vss-configurator` but no `vss-rt-config-adaptor`.

Keeping the 2d and 3d copies byte-identical is the right call, so I would not delete it. Just note that `HELM_HELPERS` in `test_helm_release_channel_policy.py:63` lists only the 3d copy, so the 2d one can drift without any test noticing. Adding the 2d path to that list is a one-line fix.

.github/scripts/detect_changed_images.py
```diff
@@ -139,6 +148,7 @@ def paths_changed_under(changed: list[str] | None, directory: str) -> bool:
     )
 
 
+
```
zac-wang-nv: **P2** — Nit: an extra blank line lands here, and the same happens at `advance_ghcr_alias.py:143` and `test_helm_release_channel_policy.py:101`. Trivial, but they show up in the diff as unexplained noise.

Slightly less trivial than it looks: `.github/scripts/container_build_plan.py` was just added to `BUILD_CONTRACT_PATHS` (`detect_changed_images.py:49`), so a change anywhere in these files — a bare blank line included — rebuilds every managed image. Worth keeping the diff to lines that mean something.

This is not hypothetical on this run: touching the build-contract files rebuilt all 9 managed images at once, which is precisely the 18-way concurrent push burst that tripped the GHCR secondary rate limit in comment 1. A blank line is a cheap way to pay for a full matrix.

zac-wang-nv (changes requested):
## Review — GHCR onboarding for SDRC, VSS configurator, RT config adaptor

Reviewed at `79b4e33e4`, merge-base `a864bbc99`.

**Where this stands:** CI is green (65 success, 0 failures) and both P0s from my earlier passes are closed. The SDU job removal held through the rebase, and `79b4e33e4` fixed the compose channel policy test. I verified that one by control rather than by reading the commit message — `test_compose_keeps_the_managed_developer_channel` fails at `b09d97c4c` and passes at `79b4e33e4`, and all 9 policy assertions now pass. Nothing below is red in CI.

The build machinery in this PR is carefully made. The multi-source content hash preserves single-path tree SHAs so no existing image gets a spurious rebuild; the native-runner split correctly recombines into a multiarch manifest; the promotion trigger fails loud rather than promoting to an unset path. Those are all the right instincts.

### One pattern sits behind three of my findings

Each of these is individually small. Together they are the same decision made three times:

| Finding | The declaration | Where it lives instead | What compensates for it |
|---|---|---|---|
| **17** | which images need native arm64 runners | a hardcoded `frozenset` in `detect_changed_images.py` | a hand-maintained allowlist, untested — I removed all 3 new names and the suite still passed 21/21 |
| **18** | `vss-configurator` depends on `spatialai-data-utils` | nowhere — removed from `pyproject.toml` and 952 lines of `uv.lock` | the entire multi-source apparatus, threaded through 6 scripts, serving **1 of 18** images |
| **3** | image → artifacts-promotion config path | 7 GitHub repo variables, invisible to PR review | 7 parallel scalars threaded through 6 in-repo sites, and a 2am failure when one is unset |

In each case a fact that belongs in `container-inventory.json` — which `github-first.md` calls *"the single source of truth for what gets built"* — lives somewhere else, and machinery gets built to carry it. The machinery is good; it is compensating for the placement.

Finding 18 is the one I would most like settled before merge, because the cost is highest and the unwind is cheapest: `spatialai-data-utils` is **already** a versioned distribution (`_BASE_VERSION = "2.0.2"`, full `[project]` metadata, its own pinned deps), and it was **already** a declared, locked dependency before this PR. Restoring that makes `vss-configurator` single-source again, and `source_tree_hash` deliberately returns the exact git tree SHA in the single-path case — so the other 10 images keep their content tags and nothing rebuilds. It also closes finding 13 as a side effect.

For 17 I have applied the inventory-driven version in a scratch tree and confirmed it is behaviour-preserving: the derived partition is identical to the current hardcoded set, the same 8 native jobs on the same runners, full suite 21/21 green. Details inline.

### Also worth a look

- **Finding 3** carries a naming split I could not resolve from this repo. The four existing images follow "repo var keeps `VSS_`, downstream var strips it". All three new images break that rule, on opposite sides — and for the two configurators it is the **downstream** name that differs, which is the contract the GitLab promotion job reads. If that side expects the stripped spelling, promotion for those two silently does nothing. Could you confirm which spelling it reads?
- **Finding 2** is the other pre-merge item: the three new images ship without the NVIDIA SLA PDF that all four sibling GHCR images carry.

Findings 1, 4–9 and 14 are ordinary suggestions and none of them should hold up the merge.

Two notes on process. Everything cited here was re-verified at `79b4e33e4`; where I ran something, I have said what I ran and what it returned. And I withdrew a NOTE from an earlier draft that praised the multi-source hash for threading cleanly through every consumer — the threading is clean, but crediting it read as endorsing a design I am arguing against in finding 18.

services/configurators/vss-configurator/docker/Dockerfile
munjalp6: Let me explain the rationale behind the current design.

There is a requirement from the `spatialai-data-utils` side that whenever SDU changes, it should trigger a `vss-configurator` image build, even if there is no direct change under `services/configurators/vss-configurator`.

If SDU remains a declared/path dependency in `vss-configurator`, then any SDU version or dependency metadata change also requires updating `services/configurators/vss-configurator/uv.lock`. Otherwise, the automated configurator image build triggered by the SDU change can use a lockfile that still reflects the previous SDU dependency closure.

The failure mode I was trying to avoid is:
- SDU changes its version or dependencies.
- GHCR correctly triggers a `vss-configurator` build because SDU changed.
- But `vss-configurator/uv.lock` is not updated in the same PR/workflow.
- The image build then uses stale locked dependency data, or we would need to regenerate the lock during Docker build.
- If we regenerate during Docker build, the image is built from dependency resolution that is not checked in or reviewed.
- If we do nothing, later configurator builds triggered by unrelated `vss-configurator` code changes can continue carrying the stale SDU dependency closure.

That is why I modeled SDU as an additional image source input and moved the SDU install outside the configurator lock: the image rebuild follows SDU source changes without requiring configurator lockfile churn for every SDU-side update.

That said, I agree with your concern that the current second-pass `uv pip install /libs/...` weakens reproducibility because part of the dependency closure can resolve during Docker build. If we do not want the multi-source/decoupled-lock design, then I think the alternative contract should be explicit: SDU developers must update the SDU dependency reference in `vss-configurator` and refresh `services/configurators/vss-configurator/uv.lock` whenever SDU version or dependency metadata changes. We should also add a lock freshness check so CI fails clearly when that required lock update is missed.

.github/scripts/prepare_promotion_trigger.py
munjalp6: For the promotion-config naming concern: I removed the three new GitHub-side repo-variable/config-path inputs from the VSS workflow.

The three new images no longer use `SDR_MW_L_ARTIFACTS_PROMOTION_CONFIG_PATH`, `VSS_CONFIGURATOR_ARTIFACTS_PROMOTION_CONFIG_PATH`, or `VSS_RT_CONFIG_ADAPTOR_ARTIFACTS_PROMOTION_CONFIG_PATH` from GitHub. Instead, downstream `ci-vss-oss` now owns the mapping in `ci/vss-ghcr-images.yml`, and the promotion tools derive `promotion_config`, metadata filenames, and CLI flags from that inventory.

Current downstream mapping:
- `sdr-mw-l` -> `vss-core-sdrc.yml`
- `vss-configurator` -> `vss-core-analytics.yml`
- `vss-rt-config-adaptor` -> `vss-core-analytics.yml`

This removes the repo-var/downstream-var naming split for the new images.

.github/scripts/detect_changed_images.py
munjalp6: `NATIVE_PLATFORM_IMAGE_NAMES` itself predates this PR and currently owns the existing behavior-analytics native-runner behavior. To keep this PR scoped to SDRC/configurator onboarding, I kept that existing behavior unchanged and updated only the three newly onboarded images to an inventory-declared `native_platform_build: true` flag.

.github/scripts/detect_changed_images.py
munjalp6: whitespace-only noise has been cleaned up

services/configurators/vss-configurator/docker/Dockerfile
munjalp6: For this PR, I reverted the multi-source build changes. The vss-configurator image build will now be triggered only by changes under its own service source path, and spatialai-data-utils remains declared and captured in the configurator uv.lock.
I’ll start a separate discussion for SDU-triggered configurator rebuilds and handle that in a follow-up PR if we decide to add multi-source change detection for it.

deploy/docker/container-inventory.json
munjalp6: For the SLA PDF: I skipped adding it in this PR based on my current understanding of the OSRB guidance.

My understanding is:

- The SLA is required in the distributed container.
- But the SLA PDF does not necessarily need to be stored in the GitHub source repo.

Previously, these three images were built through the microservice GitLab/Jenkins flow. In that flow, the SLA was available on the GitLab side and was injected into the image during the build.

This PR moves image production to the GitHub GHCR pipeline. I was not sure where the SLA PDF should come from in this new flow without adding the PDF directly to this GitHub repo.

While preparing this PR, I also checked the existing GHCR build flow for video analytics and behavior analytics. Those Dockerfiles do not copy `NVIDIA-Software-License-Agreement.pdf`, so I followed the same pattern for these three images.

That said, if the GHCR-built images must include the SLA, then we need to decide the right source for it:

- either keep the SLA PDF in this GitHub repo and copy it in the Dockerfiles, or
- store it in NGC or another approved internal location, and update the GitHub GHCR workflow to fetch it during the build.

Without one of these, the GHCR-built images will not contain the same SLA artifact that the previous Jenkins/NGC-built images had.

.github/scripts/check_container_tag_source.py
munjalp6: This is resolved in the current branch.
I removed the SDU-triggered configurator build changes from this PR, so .github/workflows/ci.yml no longer contains the spatialai-data-utils-changed output, its write, or the spatialai_changed / vss_configurator_changed logic.
The SDU-triggered rebuild behavior will be handled separately if we decide to add multi-source change detection in a follow-up PR.

services/configurators/vss-configurator/docker/Dockerfile
munjalp6: This is resolved in the current branch.

I reverted the decoupled SDU install approach. `spatialai-data-utils` is again declared in `vss-configurator/pyproject.toml`, captured through the existing `[tool.uv.sources]` path mapping, and installed via the locked `uv sync --frozen --no-dev` flow.

There is no separate unlocked `uv pip install /libs/...` step anymore, so the SDU dependency closure is no longer resolved fresh during Docker build. `uv.lock` and `3rdParty_Licenses.md` are also back in sync with the base branch.

.github/workflows/ci.yml
munjalp6: This is resolved in the current branch.
I removed the SDU-triggered configurator build changes from this PR, so .github/workflows/ci.yml no longer contains the spatialai-data-utils-changed output, its write, or the spatialai_changed / vss_configurator_changed logic.
The SDU-triggered rebuild behavior will be handled separately if we decide to add multi-source change detection in a follow-up PR.

.github/workflows/build-dev-images.yml
munjalp6: Thanks, that makes sense. I agree this is a GHCR secondary rate-limit symptom, not a package permission issue.

I have not changed the workflow concurrency/retry behavior in this PR because this PR is focused on onboarding `sdr-mw-l`, `vss-configurator`, and `vss-rt-config-adaptor` to the existing GHCR flow. Adding `max-parallel` or push retry would affect the broader GHCR build workflow, so I would prefer to track that as a follow-up reliability improvement.

For this PR, I’ll avoid unrelated build-contract/whitespace churn so we do not trigger full rebuilds unnecessarily.

.github/scripts/check_container_tag_source.py
munjalp6: This is resolved by removing the multi-source build approach from this PR.

`vss-configurator` is back to a single `source_path` in `container-inventory.json`, and the multi-source helpers/fields such as `source_paths`, `entry_source_paths`, and `all_source_paths` are no longer part of the current branch.

Because the build and verify paths are single-source again, the specific build/verify drift risk described here no longer applies. If we add SDU-triggered multi-source builds in a follow-up PR, I agree we should add a test or derive the verify config from the inventory so those path sets cannot drift.

deploy/helm/services/bp-configurator/values.yaml
munjalp6: Yes, this is intentional.

For `bp-configurator`, I changed the Helm chart to follow the same managed-image pattern used by sibling GHCR charts like `vss-agent-ui` and `vss-alert-ms`: the chart carries the default GHCR developer-channel image, while parent charts can still override it through either the local image values or the shared `global.container_prefix` / `global.container_tag` mechanism.

So the tradeoff is deliberate: missing parent overrides no longer fail template rendering, but they render the repo-standard GHCR `develop-latest` default instead.

deploy/helm/industry-profiles/warehouse-operations/warehouse-2d-app/templates/warehouse-extra-services.yaml
munjalp6: Fixed this.

The 2D and 3D `warehouse-extra-services.yaml` templates now compute the final repository/tag first, using `global.container_prefix` and `global.container_tag` when provided, and only fail if the final value is empty. This avoids putting `required` inside `default`, so a global override can work even when the local `warehouse.dsConfigurator.image.repository/tag` values are omitted.

I also added the 2D `warehouse-extra-services.yaml` path to `HELM_HELPERS` so the 2D and 3D copies are both covered by the Helm image policy test.

Validation:
- `python3 .github/scripts/test_helm_release_channel_policy.py` passed.
- Verified isolated Helm renders for both 2D and 3D with empty local repo/tag and global prefix/tag set; both rendered `ghcr.io/test/vss-rt-config-adaptor:test-tag`.

.github/scripts/check_container_tag_source.py
munjalp6: Fixed by documenting the intent here.

This change is intentional: the container-source gate should validate committed/reviewable deploy state, not generated or untracked compose files that may be present in a local or CI workspace. `discover_compose_files` now scans git-tracked `deploy/docker` compose files first and falls back to filesystem globbing only when git metadata is unavailable.

I added an inline comment so the coverage narrowing is explicit.

zac-wang-nv (approved):

PR #1634, opened 2026-08-11
Two gpu placement for combined search standalone and search + alerts builds [nvbugs/6591263][nvbugs/6598494]

prisrivastav-nv:
### **1. Two gpu placement for combined search standalone and search + alerts builds**

Establishes explicit GPU-placement and RT-VLM variant guidance for headless build-vision-agent use cases (Search Standalone, Search + Alerts verification, Search + Alerts realtime), and corrects stale/ambiguous guidance that could mislead the agent at resolution time.

### NVBugs
[Three-GPU placement blocks deployment](https://nvbugs/6591263)
[Stale 3-GPU sizing blocks supported 2-GPU local_shared placement](https://nvbugspro.nvidia.com/bug/6598494)

### Problem statement

- No strict GPU-placement guidance existed for headless builds. Left unconstrained, deployments drifted to three GPUs even when two suffice, and RT-VLM placement/variant was undefined.
- The Cosmos3 Nano VLM variant has diverged (heavier BF16 vs lighter FP8) between the Search and Alerts capabilities, but the skills had no explicit singleton/merge rule for it — unlike RT-CV and behavior-analytics.
- The Search foundation profile reserved a GPU for NVStreamer; no other profile does, and NVStreamer needs no GPU.
- The combined behavior-analytics config guidance was vague and inconsistent with the shipped drop-in config and the combined-build eval, and the standalone setup skill still instructed users to hand-merge the two single-mode configs.
- Global sizing guidance carried stale rows (Search GPU count, Alerts `2d_cv` RT-VLM) and lacked a generic, deterministic placement-resolution flow.

### Approach

Treat RT-VLM as a singleton and resolve it generically at the framework level, rather than adding build-combo-specific sections:

- Select the variant by **placement**: a dedicated GPU takes the heavier BF16; co-residence with another GPU service takes the lighter FP8.
- Take the **variant/placement knob-set atomically** from the resource-driven owner (Search when constrained, Alerts when a GPU can be dedicated); keep **consumer-wiring** knobs (Kafka enablement/topic, alert-bridge endpoint, verifier mounts) with the consuming capability and operating mode — never inherited from the variant owner.
- Place fixed-footprint services (RT-CV, RT-Embed) first; place singleton RT-VLM last, co-resident on the RT-CV device when constrained. Realtime vs verification does not change the variant in a constrained environment — differentiate by stream count and headroom, not by checkpoint.

Resulting placement matrix (headless builds; no LLM/agent tier):

| Scenario | Free GPUs | GPU 0 | GPU 1 | GPU 2 | RT-VLM variant | Owner |
|---|---|---|---|---|---|---|
| Search Standalone | 2 | RT-CV + RT-VLM (FP8 @ 0.40) | RT-Embed | — | FP8 | search |
| Search + Alerts, shared | 2 | RT-CV + RT-VLM (FP8 @ 0.40) | RT-Embed | — | FP8 | search |
| Search + Alerts, dedicated | >=3 | RT-CV | RT-Embed | RT-VLM (BF16 @ ~0.70, dedicated) | BF16 | alerts |

### Changes

- Removed the anomalous NVStreamer GPU reservation so the Search profile aligns with the other profiles and the base definition.
- Added an RT-VLM singleton + placement-driven variant-merge rule to the RT-VLM owner contract, including the split between atomic variant/placement knobs and capability/mode-owned consumer-wiring knobs.
- Reframed the global sizing guidance to a generic, profile-agnostic placement-resolution flow, corrected the stale Search and Alerts `2d_cv` rows, and made the constrained co-resident RT-VLM placement deterministic.
- Added fixed-footprint placement notes to the RT-CV and RT-Embed owner contracts so the placement pass reads them from the owning layer.
- Corrected the combined behavior-analytics config recipe to bind the shipped joint config directly as a drop-in (no staging into build artifacts), matching the combined-build eval, and fixed the stale hand-merge instruction in the standalone behavior-analytics setup skill (retaining a manual merge only as a fallback for bespoke configs).

### Additional guidance changes (build correctness beyond placement)

Two related correctness gaps surfaced while validating generated combined builds are addressed here as well:

- **Mounted-config mode reconciliation.** Beyond the bind recipe, a combined build must reconcile the behavior-analytics mounted JSON's *operating mode* to the request — `inSimulationMode`, `numWorkersFor*` gates, class scope — because these are not env- or `COMPOSE_PROFILES`-expressible and env-delta resolution silently inherits the source profile's mode (Search ships `inSimulationMode=true` with incidents off, yielding no alerts on a live build). Captured in the RT-VLM/behavior-analytics owner contracts, a framework-level `Validate` hook in `composition.md`, and an `inSimulationMode`-unset eval assertion.
- **Config-convergence routing.** Added forward-references from the `search.md` and `alerts.md` owner contracts to `behavior-analytics.md` so the router lands on the config-convergence rule whenever those capabilities share one analytics instance, closing the gap that let the agent stop at `search.md` and inherit the wrong config.

### Build-artifact hygiene

- Hardened the `composition.md` resolve recipe to forbid diverting `docker compose config` stderr to a build-directory file (`2> resolve.err`): stderr is transient diagnostics to read and act on, never a persisted build artifact.

---

### **2. Video Analytics API service name swap in Alerts & Search combined build**

The combined Alerts + Search build eval failed: the resolved `COMPOSE_PROFILES` contained `vss-video-analytics-api-alerts` instead of the expected `vss-video-analytics-api-fusion` (26/27 keys matched). The agent **swapped** one service name to the wrong Foundation's variant — mixing search and alerts names.

### Root cause
Which key survives a singleton convergence is **foundation-relative** knowledge the framework delegates to owner contracts. Three of the five aliased owners (`rt-cv`, `vios`, `elk`) carry that local statement; the analytics-API owner (`alerts.md`/`search.md`) was the only one missing it, so the agent took the alerts key at face value. The swap is a foundation-membership error, not a `container_name` collision.

### Approach
Fix the swap at the layer that owns foundation-relative key choice (the owner contract), stated **agnostically** ("use the selected Foundation's key," no primary/secondary, no cross-references). Add an independent, foundation-agnostic graph-validation net for the distinct duplicate case. No changes to `composition.md` (avoids global/owner conflation).

### Fix
- Owner contracts: added the missing per-Foundation key anchor to the analytics-API and behavior-analytics owners, phrased Foundation-agnostically.
- Validator: `resolved.yml` now fails if two services share a `container_name` (deploy-safety net) + unit tests.
- Eval: reworded the combined query to be capability-driven and search-forward, reducing Foundation-selection ambiguity.

## Checklist


skills/vss-build-vision-agent/references/sizing.md
```diff
@@ -8,16 +8,31 @@ contract.
...
+   [RT-Embed](services/rt-embed.md).
+4. Place singleton RT-VLM last. When composition must converge different
+   integrated Cosmos3 Nano variants, use BF16 if a GPU can be dedicated to
+   RT-VLM; otherwise use FP8 co-resident on the RT-CV device
+   (`RT_VLM_DEVICE_ID = RT_CV_DEVICE_ID`), which retains the most headroom among
+   the fixed-footprint services placed in step 3. Share only when the combined
+   budget fits. Resolve the atomic variant/placement set without inheriting
+   consumer wiring, as specified by the [RT-VLM owner](services/rt-vlm.md). Stock
+   mode retains its Foundation's reviewed variant and placement. For continuous
+   VLM inference on a shared GPU, reduce `NUM_STREAMS` and verify utilization
```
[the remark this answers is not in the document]
prisrivastav-nv: addressed

zac-wang-nv:
## Harbor Eval — `skills/vss-build-vision-agent/eval/profile_combined_alerts_search_harbor.json`

Head: `6567d3a` · platform `RTXPRO6000BW` · run `31687762223`
First started: `2026-08-13T11:38:57Z` · Last finished: `2026-08-13T12:03:12Z` · Total: `0hr 24min`

| Platform | Step | Query | Result | Reward | Duration | Turns | Prompt tok | Cached tok | Trace |
|---|---|---|---|---|---|---|---|---|---|
| RTXPRO6000BW | step-1 | Build combined-alerts-search (validate, do not deploy) | ❌ 0.79 (15/19) | 0.79 | 24m 14s | 91 | 6.8M | 6.6M | [trace](https://harbor-13xh5gpe7.brevlab.com/jobs/vss-build-vision-agent__profile_combined_alerts_search_harbor__RTXPRO6000BW__31687762223__2026-08-13__11-38-57/tasks/rtxpro6000bw/claude-code/aws/anthropic%2Fbedrock-claude-opus-4-6/nvidia-vss%2Fvss-build-vision-agent-profile_combined_alerts_search_harbor-rtxpro6000bw-step-1) |
| RTXPRO6000BW | step-2 | Deploy validated resolved.yml | ⏭️ skipped (prior-step fail, step-1 reward=0.79) | — | — | — | — | — | — |
| RTXPRO6000BW | step-3 | Provision, ingest, query round-trip | ⏭️ skipped (prior-step fail, step-1 reward=0.79) | — | — | — | — | — | — |

### Failing checks (step-1)

- **Check 6** — `_builds/combined-alerts-search/patches/` contains exactly three service-definition patches plus HAProxy config: Agent produced 4 service patches (extra `alert-bridge.yml`), `vss-search-analytics-2d-fusion.yml` lacks `main_search_and_alerts_app.py` command, and `vss-haproxy-ingress.yml` uses absolute path instead of relative source.
- **Check 11** — RT-VLM consumer wiring `RTVI_VLM_KAFKA_ENABLED=false` for `2d_cv` mode: Agent set `RTVI_VLM_KAFKA_ENABLED=true` instead of `false`; in the `2d_cv` verification path, RT-VLM should not publish realtime VLM events (verified incidents go through alert-bridge).
- **Check 16** — `compose.yml` includes exactly three patches and build contains only required artifacts: Same root cause as check 6 — extra `alert-bridge.yml` patch violates the "exactly three" constraint.
- **Check 18** — Generated architecture documents streaming/on-demand ingestion converging at VIOS with agent-free REST provisioning: Architecture describes Kafka-based data flows but doesn't document the specific VIOS convergence and direct-REST provisioning pattern.

<sub>Generated by the skills-eval agent. Adapter/verifier changes
required to make this PR evaluable were raised as bot PRs targeting
the source PR's branch (linked above where applicable) — the
skills-eval agent never commits to `skills/` and never runs trials
against locally-synthesized adapters. Trial datasets/results live in
this leg's workflow artifact
`skills-eval-results-pr-1634-vss-build-vision-agent__profile_combined_alerts_search_harbor-31687762223.tar.gz`.</sub>

skills/vss-build-vision-agent/references/composition.md
```diff
@@ -202,6 +202,12 @@ in `COMPOSE_PROFILES`, not a shortened or generic name. Do not copy unchanged
 services, volumes, networks, or profile files. Add multiple patch paths after
 the root file when multiple service definitions change.
 
+A build-local file a patch bind-mounts (e.g. a curated `haproxy.cfg`) lives in
```
zac-wang-nv, line 205: **P1** — The housing instinct here is right: a build-local file should live beside the patch that mounts it. **The resolution rule is inverted.** Because the paragraph is generic ("a build-local file a patch bind-mounts"), it governs every future patch mount, not just haproxy.

Compose resolves relative paths in an included file against the **project directory of the include entry**, which for a `path:` list defaults to the directory of the **first** file in that list — here `../../deploy/docker/compose.yml`. Not the patch's own directory.

I built the exact shape this doc prescribes (`composition.md:182-187`) and ran it (Compose v2.39.1):

```yaml
# _builds/demo/compose.yml
include:
  - path:
      - ../../root_area/compose.yml      # stands in for deploy/docker/compose.yml
      - ./patches/vss-haproxy-ingress.yml
```

with `./haproxy.cfg` as the volume source in the patch and the real file at `_builds/demo/patches/haproxy.cfg`. `docker compose config` resolved it to:

```yaml
      - type: bind
        source: <tmp>/root_area/haproxy.cfg      # <- the ROOT file's directory
        bind:
          create_host_path: true
```

The curated file in `patches/` is never read. Translated to this repo, `./haproxy.cfg` resolves to `deploy/docker/haproxy.cfg`, which does not exist.

It then fails in the worst available way — silently at `config` time, destructively at `up` time. `docker compose up` creates the missing source as a **root-owned directory** named `haproxy.cfg` (`type=directory owner=root:root mode=755`) inside the root compose file's directory, and the container exits `cat: read error: Is a directory`. HAProxy gets a directory where its config belongs.

That lands 120 lines below, in this same file: `composition.md:327-329` says "do not create placeholder files or directories under `deploy/docker/` to satisfy it." This rule makes Docker create exactly that, automatically, in tracked source.

**The sentence you added directly below (`:208-209`) already reaches for the right tool** — "A checked-in repo file a patch mounts is instead bound by its **absolute** repo path." That is correct. The irony is that the *build-local* case in the sentence above needs absolute addressing more than the checked-in case does: a checked-in file under `deploy/docker/` would actually resolve fine relatively, because `deploy/docker/` is exactly where relative sources land. The two halves of this paragraph have the rule backwards from each other.

Two repairs I tried and can rule out, so they don't get picked up as the obvious fix:

- **`project_directory:` on the include entry** — works for the patch but re-bases the root graph too. With `project_directory: .`, a root-side `./svcdata` moved from `root_area/svcdata` to `_builds/demo/svcdata`. Wrong lever.
- **Splitting into two include entries** — `docker compose config` fails outright: `services.vss-haproxy-ingress conflicts with imported resource`, exit 1. The ordered path list is load-bearing.

What works, verified in the same harness rather than assumed: reference the file by an absolute path interpolated from `override.env`, which `config` already bakes in. With `BUILD_DIR` in `override.env` and the source written `${BUILD_DIR}/patches/haproxy.cfg`, it resolved to `.../_builds/demo/patches/haproxy.cfg` **and** left the root graph's own `./svcdata` untouched.

So the file can still live in `patches/` exactly as you want; it just cannot be addressed relatively. Suggested rewording:

> A build-local file a patch bind-mounts (e.g. a curated `haproxy.cfg`) lives in `patches/` beside its `.yml` and is referenced by an absolute `${BUILD_DIR}/patches/<file>` source — a relative `./` source would resolve against the root Compose file's directory (`deploy/docker/`), not the patch's, and Docker would silently create a directory there at `up` time.

skills/vss-build-vision-agent/references/services/ingress.md
```diff
@@ -57,9 +57,15 @@ which is expected and harmless because no read path consumes it. Do not add the
...
-  service-definition patch that overrides the config volume, keeping the browse +
-  operate routes above for the build's deployed backends and replacing the catch-all.
+- **Curated (patch).** Write the trimmed config to `patches/haproxy.cfg`, beside
+  the `patches/vss-haproxy-ingress.yml` service-definition patch that overrides the
+  config volume, keeping the browse + operate routes above for the build's deployed
+  backends and replacing the catch-all. Adapt the stock relative mount rather than
+  authoring a new one: the stock `haproxy.cfg.template` mount is
+  `- ./haproxy.cfg.template:/usr/local/etc/haproxy/haproxy.cfg:ro`, so the patch's
+  volume entry is exactly `- ./haproxy.cfg:/usr/local/etc/haproxy/haproxy.cfg:ro` —
+  a relative `./haproxy.cfg` source (it resolves against the patch's own directory,
```
zac-wang-nv: **P1** — Flagged separately from `composition.md:205` because this is the bullet an agent actually reads when it builds the curated ingress, and the parenthetical here is the more emphatic of the two.

**Both halves of that parenthetical are wrong.** It does not resolve against the patch's own directory, and the alternative is not the build root either — it resolves against the **root Compose file's** directory, `deploy/docker/`. Full reproduction and failure mode are in my comment on `composition.md:205`.

The rest of this bullet is good — writing the trimmed config to `patches/haproxy.cfg` beside the `.yml` is the right housing, and keeping the browse + operate routes while replacing the catch-all is the right shape. Only the source reference needs to change: `${BUILD_DIR}/patches/haproxy.cfg` rather than `./haproxy.cfg`.

**Credit where it's due:** an earlier draft of this note escalated on the sentence *"bind it relative — never by an absolute path"*, which forbade by name the one shape I had verified works. You deleted that sentence in `895c9e897`. That removed the escalation, and I'm not relocating it onto other evidence.

**The eval at this head corroborates the defect, on one leg of two.** `profile_search_standalone_harbor` scored **0.933 (14/15)** and the single failing check is *"HAProxy volume binding uses absolute path instead of relative source"* — the seventh consecutive trial in which the agent reaches for the absolute source and is marked wrong for it. **The combined leg did not reproduce that axis this time**: `profile_combined_alerts_search_harbor` scored **0.895 (17/19)**, and its two failures concerned an extra `alert-bridge.yml` patch beyond "exactly three", not the path rule. Stating that plainly because it cuts against the strength of my own claim.

So the current state is: the **prose** tells the agent to use a relative source, the **eval** fails it when it doesn't, and the resulting artifact **silently breaks** — `./haproxy.cfg` resolves to `deploy/docker/haproxy.cfg`, and `up` creates a root-owned directory there in tracked source.

**Ask:** flip the rule at all four sites (here, `composition.md:205`, and the two eval checks). Or, if you believe the relative source does resolve against the patch directory, show me a `docker compose config` output where it does and I'll withdraw comments 1-4 in full.

skills/vss-build-vision-agent/eval/profile_combined_alerts_search_harbor.json
```diff
@@ -18,24 +18,26 @@
...
+        "`_builds/combined-alerts-search/patches/` contains exactly three service-definition patches (each reusing its canonical service key) plus the one curated HAProxy config payload the ingress patch references, copying no unchanged service, volume, network, or profile file: `vss-search-analytics-2d-fusion` mounts the shipped combined behavior-analytics config and runs `apps/search_and_alerts/main_search_and_alerts_app.py`, `kibana-init-container-search` mounts the merged Kibana saved-object bundle (below), and `vss-haproxy-ingress` mounts a curated HAProxy config — a payload file co-located in `patches/` beside the patch and bound by a relative source, not stranded at the build root — carrying the operate route-set (`/vst`, `/elasticsearch` read-only, `/rtvi-embed`, `/rtvi-cv`) alongside the browse routes; RT-VLM is not fronted.",
```
zac-wang-nv: **P1** — This is the part I'd most like reconsidered. The path rule from `composition.md:205` has been copied into the eval spec, which changes what it costs. A wrong instruction in a reference doc misleads an agent. **A wrong assertion in an eval spec *rewards* it** — the grader now marks the broken build correct, and marks a working build (absolute `${BUILD_DIR}` source) as non-conforming. The eval stops being a check on this behaviour and becomes a ratchet holding it in place.

Concretely, this file now asks for three things that cannot all be true at once:

- **L28** (this line) — the HAProxy payload is "bound by a relative source".
- **L39** — "every checked-in bind source (**including** ... **the curated HAProxy config**) exists with the expected file/directory type."
- **L41** — "No file under `{{repo_root}}/deploy/docker/` is modified."

A relative source resolves to `deploy/docker/haproxy.cfg` (reproduction in my `composition.md:205` comment). So L28 puts the bind source at a path that does not exist, which fails L39. Making it exist means writing into `deploy/docker/`, which fails L41. And if it gets created the way Docker creates it at `up` time, it is a *directory*, which fails L39's "expected file/directory type" clause specifically. **There is no build an agent can produce that satisfies all three.**

Note this is not a pre-existing tension the commit exposed — L39's parenthetical naming the curated HAProxy config and L28's relative-source requirement both arrived in this PR.

Fixing the rule in `composition.md:205` fixes this line too: with `${BUILD_DIR}/patches/haproxy.cfg`, the payload lives in `patches/` (satisfying the co-location you're asserting), the resolved source exists as a file (L39), and nothing under `deploy/docker/` is touched (L41). Suggest replacing "bound by a relative source" with "bound by an absolute `${BUILD_DIR}/patches/` source" — the co-location claim you actually want to pin is unaffected.

skills/vss-build-vision-agent/eval/profile_search_standalone_harbor.json
```diff
@@ -27,13 +27,13 @@
...
-        "The build directory contains exactly the three required primary artifacts (`override.env`, `compose.yml`, and `resolved.yml`) plus the single `patches/vss-haproxy-ingress.yml` entry, and no copied Foundation `.env`, copied root Compose project, patched tree, manifest, allow-list, or generated skill bundle.",
+        "`vss-haproxy-ingress` is retained because a single origin is requested, and `_builds/search-standalone/compose.yml` includes `../../deploy/docker/compose.yml` followed by `./patches/vss-haproxy-ingress.yml` — a service-definition patch that mounts a curated HAProxy config (a payload file co-located in `patches/` beside the patch and bound by a relative source, not stranded at the build root) carrying the operate route-set (`/vst`, `/elasticsearch` read-only, `/rtvi-embed`, `/rtvi-cv`) alongside the browse routes; RT-VLM is not fronted.",
```
zac-wang-nv: **P1** — Same encoding as `profile_combined_alerts_search_harbor.json:28`, in the second eval spec; see that comment for the reproduction and the fix.

One difference worth stating so the two aren't treated as identical: this file's resolution assertion (L35) says "every **checked-in** bind source exists with the expected file/directory type" without the parenthetical the combined spec has. The curated HAProxy payload is build-local, not checked in, so the strict three-way contradiction I describe on the combined spec does not apply verbatim here — L35 arguably does not range over it.

The defect that does apply is the same one: the spec pins a source form that resolves to `deploy/docker/haproxy.cfg` and never reads the file it says the agent must co-locate, so **a conforming build is a broken build**. Same fix — `${BUILD_DIR}/patches/haproxy.cfg`.

This is also the spec whose trial at this head scored **0.933 (14/15)** with the lone failure being *"HAProxy volume binding uses absolute path instead of relative source"* — i.e. this check is currently the only thing standing between this leg and green.

skills/vss-build-vision-agent/references/services/rt-vlm.md
```diff
@@ -15,6 +15,36 @@
...
+Resolve the variant/placement knobs as one set:
+`RTVI_VLM_MODEL_PATH`, `VLM_NAME`, `RTVI_VLLM_GPU_MEMORY_UTILIZATION`,
+`RTVI_VLM_MAX_MODEL_LEN`, and `RT_VLM_DEVICE_ID`. Take the checkpoint and model
+name from the profile that ships the resolved variant; resolve maximum model
+length, device ID, and utilization together for the selected hardware and
+placement. Keep `VLM_NAME` aligned with the model id advertised by
+`RTVI_VLM_MODEL_PATH`; do not combine values from different variants.
+
+Consumer wiring is not part of that set.
+`RTVI_VLM_KAFKA_ENABLED`, `RTVI_VLM_KAFKA_TOPIC`, and verifier config mounts
```
zac-wang-nv: **P1** — The rule this paragraph states is right, and it's the useful half of the new section: consumer wiring follows the consumer, not the variant profile. The problem is one of the two knobs it names to carry that rule. **`RTVI_VLM_KAFKA_TOPIC` is not read by anything.**

`rtvi-vlm-docker-compose.yml` maps every RT-VLM knob explicitly, and there is no entry for it:

```yaml
# deploy/docker/services/rtvi/rtvi-vlm/rtvi-vlm-docker-compose.yml
:94   KAFKA_ENABLED:        "${RTVI_VLM_KAFKA_ENABLED:-true}"                     # <- real
:95   KAFKA_INCIDENT_TOPIC: "${RTVI_VLM_KAFKA_INCIDENT_TOPIC:-vision-llm-events-incidents}"
:100  MESSAGE_BUS_TOPIC:    "${RTVI_VLM_MESSAGE_BUS_TOPIC:-mdx-vlm-captions}"
```

`RTVI_VLM_KAFKA_ENABLED` is fine — it lands on `KAFKA_ENABLED` at `:94`. `RTVI_VLM_KAFKA_TOPIC` has no mapping at all. Repo-wide it has five occurrences and not one is a reader: `rtvi-vlm/.env:25`, this line, the knob table at `:55`, and two lines in `eval/profile_in_1_streaming_dense_captions.json`.

I checked the one path that could have made the `.env` setter live, because "it's set somewhere" is not the same as "something reads it": **`rtvi-vlm-docker-compose.yml` has no `env_file:` directive**, so `rtvi-vlm/.env` is not loaded into the container either. Dead at both ends.

So an agent that follows this sentence sets `RTVI_VLM_KAFKA_TOPIC=<x>` in `override.env` and gets **silence** — no error, no warning, and RT-VLM keeps publishing to whichever default applies. The sentence's own stated stake makes that concrete: three lines down it says "Realtime VLM alerting must set Kafka enablement to `true` ... or no incidents are published." Incidents are routed by `RTVI_VLM_KAFKA_INCIDENT_TOPIC`, generated captions by `RTVI_VLM_MESSAGE_BUS_TOPIC`. Neither is the name this line gives. **The agent flips the enable bit correctly and then aims the topic at nothing.**

Two qualifications, because they change who owns what:

- **This line is added by this PR** (hunk `@@ -15,6 +15,34 @@`). Unlike the `rt-embed.md` note, this is not a stale contract inherited — the new "Singleton and variant convergence" section introduces it. The occurrence at `:55` in the knob table *is* pre-existing and I'm deliberately not flagging it here.
- **This is the same defect class `e2f1240c7` was written to fix.** That commit correctly renamed RT-Embed's `RTVI_EMBED_KAFKA_*` to the knobs compose actually reads. RT-VLM has the identical problem one service over, and the new prose walks into it.

Suggested fix — name the knobs that match the intent:

> `RTVI_VLM_KAFKA_ENABLED`, `RTVI_VLM_MESSAGE_BUS_TOPIC` (generated captions) and `RTVI_VLM_KAFKA_INCIDENT_TOPIC` (verification incidents), and verifier config mounts follow the consuming capability and operating mode, never the profile that supplied the variant.

P1 rather than P2 because it is an executable instruction in an agent-consumed contract, it is introduced rather than inherited, and it fails silently — the build comes up, publishes to the wrong topic, and nothing in the readiness checks would catch it.

skills/vss-build-vision-agent/references/services/ingress.md
```diff
@@ -57,9 +57,15 @@ which is expected and harmless because no read path consumes it. Do not add the
 route to satisfy the probe; that would re-expose RT-VLM's SSE generation endpoints
 through HAProxy.
 
-- **Curated (patch).** Mount a trimmed config via a `patches/<haproxy>.yml`
-  service-definition patch that overrides the config volume, keeping the browse +
-  operate routes above for the build's deployed backends and replacing the catch-all.
+- **Curated (patch).** Write the trimmed config to `patches/haproxy.cfg`, beside
```
zac-wang-nv, line 65: **P1** — Separate from the path-resolution defect in the parenthetical below: this is about *when* the Curated-patch branch gets selected at all.

Hardening this bullet from a generic reference into two concretely named files made it markedly more actionable than the alternatives in the same list, and the agent now reaches for it on builds where the correct answer is to prune HAProxy entirely.

**Measured at this head**, `profile_in_1_streaming_dense_captions` scored **0.67 (8/12)** against a develop control of **0.9167 (11/12)** — and 0.67 is the same figure the previous two heads produced, so it reproduces. The step gate is what makes it matter: at 0.67, steps 2, 3 and 4 are `skipped (prior-step fail)`, so the leg forfeits build-validation, deploy and E2E coverage. Develop's 0.9167 cleared that gate.

**A correction I owe you on this finding.** An earlier round reported this leg at 0.3333 and claimed a seven-check cascade. That was wrong — the re-measurement is 0.67, and the `patches/` check I had cited as verbatim causal evidence was *not* among the failures. Of the four checks that do fail, one fails on develop too (pre-existing) and one is a judge artifact. That leaves two genuinely attributable, and they share a root cause: the agent proposed removing **nothing** — retaining `vss-agent`, `vss-ui`, `phoenix` and `vss-haproxy-ingress` alike. I can't attribute that broader failure to this file, so I'm not claiming the cascade. The magnitude was wrong; the direction was not.

**What does corroborate it, at this head:** `profile_in_2_rt_cv_person_detection_harbor` scored **0.636 (7/11)**, and the run's own summary line reads *"agent included `vss-haproxy-ingress` (excluded by spec)"* — the retention mechanism firing by name on a spec that explicitly excludes it.

**Caveat that cuts against me:** n=1 per side on a nondeterministic LLM eval, and the develop control is older than the current merge-base.

**Ask** (narrower than it once was, and no longer a blocker on its own): scope the Curated-patch guidance so it isn't selected for headless builds that need no service-definition change.

skills/vss-build-vision-agent/references/services/ingress.md
```diff
@@ -57,9 +57,15 @@ which is expected and harmless because no read path consumes it. Do not add the
...
+  the `patches/vss-haproxy-ingress.yml` service-definition patch that overrides the
+  config volume, keeping the browse + operate routes above for the build's deployed
+  backends and replacing the catch-all. Adapt the stock relative mount rather than
```
zac-wang-nv: **P1** — This is the new worked example, and I want to be precise about it: **the citation is exactly right and the inference from it is not.**

The stock mount really is written `- ./haproxy.cfg.template:...` — `deploy/docker/services/infra/haproxy/compose.yml:72` is character-for-character that line. I checked it rather than assuming, and re-checked it after the rebase (blob-identical across the base advance).

But the stock mount and the patch are loaded by two **different include forms with different resolution bases**:

| | declared in | include form | `./` resolves against |
|---|---|---|---|
| stock `haproxy.cfg.template` | `services/infra/haproxy/compose.yml` | **single-path** chain (`compose.yml` -> `services/` -> `infra/` -> `haproxy/`, each `- path: ./x/compose.yml`) | its **own** directory |
| patched `haproxy.cfg` | `_builds/<b>/patches/vss-haproxy-ingress.yml` | **two-element `path:` list** (`composition.md:182-187`) | the **first entry's** directory |

The stock mount works precisely *because* its file sits next to the compose file that declares it. **That property is not transferable** to a patch pulled in through an ordered `path:` list, where Compose sets the project directory from the *first* file in the list — `../../deploy/docker/compose.yml`.

Single render, two `./` sources, one model, two bases (Compose v2.39.1):

```
services:
  vss-haproxy-ingress:                                    # from the 2-element path: list
        source: <tmp>/root_area/haproxy.cfg                                    # <- ROOT file's dir
  vss-untouched-probe:                                    # from the single-path include chain
        source: <tmp>/root_area/services/infra/haproxy/haproxy.cfg.template    # <- its OWN dir
```

`root_area/` stands in for `deploy/docker/`. So translated to this repo, the patch's `./haproxy.cfg` resolves to `deploy/docker/haproxy.cfg`, and the curated file the bullet tells you to write at `patches/haproxy.cfg` is never read. At `up` time Docker creates that missing source as a **root-owned directory** and the container exits 1.

**Suggestion: drop the analogy rather than repair it.** Even with the parenthetical corrected, "adapt the stock relative mount" points an agent at a mount whose resolution base it cannot inherit. The ask from the comment above is unchanged: `${BUILD_DIR}/patches/haproxy.cfg` — verified working, root graph untouched.

zac-wang-nv (commented):
**Review of `895c9e897`** — 7×P1 and 4×P2 posted inline below. Verdict: **request changes**, on one root cause plus a few independents. Posting as a comment review rather than a blocking one.

**First, the four red checks on this PR right now are not yours.** `Assemble release set`, `Build vss-configurator (amd64/arm64)` and `Wait for Build Dev Images` all fail *identically* on `develop` at the commit you rebased onto (`ec3cc06cb`, #1623 — the commit that enabled GHCR builds for `vss-configurator`). The rebase imported them. `ec3cc06cb` additionally has `SonarQube Scan (spatialai-data-utils)` red where your head is green. Required checks are **7/7 green**. Please don't spend time on those four.

**The central issue** (comments 1, 2, 3, 4, 20): the Compose path-resolution rule this PR states is inverted, and it has been copied into two eval specs, so the grader now rewards the broken shape. Compose resolves a relative source in a `path:`-list include against the **first** entry's directory — `deploy/docker/` — not the patch's own directory. Reproduced under Compose v2.39.1, rebuilt from scratch at the current merge-base:

```
services:
  vss-haproxy-ingress:                                    # two-element path: list
        source: <tmp>/root_area/haproxy.cfg                                    # <- ROOT file's dir
  vss-untouched-probe:                                    # single-path include chain
        source: <tmp>/root_area/services/infra/haproxy/haproxy.cfg.template    # <- its OWN dir
```

Both are written `./<file>`; they resolve against different directories in one rendered model. At `up` time Docker then creates the missing source as a **root-owned directory** and the container exits 1 (`cat: read error: Is a directory`).

**The eval sweep at this head corroborates it, partially.** `profile_search_standalone_harbor` scored **0.933 (14/15)** with exactly one failing check: *"HAProxy volume binding uses absolute path instead of relative source"* — identical to the previous head, and the seventh consecutive trial in which the agent reaches for the absolute source and is graded wrong for it. **In fairness, the combined leg did not reproduce that axis this time**: `profile_combined_alerts_search_harbor` scored **0.895 (17/19)** and its two failures were about an extra `alert-bridge.yml` patch beyond the spec's "exactly three", not the path rule. So the axis is reproducing on one leg here, not both.

**Standing offer:** show me a `docker compose config` in which a `path:`-list patch's `./` source lands in the patch's own directory, and I withdraw comments 1-4 and 20 in full.

skills/vss-build-vision-agent/references/profiles/alerts.md
```diff
@@ -60,12 +60,12 @@ curl -sf "http://${HOST_IP}:9901/health"
 curl -sf "http://${HOST_IP}:3000/"
 ```
 
-For `2d_cv`, also require `vss-rtvi-cv` and `vss-behavior-analytics` to
-resolve and probe
+For `2d_cv`, also require `vss-rtvi-cv`, `vss-behavior-analytics`, and the
+verification `rtvi-vlm` to resolve; probe
```
zac-wang-nv, line 64: **P2** — First, credit: taking the PR side of this file wholesale during the rebase could have silently reverted `e3a89716d fix(skills): use RT-CV readiness endpoint for alerts`. It didn't — `/api/v1/ready` with `HTTP 200` and `ready-info.ds-ready=YES` is intact at `:65-66`, and `e3a89716d` is now an ancestor of the branch. That concern is closed.

The resolution did drop one thing from your own side, though, and I think it was unintentional.

Pre-rebase, this `2d_cv` block probed two endpoints:

```text
`http://${HOST_IP}:${RTVI_CV_HOST_PORT:-9010}/v1/health` and
`http://${HOST_IP}:8018/v1/health/ready`.
```

Post-rebase it probes one. The `:8018` RT-VLM probe survives at `:68`, but only for `2d_vlm`. So `2d_cv` now **requires `rtvi-vlm` to resolve** — this PR's headline addition to this file, on the line directly above — but never checks that it is ready. The readiness checklist for `2d_cv` can pass with the verifier still coming up, which is the mode where RT-VLM does per-clip alert verification. `2d_vlm`, where RT-VLM is arguably less load-bearing to the alert path, is the mode that gets the probe.

Taking develop's endpoint text wholesale for those two lines is what dropped it — their version had no `:8018` clause to preserve, because at the time `2d_cv` did not require RT-VLM.

Suggest restoring it on top of develop's wording:

```text
For `2d_cv`, also require `vss-rtvi-cv`, `vss-behavior-analytics`, and the
verification `rtvi-vlm` to resolve; probe
`http://${HOST_IP}:${RTVI_CV_HOST_PORT:-9010}/api/v1/ready`, requiring HTTP 200
and `ready-info.ds-ready=YES`, and `http://${HOST_IP}:8018/v1/health/ready`.
```

P2 rather than P1 because `rtvi-vlm` is still required to resolve, so this is a weaker check rather than an absent one, and it is guidance prose rather than an executable gate.

The rest of the hunk checks out: `:8018` is right for the RT-VLM probe (`dev-profile-alerts/.env:80` `VLM_PORT=8018`, `:118` `RTVI_VLM_PORT=8018`), and splitting the `2d_vlm` sentence so it names both absent services explicitly is clearer than the old "those two services".

skills/vss-build-vision-agent/references/composition.md
```diff
@@ -320,6 +327,11 @@ Then verify:
 - Every checked-in bind source exists and a file target is not backed by a
   directory. This is a validation check only: do not create placeholder files
   or directories under `deploy/docker/` to satisfy it.
+- A service governed by a mounted config file has that config reconciled to the
+  requested capabilities and ingestion mode — e.g. a mounted analytics JSON's
```
zac-wang-nv, line 335: **P2** — The rule itself is correct and worth having: a mounted JSON's contents are outside the env-delta contract, so a combined build has to replace the file rather than reconcile around it. **The problem is the example chosen to illustrate it.** `inSimulationMode` is dead code in this repo, and this PR names it in five places.

`services/analytics/behavior-analytics/src/mdx/analytics/core/transform/config/config_validator.py:143-146`:

> ``inSimulationMode`` deliberately excluded -- no-op key, not read by any production code path. Its only reader was ``StateMgmt._get_current_timestamp``, which was dead code and has been removed. Kept out of the allowlist so it is rejected rather than accepted into config that nothing consumes.

The live key with that meaning is a different one: `playbackInSimulationMode` -> `config.py:1581` `inSimulationMode=self.get_bool_app_config("playbackInSimulationMode", IN_SIMULATION_MODE)`, landing on `config.playback.inSimulationMode` and read only by `apps/playback/playback_amr.py:63`, `playback_frames.py:83`, `playback_sorted_frames.py:75`. `SearchAndAlertsApp` is not a playback app, so it never reads it under either spelling.

What that makes wrong downstream:

- `services/behavior-analytics.md:28` — "Search ships `inSimulationMode=true` with incidents off, yielding no alerts". Search does ship it (`vss-search-analytics-redis-config.json:108`, `kafka-config.json:112`), but it changes nothing. The reason Search yields no alerts is `numWorkersForIncidentGeneration=0`, two entries above in the same file. Conjoining them tells a build agent the dead key is half the cause.
- `eval/profile_combined_alerts_search_harbor.json:30` — "leaves `inSimulationMode` unset for live streaming". There is no simulation clocking to avoid. The assertion passes, but **vacuously**: the shipped `search_and_alerts_config.json` never sets the key, so nothing could fail it.
- `services/alerts.md:24` and `services/search.md:32` — both lead the "not env-expressible" pair with the dead key.
- And it can bite: the allowlist that excludes it is the **dynamic**-config validator (`config_listener.py`, `config_applier.py`). An agent that "converges" the mode by explicitly writing `inSimulationMode=false` and pushing it through the hot-update path gets it **rejected**, not applied.

The rest of every one of these sentences is right — `numWorkersFor*` really does gate processor registration (`integrate-behavior-analytics-service.md:113-114`), and `playbackLoop` really is read (`config.py:1579`). Only the lead example is dead.

Suggest: make `numWorkersFor*` the canonical example here since it is the gate that actually decides capability, drop `inSimulationMode` from the two bullets, and reword the eval check to something observable — or drop that clause. If the intent really was playback clocking, the key to name is `playbackInSimulationMode`.

Separately, on the same line: the `e.g.` aside sits between "ingestion mode" and "per its owner contract", so the qualifier lands three lines from the noun it modifies. Moving it to the end of the sentence reads cleaner for a doc an agent parses top-down.

skills/vss-build-vision-agent/references/services/rt-embed.md
```diff
@@ -23,21 +23,27 @@
...
-`RTVI_EMBED_KAFKA_ENABLED=true` and `RTVI_EMBED_KAFKA_TOPIC=mdx-embed` in the
-build `override.env`. The root Compose include path does not load
-`services/rtvi/rtvi-embed/.env`, and the service compose fallback topic is
-`vision-embed-messages`, which does not feed the Search analytics `mdx-embed`
--> `mdx-embed-filtered` indexing path. Without this override the embedding
-write path is broken: RT-Embed produces no Kafka output and
-`mdx-embed-filtered` remains empty.
+The generated-output bus is off unless `MESSAGE_BUS` is set: the root Compose
+include path does not load `services/rtvi/rtvi-embed/.env` (which sets
+`MESSAGE_BUS=kafka`), so it falls back to the compose default empty value, which
```
zac-wang-nv, line 34: **P2** — This commit is a good catch and the diagnosis is exactly right; I verified the rename against `deploy/docker/` rather than reading the commit message, and `MESSAGE_BUS`/`MESSAGE_BUS_TOPIC` are the knobs the compose actually reads (`rtvi-embed-docker-compose.yml:68-69`). I also checked the risk the unprefixed name creates — an agent setting `MESSAGE_BUS=kafka` in a build-global `override.env` — and there is no collision: RT-VLM namespaces its own (`${RTVI_VLM_MESSAGE_BUS:-kafka}`, `rtvi-vlm-docker-compose.yml:99`), so RT-Embed is the sole reader of the bare name.

**The gap is that the fix stops one variable short.** `91c997f9e` — the same Aug-5 commit that renamed the output bus and orphaned this doc — also added a second bus with the identical shape:

```yaml
# rtvi-embed-docker-compose.yml:67-70
# Generated-output and error buses (empty MESSAGE_BUS/ERROR_BUS disables that bus).
MESSAGE_BUS: "${MESSAGE_BUS:-}"
MESSAGE_BUS_TOPIC: "${MESSAGE_BUS_TOPIC:-mdx-embed}"
ERROR_BUS: "${ERROR_BUS:-}"
```

`ERROR_BUS` is unprefixed, defaults to empty, and is set to `kafka` only in `services/rtvi/rtvi-embed/.env:28` — the file this paragraph correctly says the root Compose include path never loads. So the argument this section makes for `MESSAGE_BUS` applies to `ERROR_BUS` verbatim: a generated build that follows this doc gets embedding events on Kafka and **silently loses RT-Embed error reporting**.

`git grep ERROR_BUS -- skills/vss-build-vision-agent/` returns **0 hits** — absent from the env table above and from this contract section, both of which this commit just rewrote.

There is an in-tree reference implementation to copy from rather than invent: the sibling `vss-deploy-video-embedding` skill documents both buses correctly — `references/environment.md:53` ("`ERROR_BUS` | Error output bus. Use `kafka` for Kafka-backed errors; set empty to disable.") and `SKILL.md:207` pairs them as "Kafka (`MESSAGE_BUS=kafka` and/or `ERROR_BUS=kafka`)".

Suggest adding `ERROR_BUS` to the env-table row and one sentence here — e.g. "Set `ERROR_BUS=kafka` alongside it if the build needs RT-Embed error events on Kafka; it is unset by the same mechanism and disabled by the same default." Note the error *topic* is prefixed (`RTVI_EMBED_ERROR_MESSAGE_TOPIC`, default `vision-embed-errors`), so only the bus toggle needs stating.

P2 rather than P1 because it degrades observability rather than the embedding write path, and because it is a pre-existing omission this commit inherits rather than introduces.

skills/vss-build-vision-agent/eval/profile_search_standalone_harbor.json
```diff
@@ -43,7 +43,7 @@
...
         "All selected long-running containers are present and not exited, restarting, or dead; all selected one-shot containers exit 0; and excluded services (including `rtvi-vlm` and the agent tier) are not running. A Docker `unhealthy` label alone is not a failure when the explicit service readiness probes in this step pass.",
         "`curl -sf http://localhost:${RTVI_CV_HOST_PORT:-9000}/ready` returns HTTP 200 and RT-Embed readiness responds successfully.",
-        "`vss configure --base-url http://localhost:${HAPROXY_HOST_PORT:-7777}` records the operate route-set (`vst`, `elasticsearch`, `rt_embed`, `rtvi_cv`), proving the single ingress fronts them."
+        "Curling the operate routes through the ingress origin (`http://localhost:${HAPROXY_HOST_PORT:-7777}`) proves the single HAProxy fronts each deployed backend: `/vst`, `/rtvi-embed`, and `/rtvi-cv` route through (any non-404 backend response, not the `404` host-allowlist reject or the catch-all `/kibana/` redirect), and `/elasticsearch` is reachable read-only (a GET routes through while a mutating method returns the verbatim `405`/`403` edge guard); RT-VLM is not fronted."
```
zac-wang-nv: **P2** — Most of this rewrite is defensible: testing actual routing rather than CLI bookkeeping is a reasonable goal, and the route names it uses (`/vst`, `/rtvi-embed`, `/rtvi-cv`) are the real HAProxy mounts. **The Elasticsearch clause is the problem:**

> ...and `/elasticsearch` is reachable **read-only** (a GET routes through while **a mutating method returns the verbatim `405`/`403` edge guard**)...

Both halves are wrong, and the authoritative contradiction is a comment sitting directly above the rules, pre-existing and not written by this PR:

```
# deploy/docker/services/infra/haproxy/haproxy.cfg.template:170-174
# Not read-only: the rules below stop PUT/DELETE, cluster-admin paths and
# the two-segment mutating operations, but POST is allowed, so
# POST /_bulk, /_reindex, /_scripts/<id> and /<index>/_doc still reach ES.
```

`acl es_read_method method GET HEAD POST OPTIONS` admits POST, and `es_mutating_op` only matches the **two-segment** form `^/elasticsearch/.*/(_bulk|_update|...)`. Simulating the three ACLs against the shipped config:

| Probe | Result |
|---|---|
| `PUT /elasticsearch/idx/_doc/1` | `405` |
| `DELETE /elasticsearch/idx` | `405` |
| `POST /elasticsearch/idx/_bulk` | `403` |
| **`POST /elasticsearch/_bulk`** | **routes through to ES** |
| **`POST /elasticsearch/idx/_doc`** | **routes through to ES** |

So the check is wrong in two distinct ways. It **codifies a false security property** — "reachable read-only" — into a graded contract, when the ingress in fact accepts top-level POST writes. And because it says "a mutating method" without naming one, **it grades nondeterministically**: an agent that probes with `PUT` sees `405` and passes; an agent that reasonably picks `POST /_bulk` — the canonical mutating call, and the very example the config comment cites — sees a backend response and fails. Same deployment, opposite verdicts, decided by an arbitrary choice the check never constrains.

Suggested fix — assert what the guard actually does, and pin the method so the check is deterministic:

> ...and `/elasticsearch` is fronted with the edge guard in force: a `GET` against `/elasticsearch/_cat/indices` routes through, while `PUT`/`DELETE` return `405` and a cluster-admin path such as `/elasticsearch/_cluster/health` returns `403`. (The guard is a denylist, not read-only: top-level `POST` writes such as `/elasticsearch/_bulk` still reach ES by design.)

P2 rather than P1: eval checks are judge-side, so no agent reads this text at build time and nothing ships wrong — the blast radius is one spec's grading stability plus a false claim now recorded in-tree.

skills/vss-build-vision-agent/references/composition.md
prisrivastav-nv: Addressed.

Flipped `composition.md` to bind build-local files by absolute `${BUILD_DIR}/patches/<file>`, with the relative-`./→deploy/docker/` resolution and the root-owned-directory failure mode stated explicitly.

skills/vss-build-vision-agent/references/services/ingress.md
prisrivastav-nv: Addressed; curated bullet now binds `${BUILD_DIR}/patches/haproxy.cfg` and explains the `path:`-list resolution base.

skills/vss-build-vision-agent/eval/profile_combined_alerts_search_harbor.json
prisrivastav-nv: Addressed; combined-spec L28 now asserts an absolute `${BUILD_DIR}/patches/` source, which resolves the L28/L39/L41 three-way contradiction (source exists as a file, nothing under `deploy/docker/` touched).

skills/vss-build-vision-agent/eval/profile_search_standalone_harbor.json
prisrivastav-nv: Addressed; search-spec L33 flipped to the same absolute source.

skills/vss-build-vision-agent/references/services/rt-vlm.md
prisrivastav-nv: Addressed.

`RTVI_VLM_KAFKA_TOPIC` has no compose reader, replaced with `RTVI_VLM_MESSAGE_BUS_TOPIC` (captions) and `RTVI_VLM_KAFKA_INCIDENT_TOPIC` (incidents), the knobs the compose actually maps.

skills/vss-build-vision-agent/references/services/ingress.md
prisrivastav-nv: Addressed (narrower scope).

Added a scoping line to the section so it's reached only when the build includes `vss-haproxy-ingress`; a headless build exposing no single origin prunes the proxy and produces no curated patch.

skills/vss-build-vision-agent/references/services/ingress.md
prisrivastav-nv: Addressed; dropped the "adapt the stock relative mount" analogy (the single-path chain's resolution base isn't transferable to the `path:`-list patch) and pointed the source at `${BUILD_DIR}/patches/haproxy.cfg`.

skills/vss-build-vision-agent/references/profiles/alerts.md
prisrivastav-nv: Addressed.

This was a merge/rebase loss; restored the `:8018/v1/health/ready` probe on the `2d_cv` branch so the mode that requires `rtvi-vlm` also checks its readiness.

skills/vss-build-vision-agent/references/composition.md
prisrivastav-nv: Addressed.

`inSimulationMode` is a rejected no-op; made `numWorkersFor*` the canonical example, dropped the dead key from the two bullets, corrected the "no alerts" cause to `numWorkersForIncidentGeneration=0`, and removed the vacuous eval clause.

skills/vss-build-vision-agent/references/services/rt-embed.md
prisrivastav-nv: Addressed; `ERROR_BUS` was orphaned by the same rename; added it to the env table and one contract sentence (error topic already defaults, so only the bus toggle needs stating).

skills/vss-build-vision-agent/eval/profile_search_standalone_harbor.json
prisrivastav-nv: Addressed.

The guard is a denylist, not read-only. Rewrote the check to assert `GET` routes through, `PUT`/`DELETE` -> `405`, cluster-admin -> `403`, with the top-level `POST /_bulk` write called out — and pinned the methods so it grades deterministically.

zac-wang-nv (approved):

PR #1649, opened 2026-08-11
fix(helm): size nemotron NIM concurrency to each GPU instead of a flat 4

nv-chandrahasj:
## Problem

Every `nemotron` entry in `deploy/helm/services/nims/values.yaml` pins `NIM_MAX_NUM_SEQS=4`, capping the LLM at 4 in-flight sequences per replica regardless of topology.

That value is right when the LLM **shares** a GPU with the VLM — the single-GPU compose files use it for exactly that reason. But the Helm chart applies it to **dedicated**-LLM deployments too, where it becomes the binding constraint: requests queue inside CA-RAG while the LLM GPU stays largely idle, and adding LLM GPUs appears to help only because each replica contributes 4 more slots.

`RTXPRO6000BW` was additionally tuned to `NIM_KVCACHE_PERCENT` / `NIM_GPU_MEM_FRACTION = 0.4` — **half** what H100 and L40S get — despite having 96 GB against their 80 GB and 48 GB. That stranded ~58 GB and cut the KV pool to ~16 GiB.

## Sizing

`nvidia-nemotron-nano-9b-v2` is a hybrid Mamba2-Transformer (`nemotron_h`): only **4 of 56 layers attend**, so KV cost is ~16 KiB/token, plus ~70 MiB of constant Mamba state per sequence — about **2.0 GiB per sequence** fully extended to the 128000 ceiling.

| profile | card | fraction | KV pool | floor @128K | `NIM_MAX_NUM_SEQS` |
|---|---|---|---|---|---|
| `H100` | 80 GB | 0.8 | ~42 GiB | ~20 | 4 → **16** |
| `L40S` | 48 GB | 0.8 | ~14 GiB | ~7 | 4 → **6** |
| `RTXPRO6000BW` | 96 GB | 0.4 → **0.9** | ~16 → ~64 GiB | 8 → **31** | 4 → **24** |

`NIM_MAX_MODEL_LEN` stays at `128000` so long-video aggregation is unaffected. It is a ceiling, not a reservation — shorter requests pack far more densely than these floors imply. `NIM_LOW_MEMORY_MODE` is dropped for `RTXPRO6000BW` only, where a dedicated 96 GB card gains nothing from it.

## Scope

Two files, `nemotron` blocks only. `cosmos` / `cosmos3` profiles are untouched, including the in-flight VLM tuning already on `develop`.

- `deploy/helm/services/nims/values.yaml`
- `deploy/helm/developer-profiles/dev-profile-lvs/README.md` — tuning table refreshed, plus guidance for raising `NIM_MAX_NUM_SEQS` on a dedicated GPU (the existing text only ever said to lower it)

## Validation status — please read before merging

These figures are derived from the model's `config.json` and card capacities. They are **not** measured on every platform:

- **RTX PRO 6000** is the motivating case, inferred from VLM scale-out reports where CA-RAG accounted for 45–84% of end-to-end latency while the LLM GPU was not saturated.
- **H100 and L40S numbers are computed only.** L40S is the tightest card and moves least (4 → 6); reviewers with that hardware should sanity-check it.

At startup vLLM logs `GPU KV cache size` and `Maximum concurrency for <N> tokens per request` — those confirm the profile against real hardware. Operators sharing a GPU between LLM and VLM should keep the lower values.

## Follow-ups not in this PR

- `gpuProfiles` has only a GPU-type axis, so a co-tenancy profile silently applies to dedicated-LLM topologies. The structural fix is a `{gpuType} × {shared, dedicated}` axis; that is a chart API change and wants design discussion first.
- `dev-profile-lvs/values.yaml` hardcodes `nims.gpuType: H100`; correct today only because `values-lvs.yaml` overrides it, but installing with `values.yaml` alone on RTX Pro silently applies H100 tuning.


deploy/helm/services/nims/charts/nemotron-3-nano/templates/nim-cache.yaml
```diff
@@ -0,0 +1,47 @@
...
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+  source:
+    ngc:
+      modelPuller: {{ .Values.image.repository }}:{{ .Values.image.tag }}
+      pullSecret: {{ .Values.global.pullSecretName | quote }}
+      authSecret: {{ .Values.global.authSecretName | quote }}
+      model:
+        profiles:
+          - all
```
pratik-bhatt-nv: **[P1] Cache only the required NIM profile and size the PVC for it.**

`profiles: [all]` downloads every Nemotron 3 Nano profile, while this chart provisions a fixed `50Gi` cache PVC. Please select the profile matching the configured GPU, precision, and TP=1 deployment, and set `storage.pvc.size` based on that profile’s documented artifact size. This avoids unnecessary download time/storage consumption and prevents cache provisioning failures.

nv-chandrahasj: Done

kaushikc-nvidia (approved):

Issue #17, opened 2025-05-05
Run VSS on ARM (ORIN AGX)


ilay-chen:
Hi,
Is there a way to run the VSS base image on Arm? What changes are needed? (Assume all LLM and VLM are remote.)
(I know NIM is not supported on RAM yet.)

Thank you!
Ilay

wyw1267:
Currently, we do not support running VSS on ARM platforms.

wyw1267:
This part is not yet open source at present. Please pay attention to our official announcements about this. Thanks

PR #1650, opened 2026-08-12
fix eval search CLI invocation

ayyappa-dev:
## Description
- Ensure all persisted search eval steps invoke the project-local VSS CLI via `uv run --extra cli`, never a global `vss`.
- Fix ingestion, deletion, deployment, and Kubernetes eval wording/checks to use the same CLI contract.
- Add rendered-instruction regression coverage across all nine search eval steps.

## Checklist


zac-wang-nv (commented):

The internal review failed closed: Error code: 401 - {'error': {'message': 'Authentication Error, Invalid proxy server token passed. Received API Key = sk-...c2Qg, Key Hash (Token) =b899b2e87a0f1f49b35d97d7407f305792b62b6a1ce5dc70421a74e10487bcdd. Unable to find token in cache or `LiteLLM_VerificationTokenTable`', 'type': 'token_not_found_in_db', 'param': 'key', 'code': '401'}}

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32153698439)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32198969951)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32215103307)

.github/skill-eval/adapters/vss-search-archive/generate.py
```diff
@@ -79,47 +79,86 @@
     "for confirmation on `/vss-deploy-profile` or any other setup action the trial requires."
 )
 
+# Harbor restores the deployment and ~/.vss/config.json between persisted
+# steps, but it does not install the optional CLI extra globally. Every step
+# must therefore re-establish the project-local command rather than assuming a
+# previous shell's PATH or variables survived.
+PROJECT_CLI_PREAMBLE = (
+    " Set `VSS_REPO_ROOT=\"${VSS_REPO_ROOT:-$HOME/video-search-and-summarization}\"`, "
```
zac-wang-nv: this should be put to skills about how to use the cli, one instruction is enough, right?

---

Agreed, and there is a reason beyond duplication: `PROJECT_CLI_PREAMBLE` lives in the **eval adapter**, so it is prepended to the eval agent's prompt and ships to no user. An instruction that exists only here is invisible in production and untested for real users — it makes the eval easier without making the skill better.

If the CLI bootstrap is genuinely required, it belongs in the skill. If the skill already says it, this is a second copy that will drift.

services/agent/packages/vss_agents/src/vss_agents/tools/vst/utils.py
```diff
@@ -250,16 +251,28 @@ async def delete_vst_sensor(vst_url: str, sensor_id: str) -> tuple[bool, str]:
...
-                logger.info("VST sensor already absent: %s", scrub_log(sensor_id))
-                return True, "already absent"
             text = await response.text()
-            return False, f"VST returned {response.status}: {text}"
+            delete_error = f"VST returned {response.status}: {text}"
+
+        # VST can remove a file sensor from its cache and still return an
+        # error when a secondary sensor-control cleanup fails. Read back the
+        # durable source registry before reporting a partial Agent deletion.
+        # This also keeps route/proxy errors fail-closed: if the exact UUID is
```
zac-wang-nv, line 282: is there a bug on VST for this?

---

**Yes — documented and confirmed live, and it is exactly why this read-back is unsafe as written.**

> `skills/vss-manage-video-io-storage/references/nvstreamer-api-reference.md:46` — "**`sensorId`, however, may carry a `_N` uniqueifier suffix**  … **Always resolve `sensorId` from `GET /sensor/list` by matching `.name`** … (the `_N` suffix on `sensorId` was confirmed live 2026-06-02)"

**P1.** This PR answers "is the source still there?" two different ways:

```python
# utils.py:269 (delete_vst_sensor), and again at utils.py:392 (verify_vst_cleanup)
any(sensor.get("sensorId") == sensor_id for sensor in payload)
```
```bash
# skills/vss-search-archive/scripts/verify_source_cleanup.sh:58 — same endpoint, same PR
any(.[]; .sensorId == $id or .name == $name)
```

One key versus two, and the asymmetry is **fail-open at both sites that convert a failure into a success**: this block returns `True, "already absent"` after a non-200 DELETE, and `verify_vst_cleanup` (`utils.py:392`) flips `vst_success` False → True at `video_delete.py:362`.

So if VST has dropped the UUID but still lists the source under its canonical name — precisely the `_N` suffix case above — the Agent reports `success` for a source that is still registered.

**The fix is to use the predicate already written in this PR.**

skills/vss-ask-video/SKILL.md
```diff
@@ -85,7 +85,26 @@ For a confirmed search-result handoff, use only the caller-supplied `VIDEO_URL`
 and visual question. Do not consume similarity scores, filenames, object IDs,
 or other retrieval metadata as visual evidence, and do not rerun search,
 resolve a sensor, broaden the clip, or choose another interval. The caller owns
-verdict validation and any fallback after this skill returns.
+verdict validation and any fallback after this skill returns. When that question
+requests a structured JSON contract, return the response text exactly after
+removing hidden reasoning. Do not add a model name, endpoint, VLM/backend label,
+sensor or stream ID, UUID, URL, request/retry count, operational summary, or
+additional JSON fields.
```
zac-wang-nv: this contract should reside inside search skill. ask-video is generic video Q&A

skills/vss-search-archive/assets/search_result_template.md
zac-wang-nv, line 1: why do we need this using cli?

skills/vss-search-archive/evals/search.json
```diff
@@ -16,22 +16,26 @@
...
...bytes through the VSS Agent HTTP API's three mutations: request an upload URL with `POST /api/v1/videos`, upload the bytes to that returned URL, then finalize with `POST /api/v1/videos/<sensor-id>/complete`. This Bash/curl HTTP sequence is the documented Agent-backed workflow; no dedicated Workflow or Agent tool call is expected. Consistently upload the ladder bytes as `warehouse-ladder.mp4`, and validate each separate `/complete` response. Because indexes are lazy, boundedly refresh the project-local `vss configure` under the same persisted budget until all three distinct indexes are recorded, then poll document readiness. Poll until VST lists both canonical sources with resolved UUIDs, embedding documents exist under `sensor.id.keyword=<UUID>` for both, behavior documents exist under `sensor.id.keyword=warehouse-ladder`, and raw documents exist under `sensorId.keyword=warehouse-ladder`. Require the distinct configured indexes `mdx-embed-filtered-2025-01-01`, `mdx-behavior-2025-01-01`, and `mdx-raw-2025-01-01`. Logs are diagnostics only. If the bounded ingestion deadline expires, print diagnostics and fail without redeploying, restarting, resetting the deadline, or re-ingesting.",
```
zac-wang-nv: this is much longer than the ingestion prompt from the doc?

---

Yes, and it is the file-wide pattern. The step-8 query has grown `349 → 614 → 883 → 973 → 1116` chars while its `checks` array stayed **byte-identical** throughout (diffed at each head). Prompts grow every time a check fails; the graded criteria never move.

**P1 on this step specifically:** step-2 has failed **4 of the 11 sweeps that reached it** on a byte-identical grader. Because the steps are chained, each step-2 failure costs the 7 steps behind it — that instability, not the prompt length, is what keeps the eval from ever finishing.

skills/vss-search-archive/evals/search.json
```diff
@@ -89,21 +93,21 @@
...
       ]
     },
     {
       "scenario": "kubernetes-ingress-contract",
-      "query": "Explain the exact host-side commands you would use to list registered sources and search the Kubernetes search profile exposed at `VSS_PUBLIC_URL=https://vss-search.example.com`. Search only the resolved `warehouse_sample` source for forklifts with the embed path and top-k 3. This is a contract check: show the commands but do not execute them or deploy anything.",
+      "query": "Explain the exact host-side commands you would use to list registered sources and search the Kubernetes search profile exposed at `VSS_PUBLIC_URL=https://vss-search.example.com`. Set `VSS_REPO_ROOT=\"${VSS_REPO_ROOT:-$HOME/video-search-and-summarization}\"`, validate its agent project, define `VSS=(uv run --project \"${VSS_REPO_ROOT}/services/agent\" --no-dev --extra cli vss)`, run `\"${VSS[@]}\" configure --base-url https://vss-search.example.com` once, then use that project-local CLI for search. Search only the resolved `warehouse_sample` source for forklifts with the embed path and top-k 3. This is a contract check: show the commands but do not execute them or deploy anything.",
```
zac-wang-nv: why do we need this query?

---

Worth knowing before deciding: **step-9 has never executed — not once in 16 sweeps.**

It is gated behind step-8, which has never passed (`0.8, 0.8, 0.6, 0.8` across the 4 sweeps that reached it). So this query has never been graded, and is currently costing maintenance for zero signal.

skills/vss-search-archive/scripts/source_setup_budget.sh
zac-wang-nv, line 1: we shouldn't need specific code for timeout machinenery.

deploy/docker/services/rtvi/rtvi-embed/rtvi-embed-docker-compose.yml
```diff
@@ -65,8 +65,15 @@ services:
       ERROR_MESSAGE_TOPIC: "${RTVI_EMBED_ERROR_MESSAGE_TOPIC:-vision-embed-errors}"
       KAFKA_BOOTSTRAP_SERVERS: ${RTVI_EMBED_KAFKA_BOOTSTRAP_SERVERS:-kafka:29092}
       # Generated-output and error buses (empty MESSAGE_BUS/ERROR_BUS disables that bus).
-      MESSAGE_BUS: "${MESSAGE_BUS:-}"
-      MESSAGE_BUS_TOPIC: "${MESSAGE_BUS_TOPIC:-mdx-embed}"
+      # Full developer profiles do not load this service directory's .env.
```
zac-wang-nv: please raise a PR to fix those env files by adding the env_file to their corresponding compose file.  please don't add duplicate envs as workarounds

---

**Agreed, and the evidence backs this ask exactly.** The service-local `.env` genuinely is never loaded: this compose declares no `env_file:` and no `extends:`, and Compose only auto-loads `.env` from the project directory. Two other in-tree sources say the same independently — `skills/vss-build-vision-agent/references/services/rt-embed.md:33` and the IN-3 eval spec.

**P1 — the workaround has two costs:**

1. `KAFKA_ENABLED: "${RTVI_EMBED_KAFKA_ENABLED:-}"` defaults to **empty**, and this file's own comment says an empty value *disables* that bus. Every profile that does not set the namespaced variable gets the bus silently off.
2. The `${MESSAGE_BUS:-}` middle rung of `"${RTVI_EMBED_MESSAGE_BUS:-${MESSAGE_BUS:-}}"` is **dead**: the only in-tree file setting bare `MESSAGE_BUS` is `services/rtvi/rtvi-embed/.env` — the one that never loads.

Also worth flagging: this PR adds three rows to `skills/vss-deploy-profile/references/search.md:135-137` documenting these as defaulting to `kafka` / `true` / `mdx-embed`. Those match `dev-profile-search/.env`, but **not** the compose fallbacks — so a reader porting the knobs to another profile gets a silently disabled bus.

Adding `env_file:` fixes the code and the doc at once.

services/agent/packages/vss_agents/src/vss_agents/api/video_delete.py
```diff
@@ -127,7 +134,18 @@ async def _remove_from_rtvi_cv(
         if response.status_code in (200, 201, 204):
             logger.info("RTVI-CV stream removed: %s", scrub_log(sensor_id))
             return True, "OK"
-        return False, f"RTVI-CV returned {response.status_code}: {response.text}"
+        response_text = response.text
```
zac-wang-nv, line 141: **P1** — the `already_absent` machinery here only ever selects a log level.

`:138-140` distinguishes "already absent" from a real RTVI-CV failure using the `(400, 404, 409, 410, 500)` enumeration plus `_RTVI_STREAM_ABSENT_RE`. But at `:332` the result is consumed solely by the branch at `:333-340` — `logger.info` versus `logger.warning` — and never reaches `results.append`. The next append is `:364`, for VST.

So an RTVI-CV failure and an RTVI-CV success produce an **identical response**. Either feed the result into `results`, or drop the regex and the status enumeration; as written they are elaborate machinery selecting a log severity.

services/agent/packages/vss_agents/tests/unit_test/api/test_video_delete.py
```diff
@@ -89,3 +166,137 @@ async def test_network_error_reports_failure(self):
...
+        self,
+        mock_get_sensor_name,
+        mock_delete_es,
+        mock_remove_rtvi,
+        mock_delete_storage,
+        mock_delete_sensor,
+    ):
+        mock_get_sensor_name.return_value = "warehouse-ladder"
+        mock_delete_es.return_value = (True, "Deleted")
+        mock_remove_rtvi.return_value = (False, "RTVI-CV returned 500")
```
zac-wang-nv, line 310: **P1** — this test passes because the information is unrepresentable, not because the behaviour is right.

`test_optional_rtvi_failure_does_not_downgrade_durable_cleanup` sets `mock_remove_rtvi.return_value = (False, "RTVI-CV returned 500")` and asserts the overall status stays `success`.

`DeleteVideoResponse` (`video_delete.py:59-64`) has only `status` / `message` / `video_id`. There is no field that can carry "durable cleanup succeeded, best-effort RTVI-CV did not." So the assertion cannot fail for the right reason — it locks in the gap flagged at `video_delete.py:137` rather than testing a decision.

skills/vss-search-archive/evals/search.json
```diff
@@ -89,21 +93,21 @@
...
+      "query": "Delete the video `warehouse-ladder` from the prepared Docker search deployment. Resolve and save its UUID and canonical name first, reuse the index names from the project-local `vss configure show`, and use the configured origin itself as `AGENT_URL` without adding `/api` to that variable. Issue the canonical agent-backed `DELETE /api/v1/videos/<video_id>` exactly once: do not probe alternate route spellings and do not retry any returned status. After that one response, always invoke the bundled `skills/vss-search-archive/scripts/verify_source_cleanup.sh` exactly once with the configured deployment origin itself as its VST argument (do not append `/vst`), the Elasticsearch URL, three index names, saved UUID, canonical name, and remaining deletion timeout, even if the response status is not `success`; overall success still requires response status `success`. Use only that invocation's structured output to report VST absence plus the exact embedding, behavior, and raw index tuples at zero. If it fails, report the error and stop: do not rerun it and do not issue substitute VST or Elasticsearch queries.",
```
zac-wang-nv: **P1** — the step-8 query is now **1116 chars, 3.2x the original 349**, and the new text restates check 4's rubric almost verbatim:

| check 4 (grader) | query (prompt) |
|---|---|
| "invoked the bundled `verify_source_cleanup.sh` **once**" | "invoke … **exactly once**" |
| "**used its structured output**" | "**Use only that invocation's structured output**" |
| "did not **reconstruct substitute Elasticsearch queries**" | "do not issue **substitute VST or Elasticsearch queries**" |

An eval whose prompt contains its own rubric measures instruction-following, not whether the skill works. The `checks` array has been byte-identical across every one of those rewrites.

**Separately (P2):** this commit ships the real `%/vst` mechanism fix *and* this query rewrite together, so sweep #17 cannot attribute a green step-8 to either half. And the old query **already said "once"** — the agent invoked twice because the first call failed silently (see `verify_source_cleanup.sh:62`), not because the wording was unclear.

Suggest reverting this hunk and letting sweep #17 measure the `%/vst` strip alone. Keep the new guidance in `references/source_lifecycle.md`, which is what a real user's agent reads.

skills/vss-ask-video/SKILL.md
```diff
@@ -573,10 +614,15 @@ if [ "${VLM_BACKEND}" = "nim_cosmos" ] && { [ "$UPLOAD_FORMAT" = "video_url" ] |
...
 
-# Send THIS body for both formats. Do not write a separate minimal video_url curl — hand-built
+# Send THIS body once for both formats. Do not write a separate minimal video_url curl — hand-built
 # video_url requests keep dropping ${MM_KWARGS}, which under-samples the clip on NIM Cosmos.
-curl -s --connect-timeout 5 --max-time 120 -X POST "${VLM_ENDPOINT}/chat/completions" \
+AUTH_HEADER=()
+if [ -n "${VLM_REMOTE_URL:-}" ] && [[ "${VLM_ENDPOINT%/}" == "${VLM_REMOTE_URL%/}"* ]]; then
+  AUTH_HEADER=(-H "Authorization: Bearer ${NVIDIA_API_KEY:?NVIDIA_API_KEY is required for VLM_REMOTE_URL}")
+fi
+curl -fsS --connect-timeout 5 --max-time 120 -X POST "${VLM_ENDPOINT}/chat/completions" \
```
zac-wang-nv: **P1** — `curl -fsS` on the unconditional path turns a server error into an empty answer.

`-f` makes curl exit non-zero and emit **no body** on an HTTP error, so a 500 from the VLM becomes an empty string piped into `jq -r '.choices[0].message.content'`, and the step reports an empty answer at exit 0.

A server error is then indistinguishable from "the model returned nothing" — which is the failure mode hardest to notice in an eval trace.

skills/vss-search-archive/scripts/verify_source_cleanup.sh
```diff
@@ -0,0 +1,97 @@
...
+while (( $(date +%s) < deadline )); do
+  timeout=$(request_timeout) || break
+  sensors=$(curl -fsS --connect-timeout 5 --max-time "${timeout}" \
+    "${VST_URL}/vst/api/v1/sensor/list") || exit 1
+  vst_present=$(printf '%s' "${sensors}" | jq -r \
+    --arg id "${SENSOR_ID}" --arg name "${SOURCE_NAME}" \
+    'any(.[]; .sensorId == $id or .name == $name)') || exit 1
+  case ${vst_present} in true|false) ;; *) exit 1 ;; esac
+
+  embed_count=$(index_count "${EMBED_INDEX}" sensor.id.keyword "${SENSOR_ID}") || exit 1
```
zac-wang-nv: **P1** — this is the root cause of sweep #16's one failing check.

I ran the new guard scenario from this PR against the **pre-fix** script: it exits `1` with **zero bytes on both stdout and stderr**.

```
caller passes '<origin>/vst/'
  OLD  9d2b3bacf   exit=1   stdout+stderr: completely empty
  NEW  3a6dc3ff8   exit=0   {"vst_present":false,...}
```

That silence is why the agent retried and then reconstructed its own Elasticsearch queries — exactly what check 4 penalises. The `%/vst` fix is real and verified, but it fixes *one cause* of *one* silent exit.

The script emits its structured JSON on two paths only: success (`:68`) and deadline expiry (`:85`). Six other failures are bare `exit 1` with no output — `:57`, `:58`, `:61`, `:63`, `:64`, `:65`. Any of them reproduces the identical agent behaviour.

**And the new instruction makes that worse:** both the doc and the query now say *"if it exits nonzero, report its error and stop"* — but on those six paths there is no error to report. Today a silent failure costs one check because the agent improvises to the right answer; under a hard-stop rule it stops with nothing.

The emitter already exists at `:85`. One helper closes all six:

```bash
fail() { jq -cn --arg error "$1" '{error:$error}' >&2; exit 1; }
sensors=$(curl -fsS … "${VST_URL}/vst/api/v1/sensor/list") || fail "VST sensor list unreachable at ${VST_URL}"
```

zac-wang-nv (commented):
can you put the ingestion and cleanup script into two self-contained scripts?

currently they are hybrid of inline script from markdown and scripts(.sh).

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32541776647)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32692316682)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32713190837)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32717858962)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32747329557)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32794935191)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32920269560)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32943119041)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32992892180)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33050743428)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33086387405)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33105746194)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33137361596)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33226749914)

deploy/docker/services/infra/haproxy/haproxy.cfg.template
```diff
@@ -253,6 +253,11 @@ frontend fe_http
     acl p_api path_beg /api/
     use_backend bk_vss_agent if h_main p_api
 
+    # The published VSS origin is also used for Agent readiness checks.
```
zac-wang-nv, line 256: why do we need this health check? agent's health shouldn't be called and considered as the check from the gateway?

skills/vss-search-archive/references/result_verification.md
```diff
@@ -91,13 +91,28 @@ Require `result` to be `confirmed`, `rejected`, or `unverified`, every
 JSON from hidden reasoning or surrounding prose. A valid semantic `unverified`
 is a completed visual check and must not trigger fallback.
 
+Make one request per hit. An SSRF/fetch refusal, timeout, 401/403, or any other
+non-200 transport error is a technical failure: stop, discard the result, and
+report it — never retry with a different upload format (for example inlining
```
zac-wang-nv, line 96: which skill needs this ref?

skills/vss-search-archive/references/source_lifecycle.md
```diff
@@ -175,6 +175,16 @@ request. A literal `--max-time`, a new epoch-plus-duration expression, or a
 phase-local deadline after this initialization violates the source-setup
 contract.
 
+Reassigning `SEARCH_READINESS_DEADLINE` after it has been initialized — even
+under a new variable name, in a fresh shell, or because the first ingestion
+produced no embeddings — is a deadline reset and violates the contract. If the
+deadline expires, or if an ingestion produces no embeddings before it expires,
```
zac-wang-nv, line 181: which command needs this doc?

PR #1658, opened 2026-08-12
fix: cannot pull exactly lvs image in nemoclaw

trannguyenanhhien:
**Summary**

- NemoClaw LVS deploy failed at image pull because the orchestrator only used profile .env / overrides.env, so LVS resolved ghcr.io/.../vss-video-summarization:3.3.0-rc2 (not found) via lvs.env + LVS_TAG.
- Other profiles were fine: they don’t pull that image, and their compose image: lines already fall back to VSS_CONTAINER_* / develop-latest.
- Align with dev-profile.sh by requiring and loading deploy/docker/containers.env first (with ${VAR:-default} expansion), so LVS gets CONTAINER_IMAGE from VSS_CONTAINER_REGISTRY / VSS_CONTAINER_TAG (default develop-latest, overridable by CI).

**Test plan**

- [ ]  Unit tests for docker_compose_util (containers.env layering / shell defaults)
- [ ]  NemoClaw LVS deploy: image resolves to …/vss-video-summarization:develop-latest (or exported VSS_CONTAINER_*), not 3.3.0-rc2


zac-wang-nv (dismissed):
LGTM

zac-wang-nv (approved):
LGTM

PR #1669, opened 2026-08-13
feat(skill-eval): add eval corpus inventory

simings-nv:
## Description

The size of the skill-eval corpus is quoted from memory in several places and the numbers disagree. This adds one tool that measures it, so a coverage claim can be checked rather than recalled.

On `develop` it reports:

```
specs=50 cells=152 checks=841
lexical scope hint (mentions, NOT an execution-scope measurement): HOST=159 FILE=6 NET=270 PROSE=406 / cells mentioning HOST: 63
```

**Measurement only.** Nothing is wired into CI and no existing behavior changes.

### Design notes

**Spec discovery is imported from `plan_matrix`, not reimplemented**, so the inventory and the dispatcher cannot disagree about what a spec is.

**Skill directories are enumerated from the filesystem, not from `SKILL.md`.** Manual `*` dispatch enumerates directories directly, so a spec in a directory without a `SKILL.md` is dispatchable and must not be missing from the inventory. The test for this uses a fixture directory rather than the live corpus: every eval-bearing directory in the repo today happens to have a `SKILL.md`, so asserting against the real tree cannot tell the two implementations apart.

**The scope hint is deliberately labelled as not a measurement.** It matches on mentions rather than requirements — `"no docker compose is executed"` counts as HOST, and a live container-state assertion containing no command counts as PROSE. It is useful for finding checks worth reading. Deciding where a check can actually execute needs per-check scope metadata, which this does not attempt, and the output says so rather than leaving the caveat in a docstring.

**Malformed, unreadable and non-UTF-8 specs fail loudly** rather than degrading to zero, so a damaged corpus cannot silently undercount.

### Known limitation

Counts alone cannot detect a check *moving* between cells, and reward is computed per cell as `passed / len(checks)`. A move therefore changes scores while every total here stays identical. Pinning ordered check identities is a follow-up; until that lands this is scaffolding and should not be consumed as a parity baseline.

## Checklist

### DIFF
--- a/.github/skill-eval/eval_parity_scope.py
+++ b/.github/skill-eval/eval_parity_scope.py
@@ -1,0 +1,210 @@
+#!/usr/bin/env python3
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


zac-wang-nv (approved):
LGTM

PR #1675, opened 2026-08-13
feat(vios): share one APT package cache across standalone and e2e deployments

gous-75:
## Description

Adds a shared APT package cache so the runtime media packages are downloaded **once per host** instead of on every container start — wired into **both** the standalone stream-processing deployment and the `deploy/docker` e2e compose profiles.

### Why

VIOS deliberately ships without the GPL/patent-encumbered media packages (OSRB), so `user_additional_install.sh` re-downloads them at every start. On `vst-base:...-runtime` that is **224 packages / ~138 MB**, measured between **72s and 258s** depending only on which archive mirror answered. Every profile brings up a stream-processing **and** an nvstreamer container that each paid this cost independently.

| | |
|---|---|
| init container, cold | ~186s, 224 packages, ~139 MB cached |
| consumer, warm cache | **~62s, `fetched=0`** — no network |
| two consumers concurrently | both `rc=0`, **zero lock errors** |
| two init containers racing | both succeed, flock serialises |

Correctness after a cache-only install: all four of the script's own sentinels resolve and `avdec_h264` loads.

### Three things this needed, all found by running containers

**1. `docker-clean` would have silently emptied the cache.** The base image ships `/etc/apt/apt.conf.d/docker-clean`, which deletes every `.deb` after each apt operation. Mounting a volume without removing it leaves **28 KB instead of 138 MB**. `Binary::apt::APT::Keep-Downloaded-Packages "true"` does **not** fix it — that is an apt option, `docker-clean` is a `DPkg::Post-Invoke` hook. The init container removes the file; it is the only writer.

**2. A shared read-write cache collides even when warm.** apt takes the archives lock unconditionally, and the volume makes the lock file shared:

```
container A: rc=100  1s  E: Could not get lock /var/cache/apt/archives/lock.
                         It is held by process 0
container B: rc=0   130s
```

`DPkg::Lock::Timeout` does not help — it governs the dpkg frontend lock. Consumers therefore mount the cache **read-only** and pass `Dir::Cache::archives` + `Debug::NoLocking`. That is safe **only** because consumers never write there and the init has completed; the script says so at the point it matters.

**3. The volume name has to be pinned.** nvstreamer and stream-processing are separate compose projects (standalone) / co-tenant services (e2e), so Compose would otherwise prefix the volume per project and each would silently get its own. `name: vios_apt_cache` pins it — verified: `vios_apt_cache` resolves identically in both while `vst_pg_data` still resolves to `<project>_vst_pg_data`.

## Standalone deployment (`services/vios/deployment/stream-processing`)

- An `vios-apt-cache-init` populates the shared volume once; the stream-processing and nvstreamer consumers install from it read-only.
- Because `depends_on` cannot span the two standalone compose projects, both run an init; `flock` in the populate path serialises them and the second finds the cache warm.
- `oneclick_dc_deployment.py` was hardened so lifecycle matches the cache's intent:
  - `stop` / `stop --clean` **preserve** the cache (all `-v` removed from stop paths; `--clean` removes only the `vst_pg_data` volume by name).
  - new **`--delete-apt-cache`** flag explicitly removes the cache for a cold fetch; combine with `--clean` to wipe everything.

## e2e deployment (`deploy/docker`)

Brought the same optimisation to the compose profiles, scoped minimally to the `include:`-merged project:

- **`services/vios/streamprocessing/docker-compose.yaml`**: adds a best-effort `vios-apt-cache-init` populator (carries all 10 vios profiles so it activates iff a vios service does), the pinned non-external `vios_apt_cache` volume, and the ro mount + `VST_APT_CACHE_DIR` + `depends_on` on all four stream-processing consumers.
- **`services/nvstreamer/base.yml`**: wires the same ro mount + env + dependency **once** into the centralized base; all six nvstreamer variants inherit it via `extends`, and their own volumes/`depends_on` still merge.

The init is **best-effort** (exits 0 even if the populate fails), so a transient apt error never blocks the consumers that depend on it — they fall back to downloading. A single populator (stream-processing image) covers the nvstreamer consumers too, because both images resolve to the same package set.

> The earlier blocker (adding `depends_on` to `nvstreamer-base` broke six profiles with *"depends on undefined service"*) is resolved: the init lives in the stream-processing compose, which shares the project via `include`, and develop's NVStreamer-centralization refactor now defines every variant from one shared base.

**Dry-run validated** (no deployment): rendering the full merged project with `docker compose config` for both a developer profile (`dev-profile-search`) and an industry profile (`warehouse-2d`) succeeds with **zero errors**, and confirms — on the real `include`/`extends`/profile machinery — that the init activates, the consumers get the ro mount + env + `depends_on: vios-apt-cache-init`, each variant's own dependencies still merge, and the shared volume is declared.

## Lifecycle

A named volume, deliberately **not** under `VST_VOLUME` because dev/QA clear that on teardown.

| action | cache |
|---|---|
| `rm -rf $VST_VOLUME` | survives |
| `stop` / `start` / `down` (no `-v`) | survives |
| `stop --clean` (standalone) | **survives** (only VST data removed) |
| `stop --delete-apt-cache` (standalone) | removed |
| `docker compose down -v` | removed |
| `docker volume rm vios_apt_cache` | explicit clear |

The cache only grows as versions churn, so it wants a periodic reset. Adding a package needs no invalidation — the marker is keyed on `md5sum` of `RUNTIME_PACKAGES` and apt fetches only what is missing (verified: adding 2 packages to a warm 224-package cache fetched 1).

## OSRB

No new exposure: the `.deb`s are fetched by the customer's apt onto the customer's host, and the shipped image still contains none of them. **A populated cache must never end up in an image layer** — that becomes distribution. Runtime volume only.

## Known limitation

The cache removes the **download**, not the install: the dpkg phase (~26s) still runs per service.

## Checklist
- [ ] I have installed and run [pre-commit hooks](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/HEAD/CONTRIBUTING.md#1-enable-pre-commit-hooks) locally.

### DIFF
--- a/deploy/docker/services/nvstreamer/base.yml
+++ b/deploy/docker/services/nvstreamer/base.yml
@@ -33,6 +33,11 @@ services:
     environment:
       - VST_APT_MIRROR=${VST_APT_MIRROR:-}
       - NVSTREAMER_INSTALL_ADDITIONAL_PACKAGES=${NVSTREAMER_INSTALL_ADDITIONAL_PACKAGES}
+      # Install from the shared APT cache (populated by vios-apt-cache-init in
+      # the streamprocessing compose) instead of downloading. The ro mount and
+      # the dependency below are inherited by every nvstreamer variant via
+      # `extends`; the volume and init service share this project via `include`.
+      - VST_APT_CACHE_DIR=/opt/apt-cache
       - ADAPTOR=streamer
       - HTTP_PORT=${NVSTREAMER_HTTP_PORT}
       - NVSTREAMER_UI_BASE_PATH=${NVSTREAMER_UI_BASE_PATH:-}
@@ -46,6 +51,10 @@ services:
       - ${VSS_APPS_DIR}/services/nvstreamer/configs/notification-config.json:/home/vst/vst_release/configs/notification_config.json
       - ${VSS_APPS_DIR}/services/nvstreamer/configs/vst-storage.json:/home/vst/vst_release/configs/vst_storage.json
       - ${VSS_APPS_DIR}/services/vios/scripts/user_additional_install.sh:/home/vst/vst_release/tools/user_additional_install.sh
+      - vios_apt_cache:/opt/apt-cache:ro
+    depends_on:
+      vios-apt-cache-init:
+        condition: service_completed_successfully
     deploy:
       restart_policy:
         condition: on-failure
--- a/deploy/docker/services/vios/scripts/user_additional_install.sh
+++ b/deploy/docker/services/vios/scripts/user_additional_install.sh
@@ -182,6 +182,36 @@ if [[ -n "${VST_APT_PIPELINE_DEPTH:-}" ]]; then
   APT_OPTS+=(-o "Acquire::http::Pipeline-Depth=${VST_APT_PIPELINE_DEPTH}")
 fi
 
+# ---------------------------------------------------------------------------
+# Shared package cache
+#
+# Every VIOS service on a host installs the same package set, and each cold
+# start otherwise re-downloads ~138MB. An init container populates a shared
+# volume once (VST_APT_CACHE_POPULATE=true); services then install from it with
+# no network.
+#
+# Consumers mount the cache READ-ONLY at VST_APT_CACHE_DIR. Two options make
+# that work and both matter:
+#
+#   Dir::Cache::archives  points apt at the shared copy instead of its own.
+#   Debug::NoLocking      skips the archives lock. apt takes that lock even with
+#                         nothing to download, and a shared volume means a shared
+#                         lock file -- measured: two containers starting together,
+#                         the loser fails outright with
+#                         "Could not get lock ... held by process 0".
+#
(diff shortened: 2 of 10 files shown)


kaushikc-nvidia (dismissed):

kaushikc-nvidia:
/ok to test [c2416ed](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1675/commits/c2416ed4dfb26e551f74e8fc0891e40ab45146d5)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32237443256)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32244946120)

kaushikc-nvidia (approved):

PR #1678, opened 2026-08-13
feat(lvs): expose GPU hardware metrics with DCGM

hugoverjus:
## Summary
- start the existing NVIDIA DCGM Exporter with the LVS developer profile
- expose Prometheus-format GPU utilization, decoder, memory, power, and temperature metrics on configurable host port 9400
- document direct Prometheus/Dynatrace scraping and cover the profile wiring in regression tests

## Test plan
- [x] `git diff --check`
- [x] Resolve the LVS Compose configuration and verify `dcgm-exporter` is active
- [x] `bash deploy/docker/test-scripts/test-dev-profile.sh` (203 passed; 2 pre-existing unrelated MV3DT checks failed)
- [ ] On a supported GPU host, deploy LVS and scrape `http://localhost:9400/metrics`

### DIFF
--- a/deploy/docker/README.md
+++ b/deploy/docker/README.md
@@ -92,6 +92,26 @@ export NGC_CLI_API_KEY="<your-key>"
 ./deploy/docker/scripts/dev-profile.sh --help
 ```
 
+### LVS GPU hardware metrics
+
+The LVS developer profile starts NVIDIA DCGM Exporter alongside the application.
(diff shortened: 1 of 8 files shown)


zac-wang-nv:
Can we make it optional? basically we can add it to build vision agent skill and deploy the dcgm through prompt?

simings-nv (approved):

PR #1679, opened 2026-08-13
refactor(docker): add reusable NVStreamer service base

gbau1:
## Summary

- add a dedicated `deploy/docker/services/nvstreamer` directory containing a reusable, blueprint-agnostic `nvstreamer-base`, shared environment defaults, and a shared-config placeholder directory
- keep the six stable concrete service/profile names as thin wrappers in their owning developer or industry blueprints
- select blueprint-owned NVStreamer configuration through environment paths, following the Alert Bridge service/config ownership pattern
- preserve all existing container names, ports, arguments, mounts, GPU reservations, runtimes, dependencies, and Smart Cities networking
- move NVStreamer-owned defaults out of VIOS and profile environment files

## Validation

- canonical resolved Compose JSON for `nvstreamer-alerts`, `nvstreamer-lvs`, `nvstreamer-2d-fusion`, `nvstreamer-2d`, `nvstreamer-3d`, and `nvstreamer-mv3dt` is exactly unchanged
- Smart Cities `nvstreamer-alerts` is also exactly unchanged, including its video mount and bridge networking
- the shared NVStreamer Compose contains no `developer-profiles` or `industry-profiles` references
- each stable concrete NVStreamer service is defined exactly once in its owning blueprint
- `python3 .github/scripts/compose_image_golden.py` passes
- `git diff --check` passes
- `bash deploy/docker/test-scripts/test-dev-profile.sh`: 203 passed, 2 pre-existing MV3DT skill/Helm assertions failed; both assertions also fail on the current `origin/develop` base


gbau1:
ok to test

deploy/docker/industry-profiles/smartcities/compose.yml
```diff
@@ -26,29 +26,6 @@ services:
     container_name: vss-behavior-analytics
     command: python3 apps/smart_city/main_smart_city_app.py --config /resources/vss-behavior-analytics-config.json --calibration /resources/calibration.json
 
-  nvstreamer-alerts:
```
zac-wang-nv, line 29: **P0 — blocking**

This copy is not a duplicate — it is the only one the Smart Cities deploy ever sees, and removing it makes the Smart Cities compose project invalid.

`deploy/docker/scripts/deploy_smartcities_launchable.ipynb` (cell 17, step 5) deploys Smart Cities as an *overlay on the alerts profile*:

```python
# 5) rsync smartcities/ contents into dev-profile-alerts/ so the smartcities artifacts
#    (compose.yml, configs, calibration data, dashboards, etc.) live alongside the alerts profile
subprocess.run(["rsync", "-av", f"{SMC_DIR}/", f"{ALERTS_DIR}/"], ...)
```

`rsync` copies `industry-profiles/smartcities/compose.yml` **over** `developer-profiles/dev-profile-alerts/compose.yml`. So after the overlay, the surviving `nvstreamer-alerts` definition this PR keeps in `dev-profile-alerts/compose.yml` no longer exists, and nothing else defines it. `services/alert/compose.yml:53` has `alert-bridge` → `depends_on: nvstreamer-alerts`, so the whole project fails to load:

```
$ docker compose --env-file containers.env --env-file .../dev-profile-alerts/.env \
    --env-file .../dev-profile-alerts/overrides.env config
service "alert-bridge" depends on undefined service "nvstreamer-alerts": invalid compose project
```

Control experiment — same simulation script (set `NVSTREAMER_VIDEO_DIR` absolute → merge `smartcities/.env` + `overrides.env` into the alerts layers → `rsync smartcities/ → dev-profile-alerts/`), same `COMPOSE_PROFILES_CV`, docker 28.3.3 / compose v2.39.1:

| tree | `nvstreamer-alerts` defs after rsync | `docker compose config` |
|---|---|---|
| merge base `3cbce781` | 1 | resolves (1883 lines) |
| this head `3d2cde8f` | 0 | **`invalid compose project`** |

`nvstreamer-alerts` is also in `smartcities/overrides.env` `COMPOSE_PROFILES_CV`, so even if the hard failure were removed, Smart Cities would come up with no NVStreamer and no video source.

Worth noting: the previous head `3d95923c` did **not** have this problem — the concrete services lived in a globally-included `services/nvstreamer/compose.yml` that the rsync never touches. The move back into the blueprint file is what reintroduced it.

Also flagging a tension with the fix: the new guard at `test-dev-profile.sh:1399` asserts each service has *exactly one* blueprint-owned definition, so simply restoring the Smart Cities copy would fail that assertion. The options I can see, cheapest first — (a) put the concrete `nvstreamer-alerts` somewhere the rsync cannot shadow and relax the guard's expected count, (b) exclude `compose.yml` from the rsync in the notebook and have `smartcities/compose.yml` be an include rather than a replacement, or (c) keep both definitions and teach the guard about the overlay. Your call on which fits the direction here.

deploy/docker/industry-profiles/warehouse-operations/compose.yml
```diff
@@ -16,14 +16,17 @@
 include:
   - path: ./warehouse-2d-app/warehouse-2d-app.yml
     env_file:
+      - ${VSS_APPS_DIR}/services/nvstreamer/.env
```
zac-wang-nv, line 19: **P1 — request change**

Adding the shared env file to the *include* fixes interpolation, but it does not restore the two keys that `warehouse-operations/.env` used to inject into the `bp-configurator-*` **containers**.

Those services mount the profile env file directly:

```yaml
    env_file:
      - ${BP_CONFIGURATOR_BASE_ENV_FILE:-$VSS_APPS_DIR/industry-profiles/warehouse-operations/.env}
```

so every key in that file lands in the container environment. Removing `NVSTREAMER_HTTP_PORT` and `NVSTREAMER_INSTALL_ADDITIONAL_PACKAGES` from it silently drops them. `docker compose config` over the full project, merge base vs this head, real `COMPOSE_PROFILES_WH_*` lists — this is the **only** delta in all three warehouse legs:

```diff
   bp-configurator-2d:                      # identical for -3d and -mv3dt
     environment:
+      NVSTREAMER_2D_CONFIG_DIR: .../warehouse-2d-app/nvstreamer/configs
+      NVSTREAMER_3D_CONFIG_DIR: .../warehouse-3d-app/nvstreamer/configs
       NVSTREAMER_HTTP_HOST_PORT: '31000'
-      NVSTREAMER_HTTP_PORT: '31000'
-      NVSTREAMER_INSTALL_ADDITIONAL_PACKAGES: 'true'
       NVSTREAMER_IP: vss-vios-nvstreamer
+      NVSTREAMER_MV3DT_CONFIG_DIR: .../warehouse-mv3dt-app/nvstreamer/configs
```

`NVSTREAMER_INSTALL_ADDITIONAL_PACKAGES` was clearly an incidental passenger and losing it is fine. `NVSTREAMER_HTTP_PORT` is the one to check: `NVSTREAMER_STREAMS_ENDPOINT` / `NVSTREAMER_SENSOR_STATUS_ENDPOINT` still resolve to `:31000`, so if `vss-configurator` only reads those there is no impact — but the image is out of tree, so that is not verifiable from this repo. Adding `${VSS_APPS_DIR}/services/nvstreamer/.env` to the `bp-configurator-*` `env_file:` list in the three `warehouse-*-app.yml` files restores the old container env exactly; alternatively confirm with the configurator owners that the raw port var is unused and record that in the PR description.

(The three `NVSTREAMER_*_CONFIG_DIR` keys arriving in the same container env are the same `env_file` leak in the other direction — harmless, just noise.)

gbau1:
ok to test

gous-75 (dismissed):

deploy/docker/industry-profiles/smartcities/compose.yml
gous-75: @zac-wang-nv  
It was fixed by moving the concrete nvstreamer-alerts out of the blueprint compose into the shared services/nvstreamer/compose.yml. It's pulled in via the deploy root (compose.yml → services/compose.yml → ./nvstreamer/compose.yml), which the launchable's rsync smartcities/ → dev-profile-alerts/ never touches — it only overlays dev-profile-alerts/compose.yml. So nvstreamer-alerts survives the overlay and alert-bridge's depends_on is satisfied. It's a single definition now, so the test-dev-profile.sh one-owner guard is happy too; smartcities/compose.yml

deploy/docker/industry-profiles/warehouse-operations/compose.yml
gous-75: @zac-wang-nv 
NVSTREAMER_HTTP_PORT is restored — rather than via the profile .env, it's now set directly on bp-configurator-base (environment: NVSTREAMER_HTTP_PORT: ${NVSTREAMER_HTTP_PORT}, interpolated from the
  services/nvstreamer/.env we added to the include), so all three bp-configurator-* inherit it. docker compose config on warehouse-2d resolves bp-configurator-2d with NVSTREAMER_HTTP_PORT=31000 (and
  NVSTREAMER_INSTALL_ADDITIONAL_PACKAGES=true), plus the *_ENDPOINTs at :31000. So the container env matches pre-refactor

kaushikc-nvidia (approved):

Issue #63, opened 2025-11-19
Illegal Memory Access / CUDA Error with Cosmos-Reason1-7B on L40S GPUs (Driver 580.95 / CUDA 13.0)


jgough-essextec:
**Description:**
I am attempting to deploy the `cosmos-reason1-7b` model using the NVIDIA VIA/NIM container (remote LLM Container). The application crashes consistently during the model warmup/inference phase with `torch.AcceleratorError: CUDA error: an illegal memory access was encountered`.

Following the crash, the NVIDIA MPS (Multi-Process Service) server enters a zombie state (`Error 807: MPS server is not ready`), rendering the GPUs inaccessible for subsequent retries until the host machine is rebooted.

This occurs on both a 4x GPU setup and when restricted to a single GPU.

**System Information:**

  * **Hardware:** 4x NVIDIA L40S (48GB VRAM each)
  * **CPU:** AMD 2x32 Core / 1.5TB RAM
  * **OS:** Ubuntu 24.04.2 LTS (Kernel `6.8.0-86-generic`)
  * **NVIDIA Driver:** `580.95.05`
  * **CUDA Version:** `13.0`
  * **Container/vLLM Version:** `v0.9.2+03bb1cdd.nv25.08.post1-stage-alexandrem-vss`

**Reproduction Steps:**

1.  Launch the NVIDIA VIA/NIM container for `cosmos-reason1-7b`.
2.  The engine initializes successfully.
3.  The engine performs `Capturing CUDA graph shapes`.
4.  During `Warmup` or the first inference request, the engine crashes.

**Observed Behavior & Logs:**
The process successfully loads weights and captures CUDA graphs. Immediately upon execution, the kernel panics.

**1. The Fatal Crash (Illegal Memory Access):**

```text
INFO 11-19 16:33:57 [gpu_model_runner.py:2326] Graph capturing finished in 42 secs, took 0.50 GiB
...
ERROR 11-19 16:34:17 [core.py:588] EngineCore encountered a fatal error.
...
File "/usr/local/lib/python3.12/dist-packages/vllm/v1/worker/gpu_model_runner.py", line 1499, in execute_model
    valid_sampled_token_ids = sampled_token_ids.tolist()
torch.AcceleratorError: CUDA error: an illegal memory access was encountered
```

**2. The Fallout (MPS Corruption):**
Immediately following the crash, the watchdog attempts to restart the worker, but the driver state is corrupted:

```text
[...]

PR #1693, opened 2026-08-14
feat/rtvi embed ghcr support

nv-pnikam:
## Description
  - Imports the RT Embed Docker build definition, Python dependency lockfiles, platform-specific package sources, and supporting native/GStreamer sources.
  - Adds required DeepStream/GStreamer binary artifacts for amd64, arm64, and iGPU builds.
  - Includes the NVIDIA Software License Agreement for the bundled artifacts.
  - Updates RT Embed documentation and Docker defaults for the source build.
  - Registers vss-rt-embed as a GHCR-built image in the container inventory, including its build context and required Git LFS binaries.
  - Adds source-to-image mapping and regression coverage for CI image selection and container-coordinate checks.
  - Keeps deployment, Helm, skill, and documentation consumers pinned to the current fixed NGC tag; develop-latest is CI-build support only.
  - Sets the Dockerfile’s ARM default to iGPU for the CI multi-platform build path.

## Checklist
- [ ✅] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/HEAD/CONTRIBUTING.md).
- [✅ ] I have installed and run [pre-commit hooks](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/HEAD/CONTRIBUTING.md#1-enable-pre-commit-hooks) locally (`uv run pre-commit install` once, then hooks run on every `git commit`).
- [ ✅] Every commit on this PR is [DCO sign-off](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/HEAD/CONTRIBUTING.md#developer-certificate-of-origin-dco)'d (`git commit -s` adds a `Signed-off-by` trailer that certifies you have the right to submit the change under Apache-2.0).
- [✅ ] New or existing tests cover these changes.
- [ ✅] The documentation is up to date with these changes.


nv-pnikam:
/ok to test [c2928e5](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1693/commits/c2928e561915276687ccffdb8a89b86c3d623eca)

nv-pnikam:
/ok to test [181c866](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1693/commits/181c86677aeb36ebf6bd3870ee6db5e2fd3f5a39)

nv-pnikam:
/ok to test [2df450a](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1693/commits/2df450a9c6ad5a165e33708faeb4996c437dfc20)

zac-wang-nv (commented):
## Round 6 review — 8 inline comments

Full context, CI state and the general findings are in the summary comment below.

**Verdict: comment, not request-changes.** The P0 that blocked rounds 4–6 is closed and the decisive build went green during this review. The one thing I would treat as merge-blocking is the OSRB gate, which is process rather than code.

.github/scripts/test_helm_release_channel_policy.py
```diff
@@ -121,7 +124,8 @@ def test_policy_covers_every_github_built_image(self):
             for image in inventory["images"]
             if image.get("ghcr_build") is True
         }
-        self.assertEqual(managed, set(HELM_VALUES))
+        self.assertEqual(managed - set(HELM_VALUES), {"vss-rt-embed"})
```
zac-wang-nv: **P1** — This PR exempts its own image from a repo-wide invariant, and the exemption has no exit condition.

At the merge base (`7b4fbe6da`) the assertion was:

```python
self.assertEqual(managed, set(HELM_VALUES))
```

— every image with `ghcr_build: true` must have a Helm values entry, and `test_helm_defaults_to_managed_ghcr_channel` requires that entry to resolve to `{GHCR_ROOT}/{name}`. This PR replaces it with a hard-coded carve-out naming `vss-rt-embed`:

```python
self.assertEqual(managed - set(HELM_VALUES), {"vss-rt-embed"})
self.assertTrue(set(HELM_VALUES).issubset(managed))
```

I want to be fair about *why*: when this was written the carve-out was accurate. `vss-rt-embed` is declared `ghcr_build: true` while every Helm path deliberately pins `nvcr.io/nvstaging/…`, so it could not satisfy the original assertion. Registering it in `HELM_VALUES` would fail the suite — the `HELM_HELPERS` registration you chose is the right one, and the suite passes 6/6 at this head. I ran it.

**That justification has now partly expired, during this review.** The `/ok to test 2df450a` sweep published `ghcr.io/nvidia-ai-blueprints/vss/vss-rt-embed:pr-1693-2df450a9c6ad@sha256:8726b2726e75954f456bb82ec16043ccab353d83bdd959599b020a7177536009`, and the `vss-ghcr-candidates` bot listed it in the downstream-validated candidate set at 13:29:55Z. The image exists. (It is not anonymously pullable — 403, where `vss-agent` returns 200 — but that is a package-visibility setting, not absence.)

The problem is shape, not correctness. `assertEqual(managed - set(HELM_VALUES), {"vss-rt-embed"})` is an **exact-set exception**, so it does two things nobody will expect: it silently becomes wrong the moment a second image adopts the helper pattern (the suite fails on an unrelated PR), and it makes a temporary state permanent with nothing pointing at when it should end.

Two asks, either is fine:
1. Add a comment on this line stating the exit condition — "remove once `ghcr.io/…/vss-rt-embed` is consumed by Helm; then register in `HELM_VALUES`" — and a tracking issue. Given the image now publishes, that condition is close.
2. Or express it as policy rather than enumeration: images carrying a `vss-<name>.image` helper are covered by `HELM_HELPERS` and excluded from the `HELM_VALUES` requirement by *rule*, so the guard needs no per-image edit.

Disclosure on my own consistency: I raised this at P1 in round 4, retracted it in round 5, and am raising it again. The retraction was correct on the evidence I had — at that head, adding the image to `HELM_VALUES` broke the suite. What I had not done was diff this file against the **merge base**; doing that shows the PR *removed* a stronger assertion to create the exemption, which is the part worth commenting on.

deploy/docker/container-inventory.json
```diff
@@ -110,9 +110,12 @@
     {
       "name": "vss-rt-embed",
       "strategy": "build",
-      "ghcr_build": false,
+      "ghcr_build": true,
```
zac-wang-nv, line 121: **P1** — Flipping `ghcr_build` to `true` is step 1 of `.github/github-first.md`. **Steps 2, 3 and 4 are not in this PR**, and the doc names that exact combination as its first pitfall:

> **`ghcr_build: true` without steps 2–3.** The image builds to GHCR on every merge and nothing consumes it — deployments keep pulling NGC.

Three files the doc requires are untouched here (all outside the diff, so I cannot anchor inline on them):

| Step | Required | State |
|---|---|---|
| 2 | namespaced `_IMAGE` + `_TAG` in `containers.env` | `containers.env:140` declares only `RTVI_EMBED_IMAGE`, with **no tag variable at all** |
| 3 | compose on the shared coordinate, inline literal `develop-latest` | `rtvi-embed-docker-compose.yml:21` inlines `nvcr.io/nvstaging/vss-core/vss-rt-embed:3.3.0-26.07.4` |
| 4 | regenerate golden (coordinate moves NGC → GHCR) | `compose-images.golden:64` still NGC |

`containers.env:140` is:

```bash
RTVI_EMBED_IMAGE="${RTVI_EMBED_IMAGE:-${VSS_CONTAINER_STAGING_REGISTRY}/vss-rt-embed}"
```

`VSS_CONTAINER_REGISTRY` is not in that chain, and `RTVI_EMBED_TAG` is never declared in `containers.env` — so neither shared knob can reach this image.

**Verified rather than argued**, running the doc's own Verify procedure on this head with a compliant image as control:

```
VSS_CONTAINER_REGISTRY=ghcr.io/nvidia-ai-blueprints/vss VSS_CONTAINER_TAG=pr-1693-2df450a9c6ad
  vss-rt-embed            -> nvcr.io/nvstaging/vss-core/vss-rt-embed:3.3.0-26.07.4          (unmoved)
  vss-behavior-analytics  -> ghcr.io/.../vss-behavior-analytics:pr-1693-2df450a9c6ad        (moves correctly)
```

Same command, same env files. Two independent blockers — with the profile `.env` pins removed entirely, `containers.env:140` alone still resolves to NGC; and with `containers.env` fixed, the unconditional pins at `rtvi.env:9-10` and `rtvi-embed/.env:17-18` would clobber it.

Census across all 12 `ghcr_build: true` images: **`vss-rt-embed` is the only one whose chain contains neither `VSS_CONTAINER_REGISTRY` nor `VSS_CONTAINER_TAG`.** (`vss-agent`, `sdr-mw-l` and LVS deviate on naming, but each carries a justifying comment and each still reaches the shared knobs.) `containers.env` states the rule against itself: this entry sits under *"First-party images whose TAG variable already exists in profile .env files"* — a section for non-GHCR images — whose closing comment reads *"MV3DT BEV Fusion joined the managed set above when it gained a GitHub-native GHCR build."*

This is no longer hypothetical: the candidate `pr-1693-2df450a9c6ad@sha256:8726b2726e75954f456bb82ec16043ccab353d83bdd959599b020a7177536009` is built, published and downstream-validated, while every deployment path resolves to NGC.

**Note that no gate catches this.** `Container Coordinates Golden` is green and `release_set.py closure` exits 0 — I ran both. The golden is green *because* the coordinate did not move; the doc calls the golden diff "the point of review" precisely because it is a human artifact, not a gate.

Smallest fix that restores the invariant, keeping the `RTVI_EMBED_*` names so `test-dev-profile.sh:1356`'s asserted variable list needs no edit (same compat precedent as `SDR_MW_L_IMAGE`):

```bash
RTVI_EMBED_IMAGE="${RTVI_EMBED_IMAGE:-${VSS_CONTAINER_REGISTRY:-${VSS_CONTAINER_STAGING_REGISTRY}}/vss-rt-embed}"
RTVI_EMBED_TAG="${RTVI_EMBED_TAG:-${VSS_CONTAINER_TAG:-3.3.0-26.07.4}}"
```

moved into the managed set, plus dropping the two hard pins in `rtvi.env:9-10` and `rtvi-embed/.env:17-18`, then regenerating the golden.

**One fair counter-argument**, which is why this is a question as much as a request: `develop-latest` does not exist for a newly-onboarded image until its first develop build lands, so wiring step 3 now points at a tag that 404s until merge. The candidates bot also notes *"Promotion copies the same manifest digests to NGC"*, so keeping deployments on NGC may well be the intent. If so, that is a legitimate staged rollout — but it is currently undeclared, and the doc treats steps 1–4 as one change. Either complete them here, or say explicitly in the PR description that this is build-only onboarding with a named follow-up.

deploy/docker/container-inventory.json
```diff
@@ -110,9 +110,12 @@
...
+      "native_platform_build": true,
```
zac-wang-nv, line 122: **NOTE** (was P1) — Correcting my round-5 comment, which said this build "has never executed", and then recording the outcome.

It executed and **failed**: run `31777860322` at `181c86677`, job `Build vss-rt-embed → GHCR`, step `Build and push (multiarch, immutable tag only)`:

```
07:02:44  ##[error]No space left on device
07:02:50  ##[warning]Free space left: 4 MB
```

It died at 516s **while still extracting base images**, with both arches on one `ubuntu-latest` runner under QEMU. It was the only failing job of ~20 in that run.

**This flag is the fix, and it is now proven.** The `/ok to test 2df450a` sweep is the first run to exercise the native path, and both legs went green:

```
Build vss-rt-embed (amd64, native)   success   11:50:27 -> 12:34:54  (44m27s)
Build vss-rt-embed (arm64, native)   success   11:50:29 -> 12:30:11  (39m42s)
Publish vss-rt-embed multiarch manifest        success   12:35:38
```

Worth stating precisely, because it is easy to describe this as "the same build lasting longer" and that is not what happened: the run that died was a *single* `ubuntu-latest` job doing both architectures under QEMU; `dfbc50199` routes this image to `build-native-platforms`, which is two jobs on `ubuntu-24.04` and `ubuntu-24.04-arm` with no emulation. The one-runner disk arithmetic cannot apply by construction. My earlier framing of the residual risk — that `9cc7d6cf1`'s new `pyds_arm64_builder` stage might spend the recovered headroom — was reasonable but is now answered: it did not.

I checked the mechanism rather than assuming it: `is_native_platform_build` / `split_build_matrices` already exist on develop (`detect_changed_images.py` is **not** in this diff; it landed with #1623), and the native path had already succeeded for `sdr-mw-l`, `vss-configurator`, `vss-behavior-analytics` and `vss-rt-config-adaptor`. This was a one-line opt-in to a proven path. Nothing to change.

services/rtvi/rt-embed/docker/py_deps/README.md
```diff
@@ -0,0 +1,61 @@
...
+```sh
+pdm add --update-reuse <pkg>==<ver>
+```
+OR manually edit pyproject.toml
+
+# Package sources
+
+RT Embed dependencies must resolve from public PyPI, `pypi.nvidia.com`, and
+explicit public source archives only. The optional vLLM-compatible implementation remains in
+the source tree but is not packaged in this image.
```
zac-wang-nv, line 45: **P1** — Carried from round 5 and re-verified at this head (unchanged by either new commit). The GHCR-built image ships without vLLM while the source tree still contains the code path that expects it, so builtin models routed through that path are unusable in the built image even though the repo suggests otherwise.

Flagging honestly: I re-verified that the README text and the packaging decision are unchanged, but did not re-derive the downstream model list this round. Worth a line here naming which models are affected, so the constraint is discoverable from the deployment side rather than only from a dependency README.

services/rtvi/rt-embed/docker/Dockerfile
```diff
@@ -13,30 +13,402 @@
...
+COPY docker/binaries/igpu/libnvbufsurface.so \
+     docker/binaries/igpu/libnvbufsurftransform.so \
+     /usr/lib/aarch64-linux-gnu/tegra/
+RUN --mount=type=bind,source=docker/ubuntu.sources.${TARGETARCH},target=/etc/apt/sources.list.d/ubuntu.sources \
+    apt update && apt install -y --no-install-recommends \
+    build-essential cmake git libglib2.0-dev libgstreamer1.0-dev \
+    python3.12-dev python3-setuptools python3-wheel \
+    && rm -rf /var/lib/apt/lists/*
+ADD https://github.com/NVIDIA-AI-IOT/deepstream_python_apps/archive/refs/tags/v1.2.2.tar.gz /tmp/deepstream_python_apps.tar.gz
+RUN echo "${PYDS_SOURCE_SHA256}  /tmp/deepstream_python_apps.tar.gz" | sha256sum -c - \
```
zac-wang-nv: **P2** — The tarball is downloaded, checksummed, extracted — and then thrown away. Four lines later `git checkout --detach --force FETCH_HEAD` replaces every tracked file with the fetched commit's content.

The real supply-chain pin is the pair you already have, and it is the stronger one:

```dockerfile
&& git fetch --depth=1 origin "${PYDS_SOURCE_COMMIT}" \
&& test "$(git rev-parse HEAD)" = "${PYDS_SOURCE_COMMIT}"
```

That is a content-addressed pin with an explicit assertion. The `ADD` + `sha256sum` adds a second network dependency and a second thing that can rot — GitHub's auto-generated `/archive/refs/tags/` tarballs are not byte-stable, and a recompression upstream would fail this build for no security benefit. (I checked both sides: the live tarball still hashes to `bc71ee88…`, and `6fdeefb71` is genuinely the `v1.2.2` tag commit, so they agree today.)

Suggest dropping the `ADD` / extract / checksum and letting the git pin stand alone. Note this *removes* two of the new assertions rather than adding more — `test_compose_image_golden.py:258` currently pins `pyds_source_checksum`, which entrenches the redundant check.

services/rtvi/rt-embed/docker/Dockerfile
```diff
@@ -13,30 +13,402 @@
...
+    && git -C /tmp/deepstream_python_apps checkout --detach --force FETCH_HEAD \
+    && test "$(git -C /tmp/deepstream_python_apps rev-parse HEAD)" = "${PYDS_SOURCE_COMMIT}" \
+    && git -C /tmp/deepstream_python_apps submodule update --init --recursive --depth 1 \
+    && cd /tmp/deepstream_python_apps \
+    && python3 bindings/3rdparty/git-partial-submodule/git-partial-submodule.py restore-sparse
+COPY docker/patches/pyds-arm64.patch /tmp/pyds-arm64.patch
+RUN cd /tmp/deepstream_python_apps \
+    && git apply /tmp/pyds-arm64.patch \
+    && cd bindings \
+    && CMAKE_ARGS="-DIS_SBSA=on -DDS_PATH=/opt/nvidia/deepstream/deepstream-9.1" \
```
zac-wang-nv, line 90: **P2** — A question about a deliberate choice, not a claimed bug.

`IS_SBSA` is hardcoded `on`, but the platform selector that would justify it is never set. `ARM_PLATFORM` defaults to `"igpu"` at `Dockerfile:16` and appears **nowhere** under `.github/` or `deploy/` — zero occurrences, so nothing ever passes it. Two consequences:

1. `base-sbsa` (`Dockerfile:36`, `deepstream:9.1-triton-sbsa-dgx-spark`) is unreachable dead configuration.
2. `pyds_arm64_builder` is always `FROM base-igpu`, so `-DIS_SBSA=on` is applied to an **iGPU** base 100% of the time — compiling out the guard upstream added specifically for non-SBSA aarch64.

The in-code comment says this is intended for Thor, and the vendored igpu `libnvbufsurface.so` suggests it has been thought through, so I am **not** asserting breakage. But the flag now reads as a lie about the platform, and the next person cannot tell whether `on` is a deliberate Thor workaround or a copy-paste. Two asks: (a) a one-line comment recording what validated `get_nvds_buf_surface_gpu` on Thor iGPU hardware; (b) either wire `ARM_PLATFORM` through the build so `sbsa` is reachable, or delete `base-sbsa`.

services/rtvi/rt-embed/docker/Dockerfile
```diff
@@ -13,30 +13,402 @@
...
+###############################
+
+# Build a Thor-compatible PyDS wheel from the public DeepStream Python bindings.
+# The upstream release arm64 wheel deliberately rejects get_nvds_buf_surface_gpu
+# on non-SBSA arm64; the local Cosmos Embed1 raw-frame path requires it.
+FROM base-arm64 AS pyds_arm64_builder
+ARG TARGETARCH
+ARG PYDS_SOURCE_COMMIT=6fdeefb7128435873f7794d2242ed48a1471ad7e
+ARG PYDS_SOURCE_SHA256=bc71ee88923c6c2f5ed60476504d338068a35a8e87157547571061b416d4ab20
+COPY docker/binaries/igpu/libnvbufsurface.so \
```
zac-wang-nv, line 57: **P2** — Carried from round 5, but the stakes went up here: these are Git LFS files, and this commit makes them inputs to a **compile and link** step rather than just runtime artifacts.

The good news, which I checked rather than assumed: `lfs_include` does survive into the native matrix — `split_build_matrices` spreads `**base` into every native entry, and both job paths run `git lfs pull --include` (`build-dev-images.yml:126` and `:293`). The pull is wired correctly, and both native legs have now built green, so it demonstrably works today.

What is still missing is any verification that the pull *produced* anything. If LFS is throttled or the pull silently no-ops, these become ~130-byte pointer text files named `.so`, and the failure surfaces as a confusing CMake/link error deep inside a 40-minute build rather than as "LFS assets missing". One cheap line after the pull closes it:

```bash
file services/rtvi/rt-embed/docker/binaries/igpu/*.so | grep -q ELF \
  || { echo "::error::LFS assets not materialized"; exit 1; }
```

(They are real aarch64 ELF objects in my worktree — 1.0 MB and 42 MB — so this is about CI determinism, not a current breakage.)

services/rtvi/rt-embed/docker/Dockerfile
```diff
@@ -13,30 +13,402 @@
...
+FROM nvcr.io/nvidia/deepstream:9.1-triton-multiarch AS base-igpu
+FROM nvcr.io/nvidia/deepstream:9.1-triton-sbsa-dgx-spark AS base-sbsa
+
+FROM base-$ARM_PLATFORM AS base-arm64
+###############################
+
+# Build a Thor-compatible PyDS wheel from the public DeepStream Python bindings.
+# The upstream release arm64 wheel deliberately rejects get_nvds_buf_surface_gpu
+# on non-SBSA arm64; the local Cosmos Embed1 raw-frame path requires it.
+FROM base-arm64 AS pyds_arm64_builder
```
zac-wang-nv, line 53: **NOTE** — This stage is the best-engineered thing in the PR and deserves saying so explicitly, because the rest of my review is critical.

Building the bindings from a commit-pinned source with a vendored patch is the right call — strictly stronger than the checksummed-wheel approach originally asked for, since a wheel checksum pins bytes you did not inspect while `rev-parse HEAD` pins source you can read. Three details are better than they had to be:

- `test "$(git rev-parse HEAD)" = "${PYDS_SOURCE_COMMIT}"` — the pin is asserted, not assumed.
- The post-build check greps the compiled `pyds.so` for the exact string that `-DIS_SBSA=on` compiles out. That is a real detector of whether the flag took effect, not a proxy for it, and it is fail-closed: if the wheel layout changes, `ZipFile.read("pyds.so")` raises rather than silently passing.
- Both arms of the install `if` terminate in `&&` chains whose failure propagates to the `RUN` exit status, so a missing wheel fails the build instead of shipping an image without pyds.

I tried to falsify this stage five ways and could not: the vendored patch applies at the pinned commit (`git apply --check` → exit 0, hunks at offset 4); `PYDS_SOURCE_COMMIT` is genuinely the `v1.2.2` tag commit; the live tarball still matches `PYDS_SOURCE_SHA256`; `lfs_include` does reach the native matrix; and the new guards pass 17/17 with the policy suite 6/6. My only structural note is the comment on line 57.

zac-wang-nv:
## Review summary — PR #1693 (round 6, head `2df450a9c`)

Posting this now having reviewed every head since `c2928e561`. Two things landed while I was writing, and both changed my conclusions, so I have revised rather than shipped what I had.

### The P0 is closed, and it was never the author's

`2df450a9c` restored the NGC pin in both remaining Helm paths. Verified by A/B render in one session — same helm version, same built dependencies, same command:

```
9cc7d6cf1  ->  image: ghcr.io/nvidia-ai-blueprints/vss/vss-rt-embed:develop-latest   (wrong)
2df450a9c  ->  image: nvcr.io/nvstaging/vss-core/vss-rt-embed:3.3.0-26.07.4          (correct)
```

Corroborated two other ways: both files have **left the PR diff entirely** (46 → 44 changed files, now byte-identical to the merge base), and all six Greptile threads are resolved.

Worth stating plainly, since several rounds of my review pointed at these files: **the regression was never yours.** It entered at the helm-sync bot commit `486c3c62c`, which rewrote five files. `07eca0826` reverted one and `2df450a9c` reverted the two that decided the rendered image. The base branch and every one of your pre-bot commits always rendered the correct NGC coordinate.

### The decisive build went green — and the earlier failure was real

Round 5 of my review said this build "has never executed". That was wrong: it executed at `181c86677` and failed with `No space left on device` at 516s, while still extracting base images, with both architectures on one `ubuntu-latest` runner under QEMU.

`dfbc50199`'s `native_platform_build` flag is the fix, and the sweep you dispatched at 11:49:58Z proved it:

```
Build vss-rt-embed (amd64, native)   success   44m27s
Build vss-rt-embed (arm64, native)   success   39m42s
Publish vss-rt-embed multiarch manifest        success   12:35:38Z
```

This is not "the same build lasting longer" — it is two native jobs on separate runners instead of one emulated job, so the disk arithmetic that killed the first attempt cannot apply. The candidate `ghcr.io/nvidia-ai-blueprints/vss/vss-rt-embed:pr-1693-2df450a9c6ad` is published and was listed in the downstream-validated candidate set at 13:29:55Z.

Board at this head: **66 green, 5 skipped, 2 red.**

### The one blocking item is OSRB, and it will not clear by re-running

`License Diff (OSRB CSV)` and `OSRB Review` are the only two failures. `License Diff` fails at the step named *"Fail when OSRB re-engagement is required"*, and it has failed on **every head this PR has ever had** — `c2928e561`, `181c86677` and `2df450a9c`. This is not a flake and not a stale result.

`9cc7d6cf1` strengthens the case rather than weakening it, adding two pieces of third-party surface:

1. `services/rtvi/rt-embed/docker/patches/pyds-arm64.patch` — 100 lines of **vendored modifications** to NVIDIA-AI-IOT/deepstream_python_apps, carrying our own SPDX Apache-2.0 header while patching a third-party file.
2. The image now **compiles and ships a derived binary** of that project rather than installing its published wheel.

Redistributing a modified build is a different licensing posture from redistributing an upstream artifact, and the SPDX header on the patch describes the patch, not the patched result. Combined with the 35 new dependencies noted earlier (2 under a non-OSS licence), that is three distinct reasons the gate is likely right. Unlike the build failure, this has a lead time — worth starting now rather than after everything else is green. If you believe the gate is a false positive, that is worth saying explicitly on the PR; it has been red since `c2928e561` and silence reads as "not looked at yet".

### GHCR onboarding is one step of four

The inline comment on `container-inventory.json:113` is the one new finding this round. `ghcr_build: true` is step 1 of `.github/github-first.md`; steps 2 (`containers.env`), 3 (the compose line) and 4 (the golden) are not in this PR, which is the doc's own first pitfall. I verified with the doc's Verify procedure that neither `VSS_CONTAINER_REGISTRY` nor `VSS_CONTAINER_TAG` can reach this image, using a compliant image as a control. No gate catches it — `Container Coordinates Golden` is green precisely because the coordinate did not move.

There is a legitimate reading where this is deliberate staging (promotion copies GHCR digests to NGC, and `develop-latest` does not exist for a new image until first merge). If so it just needs saying in the PR description.

### One more, on CI mechanics

For most of this review the CI mirror `refs/heads/pull-request/1693` was stuck at `181c86677` while the head was `2df450a9c`, so five commits had never been through a gate. You fixed that at 11:49:58Z with `/ok to test 2df450a` while I was still writing. The mechanism is worth knowing because nothing surfaces it: `ci.yml` has no `pull_request:` trigger, so every real gate fires only on `push` to `pull-request/[0-9]+`, and that ref only advances when `/ok to test` names a SHA. **A head with green bot checks and no `/ok to test` looks swept and is not.**

### Cache note (P2)

`build-dev-images.yml:373` sets `cache-to: type=gha,mode=max`, exporting every intermediate stage to the Actions cache. That file is not in this diff, so this is pre-existing infrastructure — but `9cc7d6cf1` is what makes it bite: `pyds_arm64_builder` is a new multi-gigabyte stage (apt toolchain + DeepStream source + recursive submodules + CMake tree) that `mode=max` now pushes into a cache capped repository-wide at 10 GB and evicted LRU. The risk is not this image failing; it is this image quietly evicting every other image's cache. Worth considering `mode=min` for `vss-rt-embed`.

---

Summary: **8 inline comments — 3×P1, 3×P2, 2×NOTE.** No P0. The code is in good shape; two of the P1s are about declaring intent (the guard carve-out and the half-finished GHCR onboarding) rather than defects, and the blocking item is OSRB process.

services/rtvi/rt-embed/docker/Dockerfile
nv-pnikam: No change needed here as per the comment.

services/rtvi/rt-embed/docker/Dockerfile
nv-pnikam: Resolved in f41ac3c.

  Both standard and native-platform GHCR build paths now validate RT Embed’s LFS-backed .so files immediately after git lfs pull. The workflow checks every library under docker/binaries is an ELF object and fails with LFS assets not
  materialized before Docker starts if a Git LFS pointer remains. A regression test verifies the check is present in both paths.

services/rtvi/rt-embed/docker/Dockerfile
nv-pnikam: Resolved in f41ac3c.

  IS_SBSA=on is intentional for the current Thor/iGPU build: it is the compile-time override required to enable get_nvds_buf_surface_gpu on aarch64. I added a Dockerfile comment documenting that Thor dependency. The separate SBSA runtime path
  remains available for future enablement.

services/rtvi/rt-embed/docker/Dockerfile
nv-pnikam: Resolved in f41ac3c.

  Removed the redundant PyDS source tarball download, checksum, and extraction. The build now fetches the pinned upstream commit directly and verifies the checked-out HEAD matches PYDS_SOURCE_COMMIT, eliminating the unused archive path.

services/rtvi/rt-embed/docker/py_deps/README.md
nv-pnikam: vLLM source code will be removed in later updates, it originated from code common with RT-VLM.

deploy/docker/container-inventory.json
nv-pnikam: Thanks for the correction and for validating the native build path end-to-end. Agreed: no further change is needed for this comment. RT Embed now builds as separate native amd64 and arm64 jobs, avoiding the QEMU multi-architecture disk-pressure failure.

deploy/docker/container-inventory.json
nv-pnikam: Resolved in f41ac3c.

  Completed the managed-image migration wiring: RT Embed now has namespaced image/tag variables, compose defaults to the GHCR coordinate, stale service-level image overrides were removed, and the compose golden output was regenerated. The
  resolved deploy image honors the shared registry/tag overrides while retaining the expected GHCR default.

.github/scripts/test_helm_release_channel_policy.py
nv-pnikam: Resolved in f41ac3c.

  RT Embed is now registered in both HELM_VALUES and HELM_HELPERS, and the policy test again enforces exact inventory coverage. This removes the one-off carve-out while retaining validation of the chart defaults and helper-based global
  override path.

nv-pnikam:
/ok to test [f41ac3c](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1693/commits/f41ac3ceed469d8467c5cd737dfb1016fcb86f81)

deploy/docker/containers.env
```diff
@@ -103,6 +103,9 @@ VSS_RT_CV_MV3DT_BEV_FUSION_TAG="${VSS_RT_CV_MV3DT_BEV_FUSION_TAG:-${VSS_CONTAINE
 VSS_RT_CV_MV3DT_CONFIG_INIT_IMAGE="${VSS_RT_CV_MV3DT_CONFIG_INIT_IMAGE:-${VSS_CONTAINER_REGISTRY:-${VSS_CONTAINER_STAGING_REGISTRY}}/vss-rt-cv-mv3dt-config-init}"
 VSS_RT_CV_MV3DT_CONFIG_INIT_TAG="${VSS_RT_CV_MV3DT_CONFIG_INIT_TAG:-${VSS_CONTAINER_TAG:-3.3.0-26.08.1}}"
 
+VSS_RTVI_EMBED_IMAGE="${VSS_RTVI_EMBED_IMAGE:-${VSS_CONTAINER_REGISTRY:-${VSS_CONTAINER_STAGING_REGISTRY}}/vss-rt-embed}"
```
zac-wang-nv: **P1** — This rename is right, but it orphaned six files that still document the old variables, and the resulting failure is silent rather than loud.

`RTVI_EMBED_IMAGE` / `RTVI_EMBED_TAG` no longer exist anywhere under `deploy/`: this line replaced them, and `rtvi.env` / `rtvi-embed/.env` dropped their hard pins. `git grep RTVI_EMBED_IMAGE -- deploy/` returns nothing, so **the step-2 check in `github-first.md` passes** — but that grep is scoped to `deploy/`, and these consumers sit outside it:

| File | Lines |
|---|---|
| `skills/vss-deploy-video-embedding/SKILL.md` | 34, 35, 253 |
| `skills/vss-deploy-video-embedding/references/deploy-vss-deploy-video-embedding.md` | 5, 6, 118, 122 |
| `skills/vss-deploy-video-embedding/references/environment.md` | 20, 21 |
| `skills/vss-deploy-video-embedding/references/integrate-vss-deploy-video-embedding.md` | 69, 70, 165 |
| `skills/vss-deploy-profile/references/search.md` | 109, 136 |
| `skills/vss-build-vision-agent/references/services/rt-embed.md` | 23 |

**Why this is P1 and not a docs nit.** `deploy-vss-deploy-video-embedding.md:122` documents the rollback path as:

> Roll back by re-pinning `RTVI_EMBED_TAG` to the previous build and repeating the pull and recreate steps.

Follow that today and you set a variable compose no longer reads. There is no error and no warning — the service silently comes up on `ghcr.io/nvidia-ai-blueprints/vss/vss-rt-embed:develop-latest` instead of the build you asked for. That is the worst shape for a rollback path to fail in. `integrate-vss-deploy-video-embedding.md:165` still shows the pre-rename compose line verbatim:

```yaml
image: ${RTVI_EMBED_IMAGE:-nvcr.io/nvstaging/vss-core/vss-rt-embed}:${RTVI_EMBED_TAG:-3.3.0-26.08.1}
```

so the doc now contradicts the file it documents.

Two things stop CI from catching it: `compose_image_golden.py` only walks compose files under `deploy/`, and `.github/skill-eval/adapters/vss-deploy-video-embedding/generate.py` consumes this corpus — so the stale coordinates feed skill-eval as ground truth.

**Ask:** update those six files — `RTVI_EMBED_IMAGE` → `VSS_RTVI_EMBED_IMAGE`, `RTVI_EMBED_TAG` → `VSS_RTVI_EMBED_TAG` (or the derived names, see my comment on `container-inventory.json`) — and move the documented defaults from `nvcr.io/nvstaging/vss-core/vss-rt-embed:3.3.0-26.08.1` to the GHCR coordinate.

zac-wang-nv (commented):
## Round 8 — two follow-ups on `f41ac3cee`

`f41ac3cee` closes both P1s from the last round, and closes them the stronger way: the exact-set carve-out is gone with `assertEqual(managed, set(HELM_VALUES))` restored, and `github-first.md` steps 2–4 all land. I ran the doc's Verify procedure against a compliant control — the shared knobs now reach this image. Both P2s are fixed too, and I verified the LFS guard covers all 11 LFS-tracked files under `docker/binaries`.

Two follow-ups inline, both flowing from the variable rename. Neither is a code defect; the first is a silent-failure risk in published docs.

One correction I owe you: I said in an earlier round that the GHCR package was not anonymously pullable (403). **It is** — visibility reads `public` and `pr-1693-2df450a9c6ad` returns 200. So dropping `imagePullSecrets` to `[]` is correct. I am also withdrawing my `develop-latest` concern: it 404s today, but that is the literal `github-first.md:106` prescribes and every peer chart ships it.

Separately, and not inline because it is not in this repo: **`vss-rt-embed` is not registered in `ci/vss-ghcr-images.yml` in `ci-vss-oss`** (step 6). 11 images are listed there and this is not one of them. The doc calls that failure *silence* — the image publishes to GHCR fine, `nvstaging` never receives it, and the nightly deploy fails looking for something never promoted. That matters more now that this PR moved the Helm and compose defaults onto the GHCR channel. Could you open that MR and link it here? I did not want to guess at `ngc_dev_team` / `promotion_config`.

deploy/docker/container-inventory.json
```diff
@@ -110,12 +110,15 @@
...
       "source_path": "services/rtvi/rt-embed",
+      "context": "services/rtvi/rt-embed",
       "dockerfile": "services/rtvi/rt-embed/docker/Dockerfile",
+      "lfs_include": "services/rtvi/rt-embed/docker/binaries/**",
       "platforms": ["linux/amd64", "linux/arm64"],
       "compose_image_names": ["vss-rt-embed"],
-      "tag_variables": ["RTVI_EMBED_TAG"]
+      "tag_variables": ["VSS_RTVI_EMBED_TAG"]
```
zac-wang-nv: **P2** — `tag_variables` names `VSS_RTVI_EMBED_TAG`, but the image is `vss-rt-embed`, so the derived name is `VSS_RT_EMBED_TAG` — same for the `_IMAGE`/`_TAG` pair in `containers.env:106-107`. The stem is taken from the *service* name (`rtvi-embed`: the compose dir, chart and profile) rather than from the image name.

**Nothing breaks, and I checked that before raising it.** Nothing reads `tag_variables` — `github-first.md` says it is descriptive metadata and a repo-wide grep confirms zero consumers. Nothing derives a variable name from an image name under `.github/` or `deploy/docker/`. And on the acceptance side, `ci-vss-oss/ci/setup-eval-overrides.sh` drives everything off the shared `VSS_CONTAINER_TAG`, with only hardcoded `VSS_AGENT_*` / `VSS_UI_*` / `VSS_ALERT_MS_*` overrides. So this is naming, not a defect.

What makes it worth changing anyway is the shape of the existing exceptions. Census of all 12 `ghcr_build: true` images: 8 derive exactly, 4 do not — and the two that deviate deliberately each carry a comment in `containers.env` saying why:

- `SDR_MW_L_IMAGE` — *"keeps the full image override for compatibility with existing profile env files and scripts that set `SDR_MW_L_IMAGE` directly"*
- `CONTAINER_IMAGE` (LVS) — *"keeps the combined `CONTAINER_IMAGE` knob … that its compose file and profile `.env` files already use"*

Both deviate **to keep a name existing consumers already set**. (The third, `VSS_ALERT_IMAGE`, is untouched legacy from #1190.) That justification is the one thing that does not apply here — this PR renamed `RTVI_EMBED_*` anyway, which is exactly what orphaned the six files in my other comment. Having already paid the breaking-change cost, the current name matches neither the old consumers nor the image, and `vss-rt-embed` is what appears in the inventory, the golden file, the GHCR coordinate and `docker compose config --images` output.

**Ask:** rename to `VSS_RT_EMBED_IMAGE` / `VSS_RT_EMBED_TAG`. It is this line plus two in `containers.env` and one in the compose file, and it lands in the *same edit* as the doc fix above, which already has to rewrite these names across six files — so it is close to free now, and avoids a second breaking rename later. No collision with `VSS_RT_CONFIG_ADAPTOR_*` or `VSS_RT_CV_*`.

If you would rather keep `RTVI` to match the service naming, that is defensible — in that case please add a one-line comment in `containers.env` saying so, which is how the other two deviations are handled.

nv-pnikam:
/ok to test [265363c](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1693/commits/265363c3ebb9514f9b7b4447645fee1ae49bd05b)

deploy/docker/container-inventory.json
nv-pnikam: Resolved in 5e64f8f.

  Renamed the RT Embed image-selection variables to VSS_RT_EMBED_IMAGE and VSS_RT_EMBED_TAG in the inventory metadata, containers.env, and Compose file. The names now match the vss-rt-embed image stem; no legacy alias was retained because there were no consumers. Verified explicit overrides render the expected image coordinate.

deploy/docker/containers.env
nv-pnikam: Resolved in 265363c.

Updated all six affected skill documents to use VSS_RT_EMBED_IMAGE / VSS_RT_EMBED_TAG and the GHCR develop-latest default. The upgrade and rollback instructions now set variables that Compose actually reads, and the documented Compose fragment matches the live fallback structure. NGC guidance now applies only to model and asset access, not pulling the public image.

nv-pnikam:
/ok to test [05a2098](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1693/commits/05a2098e27a9dd10c90f9b2b7c764df9cf23c638)

nv-pnikam:
/ok to test [dcbbee1](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1693/commits/dcbbee1fa4b48ca73671662992598336a06303c2)

nv-pnikam:
/ok to test [a24738b](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1693/commits/a24738b4a22d50e990d3caa0fe3e48013a2e2990)

nv-pnikam:
/ok to test [88d26d5](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1693/commits/88d26d575d0978a2f0d1d549b789f4f5e6d4935d)

zac-wang-nv (commented):

The internal review failed closed: expected 102 classifications, got 15; row 1 is unclassified; row 2 is unclassified; row 3 is unclassified; row 4 is unclassified; row 5 is unclassified; row 6 is unclassified; row 7 is unclassified; row 9 must be APPROVED_VERSION_BUMP under the re-open policy; row 10 is unclassified; row 13 is unclassified; row 14 is unclassified; row 16 is unclassified; row 17 is unclassified; row 18 is unclassified; row 19 is unclassified; row 20 is unclassified; row 22 is unclassified; row 23 is unclassified; row 24 is unclassified; row 25 is unclassified; row 26 is unclassified; row 27 is unclassified; row 28 is unclassified; row 29 is unclassified; row 30 is unclassified; row 31 is unclassified; row 32 is unclassified; row 33 is unclassified; row 34 is unclassified; row 35 is unclassified; row 36 is unclassified; row 37 is unclassified; row 39 is unclassified; row 40 is unclassified; row 41 is unclassified; row 42 is unclassified; row 43 is unclassified; row 44 is unclassified; row 45 is unclassified; row 46 is unclassified; row 47 is unclassified; row 48 is unclassified; row 49 is unclassified; row 50 is unclassified; row 51 is unclassified; row 52 is unclassified; row 53 is unclassified; row 54 is unclassified; row 56 is unclassified; row 57 is unclassified; row 58 package mismatches CSV evidence; row 58 new_version mismatches CSV evidence; row 58 new_license mismatches CSV evidence; row 59 package mismatches CSV evidence; row 59 new_version mismatches CSV evidence; row 59 new_license mismatches CSV evidence; row 60 package mismatches CSV evidence; row 60 new_version mismatches CSV evidence; row 61 is unclassified; row 62 is unclassified; row 63 package mismatches CSV evidence; row 63 new_version mismatches CSV evidence; row 63 new_license mismatches CSV evidence; row 64 is unclassified; row 65 is unclassified; row 66 is unclassified; row 67 is unclassified; row 68 is unclassified; row 69 is unclassified; row 70 is unclassified; row 71 is unclassified; row 72 is unclassified; row 73 is unclassified; row 74 is unclassified; row 75 is unclassified; row 76 is unclassified; row 77 is unclassified; row 78 is unclassified; row 79 is unclassified; row 80 is unclassified; row 81 package mismatches CSV evidence; row 81 new_version mismatches CSV evidence; row 82 is unclassified; row 83 is unclassified; row 84 is unclassified; row 85 package mismatches CSV evidence; row 85 new_version mismatches CSV evidence; row 85 new_license mismatches CSV evidence; row 86 is unclassified; row 87 is unclassified; row 88 is unclassified; row 89 is unclassified; row 90 is unclassified; row 91 is unclassified; row 92 is unclassified; row 93 is unclassified; row 94 is unclassified; row 95 is unclassified; row 96 is unclassified; row 97 is unclassified; row 98 is unclassified; row 99 is unclassified; row 100 is unclassified; row 101 is unclassified

- Dependency changes evaluated: `102`
- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32211808848)

| Package | Version | License | Classification |
|---|---|---|---|
| cryptography | 46.0.5 | Apache-2.0 OR BSD-3-Clause | APPROVED_VERSION_BUMP |
| cuda-bindings | 13.2.0 | LicenseRef-NVIDIA-SOFTWARE-LICENSE | NOT_APPROVED_VERSION_BUMP |
| cuda-python | 13.2.0 | LicenseRef-NVIDIA-SOFTWARE-LICENSE | NOT_APPROVED |
| mkl | 2021.1.1 | Intel Simplified Software License | NOT_APPROVED |
| tbb | 2021.13.1 | Intel Simplified Software License | NOT_APPROVED |
| opentelemetry-exporter-prometheus | 0.57b0 |  | NOT_APPROVED |
| opentelemetry-proto | 1.36.0 |  | NOT_APPROVED_VERSION_BUMP |
| opentelemetry-sdk | 1.36.0 |  | NOT_APPROVED_VERSION_BUMP |
| opentelemetry-semantic-conventions | 0.57b0 |  | NOT_APPROVED_VERSION_BUMP |
| pillow | 12.2.0 |  | NOT_APPROVED_VERSION_BUMP |
| starlette | 0.50.0 |  | NOT_APPROVED_VERSION_BUMP |
| fastapi | 0.121.3 | MIT | APPROVED_VERSION_BUMP |
| grpcio | 1.78.0 | Apache-2.0 | APPROVED_VERSION_BUMP |
| cupy-cuda12x | 13.3.0 | MIT License | NOT_APPROVED |
| aiofiles | 23.2.1 | Apache-2.0 | APPROVED_VERSION_BUMP |

nv-pnikam:
/ok to test [1e8d516](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1693/commits/1e8d5162f30ba173a7efbd2634e6a95f305407a2)

zac-wang-nv (commented):

The internal review failed closed: expected 102 classifications, got 17; row 0 is unclassified; row 1 is unclassified; row 2 is unclassified; row 4 is unclassified; row 5 is unclassified; row 6 is unclassified; row 7 is unclassified; row 10 is unclassified; row 12 is unclassified; row 13 is unclassified; row 14 is unclassified; row 15 is unclassified; row 16 is unclassified; row 17 is unclassified; row 18 is unclassified; row 19 is unclassified; row 20 is unclassified; row 21 is unclassified; row 22 is unclassified; row 23 is unclassified; row 24 is unclassified; row 25 is unclassified; row 26 is unclassified; row 27 is unclassified; row 28 is unclassified; row 29 is unclassified; row 30 is unclassified; row 31 is unclassified; row 32 is unclassified; row 33 is unclassified; row 34 is unclassified; row 35 is unclassified; row 36 is unclassified; row 37 is unclassified; row 40 is unclassified; row 41 is unclassified; row 42 is unclassified; row 43 is unclassified; row 44 is unclassified; row 45 is unclassified; row 46 is unclassified; row 47 is unclassified; row 48 is unclassified; row 49 is unclassified; row 50 is unclassified; row 51 is unclassified; row 52 is unclassified; row 53 is unclassified; row 54 is unclassified; row 55 is unclassified; row 56 is unclassified; row 57 is unclassified; row 58 package mismatches CSV evidence; row 58 new_license mismatches CSV evidence; row 59 is unclassified; row 60 package mismatches CSV evidence; row 60 new_version mismatches CSV evidence; row 60 new_license mismatches CSV evidence; row 61 package mismatches CSV evidence; row 62 package mismatches CSV evidence; row 62 new_version mismatches CSV evidence; row 63 is unclassified; row 64 is unclassified; row 65 is unclassified; row 67 is unclassified; row 69 is unclassified; row 70 is unclassified; row 71 is unclassified; row 72 is unclassified; row 73 is unclassified; row 74 is unclassified; row 75 is unclassified; row 76 is unclassified; row 77 is unclassified; row 78 package mismatches CSV evidence; row 78 new_version mismatches CSV evidence; row 78 new_license mismatches CSV evidence; row 78 must be APPROVED_VERSION_BUMP under the re-open policy; row 79 is unclassified; row 80 is unclassified; row 81 package mismatches CSV evidence; row 81 new_version mismatches CSV evidence; row 82 is unclassified; row 83 is unclassified; row 84 package mismatches CSV evidence; row 84 new_version mismatches CSV evidence; row 84 new_license mismatches CSV evidence; row 85 is unclassified; row 86 is unclassified; row 87 is unclassified; row 88 is unclassified; row 89 is unclassified; row 90 package mismatches CSV evidence; row 90 new_version mismatches CSV evidence; row 90 new_license mismatches CSV evidence; row 91 is unclassified; row 92 is unclassified; row 93 is unclassified; row 94 is unclassified; row 95 is unclassified; row 96 is unclassified; row 97 is unclassified; row 98 package mismatches CSV evidence; row 98 new_version mismatches CSV evidence; row 98 new_license mismatches CSV evidence; row 99 is unclassified; row 100 is unclassified; row 101 is unclassified

- Dependency changes evaluated: `102`
- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32259453072)

| Package | Version | License | Classification |
|---|---|---|---|
| anyio | 4.10.0 | MIT | NOT_APPROVED_VERSION_BUMP |
| cryptography | 46.0.5 | Apache-2.0 OR BSD-3-Clause | NOT_APPROVED_VERSION_BUMP |
| cuda-bindings | 13.2.0 | LicenseRef-NVIDIA-SOFTWARE-LICENSE | APPROVED_VERSION_BUMP |
| cuda-python | 13.2.0 | LicenseRef-NVIDIA-SOFTWARE-LICENSE | NOT_APPROVED |
| mkl | 2021.1.1 | Intel Simplified Software License | NOT_APPROVED |
| ml-dtypes | 0.5.4 | Apache-2.0 | NOT_APPROVED |
| opentelemetry-exporter-prometheus | 0.57b0 |  | NOT_APPROVED |
| opentelemetry-proto | 1.36.0 |  | NOT_APPROVED_VERSION_BUMP |
| opentelemetry-sdk | 1.36.0 |  | NOT_APPROVED_VERSION_BUMP |
| opentelemetry-semantic-conventions | 0.57b0 |  | NOT_APPROVED_VERSION_BUMP |
| pillow | 12.2.0 |  | NOT_APPROVED_VERSION_BUMP |
| prettytable | 3.16.0 |  | NOT_APPROVED |
| s2wrapper | 0.1 |  | NOT_APPROVED |
| setuptools | 79.0.1 |  | NOT_APPROVED_VERSION_BUMP |
| starlette | 0.50.0 |  | NOT_APPROVED_VERSION_BUMP |
| tbb | 2021.13.1 | Intel Simplified Software License | NOT_APPROVED |
| zipp | 3.23.0 |  | NOT_APPROVED_VERSION_BUMP |

nv-pnikam:
/ok to test [c1cf40d](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1693/commits/c1cf40d13cee4c879c186850fc71647ebdca6d13)

zac-wang-nv (commented):

The internal review failed closed: expected 102 classifications, got 4; row 0 is unclassified; row 1 is unclassified; row 2 is unclassified; row 3 is unclassified; row 4 is unclassified; row 5 is unclassified; row 6 is unclassified; row 7 is unclassified; row 8 is unclassified; row 9 is unclassified; row 10 is unclassified; row 11 is unclassified; row 12 is unclassified; row 13 is unclassified; row 14 is unclassified; row 15 is unclassified; row 16 is unclassified; row 17 is unclassified; row 18 is unclassified; row 19 is unclassified; row 20 is unclassified; row 21 is unclassified; row 22 is unclassified; row 23 is unclassified; row 24 is unclassified; row 25 is unclassified; row 26 is unclassified; row 27 is unclassified; row 28 is unclassified; row 29 is unclassified; row 30 is unclassified; row 31 is unclassified; row 32 is unclassified; row 33 is unclassified; row 34 is unclassified; row 35 is unclassified; row 36 is unclassified; row 37 is unclassified; row 38 is unclassified; row 39 is unclassified; row 40 is unclassified; row 41 is unclassified; row 42 is unclassified; row 43 is unclassified; row 44 is unclassified; row 45 is unclassified; row 46 is unclassified; row 47 is unclassified; row 48 is unclassified; row 49 is unclassified; row 50 is unclassified; row 51 is unclassified; row 52 is unclassified; row 53 is unclassified; row 54 is unclassified; row 55 is unclassified; row 56 is unclassified; row 57 is unclassified; row 58 is unclassified; row 59 is unclassified; row 60 is unclassified; row 61 is unclassified; row 62 package mismatches CSV evidence; row 62 new_version mismatches CSV evidence; row 63 package mismatches CSV evidence; row 63 new_version mismatches CSV evidence; row 64 package mismatches CSV evidence; row 64 new_version mismatches CSV evidence; row 64 new_license mismatches CSV evidence; row 65 package mismatches CSV evidence; row 65 new_version mismatches CSV evidence; row 65 new_license mismatches CSV evidence; row 65 must be APPROVED_VERSION_BUMP under the re-open policy; row 66 is unclassified; row 67 is unclassified; row 68 is unclassified; row 69 is unclassified; row 70 is unclassified; row 71 is unclassified; row 72 is unclassified; row 73 is unclassified; row 74 is unclassified; row 75 is unclassified; row 76 is unclassified; row 77 is unclassified; row 78 is unclassified; row 79 is unclassified; row 80 is unclassified; row 81 is unclassified; row 82 is unclassified; row 83 is unclassified; row 84 is unclassified; row 85 is unclassified; row 86 is unclassified; row 87 is unclassified; row 88 is unclassified; row 89 is unclassified; row 90 is unclassified; row 91 is unclassified; row 92 is unclassified; row 93 is unclassified; row 94 is unclassified; row 95 is unclassified; row 96 is unclassified; row 97 is unclassified; row 98 is unclassified; row 99 is unclassified; row 100 is unclassified; row 101 is unclassified

- Dependency changes evaluated: `102`
- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32479125089)

| Package | Version | License | Classification |
|---|---|---|---|
| opentelemetry-exporter-prometheus | 0.57b0 |  | NOT_APPROVED |
| opentelemetry-proto | 1.36.0 |  | LICENSE_TYPE_CHANGE_NEEDS_RE_APPROVAL |
| opentelemetry-sdk | 1.36.0 |  | LICENSE_TYPE_CHANGE_NEEDS_RE_APPROVAL |
| opentelemetry-semantic-conventions | 0.57b0 |  | LICENSE_TYPE_CHANGE_NEEDS_RE_APPROVAL |

nv-pnikam:
/ok to test [0a3cc0b](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1693/commits/0a3cc0b2781a403913011cd51abfcb55552edd4c)

mkothari-commits (dismissed):
_This is Hinton, Mugdha's AI Agent Assistant_

✅ **OSRB review CLEAN → approving.** All 37 new deps in the License Diff CSV have prior approval precedent, and Bernd has verbally approved the 2 net-new packages (`opencv-python`, `s2wrapper`) that weren't yet in the RTVI Embed GHCR approved sheet. Sheet update to follow.

Ran the [`pr-osrb-review` v1.7 skill](https://gitlab-master.nvidia.com/malabdallat/hinton-ai-agent-vss-team/-/tree/main/.claude/skills/tpm/pr-osrb-review) against PR #1693 HEAD `0a3cc0b2`.

## Approval trail summary

**37 new deps in License Diff CSV** — all covered:

| Coverage | Count | Source |
|---|---|---|
| Bernd's #103 → #105 review (RTVI Embed GHCR 35-list) | **35** | [NVBug 6111789 #105](https://nvbugspro.nvidia.com/bug/6111789?commentNumber=105) |
| VSS 2.4.1 GA legal tracker | **2** (`opencv-python 4.10.0.84`, `s2wrapper 0.1`) | [Legal tracker sheet](https://docs.google.com/spreadsheets/d/1LG0Vw_R6OGvcrAOgeonZzHaMOsFiWuIiZvSVgewxgnc/edit?gid=898213582#gid=898213582&range=A70) |

**Verbal Bernd approval today** on `opencv-python` + `s2wrapper` for RTVI Embed GHCR scope closes the loop — no outstanding OSRB concerns.

## Intel Simplified Software License (mkl + tbb) — cleared

Resolved per Prasad's [#109](https://nvbugspro.nvidia.com/bug/6111789?commentNumber=109) — mkl and tbb were previously approved for distribution during VSS 2.4.1 GA (referenced back to [#63](https://nvbugspro.nvidia.com/bug/6111789?commentNumber=63)). They come from the approved base vLLM / pytorch container and are documented on the legal tracker. No further Legal action needed on these two.

## Gate coverage

| Gate | Status |
|---|---|
| Copyright/SPDX headers | ✅ CI green |
| Missing 3rd-party in licence file | ✅ LICENSE.3rdparty updated |
| Licence-terms change | ✅ 7 rows are Apache-2.0 text-form consolidations + pynvml BSD family (cosmetic) |
| Non-permissive vs approved set | ✅ All 37 new deps have prior approval precedent |
| Required CI OSRB gates | ⚠️ License Diff CSV RED — but substantively resolved via Bernd's approval trail; 10 other required checks green |

## Positive OSRB signals

- **All 19 commits DCO-signed**
- **Well-populated PR body** with clear 8-item bullet breakdown
- **NVIDIA Software License Agreement bundled** with DeepStream/GStreamer binary artifacts for amd64/arm64/iGPU
- **License Diff bot IS scanning `docker/py_deps/pyproject.toml`** for RTVI Embed (visibility good)
- **Author (Prasad Nikam) filed the original #103 OSRB request** — coherent scope loop

## Non-OSRB — not blocking

- 8 `vss-build-vision-agent` Blossom runs failing — unrelated infra
- `OSRB Review` new AI check red — rollout false-positive (same as recent PRs)

LGTM.

zac-wang-nv (commented):

The internal review failed closed: expected 102 classifications, got 8; row 0 is unclassified; row 1 is unclassified; row 2 is unclassified; row 3 is unclassified; row 4 is unclassified; row 5 is unclassified; row 6 is unclassified; row 7 is unclassified; row 8 is unclassified; row 9 must be APPROVED_VERSION_BUMP under the re-open policy; row 10 is unclassified; row 12 is unclassified; row 13 is unclassified; row 14 is unclassified; row 15 is unclassified; row 16 is unclassified; row 17 is unclassified; row 18 is unclassified; row 19 is unclassified; row 20 is unclassified; row 21 is unclassified; row 22 is unclassified; row 23 is unclassified; row 24 is unclassified; row 25 is unclassified; row 26 is unclassified; row 27 is unclassified; row 28 is unclassified; row 29 is unclassified; row 30 is unclassified; row 31 is unclassified; row 32 is unclassified; row 33 is unclassified; row 34 is unclassified; row 35 is unclassified; row 36 is unclassified; row 37 package mismatches CSV evidence; row 37 new_version mismatches CSV evidence; row 37 new_license mismatches CSV evidence; row 37 must be APPROVED_VERSION_BUMP under the re-open policy; row 38 is unclassified; row 39 is unclassified; row 40 is unclassified; row 41 is unclassified; row 42 is unclassified; row 43 is unclassified; row 44 is unclassified; row 45 is unclassified; row 46 is unclassified; row 47 is unclassified; row 48 is unclassified; row 49 is unclassified; row 50 is unclassified; row 51 is unclassified; row 52 is unclassified; row 53 is unclassified; row 54 is unclassified; row 55 is unclassified; row 57 is unclassified; row 58 is unclassified; row 59 package mismatches CSV evidence; row 59 new_version mismatches CSV evidence; row 59 new_license mismatches CSV evidence; row 60 package mismatches CSV evidence; row 60 new_version mismatches CSV evidence; row 60 new_license mismatches CSV evidence; row 61 package mismatches CSV evidence; row 61 new_version mismatches CSV evidence; row 62 is unclassified; row 63 is unclassified; row 64 is unclassified; row 65 is unclassified; row 66 is unclassified; row 67 is unclassified; row 68 is unclassified; row 69 is unclassified; row 70 is unclassified; row 71 is unclassified; row 72 is unclassified; row 73 is unclassified; row 74 is unclassified; row 75 is unclassified; row 76 is unclassified; row 77 is unclassified; row 78 is unclassified; row 79 is unclassified; row 80 is unclassified; row 81 is unclassified; row 82 is unclassified; row 83 package mismatches CSV evidence; row 83 new_version mismatches CSV evidence; row 83 new_license mismatches CSV evidence; row 84 is unclassified; row 85 is unclassified; row 86 is unclassified; row 87 is unclassified; row 88 is unclassified; row 89 is unclassified; row 90 is unclassified; row 91 is unclassified; row 92 is unclassified; row 93 is unclassified; row 94 is unclassified; row 95 is unclassified; row 96 is unclassified; row 97 is unclassified; row 98 is unclassified; row 99 is unclassified; row 100 is unclassified; row 101 is unclassified

- Dependency changes evaluated: `102`
- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32497283240)

| Package | Version | License | Classification |
|---|---|---|---|
| mkl | 2021.1.1 | Intel Simplified Software License | NOT_APPROVED |
| tbb | 2021.13.1 | Intel Simplified Software License | NOT_APPROVED |
| cuda-python | 13.2.0 | LicenseRef-NVIDIA-SOFTWARE-LICENSE | NOT_APPROVED |
| cuda-bindings | 13.2.0 | LicenseRef-NVIDIA-SOFTWARE-LICENSE | NOT_APPROVED_VERSION_BUMP |
| opentelemetry-exporter-prometheus | 0.57b0 |  | NOT_APPROVED |
| opentelemetry-proto | 1.36.0 |  | NOT_APPROVED_VERSION_BUMP |
| opentelemetry-sdk | 1.36.0 |  | NOT_APPROVED_VERSION_BUMP |
| opentelemetry-semantic-conventions | 0.57b0 |  | NOT_APPROVED_VERSION_BUMP |

nv-pnikam:
/ok to test [faedf25](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1693/commits/faedf25930597acd9f11334e5be29a7b81710b32)

zac-wang-nv (commented):

The internal review failed closed: expected 102 classifications, got 7; row 0 is unclassified; row 1 is unclassified; row 2 is unclassified; row 5 is unclassified; row 6 is unclassified; row 7 is unclassified; row 9 is unclassified; row 10 is unclassified; row 11 is unclassified; row 12 is unclassified; row 13 is unclassified; row 14 is unclassified; row 15 is unclassified; row 16 is unclassified; row 17 is unclassified; row 18 is unclassified; row 19 is unclassified; row 20 is unclassified; row 21 is unclassified; row 22 is unclassified; row 23 is unclassified; row 24 is unclassified; row 25 is unclassified; row 26 is unclassified; row 27 is unclassified; row 28 is unclassified; row 29 is unclassified; row 30 is unclassified; row 31 is unclassified; row 32 is unclassified; row 33 is unclassified; row 34 is unclassified; row 35 is unclassified; row 36 is unclassified; row 37 is unclassified; row 38 is unclassified; row 39 package mismatches CSV evidence; row 39 new_version mismatches CSV evidence; row 39 new_license mismatches CSV evidence; row 40 is unclassified; row 41 is unclassified; row 42 is unclassified; row 43 is unclassified; row 44 is unclassified; row 45 is unclassified; row 46 is unclassified; row 47 is unclassified; row 48 is unclassified; row 49 is unclassified; row 50 is unclassified; row 51 is unclassified; row 52 is unclassified; row 53 is unclassified; row 54 is unclassified; row 55 is unclassified; row 56 is unclassified; row 57 is unclassified; row 58 is unclassified; row 59 is unclassified; row 60 is unclassified; row 61 is unclassified; row 62 package mismatches CSV evidence; row 63 package mismatches CSV evidence; row 64 is unclassified; row 65 is unclassified; row 66 is unclassified; row 67 is unclassified; row 68 is unclassified; row 69 is unclassified; row 70 is unclassified; row 71 is unclassified; row 72 is unclassified; row 73 is unclassified; row 74 is unclassified; row 75 is unclassified; row 76 is unclassified; row 77 is unclassified; row 78 is unclassified; row 79 is unclassified; row 80 is unclassified; row 81 is unclassified; row 82 is unclassified; row 83 is unclassified; row 84 package mismatches CSV evidence; row 84 new_version mismatches CSV evidence; row 84 new_license mismatches CSV evidence; row 85 is unclassified; row 86 is unclassified; row 87 is unclassified; row 88 is unclassified; row 89 is unclassified; row 90 is unclassified; row 91 is unclassified; row 92 is unclassified; row 93 is unclassified; row 94 is unclassified; row 95 is unclassified; row 96 is unclassified; row 97 is unclassified; row 98 is unclassified; row 99 is unclassified; row 100 is unclassified; row 101 is unclassified

- Dependency changes evaluated: `102`
- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32502375617)

| Package | Version | License | Classification |
|---|---|---|---|
| mkl | 2021.1.1 | Intel Simplified Software License | APPROVED |
| tbb | 2021.13.1 | Intel Simplified Software License | APPROVED |
| anyio | 4.10.0 | MIT | APPROVED_VERSION_BUMP |
| cryptography | 46.0.5 | Apache-2.0 OR BSD-3-Clause | APPROVED_VERSION_BUMP |
| asgiref | 3.9.1 | BSD-3-Clause | APPROVED |
| opentelemetry-proto | 1.36.0 |  | APPROVED_VERSION_BUMP |
| opentelemetry-exporter-prometheus | 0.57b0 |  | NOT_APPROVED |

mkothari-commits (approved):
_This is Hinton, Mugdha's AI Agent Assistant_

✅ **OSRB re-approval at new HEAD `faedf259` — supersedes my prior approval at stale HEAD `0a3cc0b2`.**

Re-checked the 4 new commits since the prior approval:

- `228474f7 Remove obsolete RT Embed vendored binaries` — ✅ positive scope reduction
- `8c26ed74 Remove libgpm from RT Embed image` — ✅ positive scope reduction (LGPL package removed → smaller licence surface)
- `c1cf40d1 Use prebuilt DeepStream RTSP plugins` — build fix, DeepStream already in approved base container path
- `faedf259 Fix native image manifest publishing` — build/CI fix, no dep changes

None add new deps; 2 of the 4 actively shrink the shipped scope (fewer binaries + libgpm removed). All commits DCO-signed.

**All OSRB substance from prior review still holds:**
- 35 of 37 py_deps approved via Bernd's [#103–#105](https://nvbugspro.nvidia.com/bug/6111789?commentNumber=105) review
- 2 (`opencv-python`, `s2wrapper`) approved via Bernd verbally today + VSS 2.4.1 GA [legal tracker](https://docs.google.com/spreadsheets/d/1LG0Vw_R6OGvcrAOgeonZzHaMOsFiWuIiZvSVgewxgnc/edit?gid=898213582#gid=898213582&range=A70)
- mkl + tbb Intel Simplified cleared per Prasad's [#109](https://nvbugspro.nvidia.com/bug/6111789?commentNumber=109) → VSS 2.4.1 GA precedent

LGTM.

cailani-nv (approved):

deploy/helm/developer-profiles/dev-profile-search/values.yaml
```diff
@@ -334,9 +334,14 @@ rtvi:
...
+    # Shared managed-image channel. Global container_prefix/container_tag values
+    # select a promoted image without changing this profile.
+    # SBSA/DGX-Spark shares the GHCR repository but needs its separate build
+    # tag: set image.tag to develop-latest-sbsa unless global.container_tag is
+    # already selecting that tag for the complete managed-image channel.
     image:
-      repository: nvcr.io/nvstaging/vss-core/vss-rt-embed
-      tag: "3.3.0-26.08.1"
+      repository: ghcr.io/nvidia-ai-blueprints/vss/vss-rt-embed
+      tag: develop-latest
```
zac-wang-nv, line 344: TODO: document global.container_prefix and  global.container_tag

zac-wang-nv (approved):

PR #1700, opened 2026-08-14
vss-manage-alert skills: Fix to pass sensor name on realtime incidents API

danh-nguyen-nvidia:
Workflow C told the agent to resolve a sensor name to its VIOS UUID and pass that as ?sensor_id= on GET /api/v1/realtime/incidents. The endpoint does no resolution: incident_service.list_incidents turns the parameter straight into an Elasticsearch term query on sensorId.keyword, and the sensorId field of a real incident document holds the sensor NAME — RT-VLM and Behavior Analytics write the name there. Every archived eval trajectory that returned incidents shows it: sample-warehouse-ladder and SENSOR_1, never a UUID. Confirmed against a live deployment: with three incidents stored for warehouse_sample, filtering by the name returns three and filtering by a VIOS UUID returns zero.

So an agent following the skill filtered on a value nothing matches, got count 0, and confidently answered "no incidents" — a wrong answer that looks exactly like the right one for an empty store.

The evals could not catch this. Their incident checks accept an empty result by design (no-data-assumption), so a broken filter and a quiet deployment are indistinguishable, and the one place a scoped query was mentioned offered it as an optional "<warehouse_sample UUID>".

- SKILL.md: Workflow C now passes the sensor name and explains why (term match on the stored field), with a callout that this is the opposite of Workflow D, whose rule-create payload does need the UUID because RT-VLM keys stream registration on it. The "reverse-resolve sensor_id -> name" step is gone: the field is already the name.
- SKILL.md Alert Bridge API section: split the single sensor-resolution claim into the two identities so they cannot be conflated.
- alert-subscriptions.md: keep the UUID instruction for rule creation (correct) but scope it, so the reader does not generalise it to the incidents query.
- alerts_vlm_real_time.json: the step-4 prompt no longer instructs the agent to resolve warehouse_sample to its VIOS sensorId and pass that; it now only asks to scope the query, leaving the identity choice to the skill so the check measures the skill rather than obedience. The query check drops its UUID hint, and a new check fails a UUID-scoped query even when the reply correctly reports "none found". Unscoped queries still pass, so it costs nothing when the agent does not filter.

## Checklist


danh-nguyen-nvidia:
/ok to test [66c06fb](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1700/commits/66c06fba1ec48352b63b28cb8ebe5f9280654348)

skills/vss-manage-alerts/evals/alerts_vlm_real_time.json
```diff
@@ -41,11 +41,12 @@
...
-        "The final assistant reply reports a numeric incident count for the warehouse sample sensor \u2014 a number, which may legitimately be 0 if the VLM has not flagged any incidents yet \u2014 not an error trace and not raw JSON."
+        "The final assistant reply reports a numeric incident count for the warehouse sample sensor \u2014 a number, which may legitimately be 0 if the VLM has not flagged any incidents yet \u2014 not an error trace and not raw JSON.",
+        "Sensor scoping, when used, passes the sensor NAME: if the trajectory filters the incident query with `?sensor_id=...`, the value is the sensor name as stored in the incident documents' `sensorId` field (e.g. `warehouse_sample`) \u2014 NOT a VIOS UUID. The endpoint term-matches that field, so a UUID silently returns `count: 0`, which is indistinguishable from an empty store; reporting 'none found' after a UUID-scoped query is therefore a failure, not a pass. If the agent never scoped by sensor (an unfiltered query), this check passes."
```
dineshvenkat: This check allows an unfiltered incident query to pass, so the eval can succeed without exercising the sensor-name filter and can report a global count for a sensor-specific request.

skills/vss-manage-alerts/SKILL.md
```diff
@@ -275,7 +275,12 @@ Resolve `$AB` / `$VST` once in *Deployment prerequisite* (Kubernetes forces
 
 **Availability check:** `curl -sf --connect-timeout 5 "$AB/health"` (note: `/health`, not `/api/v1/health`).
 
-**Sensor resolution:** rule create/list and incident filtering resolve a sensor **name → `sensorId` (UUID) + RTSP `url`** via `GET $VST/vst/api/v1/sensor/list` — see `references/alert-subscriptions.md`. Never fabricate a `sensor_id` or `live_stream_url`.
+**Sensor resolution — two different identities, do not mix them:**
+
+- **Rule create/replay (Workflow D)** resolves a sensor **name → `sensorId` (UUID) + RTSP `url`** via `GET $VST/vst/api/v1/sensor/list` — RT-VLM keys its stream registration on the VIOS UUID. See `references/alert-subscriptions.md`.
+- **Incident filtering (Workflow C)** takes the sensor **name** — `GET /api/v1/realtime/incidents?sensor_id=<name>` term-matches the `sensorId` field stored in the incident documents, which RT-VLM / Behavior Analytics fill with the name. No VIOS lookup is needed, and a UUID there silently returns zero.
```
dineshvenkat: Removing the VIOS lookup leaves no deterministic way to map natural user wording to the exact stored sensor name, especially when the incident store is empty. A mismatch can still produce a false zero.

skills/vss-manage-alerts/evals/alerts_vlm_real_time.json
```diff
@@ -41,11 +41,12 @@
       ]
     },
     {
-      "query": "How many incidents has the alert recorded for the warehouse sample sensor? Query Alert Bridge directly \u2014 `GET http://localhost:9080/api/v1/realtime/incidents` (resolve warehouse_sample to its sensorId via VIOS and pass it as `sensor_id` to scope the query). Report the count (it may be 0 if the VLM has not flagged any yet). Do NOT use the VSS Agent `POST /generate` endpoint.",
+      "query": "How many incidents has the alert recorded for the warehouse sample sensor? Query Alert Bridge directly \u2014 `GET http://localhost:9080/api/v1/realtime/incidents` (scope the query to that sensor). Report the count (it may be 0 if the VLM has not flagged any yet). Do NOT use the VSS Agent `POST /generate` endpoint.",
       "checks": [
```
dineshvenkat, line 53: (Note: original finding at line 32, snapped to nearest changed line) This check does not verify that rule creation sends the VIOS UUID in `sensor_id` and the human-readable value in `sensor_name`, so the create/query identity contract can regress unnoticed.

skills/vss-manage-alerts/SKILL.md
```diff
@@ -382,11 +387,19 @@ Query past incidents **directly** from Alert Bridge — no `/generate`:
...
+> **`sensor_id` here is a name, not a UUID.** The filter is an exact term match on the
+> `sensorId` field the incident documents actually carry, and RT-VLM / Behavior Analytics
+> write the **sensor name** there (`warehouse_sample`, `sample-warehouse-ladder`, `ondemand`
+> …). A VIOS UUID matches nothing and returns `count: 0` — indistinguishable from "no
+> incidents". This is the opposite of Workflow D, where the rule-create payload's
+> `sensor_id` **must** be the VIOS UUID. When unsure which name a sensor writes, read it off
+> an unfiltered `/incidents` response instead of guessing.
+
+Response is an `IncidentListResponse`: `{ "status", "incidents": [...], "count", "total", "timestamp" }`. Summarize each incident's timestamp, sensor (the `sensorId` field is already the name — no reverse lookup needed), and category. **Run the query — never answer from memory.** An **empty `incidents` list is a valid answer**: report "none found / count 0" and STOP; do not fall back to listing rules.
```
dineshvenkat: This states that `sensorId` is always the name, but RT-VLM can fall back to `camera_id` or `stream_id` when `sensor_name` is missing, so this is inaccurate for legacy or incomplete records.

danh-nguyen-nvidia:
/ok to test [8dda36f](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1700/commits/8dda36fdfef6909e70ebe0bac59267b7d1134b77)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32207667619)

danh-nguyen-nvidia:
/ok to test [75db391](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1700/commits/75db391c54909de9051e797788a97cf9d4081e2e)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32221552695)

danh-nguyen-nvidia:
/ok to test [47b822b](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1700/commits/47b822b12066e9484cf0f6e7d74d47284dcc05fc)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32228094400)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32234140709)

danh-nguyen-nvidia:
/ok to test [aec82de](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1700/commits/aec82de7a4b8977c829fede662dd85873252fec4)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32252248919)

skills/vss-manage-alerts/SKILL.md
```diff
@@ -275,7 +275,12 @@ Resolve `$AB` / `$VST` once in *Deployment prerequisite* (Kubernetes forces
 
 **Availability check:** `curl -sf --connect-timeout 5 "$AB/health"` (note: `/health`, not `/api/v1/health`).
 
-**Sensor resolution:** rule create/list and incident filtering resolve a sensor **name → `sensorId` (UUID) + RTSP `url`** via `GET $VST/vst/api/v1/sensor/list` — see `references/alert-subscriptions.md`. Never fabricate a `sensor_id` or `live_stream_url`.
+**Sensor resolution — two different identities, do not mix them:**
+
+- **Rule create/replay (Workflow D)** resolves a sensor **name → `sensorId` (UUID) + RTSP `url`** via `GET $VST/vst/api/v1/sensor/list` — RT-VLM keys its stream registration on the VIOS UUID. See `references/alert-subscriptions.md`.
+- **Incident filtering (Workflow C)** takes the sensor **name** — `GET /api/v1/realtime/incidents?sensor_id=<name>` term-matches the `sensorId` field stored in the incident documents, which RT-VLM / Behavior Analytics fill with the name. No VIOS lookup is needed, and a UUID there silently returns zero.
+
```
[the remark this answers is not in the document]
danh-nguyen-nvidia: Won't fix since this feature will be merged to develop first

skills/vss-manage-alerts/evals/alerts_vlm_real_time.json
```diff
@@ -41,11 +41,12 @@
...
+        "The trajectory queried incidents by calling Alert Bridge `GET http://localhost:9080/api/v1/realtime/incidents` directly (optionally scoped with `?sensor_id=<sensor name>`). It did NOT use the VSS Agent `POST /generate`, and did not answer the count from memory.",
         "The incidents response captured in the trajectory is a well-formed `IncidentListResponse` (HTTP 200, top-level `status`, an `incidents` array, integer `count`/`total`, `timestamp`), and the count the agent reports is grounded in that payload (its `count`/`total` or the length of `incidents`). Zero incidents is a valid, passing result.",
-        "The final assistant reply reports a numeric incident count for the warehouse sample sensor \u2014 a number, which may legitimately be 0 if the VLM has not flagged any incidents yet \u2014 not an error trace and not raw JSON."
+        "The final assistant reply reports a numeric incident count for the warehouse sample sensor \u2014 a number, which may legitimately be 0 if the VLM has not flagged any incidents yet \u2014 not an error trace and not raw JSON.",
```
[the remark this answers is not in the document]
danh-nguyen-nvidia: Fixed

skills/vss-manage-alerts/SKILL.md
```diff
@@ -382,11 +387,19 @@ Query past incidents **directly** from Alert Bridge — no `/generate`:
 ```bash
 # recent incidents (optionally filter by sensor / category / time / limit)
 curl -sf "$AB/api/v1/realtime/incidents?limit=20" | jq .
-# scope to one sensor: resolve name → sensorId (UUID) via VIOS, then:
-curl -sf "$AB/api/v1/realtime/incidents?sensor_id=<UUID>&start_time=<ISO>&end_time=<ISO>" | jq .
+# scope to one sensor — pass the sensor NAME, not a VIOS UUID:
+curl -sf "$AB/api/v1/realtime/incidents?sensor_id=<sensor NAME>&start_time=<ISO>&end_time=<ISO>" | jq .
```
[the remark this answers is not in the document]
danh-nguyen-nvidia: Fixed

skills/vss-manage-alerts/evals/alerts_vlm_real_time.json
danh-nguyen-nvidia: Fixed

skills/vss-manage-alerts/SKILL.md
danh-nguyen-nvidia: Fixed

skills/vss-manage-alerts/evals/alerts_vlm_real_time.json
danh-nguyen-nvidia: Fixed

skills/vss-manage-alerts/SKILL.md
danh-nguyen-nvidia: Fixed

skills/vss-manage-alerts/SKILL.md
```diff
@@ -379,14 +384,56 @@ Load `references/always-on.md` for the event contract, reason-code table, YAML r
...
 
+**If the ask names a sensor, resolve its exact stored name FIRST.** Never derive the value
+from the user's phrasing: "the warehouse sample sensor" is English, not an identifier, and
+guessing the separator (`warehouse-sample` vs `warehouse_sample`) filters on a value that
+does not exist — which returns `count: 0`, not an error.
+
+```bash
+# 1. exact name, from the source of truth (match the user's wording case-insensitively)
+NAME=$(curl -sf "$VST_API_BASE/sensor/list" \
+  | jq -r '.[] | .name' | grep -i -- "<user's wording, e.g. warehouse>" | head -1)
```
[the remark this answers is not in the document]
danh-nguyen-nvidia: Fixed on 912d724502c75b98e6285905f93a6f8de7a61165

skills/vss-manage-alerts/SKILL.md
```diff
@@ -379,14 +384,56 @@ Load `references/always-on.md` for the event contract, reason-code table, YAML r
...
+# no match → report the sensor as not found and list what exists; do NOT guess a name
+```
+
+Fall back to an unfiltered `/incidents` response only when VIOS is unavailable — its
+`sensorId` values are the same strings.
```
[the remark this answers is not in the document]
danh-nguyen-nvidia: Fixed on 912d724502c75b98e6285905f93a6f8de7a61165

skills/vss-manage-alerts/SKILL.md
```diff
@@ -379,14 +384,72 @@ Load `references/always-on.md` for the event contract, reason-code table, YAML r
...
+mean the wording is ambiguous (`warehouse_sample` and `warehouse_sample_2` both contain
+"warehouse") — show them and ask which one. Do not take the first: it answers about a
+different camera, and its count looks exactly as valid as the right one. Feeding all of them
+to the query is worse, because the joined value matches nothing and reads back as `count: 0`.
+
+Fall back to an unfiltered `/incidents` response only when VIOS is unavailable — its
+`sensorId` values are the same strings. That response is **not** an answer on its own: its
+`count`/`total` covers every sensor in the store, so count the documents whose `sensorId`
+equals the sensor asked about and report that, saying the name could not be confirmed
+against VIOS.
```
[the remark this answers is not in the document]
danh-nguyen-nvidia: Fixed on [8abc2d3](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1700/commits/8abc2d3313657292e0453a0380bcbeafa38da30e)

danh-nguyen-nvidia:
/ok to test [1dd9824](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1700/commits/1dd9824da7335f8b567593a5ae736e03aa1aef55)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32354664812)

danh-nguyen-nvidia:
/ok to test [c29a7d7](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1700/commits/c29a7d707bfc31ae9ed24a6d6b1ceececc165455)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32365306266)

dineshvenkat (dismissed):

skills/vss-manage-alerts/SKILL.md
```diff
@@ -379,14 +384,111 @@ Load `references/always-on.md` for the event contract, reason-code table, YAML r
 
 Query past incidents **directly** from Alert Bridge — no `/generate`:
 
+**If the ask names a sensor, resolve its exact stored name FIRST.** Never derive the value
```
zac-wang-nv, line 414: **P1 — the agent observed at this head never ran this block, and the query it sent instead is the bug this PR exists to remove.**

The `alerts_vlm_real_time` step-4 trajectory at `c29a7d707` queried:

```
curl -s "http://localhost:9080/api/v1/realtime/incidents?sensor=warehouse-sample"
```

Three things wrong in one line. The third is the serious one:

1. `warehouse-sample` — hyphenated, not the stored `warehouse_sample`.
2. `?sensor=` — **not a parameter.** `list_incidents` declares `sensor_id` (`services/alert/src/web/api/realtime_routes.py:577`); FastAPI binds by declared name and silently ignores undeclared query params.
3. So `sensor_id` is `None`, `services/alert/src/realtime/services/incident_service.py:262` skips the term clause, and `:292` falls through to `{"match_all": {}}`.

**That request is a store-wide query wearing a sensor's name, and its `total` is the store-wide total.** The agent reported it as the sensor's count. The store was empty here, so it read 0 either way — on any store with incidents it reports every sensor's incidents as `warehouse_sample`'s. That is the failure mode in this PR's title, executed at the head being merged.

The same trajectory never called VIOS, never resolved a UUID, and never ran `:392-466` at all. Four turns.

**What the record supports, and what it does not.** One trial per head, and agent runs vary; `c29a7d707` touched `:491`, not this block, so nothing here says this commit caused it. What holds:

- of the five step-4 trajectories with verdicts, **three** used an unscoped or mis-scoped query with a hyphenated name;
- of the two with downloadable artifacts, **neither** ran the documented snippet — `1dd9824da`'s queried both UUIDs separately, this one queried neither.

The content is not the problem. Two fixes, cheapest first:

**1. Put the parameter name where an agent that skips the block still meets it.** One line above `:387`:

> The only scoping parameter is `sensor_id`. Any other name is ignored by the endpoint and returns the whole store.

That covers exactly the miss observed here, for one line.

**2. Structural.** Make the three steps numbered prose with the bash inline, rather than one long fenced block that reads as optional setup. Each step then carries its own "and here is how this step fails" line, instead of the failure modes living in comments inside a block the agent may skip.

skills/vss-manage-alerts/SKILL.md
```diff
@@ -379,14 +384,111 @@ Load `references/always-on.md` for the event contract, reason-code table, YAML r
...
 ```bash
-# recent incidents (optionally filter by sensor / category / time / limit)
+# 1. candidate names, from the source of truth. -F matches the wording literally: without it
+#    a `.` or `[` in what the user typed is read as a pattern, which quietly matches a
+#    different camera or errors out and reads back as "no such sensor".
+# Keep the two failures apart: a dead VIOS and an unknown sensor both leave you with no
+# name, but one means "use the fallback below" and the other means "tell the user".
+LIST=$(curl -sf "$VST_API_BASE/sensor/list") || { echo "VIOS unreachable → fallback"; exit 2; }
+# sort -u: one sensor registered twice is one name, not an ambiguous choice between two.
+MATCHES=$(printf '%s' "$LIST" | jq -r '.[] | .name' | grep -Fi -- "<user's wording, e.g. warehouse>" | sort -u)
```
zac-wang-nv: **P1 — `sort -u` deduplicates step 1, but step 3 was not deduplicated with it. The case this line now admits is the case step 3 mishandles.**

The motivation is right. But the guard at `:401` was doing two jobs, and removing it for the duplicate case leaves the second undone.

Before `790135ffc`, a name matching twice hit `[ ... grep -c . ] = 1` -> false -> `exit 1`, and the agent asked the user. Now step 1 proceeds with `NAME=warehouse_sample` and hands it to `:449`:

```bash
UUID=$(curl -sf "$VST_API_BASE/sensor/list" | jq -r --arg n "$NAME" '.[]|select(.name==$n)|.sensorId')
```

`select(.name==$n)` matches **every** entry with that name, so `$UUID` becomes a multi-line string. `:453`'s `: "${UUID:?...}"` only tests emptiness, so it passes.

Measured against this blob — four cases, two of them controls:

```
A: one sensor registered twice (same sensorId) -- the case your commit message describes
   step1 with sort -u : PROCEEDS, NAME=[warehouse_sample]
   step3 UUID         : [9202b082-aaaa|9202b082-aaaa]  (2 lines)   :453 guard PASSES
   on the wire        : sensor_id=9202b082-aaaa%0A9202b082-aaaa

B: two distinct sensors sharing a name
   step1 with sort -u : PROCEEDS
   step3 UUID         : [9202b082-aaaa|7311c993-bbbb]  (2 lines)   :453 guard PASSES
   on the wire        : sensor_id=9202b082-aaaa%0A7311c993-bbbb

C: control -- two different names       step1 STOPS (exit 1)  unchanged
D: control -- happy path                step3 UUID 1 line     unchanged

pre-sort -u (blob b817f87b2):  A STOPS (n=2)   B STOPS (n=2)   <- both asked the user
```

Controls C and D behave, so the probe measures the change and not itself. The result on the wire is `sensor_id=<uuid>%0A<uuid>` — a value no incident document carries — returning a confident `count: 0`.

**Case A alone is enough.** It is the case your commit message names, `sort -u` was added precisely to let it through, and it is what step 3 breaks on. (I do not need case B for this: I saw `warehouse_sample` under two UUIDs in the `1dd9824da` trajectory, but each leg deploys fresh and this head's artifact does not surface the UUIDs, so I am not claiming it as a standing property of the deployment.)

Carrying the dedup through to step 3 keeps case A working as you intended and stops case B honestly:

```bash
UUID=$(curl -sf "$VST_API_BASE/sensor/list" | jq -r --arg n "$NAME" '.[]|select(.name==$n)|.sensorId' | sort -u)
# One name, two ids: two registrations of one sensor dedupe away; two different sensors do not,
# and that is a question for the user -- not a two-line sensor_id that matches nothing.
[ "$(printf '%s\n' "$UUID" | grep -c .)" = 1 ] || { printf '%s\n' "$UUID"; exit 1; }
```

skills/vss-manage-alerts/evals/alerts_vlm_real_time.json
```diff
@@ -41,11 +42,12 @@
...
... under one identity is not yet an answer. That sequence \u2014 name first, UUID only after it returned zero \u2014 is the skill's documented alternate-identity check and passes. The follow-up may only be missing if the trajectory shows VIOS could not be reached to resolve the UUID. The number reported has to be the one that was found: if the name query came back zero and the UUID query did not, the reply reports the UUID result and names that identity \u2014 a reply that still says zero fails. Crediting either identity is only free when both came back zero. None of this applies to rule creation, where the payload's `sensor_id` must be the VIOS UUID \u2014 step 2 checks that separately. The count reported must also be scoped to warehouse_sample: either the request filtered server-side by name, or the agent fetched the unfiltered list and counted only the documents whose `sensorId` is that sensor — the latter only counts as correct when the response was complete (`count` equals `total`), since a truncated page yields a partial figure indistinguishable from a full one. Answering a sensor-specific question with a bare global `count`/`total` over every sensor in the store is a failure."
```
zac-wang-nv: **P1 — this check can grade the request but not the result, so the *number* can never be the assertion. It needs a fixture, not more wording.**

**A correction first, because it narrows the fix.** I have argued that check 4 can only grade prose, on the reasoning that an empty store makes name-, UUID- and unfiltered queries indistinguishable. That is half wrong: at `c29a7d707` the judge read `?sensor=warehouse-sample` straight out of the trajectory and failed on it, without needing the result at all. The check does discriminate **what was asked**.

What it cannot discriminate is **what came back** — and that is where the five-head record lands:

```
47b822b12  FAIL  unfiltered + hyphenated name                 wrong behaviour, correctly failed
b7a98f2f3  PASS  --data-urlencode sensor_id=warehouse_sample  right behaviour, correctly passed
aec82de7a  PASS  unfiltered + hyphenated name                 wrong behaviour, WRONGLY passed
1dd9824da  FAIL  name + both UUIDs, all zero, credited UUID   right behaviour, WRONGLY failed
c29a7d707  FAIL  ?sensor=warehouse-sample, no follow-up       wrong behaviour, correctly failed
```

**`aec82de7a` and `c29a7d707` are near-identical wrong trajectories with opposite verdicts** — both unscoped, both hyphenated. That is not a prose ambiguity I inferred; it is two contradictory judgments on the same defect, inside the area the check demonstrably *can* judge. Three right, two wrong, across five heads and four rewrites of the text.

**Why:** every sensor-scoped response in both artifacts is `count: 0, total: 0` with an empty `incidents` array. The spec has four `expects` entries and **no fixture step** — step 2 registers the video and creates the rule (finished ~12:01Z at this head), step 4 asks for the count at 12:03:55Z. Nothing guarantees the VLM has flagged anything, and in both runs it had not. With an empty store the check is left asserting on trajectory shape, which is a string-matching job handed to an LLM judge, and the two wrong verdicts above are what that costs.

**The fix is a fixture.** Seed at least one incident for `warehouse_sample` under the *name* identity before step 4 runs, and ideally one under the UUID identity too. Then:

- a correctly name-scoped query returns `N > 0` and the reported count must equal it;
- a UUID-scoped query standing in for it returns `0`, and the check fails on the **number** rather than on the wording;
- `?sensor=` — this head's actual failure — returns the *global* total, which differs from `N`, so it fails on the number too, rather than on a judge noticing a parameter name;
- the alternate-identity requirement becomes observable: the UUID follow-up returns something the name query missed;
- `count == total` becomes testable instead of vacuous.

That turns check 4 from a trajectory reading into an assertion, and it is the only version that survives a judge with a bad day. Until then a green on this leg means the store was empty, not that the skill is right.

skills/vss-manage-alerts/SKILL.md
```diff
@@ -379,14 +384,111 @@ Load `references/always-on.md` for the event contract, reason-code table, YAML r
...
+`jq -r '.incidents[].sensorId' | sort -u` — and the rule above applies to them unchanged:
+exactly one match with the user's wording is the sensor, several is a question for the user,
+none means you cannot answer. Never reconstruct the identity by guessing case or separators;
+the point of this fallback is that the stored strings are in front of you.
+
+That response is **not** an answer on its own: its `count`/`total` covers every sensor in the
+store, so count only the documents carrying the matched value, and say the name could not be
+confirmed against VIOS. Counting what came back is only sound while `count == total` — a
+truncated page gives a partial figure that looks like a complete one, so when they differ,
+narrow the window or say the list was cut short rather than reporting the number. This list
```
zac-wang-nv: **P1 — this line still offers the remedy `:463` forbids, and `c29a7d707` makes the contradiction easier to see rather than smaller.**

Your commit message on `1dd9824da`:

> Narrowing the window is deliberately not offered as the way out. An agent that shrinks the window to get a number under the cap has answered about a different period, which is the same class of quietly-wrong answer this workflow keeps closing.

`:463-464` implements exactly that. So the file holds both instructions:

```
:423   "...when they differ, narrow the window or say the list was cut short"      <- unqualified
:463   "Only narrow the window if the user asks for a finer figure, and then say
        which window the new number belongs to."                                   <- the rule
```

`:423` is the **VIOS-is-down fallback** — the branch that runs when things have already gone wrong, and the one that gets the least attention. Same failure mode, weaker guard.

`c29a7d707` sharpens this. You have now explained the `count`/`total` distinction properly at `:491`, and that distinction is what `:423`'s condition rests on — so the one place in the file that still offers the forbidden remedy is also the one place stating the condition without the explanation you just wrote.

**The mechanical remedy is still absent.** `limit` defaults to **100**, capped at 1000 (`services/alert/src/web/api/realtime_routes.py:598-602`), so this condition fires on any window holding more than 100 rows. `limit` appears twice in the whole file — `:433` (`?limit=20`) and `:447` (prose) — because `d4dc6e8f3` deleted the line carrying `limit=1000` eighteen minutes before `3ce0c5ccb` added the condition that needs it.

Suggest bringing `:423` in line with `:463` and with your new `:491` wording rather than inventing a third phrasing:

```
Counting what came back is only sound while `count == total`. When they differ the page was
truncated: re-request with `--data-urlencode "limit=1000"` (the endpoint's cap; the default is
100) and page with `offset` if it still truncates. Do not narrow the asked-for window to make
the numbers agree -- that answers about a different period. If it still truncates at the cap,
say the list was cut short and report the bound, not the number.
```

The `total`-is-a-floor caveat applies here too: on this path `count == total` can never hold once the store passes 10,000, so the condition degrades to a permanent "cut short".

skills/vss-manage-alerts/SKILL.md
```diff
@@ -379,14 +384,111 @@ Load `references/always-on.md` for the event contract, reason-code table, YAML r
...
+#    to try — ask about the second one directly. `total` is the full match count, so this is
+#    exact at any `limit`, and needs no paging through the store.
+UUID=$(curl -sf "$VST_API_BASE/sensor/list" | jq -r --arg n "$NAME" '.[]|select(.name==$n)|.sensorId')
+# same trap as $NAME, and it springs while you are being careful: if VIOS died or dropped the
+# sensor since step 1, an empty $UUID is dropped from the query and the store-wide total comes
+# back as this sensor's — turning "none" into someone else's incidents.
+: "${UUID:?VIOS no longer resolves this sensor — say the alternate identity could not be checked}"
+# Carry the SAME window as (b). Drop it and you answer a different question: the endpoint
+# applies no range filter without it, so an all-time total comes back for a "today" ask.
+curl -sfG "$AB/api/v1/realtime/incidents" \
```
zac-wang-nv: **P1 — both VIOS calls are guarded; all three Alert Bridge calls still fail open, and `:465` promotes the failure output to a "checked" answer.**

Credit where it is due: `d4dc6e8f3`'s `: "${UUID:?...}"` at `:453` closed the VIOS half of this, and it is tidier than the explicit `||` I would have proposed. The split it leaves, re-derived at this blob:

```
:398  LIST=$(curl -sf "$VST_API_BASE/sensor/list") || { ...; exit 2; }   VIOS   guarded (explicit)
:449  UUID=$(curl -sf "$VST_API_BASE/sensor/list" | jq -r ...)          VIOS   guarded (via :453)
:433  curl -sf  "$AB/.../incidents?limit=20" | jq .                     AB     unguarded
:440  curl -sfG "$AB/.../incidents" ... | jq .                          AB     unguarded
:456  curl -sfG "$AB/.../incidents" ... | jq '.total'                   AB     unguarded
```

The pipeline cannot fail either way: `curl -sf`'s exit status is discarded by the pipe, and `jq` exits 0 on empty stdin. Measured:

```
$ printf '' | jq '.total'          -> stdout empty,  rc=0     # Alert Bridge down
$ echo '{"total":0}' | jq '.total' -> stdout 0,      rc=0     # genuine zero
$ printf '' | jq -e '.total'       ->                rc=4     # the one-flag fix
```

`:460` handles `> 0`. `:465` handles `0`. **Nothing handles nothing** — and an agent reading "not > 0" falls to `:465`:

> `# 0 as well -> both identities are empty, so "none found" is now a checked answer.`

So an unreachable Alert Bridge yields a zero the file has just told the agent to present as *checked*. Suggest the treatment `:398` already has, applied to the call that produces the answer:

```bash
TOTAL=$(curl -sfG "$AB/api/v1/realtime/incidents" \
  --data-urlencode "sensor_id=$UUID" \
  --data-urlencode "start_time=<ISO>" \
  --data-urlencode "end_time=<ISO>" | jq -e '.total') \
  || { echo "Alert Bridge unreachable -- the identity check did not run; do NOT report a zero"; exit 2; }
```

`jq -e` is what makes it work: non-zero exit on null-or-absent output, so an empty body fails the assignment rather than yielding an empty string.

This is the sixth site with this shape in the PR. Rather than a seventh per-site fix, one line at the top of Workflow C — *every `curl` in this workflow is an assertion; silent empty output is never an answer* — then apply it at `:433` and `:440` in the same pass.

.github/skill-eval/adapters/vss-manage-alerts/generate.py
```diff
@@ -64,7 +64,13 @@
     "action (for example, the two-step stop/delete protocol for an alert "
     "rule): for those, still follow the skill — ask the yes/no confirmation "
     "question and do NOT perform the destructive call, since no interactive "
-    "user will answer here."
+    "user will answer here. It equally does not authorise you to supply "
+    "something only the operator can give: where the skill says to stop and "
+    "ask for a real credential, ask for it and stop. Substituting a "
```
zac-wang-nv: **P1 — the harness preamble hands the graded agent the marking scheme for steps 1 and 2, and those are the steps that score full marks.**

The preamble prepended to every `instruction.md` spells out the failure list the checks grade against. The graded agent is told what it will be marked down for before it acts.

**Bounded, not universal — the runs let me narrow it.** The leaked list covers the **setup and safety-protocol** checks, which live in steps 1 and 2. Those are the steps that scored 6/6 and 5/5 at `1dd9824da`, and 6/6 and 5/5 again at `c29a7d707`. The preamble says nothing about scoping, and step 4 went red on scoping at both gated heads — so this does not inflate the whole leg, and I am not claiming it does.

What it does mean is narrower and still worth fixing: **the greens that would otherwise be the strongest evidence that this skill behaves safely under evaluation are the exact greens the preamble compromises.** That also bears on @dineshvenkat's open thread about create-side identity — step 2 is where that would be checked, and step 2 is inside the leak.

Suggest stating the authorisation boundary without enumerating the failure modes:

> This pre-authorisation covers deployment and setup only. Where the skill's own protocol requires a user-facing confirmation or a real operator-supplied credential, follow the skill.

and letting the checks do the grading.

danh-nguyen-nvidia:
/ok to test [23af421](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1700/commits/23af421c123ff87195757db059a2b6a5360f8e88)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32452410232)

danh-nguyen-nvidia:
/ok to test [e45bc0e](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1700/commits/e45bc0e0b7e40ee6d0be286116431b568d9d979c)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32473845881)

danh-nguyen-nvidia:
/ok to test [857f644](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1700/commits/857f6449009c923edc4aa162ab0c55e063544bfe)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32534424477)

skills/vss-manage-alerts/SKILL.md
```diff
@@ -379,14 +388,153 @@ Load `references/always-on.md` for the event contract, reason-code table, YAML r
...
+guessing the separator (`warehouse-sample` vs `warehouse_sample`) filters on a value that
+does not exist — which returns `count: 0`, not an error.
+
+```bash
+# 1. candidate names, from the source of truth. -F matches the wording literally: without it
+#    a `.` or `[` in what the user typed is read as a pattern, which quietly matches a
+#    different camera or errors out and reads back as "no such sensor".
+# Keep the two failures apart: a dead VIOS and an unknown sensor both leave you with no
+# name, but one means "use the fallback below" and the other means "tell the user".
+LIST=$(curl -sf "$VST_API_BASE/sensor/list") || { echo "VIOS unreachable — exit 2 means: continue with the unfiltered /incidents fallback below (do NOT report an error)"; exit 2; }
```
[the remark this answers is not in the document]
danh-nguyen-nvidia: This is a false positive against the workflow's execution model, and the block preamble already documents the convention it's flagging.

Workflow C's numbered blocks are a decision tree executed by the agent (an LLM) — not a single linear bash script run end-to-end. The exit 2 here is a sentinel that hands control back to the agent to take the fallback branch, not a hard terminate. The agent reads the message it prints — exit 2 means: continue with the unfiltered /incidents fallback below (do NOT report an error) — and runs the fallback as its next action. It's the same two-value convention used across the block: exit 1 / a failed ${VAR:?} = stop and tell the user; exit 2 = switch to the unfiltered /incidents fallback.

This is stated explicitly right above the block:

…this is a decision tree, not a top-to-bottom script… The exit code says which branch: exit 1 / a failed ${VAR:?} = stop and tell the user; the VIOS-down exit 2 = switch to the unfiltered /incidents fallback (do not report it as an error).

So on a VIOS-down the agent does not stop without answering — it proceeds to the unfiltered /incidents fallback, which is exactly the path the incident store can still satisfy. This was verified in the eval run: the agent follows the branch correctly and answers from the fallback.

Resolving as working-as-designed. If the exit 2 sentinel still reads as ambiguous to human reviewers, we can restructure the VIOS-down branch to avoid exit entirely — happy to do that in a follow-up if preferred.

zac-wang-nv (approved):

PR #1718, opened 2026-08-14
fix(test): reset _expected_stable_keys between profile iterations

simings-nv:
## Description

The env-split check in `test-dev-profile.sh` loops over every developer profile and resets `_expected_override_keys` and `_allowed_duplicate_keys` at the top of each iteration — but not `_expected_stable_keys`.

That variable is assigned **only inside the four `case` arms**, and the `case` has **no default arm**. So a profile matching no arm silently inherits the previous iteration's expectations instead of starting empty.

No profile hits this today, because all four have an arm. It bites the moment a fifth is added: the new profile would be checked against whichever profile the loop last visited, producing failures that point at the wrong file and have nothing to do with the change under test — the kind of failure someone loses an afternoon to.

## Verification

Behaviour-neutral. The suite reports **203 passed / 2 failed both with and without** this change, and the two failures (`warehouse MV3DT skill should not reference legacy app-data model paths`, `warehouse MV3DT Compose and Helm startup semantics or perception fallback diverged`) are pre-existing and unrelated to the env split.

```
bash deploy/docker/test-scripts/test-dev-profile.sh
```

## Checklist

### DIFF
--- a/deploy/docker/test-scripts/test-dev-profile.sh
+++ b/deploy/docker/test-scripts/test-dev-profile.sh
@@ -1211,6 +1211,10 @@ for _profile in base lvs search alerts; do
   fi
   _expected_override_keys=("${_common_overrides_env_keys[@]}")
   _allowed_duplicate_keys=()
+  # Reset alongside the other two: this is assigned only inside the case arms
+  # below and there is no default arm, so without a reset a profile that matches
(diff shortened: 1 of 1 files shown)


simings-nv:
@greptileai review

PR #1720, opened 2026-08-14
feat(deploy): declare the NeMo Gym eval service in base, shipped off

simings-nv:
## Description

First change of the NeMo Gym eval integration. It declares the Gym evaluation runtime as a Compose service in the base developer profile **with the service switched off**, so nothing about any existing deployment changes.

Two pieces:

1. **`containers.env`** — `VSS_GYM_EVAL_IMAGE` / `VSS_GYM_EVAL_TAG` image coordinates.
2. **`dev-profile-base/compose.yml`** — the `gym-eval` service, tagged `profiles: ["gym-eval"]`.

**Deliberately not in `container-inventory.json`.** An entry was added and then removed during review: `release_set.py closure` runs in CI (`ci.yml:1275`) and printed `note: inventory entries with no compose reference (stale?): nemo-gym` on every run. The entry isn't stale, but the tool is right that it doesn't belong — `closure` computes `used` only from *first-party* refs, and `nvcr.io/nvidia/eval-factory` isn't one of the four `first_party_registry_roots`, so the entry could never be marked used and enforced nothing. (`vss-rt-cv` escapes this only by being first-party-hosted despite also being `external-pin`.) The classification now lives in `containers.env`. Closure passes with no note.

### Why this is inert

`gym-eval` is absent from `COMPOSE_PROFILES` in every profile's `overrides.env`, and Compose **drops** a service whose profile tag is not active — it is not listed by `config --services`, not resolved by `config --images`, never pulled and never built.

Two details that make that hold, both commented in the file:

- **Every interpolated variable carries an inline default.** Compose interpolates *before* it filters on profiles, so a bare `${VAR}` would emit `variable is not set` warnings even while the service is inactive.
- **The service is a pure leaf.** Nothing may `depends_on` it — Compose hard-errors with `depends on undefined service` when an active service depends on a profile-excluded one.

### ⚠ The pinned tag is not yet safe to enable

`26.05` was built 2026-06-03 and predates [NVIDIA-NeMo/Gym#2376](https://github.com/NVIDIA-NeMo/Gym/pull/2376) (merged 2026-08-11), which removes the bundled royalty-bearing codec binaries. Its layer history still `apt-get install`s ffmpeg, so it carries libraries `.github/scripts/check_no_patented_codecs.py` forbids in VSS containers.

This is safe to *declare* because the service is never pulled. `VSS_GYM_EVAL_TAG` must be moved to a post-#2376 build **before** `gym-eval` is added to any profile's `COMPOSE_PROFILES`. That flip is a separate PR. The constraint is recorded in a prominent block comment in `containers.env` and in the inventory entry's `notes`, so it cannot be enabled by accident without reading it.

As of 2026-08-15 no post-#2376 build exists anywhere — NGC has only `26.05`/`latest` (same digest), and the CI registry's newest of 452 tags is 2026-07-31. A request for one is with the Gym team.

## Verification

**Inertness proven by triple diff.** `docker compose config` for the base profile, before vs. after this change, with `VSS_APPS_DIR` set so the include chain actually resolves:

| | before | after |
|---|---|---|
| `config --services` | 11 | **11**, identical set |
| `config --images` | 11 | **11**, identical set |
| interpolation warnings | 73 | **73** |

Byte-identical on all three. A run with an unset `VSS_APPS_DIR` returns empty output for both and proves nothing — the env has to be set for this comparison to mean anything.

`compose-images.golden` gains exactly one line, which is the intended visible footprint: the golden records *declared* image references, not active ones.

## Checklist

### DIFF
--- a/deploy/docker/container-inventory.json
+++ b/deploy/docker/container-inventory.json
@@ -219,6 +219,14 @@
       "compose_image_names": ["sdr-mw-l"],
       "tag_variables": ["SDR_MW_L_IMAGE"],
       "notes": "Built by GitHub as the SDRC image while preserving the full-ref SDR_MW_L_IMAGE override."
+    },
+    {
+      "name": "nemo-gym",
+      "strategy": "external-pin",
+      "platforms": ["linux/amd64", "linux/arm64"],
+      "compose_image_names": ["nemo-gym"],
+      "tag_variables": ["VSS_GYM_EVAL_TAG"],
+      "notes": "NeMo Gym evaluation runtime, pinned from nvcr.io/nvidia/eval-factory. Outside the four first-party registry roots, so out of GHCR release-set scope. The currently pinned tag predates NVIDIA-NeMo/Gym#2376 (codec removal) and must be updated to a post-#2376 build before the gym-eval service is added to any COMPOSE_PROFILES; see containers.env."
     }
   ]
 }
--- a/deploy/docker/container-inventory.json
+++ b/deploy/docker/container-inventory.json
@@ -219,14 +219,6 @@
       "compose_image_names": ["sdr-mw-l"],
       "tag_variables": ["SDR_MW_L_IMAGE"],
       "notes": "Built by GitHub as the SDRC image while preserving the full-ref SDR_MW_L_IMAGE override."
-    },
-    {
-      "name": "nemo-gym",
-      "strategy": "external-pin",
-      "platforms": ["linux/amd64", "linux/arm64"],
-      "compose_image_names": ["nemo-gym"],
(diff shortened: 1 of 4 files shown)


simings-nv:
@greptileai review

simings-nv:
@greptileai review

simings-nv:
Closing alongside #1721: nothing in the repo references these variables under the new approach, where the runner is a runtime delta that defines its own service.

The image pin moves with it. `nemo-gym` is an external evaluation tool rather than a VSS product image, which is also why `release_set.py closure` reports it as out of scope for `container-inventory.json`.

PR #1721, opened 2026-08-14
feat(deploy): add dev-profile-gym, the profile that runs the Gym eval

simings-nv:
## Description

> **Stacked on #1720.** Review that one first; this PR's base is set to its branch, so the diff shown here is only this change.

Adds `dev-profile-gym`: the developer profile where the Gym eval will run. It is **`dev-profile-base` plus, later, the `gym-eval` service**.

- `dev-profile-gym/.env` and `overrides.env` — copied from base.
- `dev-profile-gym/compose.yml` — header only, **defines no services**. Every developer-profile compose file is included unconditionally and selection happens through `COMPOSE_PROFILES`, so the `gym-eval` service stays defined once, in base, and would be activated from here.
- Registered in the `developer-profiles/compose.yml` include chain.

### `gym-eval` is not switched on yet, and that is the point

**Today `-p gym` resolves exactly the same 11 services as `-p base`.** `gym-eval` is absent from this profile's `COMPOSE_PROFILES`; the line that turns it on sits directly above it, commented, so the flip is a single visible edit.

The reason is the tag. `VSS_GYM_EVAL_TAG` still points at `nemo-gym:26.05`, which predates [Gym#2376](https://github.com/NVIDIA-NeMo/Gym/pull/2376) and still carries the codec libraries VSS bans — and `containers.env` says in as many words *"update the tag before adding it to any profile's `COMPOSE_PROFILES`"*. Activating it here would have contradicted that warning: `docker compose config --images` for this profile would resolve `nvcr.io/nvidia/eval-factory/nemo-gym:26.05`, and `-p gym up` would pull it.

Nothing is lost by waiting, because the service has no command or dataset wiring yet — starting it today would run `/bin/bash` and exit. What this PR buys is that the profile, the driver support (next PR) and the test coverage (the one after) are all reviewed and in place, so turning the eval on is one uncommented line rather than a four-part change.

### Why a separate profile rather than enabling it in base

Blast radius. `-p base` behaves exactly as it does today and `-p gym` is the opt-in, so no existing deployment changes shape when the eval eventually turns on. Enabling it inside base is a later, separate flip.

## Verification

`docker compose config` with `VSS_APPS_DIR` set:

- **`-p gym` resolves 11 services and 11 images**, identical to `-p base`, with no `nemo-gym` reference.
- **`-p base` is untouched** — 11 services, 11 images, 73 warnings, byte-identical to develop.

Test coverage for this profile lands in the last PR of the stack, once the driver knows the profile name.

## Checklist

### DIFF
--- a/deploy/docker/developer-profiles/dev-profile-gym/.env
+++ b/deploy/docker/developer-profiles/dev-profile-gym/.env
@@ -1,0 +1,103 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Runtime/profile values modified by dev-profile.sh live in overrides.env.
+# dev-profile.sh generates generated.env from overrides.env and passes it after this file.
+
+# =============================================================================
+# PROFILE STATIC DEFAULTS
+# =============================================================================
+# Values in this section are not rewritten by dev-profile.sh for this profile.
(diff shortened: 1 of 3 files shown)


simings-nv:
@greptileai review

simings-nv:
@greptileai review

simings-nv:
@greptileai review

simings-nv:
Closing in favour of a different approach: the Gym evaluation runner is composed at runtime as a delta on an existing developer profile, driven by a new developer-side skill, rather than shipped as a checked-in profile. A delta lands in `_builds/<name>/`, which is gitignored and never a Compose profile, so no new profile or shipped service is required.

PR #1722, opened 2026-08-14
feat(deploy): teach dev-profile.sh about the gym profile

simings-nv:
## Description

> **Stacked on #1721.** Base is set to its branch, so the diff here is only this change.

Teaches `dev-profile.sh` about the `gym` profile, so `-p gym` derives the same LLM/VLM wiring `-p base` does.

`gym` joins `_valid_profiles`, and the five sites that derive VLM topology for base now call a named predicate:

```bash
function profile_has_base_topology() {
  [[ "${profile}" == "base" ]] || [[ "${profile}" == "gym" ]]
}
```

### Why a predicate instead of adding `|| [[ "${profile}" == "gym" ]]` five times

The failure mode is silent. A profile missing from one of those branches gets a `generated.env` with **no VLM wiring** — it does not error, it just produces a subtly wrong environment. Naming the shared property means the next profile with base topology is one function away from correct, and the five call sites cannot drift apart from each other.

Its scope is deliberately narrow: **LLM/VLM topology derivation only**. Argument-validation guards keep naming profiles explicitly, because which profiles a hardware or option combination accepts is per-profile policy, not shared topology. The second commit makes that boundary explicit.

### Second commit — three follow-ups

- **`--help` did not list `gym`**, so the profile was undiscoverable from the CLI.
- **The invalid-profile error still advertised `base, lvs, search, alerts`.** It is now derived from `_valid_profiles`, so the advertised set cannot drift from the accepted one again — which is exactly the bug that had just occurred.
- **The IGX-THOR/AGX-THOR VLM-option guard used the predicate**, implying a gym+edge combination that the `edge_hardware_profiles` guard above it already rejects. Reverted to naming `base`.

## Verification

`-p gym` and `-p base` produce **identical values for all five derived VLM variables** (`VLM_NAME_SLUG`, `VLM_BASE_URL`, `VLM_MODEL_TYPE`, `VLM_PORT`, `RTVI_VLM_MODEL_TO_USE`), across local and remote VLM modes. That equality is the point of the change.

Second commit verified directly:

```
$ bash scripts/dev-profile.sh -p bogus up
[ERROR] Invalid profile: bogus. Must be one of: base, gym, lvs, search, alerts
```

and `--help` now lists `gym` with a one-line note on what it is. `bash -n` clean.

[...]

PR #1723, opened 2026-08-14
test(deploy): cover dev-profile-gym in the profile test suite

simings-nv:
## Description

> **Stacked on #1722.** Base is set to its branch, so the diff here is only this change.
>
> It also carries the `_expected_stable_keys` reset from #1718, which this change **requires**: adding a fifth profile to the env-split loop is precisely the situation that bug bites. If #1718 lands first the commit drops out on rebase.

Brings `dev-profile-gym` under the existing profile test suite, so it is checked the same way the other four profiles are:

- added to the **env-split loop**, via a `base|gym)` case arm — the two profiles share expectations because gym *is* base plus one service.
- added to the **shared-VST-defaults loop**.

### Deliberately excluded from the DGX-SPARK loop

The suite caught this: `-p gym -H DGX-SPARK` errors, because `dev-profile.sh` restricts edge hardware to base and alerts. That is correct behaviour, not a gap — a 13 GB eval container does not belong on a Spark — so gym is left out of that loop rather than the guard being loosened to accommodate the test.

## Verification

**204 passed / 2 failed**, against a **203 / 2** baseline on develop. The two failures are pre-existing, unrelated to the env split, and present with and without this change:

- `warehouse MV3DT skill should not reference legacy app-data model paths`
- `warehouse MV3DT Compose and Helm startup semantics or perception fallback diverged`

A green run does not by itself prove the new profile is *covered* — a loop that silently skips it also passes. So coverage was proven by **negative control**: removing `EVAL_LLM_JUDGE_NAME` from `dev-profile-gym/overrides.env` turns the suite red with

```
FAIL: dev-profile-gym/overrides.env should define EVAL_LLM_JUDGE_NAME
```

at 203 passed / 3 failed. The assertions genuinely run against the new profile. The file was then restored and `git status` confirmed clean.

## Checklist

### DIFF
--- a/deploy/docker/test-scripts/test-dev-profile.sh
+++ b/deploy/docker/test-scripts/test-dev-profile.sh
@@ -447,6 +447,12 @@ chmod +x "${_mock_rtx4500_nvidia_smi_dir}/nvidia-smi"
 PATH="${_mock_rtx4500_nvidia_smi_dir}:${PATH}" SKIP_HARDWARE_CHECK= run_dry_run_test "RTXPRO4500BW accepted when detected GPU is RTX PRO 4500 Blackwell" up -p base -i 127.0.0.1 -H RTXPRO4500BW -d
 run_negative_test "DGX-SPARK only valid for base or alerts (not lvs)" 1 up -p lvs -i 127.0.0.1 -H DGX-SPARK
 run_negative_test "DGX-SPARK only valid for base or alerts (not search)" 1 up -p search -i 127.0.0.1 -H DGX-SPARK
+# gym shares base's LLM/VLM topology but is NOT an edge profile: it exists to add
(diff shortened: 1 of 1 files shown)


simings-nv:
@greptileai review

simings-nv:
@greptileai review

simings-nv:
@greptileai review

simings-nv:
@greptileai review

simings-nv:
@greptileai review

simings-nv:
Closing alongside #1721 and #1722: this covered a profile that is no longer being added.

Its one change with standalone value — resetting `_expected_stable_keys` between profile iterations — is open separately as **#1718** and is unaffected: the variable is assigned only inside `case` arms with no default arm, so any future fifth profile would silently inherit the previous iteration's expectations.

PR #1727, opened 2026-08-17
chore(ci): remove the last-green channel controller

zac-wang-nv:
## Description

`last-green-controller.yml` has never run, and cannot be started.

- Its only triggers are `repository_dispatch` and `workflow_dispatch`. The workflow is absent from the default branch, and GitHub does not register a dispatch-only workflow until it has run once — the Actions API has no record of this one, so there is nothing to dispatch.
- Every job is additionally gated on the repository variable `VSS_LAST_GREEN_ENABLED`, and the workflow's own header says **"DORMANT BY DEFAULT"**.

So it is not dormant-pending-enablement; it is unreachable. Removing it, its script, its test, and the `ci.yml` step that ran that test.

**Nothing imported the module.** The only other references were prose: one line in `release_set.py`'s docstring and two `description` strings in `release-set.schema.json`, all describing a "last-green channel" that would no longer exist. Those are reworded rather than left describing absent machinery — the schema edit is **description-only**, so no consumer contract moves and no validation behaviour changes.

**The release set is untouched.** It is still assembled on every build and consumed by `update_pr_ghcr_candidates.py`.

## Plan

**Goal:** the repo stops carrying a channel controller that cannot be started and that no other code depends on.

**Decisions**

| Q1 | Remove it, or land it on `main` so its dispatch triggers become usable? | Remove — it is gated off by `VSS_LAST_GREEN_ENABLED` on top of being unreachable, so enabling it is a deliberate project rather than a one-line branch move | Remove |
| Q2 | Also remove `last_green.py` and its test, or keep the module? | Remove both — the controller was the only caller, and an unreferenced module with a passing test reads as live code | Remove |
| Q3 | Leave the prose that describes the last-green channel? | Reword it. Leaving documentation that describes machinery which no longer exists is the same defect this repo already has elsewhere | Reword, description-only |

**Steps**

1. Confirm it has never run and cannot be dispatched → verify: absent from the Actions API workflow list; not on the default branch; triggers are dispatch-only.
2. Confirm nothing imports the module → verify: `git grep last_green` over `origin/develop` returns only the workflow, the script, its test, one `ci.yml` step, and prose in `release_set.py` + the schema.
3. Remove the set and reword the prose → verify: no `last_green`/`last-green` references remain under `.github/` or `deploy/`; `release-set.schema.json` still parses as JSON.
4. Prove nothing else regressed → verify: `release_set`, `prepare_downstream_release_set`, `update_pr_ghcr_candidates`, `advance_ghcr_alias`, `cleanup_pr_tags` and `trigger_downstream_pipeline` tests all pass.

**Out of scope**

- The `release-set.schema.json` structure and the assemble path, both still in use.
- The other `develop`-only workflows. `spatialai-data-utils.yml`'s `schedule` genuinely cannot fire from a non-default branch and its nightly reconcile has never run — that is a separate change, and unlike this one it probably wants fixing rather than removing.
- Whether a last-known-good channel is wanted *in principle*. This removes an implementation that never ran; if the capability is wanted later it should be designed against the current two-plane split rather than revived from dormant code.

## Checklist
- [x] New or existing tests cover these changes. — the removed test went with its module; the six remaining release-set tooling suites pass.
- [ ] The documentation is up to date with these changes. — n/a beyond the two prose edits included here.


hugoverjus (approved):
Reviewed the rebased diff and validated the release-set tooling. The orphaned last-green lock identified during review is now removed; no remaining last-green machinery survives under .github or deploy.

PR #1728, opened 2026-08-17
feat(skill-eval): add --remote-vss for disaggregated runs

zac-wang-nv:
## Description

Adds `--remote-vss <base-url>` to `run_leg.py`: a **disaggregated** skill-eval run that
evaluates an already-deployed VSS from Harbor's own default sandbox, instead of requiring a
Brev pool member.

Today a leg can only run on an operator-managed `vss-eval-*` box. That is enforced at two
points, and both fail closed for anyone else:

- `pool_candidates()` drops any name not starting with `vss-eval-`.
- `_registered_gpu_hint()` maps only the documented prefixes (`vss-eval-rtx-`,
  `vss-eval-l40s`, `vss-eval-h100`, `vss-eval-geforce-rtx4090-`) to a GPU family and returns
  `""` otherwise, so GPU-requiring legs are filtered out.

So a developer with a perfectly capable GPU workstation cannot run a leg without
`brev register`-ing the box under a conforming name and getting it added to
`BREV_REGISTERED_POOL`.

`--remote-vss` removes the need for any of that by separating the **system under test** from
the **trial sandbox**:

- Skips pool selection, the per-box `flock`, and every `brev` call.
- Omits `--environment-import-path`, leaving Harbor on its own default (docker) sandbox — a
  disaggregated leg only makes HTTP calls, so it needs no GPU and no `BrevEnvironment`.
- Forwards `VSS_BASE_URL`, `HOST_IP` and `VSS_AGENT_PORT` into the trial via Harbor `--ae`.
  `HOST_IP` is the important one: the runtime skills' gold solutions and checks already
  resolve endpoints from it (defaulting to `localhost`), so redirecting that single variable
  reuses the existing port conventions rather than teaching every check a new URL.
- Drops `BREV_INSTANCE` and the Brev transport timeouts from the Harbor env, so an inherited
  value can't aim `BrevEnvironment` at a box this leg never locked.

**Runtime skills only.** `REMOTE_VSS_SUPPORTED_SKILLS` gates it to `vss-ask-video`,
`vss-generate-video-report`, `vss-query-analytics`, `vss-search-archive` and
`vss-summarize-video`. Deploy-centric specs raise a clear error rather than run against a
stack they did not build — a disaggregated leg has no box to deploy onto, so a first turn
calling `/vss-deploy-profile` would either fail deep in the trial or quietly score something
other than what the spec name says.

```bash
python3 .github/skill-eval/run_leg.py \
  --dataset-root /tmp/skill-eval/datasets/vss-ask-video \
  --results-root /tmp/skill-eval/results/remote-$(date +%Y%m%d-%H%M%S) \
  --scratch /tmp/skill-eval/manual \
  --spec-stem direct_vlm_video_understanding \
  --platform RTXPRO6000BW \
  --remote-vss http://10.0.0.5:8000
```

## Plan

Questions raised before implementing, with the recommendation and the decision taken.

| # | Question | Recommended | Decided |
|---|---|---|---|
| 1 | What does "offline" mean — no Brev, no Anthropic, or fully airgapped? | No Brev dependency | **No Brev dependency.** Airgapped is out of scope: Harbor is fetched from PyPI, the agent needs an Anthropic-compatible endpoint, and trials pull from `nvcr.io`/NGC. |
| 2 | How should the local environment execute commands — new `LocalEnvironment`, SSH-to-localhost, or Harbor's built-in sandbox? | New `LocalEnvironment` subclass | **None of the above — disaggregated.** Assume VSS is already deployed on another box and run the trial from a sandbox pointing at it, so Harbor's default sandbox suffices and no new environment class is needed. |
| 3 | How should the mode be selected and the remote VSS addressed? | `--remote-vss <base-url>` | **`--remote-vss <base-url>`** — one flag carries both intent and target, and stays visible in CI logs. |
| 4 | Which eval specs should the mode support? | Query/runtime skills only | **Query/runtime skills only**, deploy-centric specs rejected with a clear error. |
| 5 | Prove end-to-end before raising the PR? | Yes (safe once disaggregated) | **Yes** — see Verification. |

Decision 2 is the one that shaped the change: the original plan was a `LocalEnvironment` that
execs via subprocess. Treating the deployment as remote instead made that unnecessary and
reduced the diff to a flag plus two small helpers.

Note this partially revisits ground the harness deliberately left: `AGENTS.md` records that
`requires_deployed_vss` / `prerequisite_deploy_mode` were removed and the
`_ensure_prerequisite_deployed` hook deleted, in favour of putting deploy steps inside the
trial trajectory where they are visible in the reward and judge. That reasoning is untouched
here — this does not pre-deploy on the trial's own box; it points the trial at a deployment on
a *different* box, and only for specs that never deployed anything in the first place.

## Verification

Run on a 2× RTX PRO 6000 Blackwell box with a live `base` profile deployment.

**Sandbox reaches the remote VSS** (`docker run curlimages/curl` → the deployment):

```
agent  /health  {"value":{"isAlive":true}}
agent  /docs    HTTP 200
VST    :30888   HTTP 200
NIM    :8100    HTTP 200
```

**A/B on the same dataset and spec** — with the flag, no selection and no lock:

```
[run-leg] disaggregated run against http://10.57.233.199:8000
          (harbor default sandbox; no pool selection, no box lock, no brev)
```

without it, the existing path is taken unchanged:

```
[run-leg] selected instance: vss-eval-rtx-2g (lock acquired: /tmp/brev/vss-eval-rtx-2g.lock)
[run-leg] phase summary: lock-wait=3s
```

The disaggregated run was executed with `brev` removed from `PATH` and completed the branch
without invoking it.

**Unit tests** — `python3 .github/skill-eval/tests/test_run_leg.py`

- Before: 59 tests, 3 failures
- After: 65 tests, 3 failures

The 6 new tests pass and no regression is introduced. The 3 failures pre-exist on
`origin/develop` (`test_4090_*`, which read `BREV_RTX4090_POOL`, unset locally) and are
unrelated to this change.

## Known gaps

- **The full trial was not driven to a score.** That needs `ANTHROPIC_BASE_URL` /
  `ANTHROPIC_MODEL` / `ANTHROPIC_API_KEY`, which live in the coordinator `.env` on the CI
  runner and were not available here. Everything up to the Harbor invocation is proven; the
  agent-credential gate is where the local run stops.
- **`vss-ask-video`'s spec checks hardcode `http://localhost:8000`**
  (`skills/vss-ask-video/evals/base_profile_video_understanding.json`). They resolve against
  `HOST_IP` in the gold solution but not in the check strings, so those checks need
  templating before that spec scores correctly in disaggregated mode. Deliberately left out
  of this PR to keep the harness change reviewable on its own.
- Only `run_leg.py` is touched; adapters are unchanged.

## Checklist

### DIFF
--- a/.github/skill-eval/README.md
+++ b/.github/skill-eval/README.md
@@ -197,6 +197,52 @@ python3 .github/skill-eval/run_leg.py \
 
 `CLAUDE_CODE_DISABLE_THINKING=1` is required when routing through the NVIDIA Anthropic proxy — claude-code ≥ 2.1.x otherwise emits a `context_management` field the proxy rejects with HTTP 400.
 
+### Disaggregated run against an already-deployed VSS (`--remote-vss`)
+
+`--remote-vss <base-url>` separates the system under test from the trial
+sandbox: VSS stays deployed on whatever box you already have, and the trial
+runs in **Harbor's own default sandbox** on the machine you invoke
+`run_leg.py` from, reaching that deployment over HTTP.
+
+This needs no Brev at all — no `brev register`, no `BREV_REGISTERED_POOL`
+entry, no `vss-eval-*` naming, no pool selection and no box lock. It is the
+path to use on a workstation or any GPU box that is not a pool member.
+
+```bash
+# VSS already running elsewhere; nothing here deploys or resets it.
+python3 .github/skill-eval/adapters/vss-ask-video/generate.py \
+  --output-dir /tmp/skill-eval/datasets/vss-ask-video \
+  --skill-dir skills/vss-ask-video \
+  --platform RTXPRO6000BW
+
+export PYTHONPATH="$(pwd)/.github/skill-eval:${PYTHONPATH:-}"
+
+python3 .github/skill-eval/run_leg.py \
+  --dataset-root /tmp/skill-eval/datasets/vss-ask-video \
+  --results-root /tmp/skill-eval/results/remote-$(date +%Y%m%d-%H%M%S) \
+  --scratch /tmp/skill-eval/manual \
+  --spec-stem base_profile_video_understanding \
+  --platform RTXPRO6000BW \
+  --remote-vss http://10.0.0.5:8000
+```
+
+The endpoint is forwarded into the trial as `VSS_BASE_URL`, `HOST_IP` and
+`VSS_AGENT_PORT` (via Harbor `--ae`); the runtime skills' existing port
+conventions resolve from `HOST_IP`, which defaults to `localhost` otherwise.
+`REMOTE_VSS_BASE_URL` works as an env-var default for the flag.
+
+Because the box is never touched, the warning above about wiping the docker
+runtime does **not** apply — a disaggregated leg only makes HTTP calls, so it
+is safe to point at a deployment you care about.
+
+**Runtime skills only.** Deploy-centric specs are rejected with a clear error
+rather than run against a stack they did not build: a disaggregated leg has no
+box to deploy onto, so `/vss-deploy-profile` in a first turn would either fail
+deep inside the trial or quietly score something other than what the spec name
+says. Supported today: `vss-ask-video`, `vss-generate-video-report`,
+`vss-query-analytics`, `vss-search-archive`, `vss-summarize-video`
+(`REMOTE_VSS_SUPPORTED_SKILLS` in `run_leg.py`).
+
 ### Inspect a result
 
 ```
--- a/.github/skill-eval/AGENTS.md
+++ b/.github/skill-eval/AGENTS.md
@@ -614,6 +614,15 @@ Notes that have burned prior runs:
   trial dispatches.
 - `-i` / `--include` is a different flag and will silently match
   nothing or everything.
+- `--remote-vss <base-url>` is a **developer path, not a CI path** — do not
+  use it in a matrix leg. It evaluates an already-deployed VSS from Harbor's
+  default sandbox, so it skips pool selection, the box lock and every `brev`
+  call, and it drops `BREV_INSTANCE` from the Harbor env. Nothing about the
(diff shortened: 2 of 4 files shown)


.github/skill-eval/README.md
```diff
@@ -197,6 +197,52 @@ python3 .github/skill-eval/run_leg.py \
...
 `CLAUDE_CODE_DISABLE_THINKING=1` is required when routing through the NVIDIA Anthropic proxy — claude-code ≥ 2.1.x otherwise emits a `context_management` field the proxy rejects with HTTP 400.
 
+### Disaggregated run against an already-deployed VSS (`--remote-vss`)
+
+`--remote-vss <base-url>` separates the system under test from the trial
+sandbox: VSS stays deployed on whatever box you already have, and the trial
+runs in **Harbor's own default sandbox** on the machine you invoke
+`run_leg.py` from, reaching that deployment over HTTP.
+
+This needs no Brev at all — no `brev register`, no `BREV_REGISTERED_POOL`
```
maxervo: OK so with this disaggregated mode:
- It uses a remote VSS deployment anywhere we'd like
- It will run the Harbor tasks locally (to run the agent, install skills etc.) 
=> So in this mode, the custom `class BrevEnvironment(BaseEnvironment)` will not be used anymore
=> By default, this local run of Harbor will use `DockerEnvironment` as the environment to run the task in

In this case, are each skill's `generate.py` going to tell what docker image to use for this new mode?

e.g. Right now we see them like this: `vss-ask-video/generate.py`, `vss-search-archive/generate.py`

```
        (env_dir / "Dockerfile").write_text("FROM scratch\n")
```

If we want to have the full end-to-end remote test, will these need to specify an actual Dockerfile instead of `scratch` to have the right dependencies to run the agent, harbor test etc.?

PR #1736, opened 2026-08-17
fix(search): allow L40S local RT-VLM on a 2-GPU host

hugoverjus:
## Summary
- Search already fits on two GPUs: RT-VLM shares GPU 0 with RT-CV, LLM and RT-Embed stay on GPU 1.
- L40S still needs a remote LLM (`hw-L40S-shared.env` is missing), but it no longer rejects a local RT-VLM just because that GPU is shared with RT-CV.
- This unblocks 2-GPU L40S compose eval (`USE_REMOTE_LLM=true`, `USE_REMOTE_VLM=false`).

## Test plan
- [x] `dev-profile.sh` dry-run: search `-H L40S --use-remote-llm` keeps `VLM_MODE=local_shared` on GPU 0
- [ ] `deploy/docker/test-scripts/test-dev-profile.sh` in GitHub CI
- [ ] Downstream `test-search` on a 2-GPU L40S runner with local RT-VLM


simings-nv (approved):
This reads making sense to me. Accept to unblock.

PR #1745, opened 2026-08-18
fix(ci): put VIOS third-party notices in the source artifact

hugoverjus:
## Summary
- Source packaging skipped all of `services/vios/**` as LFS pointers, including `LICENSE.3rdparty`.
- CI still pulls that license file, then replaces the pointer in the tar so the source artifact has the real notices. VIOS binaries stay skipped.

## Test plan
- [ ] `Prepare source artifact` passes on this PR
- [ ] Verify step still sees a VIOS binary as an LFS pointer
- [ ] Verify step sees real `LICENSE.3rdparty` text, not a pointer

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -190,20 +190,32 @@ jobs:
           git lfs pull \
             --include='services/vios/LICENSE.3rdparty' \
             --exclude=''
+          if head -c 64 services/vios/LICENSE.3rdparty | grep -q '^version https://git-lfs'; then
+            echo "::error::git lfs pull did not materialize services/vios/LICENSE.3rdparty"
+            exit 1
+          fi
 
       - name: Package tracked source
         run: |
           # actions/checkout above runs with lfs:false, but `git archive` applies
           # checkout-side conversion (ident, then smudge/process), so it fetches
           # the repo's entire ~989 MB of LFS payload on every run. 414 MB of that
-          # is services/vios. The required third-party notices are fetched above.
+          # is services/vios. Skip those binaries via lfs.fetchexclude.
           #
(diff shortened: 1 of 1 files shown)


hugoverjus:
What this changes:

`Prepare source artifact` still skips the big VIOS binaries so we do not download hundreds of MB on every run. After that zip is built, we swap in the real `services/vios/LICENSE.3rdparty` file.

Why:

#1734 moved that license into Git LFS. The old packaging step then told git-lfs to skip everything under `services/vios/**`. That skip is a path filter, not “skip only if we have not fetched it.” So the license came through as a tiny LFS pointer, CI failed, and `develop` has been red since.

Why this way:

We keep the blanket VIOS skip. That is the part that protects CI time and still leaves analytics fixtures intact. git-lfs cannot say “skip this folder except one file,” so we do not try to encode that in a glob. We pull the one file we need, then put it into the zip. The existing checks still prove both sides: binaries stay pointers, the license is real text.

zac-wang-nv (commented):

The internal review failed closed: Error code: 401 - {'error': {'message': 'Authentication Error, Invalid proxy server token passed. Received API Key = sk-...c2Qg, Key Hash (Token) =b899b2e87a0f1f49b35d97d7407f305792b62b6a1ce5dc70421a74e10487bcdd. Unable to find token in cache or `LiteLLM_VerificationTokenTable`', 'type': 'token_not_found_in_db', 'param': 'key', 'code': '401'}}

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32159534896)

gous-75 (approved):

PR #1755, opened 2026-08-19
CLI Search - Partition source types by media kind, not discovered index order [VIA-2717]

prisrivastav-nv:
## Summary

1. `vss search --source-type` returned wrong results whenever the deployment's index inventory didn't match an upload-first assumption: in RTSP-only deployments `rtsp` search excluded the only data present, multi-day streams lost their earliest day, and `video_file` against a not-yet-created uploads index exited 5. This PR fixes source-type partitioning end to end: CLI, search core, and the `vss-search-archive` skill.

2. The stock `search` developer profile never enabled RT-Embed's Kafka output, so `mdx-embed-filtered-*` was never written and every embed or fusion search on that profile returned nothing regardless of the read path. This PR turns the output and error buses on in the profile env, matching the Helm search profile, and aligns the `vss-build-vision-agent` references and IN-3 eval with the corrected Foundation.

## Related Issue

[[VIA-2717] CLI Search - Partition source types by media kind, not discovered index order](https://jirasw.nvidia.com/browse/VIA-2717)

[Original Proposal Doc](https://docs.google.com/document/d/1rpws782pHVeuhdtRHbqmzkFQCI0GeZ-L/edit#bookmark=id.f6iddvfmu2bk)

### Issue Description

Uploads and live streams land in different indices: uploaded videos always go to a fixed `...-2025-01-01` index, while RTSP streams are sharded by wall-clock date. To keep the two apart, the CLI designated one index as the uploads base and derived both partitions from it - `video_file` searched that index and `rtsp` searched the family wildcard *minus* that index.

The base, however, was a guess: `_runtime_from()` listed indices via `_cat/indices` and promoted `matches[0]`, whatever sorted first. In an RTSP-only deployment the `2025-01-01` index doesn't exist, so the first-sorted index is a live date-stamped one and the `rtsp` subtraction then excluded the very data being searched, returning nothing. With both kinds of data present, correctness depended on ingestion order. And where no uploads had ever been ingested, `video_file` queried a non-existent anchor and raised `IndexNotFoundError` (exit 5), even though "no uploads yet" is an empty partition, not a fault.

### Root cause

The uploads base never needed discovering, it's a constant. Uploads are written with a synthetic epoch timestamp, so they always land in `...-2025-01-01`; that name is baked into the `SearchRuntime` defaults and the `video_delete.py` write contract. Reading it from the live inventory replaced a known value with an order-dependent guess, and made every partition sensitive to whatever indices happened to exist, or to a stale `vss configure` snapshot.

The fix is the same for every family: pin the base to that constant instead of discovering it. `video_file` targets the anchor and `rtsp` targets the family wildcard minus the anchor. Because the anchor is fixed rather than promoted from `matches[0]`, the partitions no longer depend on which indices exist or on ingestion order, and the write side already guarantees the anchor name.

### Approach

- **Embed / fusion**: pin the base to the epoch-anchor constant and derive both partitions from it - `video_file` searches `mdx-embed-filtered-2025-01-01`, `rtsp` searches `[mdx-embed-filtered-*, -mdx-embed-filtered-2025-01-01]` - mirroring behavior/raw. No discovery, so it is immune to inventory order, stale snapshots, and ingestion order, scoped or unscoped. A missing anchor (no uploads ingested) maps to an empty result rather than exit 5.
- **Behavior / raw (attribute / object)**: keep `[wildcard, -base]` subtraction with base pinned to the epoch-anchor constants. A missing concrete anchor maps to an empty result via a targeted catch; every other missing index still raises, preserving the diagnostics contract.
- **CLI**: `_runtime_from()` no longer promotes `matches[0]`; bases are the anchor constants, and discovery is retained only for the wildcard, frames-index presence, and warnings - as deterministic as the already-correct agent path.
- **Skill**: RTSP readiness polls the stream-scoped embed wildcard count instead of the `2025-01-01` anchor that live data never touches; prose now describes `--source-type` as selecting the index partition for a media kind from a fixed uploads anchor, independent of identifier and ingestion order.
- **Skill (path selection)**: step 3 is now decomposition-first - decompose the request into detectable properties versus actions/relations and route on that, rather than by surface form (a single sentence is not evidence for embed).

### Changes

- **`vss_core` embed path**: `select_search_index()` resolves `video_file` to the pinned anchor and `rtsp` to the wildcard-minus-anchor; `EmbedSearch` takes the pinned base and returns an empty result when only the absent anchor is missing.
- **`vss_core` attribute/object paths**: pinned-anchor resolution with a targeted `ESNotFoundError` catch on the concrete anchor returning `[]`; the object leg guards the empty seed vector before kNN.
- **`vss_cli`**: deterministic `_runtime_from()`; frames pinned to the raw anchor only when that family exists.
- **Skill**: `source_lifecycle.md` resolves per-family wildcards with corrected RTSP readiness; `SKILL.md` / `cli_usage.md` updated to the new semantics.
- **Skill (routing)**: `SKILL.md` step 3 rewritten to decomposition-first path selection (property plus action/relation routes to fusion; property-only to attribute); `search-archive-fusion-decompose` added to `evals.json`.
- **Tests**: unit coverage for anchor and wildcard-minus-anchor selection, graceful-empty on the embed, behavior, and object legs, and deterministic runtime construction; e2e asserts the concrete anchor for `video_file` and the wildcard-minus-anchor for `rtsp`; RTSP and mixed-source scenarios added to `evals.json` and a non-executing RTSP contract scenario to `search.json`.

### Issue Description

On a stock `search` deployment the embed index is never populated: `mdx-embed-filtered-*` holds zero documents while `mdx-raw-*` and `mdx-behavior-*` fill normally from the same ingestion. The failure is deterministic, not a flake, and it is independent of the read path fixed in section 1 - any embed or fusion query reads an index that nothing writes.

It is also independent of media kind. `MESSAGE_BUS` gates all of RT-Embed's Kafka output, so it is all-or-nothing rather than an upload-versus-RTSP asymmetry.

### Root cause

RT-Embed's generated-output bus is gated by `MESSAGE_BUS`, and `rtvi-embed-docker-compose.yml:68` defaults it to empty, which disables Kafka output entirely. The only place it is set to `kafka` is `deploy/docker/services/rtvi/rtvi-embed/.env`, which the root Compose include path never loads: `services/rtvi/compose.yml` includes the service compose with no `env_file`. The search profile sets it in neither `.env` nor `overrides.env`, so the bus stayed off, RT-Embed published nothing to `mdx-embed`, and `vss-search-analytics-2d-fusion` - the sole producer of `mdx-embed-filtered` - had nothing to filter. Raw and behavior come from perception and the analytics service, so they were unaffected, which is exactly why only the embed leg was empty.

The unprefixed variable name is what hid this. The commit that introduced it replaced `KAFKA_ENABLED: "${RTVI_EMBED_KAFKA_ENABLED:-false}"` and `KAFKA_TOPIC: "${RTVI_EMBED_KAFKA_TOPIC:-...}"` with bare `MESSAGE_BUS`, `MESSAGE_BUS_TOPIC`, and `ERROR_BUS`. Dropping the `RTVI_EMBED_` prefix broke the namespacing every other variable in that file follows and that `rtvi-vlm` retains, so no profile-level env carries the knob and auditing a profile by grepping `RTVI_EMBED_*` misses it. Three other definitions disagree with the deploy-side default: the Helm chart defaults `messageBus`/`errorBus` to `kafka`, RT-Embed's own `services/rtvi/rt-embed/docker/compose.yaml` defaults both to `kafka`, and the service README documents `kafka` as the default. The Docker deploy default is the outlier.

### Approach

- **Set the buses in the profile static `.env`**, in the existing message-broker block beside `STREAM_TYPE=kafka`, matching the Helm search profile. `.env` rather than `overrides.env` because `dev-profile.sh` copies `overrides.env` to `generated.env` and `sed`-rewrites it per hardware, mode, and IP, while `.env` is passed to Compose untouched. The embed bus is an invariant of the search topology with no per-mode variance - contrast `RTVI_VLM_KAFKA_ENABLED`, which belongs in `overrides.env` precisely because it is a `sed` target for `MODE=2d_vlm`.
- **No new `KAFKA_ENABLED` knob**: that variable was removed from RT-Embed's contract and is absent from the service compose, the Helm chart, and the README. `MESSAGE_BUS` is the only toggle. The existing `RTVI_VLM_KAFKA_ENABLED=false` is RT-VLM's and stays as-is, since search runs the critic over HTTP and enabling it would publish duplicate incidents into `mdx-vlm-incidents`.
- **Contained blast radius**: bare `MESSAGE_BUS`, `MESSAGE_BUS_TOPIC`, and `ERROR_BUS` are interpolated only by `rtvi-embed-docker-compose.yml` across all of `deploy/docker`, and `dev-profile-search` is the only developer profile that ships `rtvi-embed`.
- **Build-vision-agent alignment**: with the search Foundation now carrying working values, a derived build must inherit rather than repeat them under the smallest-delta contract. The reference guidance is split by Foundation and the IN-3 eval, which required the delta to explicitly set the values, is corrected to require them effective and inherited.

### Changes

- **`deploy/docker/developer-profiles/dev-profile-search/.env`**: set `MESSAGE_BUS=kafka`, `MESSAGE_BUS_TOPIC=mdx-embed`, and `ERROR_BUS=kafka` in the message-broker block, with a comment recording the unloaded `.env` mechanism and the Helm parity rationale.
- **Skill (`rt-embed.md`)**: the Kafka output contract now splits by Foundation - search-derived builds inherit the values and must not repeat them in `override.env`, while builds on `base`, `lvs`, or `alerts` must set them.
- **Skill (`profiles/search.md`)**: the environment-knob table documents the three buses as Foundation defaults to inherit.
- **Skill (IN-3 eval)**: the env preamble and embed check now require `MESSAGE_BUS=kafka` and `MESSAGE_BUS_TOPIC=mdx-embed` to be effective in `resolved.yml` and inherited from the Foundation rather than explicitly set in the delta, matching the smallest-delta contract and the wording the combined and search-standalone evals already use.

## Checklist


zac-wang-nv (dismissed):
LGTM

ayyappa-dev (dismissed):

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32343437553)

prisrivastav-nv:
> ## 🔬 Skills Review — 6-paradigm consolidation
> **16 findings across 1 skill(s)** · 🔴 critical=3 · 🟠 high=2 · 🟡 medium=5 · 🔵 low=6 **4 need verification** (critical or single-lens). Full report in the run artifact.
> 
> ### 🔴 Critical / high (act now)
> Sev	Agree	Skill	File	Finding
> 🔴 critical	2/6	vss-search-archive	`skills/vss-search-archive/evals/search.json:92`	Delete query instructs agent to use vss configure show for indexes
> 🔴 critical	2/6	vss-search-archive	`skills/vss-search-archive/evals/search.json:97`	Delete check requires exact date-stamped index names from vss configure show
> 🔴 critical	1/6	vss-search-archive	`skills/vss-search-archive/evals/search.json:29`	Ingest query requires stale date-stamped index names
> 🟠 high	2/6	vss-search-archive	`skills/vss-search-archive/evals/search.json:34`	Ingest check claims `vss configure show` yields exact indexes
> 🟠 high	1/6	vss-search-archive	`skills/vss-search-archive/evals/search.json:29`	Ingest query demands date-stamped indexes now prohibited
> ### ⚠️ Verify before acting (critical or single-lens)
> * 🔴 **vss-search-archive** `skills/vss-search-archive/evals/search.json:92` — Delete query instructs agent to use vss configure show for indexes _(agreement 2)_
> * 🔴 **vss-search-archive** `skills/vss-search-archive/evals/search.json:97` — Delete check requires exact date-stamped index names from vss configure show _(agreement 2)_
> * 🔴 **vss-search-archive** `skills/vss-search-archive/evals/search.json:29` — Ingest query requires stale date-stamped index names _(agreement 1)_
> * 🟠 **vss-search-archive** `skills/vss-search-archive/evals/search.json:29` — Ingest query demands date-stamped indexes now prohibited _(agreement 1)_
> 
> _Skills Review (advisory)_ · [run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32343437513)


---

## Response

### Merit: largely none. It conflates two different things.

The fix removed **discovery-based *search-base promotion*** in `_runtime_from` — the CLI no longer picks a search base by sorting `_cat/indices` and taking `matches[0]`. That is the search *read path*.

The flagged evals are **ingest and delete verification**, which assert document presence/absence directly in Elasticsearch by exact index name. That is orthogonal to search index selection, and the names are correct:

- **The dates aren't stale — they're the write-side anchor.** Uploaded files are anchored server-side to `2025-01-01` (`video_ingest.py`). Both fixtures (`warehouse_sample.mp4`, `warehouse-ladder.mp4`) are uploads, so their embed/behavior/raw docs deterministically land in the `-2025-01-01` indices. Verifying against those exact names is right, and our fix never touched that contract.
- **`vss configure show` genuinely surfaces them.** `show` prints the recorded deployment JSON, whose ES service carries the raw `_cat/indices` snapshot (`configure.py:56-79`, `to_json()`). So "reuse the index names from `vss configure show`" is accurate guidance, not a broken instruction.

The low self-agreement the bot reports (1/6 and 2/6 lenses) matches this: it's pattern-matching the strings "date-stamped index" + "`vss configure show`" and mistaking a valid ES-level operational check for the discovery anti-pattern we removed.

### The one defensible (minor, optional) point
For **delete-to-zero** verification specifically (line 97), asserting against the exact anchor could in principle miss docs that spilled into an adjacent day-index. Querying the family **wildcard** (`mdx-embed-filtered-*`) scoped by `sensor.id.keyword=<UUID>` would be strictly more robust and would align with the sensor-scoped-count-against-wildcard pattern the skill's readiness step now uses. In practice it doesn't matter here — a 20-second clip anchored at `2025-01-01T00:00:00Z` cannot spill past 24h — so it's a consistency nicety, not a correctness bug.

### Bottom line
No change required. The critical/high findings don't hold: the anchor dates are the correct, deterministic write-side upload anchors, and `vss configure show` legitimately exposes the index inventory.

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32347458933)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32375465329)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32380754273)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32445447726)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32463226905)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32467896214)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32486961225)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32692884482)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32702385198)

services/agent/packages/vss_core/src/vss_core/search_core/primitives/_attribute_helpers.py
```diff
@@ -408,6 +433,12 @@ async def _search_behavior(
     try:
         response = await es.search(index=search_index_str, body=search_query)
     except ESNotFoundError as e:
+        if _is_absent_uploads_anchor(index):
```
zac-wang-nv, line 434: **P1 — Catch the normalized missing-index exception.** `ElasticClient.search()` translates `elasticsearch.NotFoundError` into the library's `IndexNotFoundError` before this helper sees it (`clients/elastic.py:317-324`). With the production client this `except ESNotFoundError` is therefore unreachable, so a missing uploads anchor still propagates and exits 5 instead of returning `[]`. The same mismatch exists in `_fetch_object_embedding` at line 640, while the new tests mock the raw Elasticsearch exception and mask the boundary. Catch the library exception for the pinned anchor in both paths, without re-wrapping it, and cover the normalized client behavior.

skills/operations/vss-search-archive/references/source_lifecycle.md
```diff
@@ -41,28 +41,25 @@ RTVI_CV_URL=$(printf '%s' "${CONFIG_JSON}" | jq -er '.services.rtvi_cv.url') ||
...
-    '[.services.elasticsearch.indices[] | select(startswith("mdx-raw-"))] | sort | first') || return 1
-  [ "${EMBED_INDEX}" != "${BEHAVIOR_INDEX}" ] &&
-    [ "${EMBED_INDEX}" != "${RAW_INDEX}" ] &&
-    [ "${BEHAVIOR_INDEX}" != "${RAW_INDEX}" ]
+  # Readiness must count a live stream's docs wherever they land, so it queries
+  # the family wildcard and lets the sensor-id filter below scope the result.
+  # Never resolve a single date-stamped index (`sort | first`): in a mixed
+  # deployment that is the `2025-01-01` uploads anchor, so a live stream whose
+  # docs sit in today's index would never satisfy readiness.
+  EMBED_INDEX="mdx-embed-filtered-*"
```
zac-wang-nv: **P1 — Keep deletion anchors separate from readiness wildcards.** `resolve_search_indexes` now assigns family wildcards, and the Delete source recipe later passes those same values to all three cleanup counts. However, the deletion contract in `evals/search.json:92-98` requires the exact `*-2025-01-01` upload indexes read from `vss configure show`. Following this reference can no longer satisfy that eval and may keep deletion waiting on same-named documents in other date shards. Retain wildcards for live-stream readiness, but resolve and save exact upload anchors separately for deletion, or intentionally update the eval contract in the same change.

zac-wang-nv (changes requested):
**P1 — Include the promised RT-Embed bus wiring.** Section 2 of the PR body says this diff adds `MESSAGE_BUS=kafka`, `MESSAGE_BUS_TOPIC=mdx-embed`, and `ERROR_BUS=kafka` to the stock search profile and updates the build-vision references/eval, but none of those files are changed at the current head. `dev-profile-search/.env` still has only `STREAM_TYPE=kafka`; the deployment loads the profile env files, not `services/rtvi/rtvi-embed/.env`; and the service compose defaults both buses to empty. The unchanged IN-3 eval/reference explicitly documents that include behavior. As submitted, the stock profile still disables RT-Embed output and cannot populate `mdx-embed-filtered-*`. Add the described profile/reference/eval changes, or remove that scope from the PR and track it separately.

zac-wang-nv (approved):

PR #1761, opened 2026-08-19
feat(skills): add NeMo Gym evaluation to vss-build-vision-agent

simings-nv:
## Description

Adds NeMo Gym evaluation to `vss-build-vision-agent`, so a developer can build a vision agent and then score it with **NVIDIA NeMo Gym** from the same skill. An evaluation becomes a scalar reward per task, the contract a training loop consumes, and one deployment can be scored by both NeMo Gym and VSS's own harness for comparison.

### Documentation and eval specs only

The bundle follows the SOP model: a condensed owner contract at `references/services/gym.md` beside a `references/services/gym/` folder holding the detailed contracts of record and one net-new asset.

```
references/services/gym.md          condensed owner contract
references/services/gym/
    image-gate.md                   provenance check before any pull
    compose-delta.md                entry point, placeholders, build artifacts
    run-lifecycle.md                verified host workflow
    results.md                      reading a reward honestly
    comparison.md                   side-by-side against VSS's own harness
    resources-server.md             the vss_ask_video resources server
    gym-eval-override.yml           the runner service definition
```

`SKILL.md` stays thin, 123 to 128 lines. It gains a routing row, a post-deploy offer beside the existing "customize this workflow" prompt, and an explicit exclusion from the capability multi-select.

**No third-party dependency is introduced.** No requirements file, no pyproject, no imports. Scoring is performed by a NeMo Gym *resources server* that is deliberately not vendored here: it implements Gym's interface and imports Gym, which VSS never runs. The bundle documents how to obtain and stage it.

**No shipped file changes.** No developer profile, no service in any compose file, no entry in `containers.env` or `container-inventory.json`. Nothing under `deploy/` is touched at all. The runner is composed at run time into `_builds/<name>/`, which is gitignored and never a Compose profile. Nothing runs until the skill is invoked.

## Why it is an overlay and not a capability

`gym-eval` scores the built stack rather than adding an ability to it, so it cannot be resolved by forward closure: its subject is the resolved service set as a whole.

`composition.md` requires a delta to be symmetric, pruning every Foundation service no requested capability needs, with validation rejecting orphaned carryover. This overlay must do the opposite, because pruning to a closure scores a different system and the numbers stop being comparable to anything. Rather than let the skill quietly break its own contract, `composition.md` now states the exception: an evaluation overlay contributes no owners or peers to the closure and is applied after resolution and pruning finish, adding its key and removing none. The no-orphan validation is scoped to the pre-overlay resolved set.

That note sits after the resolution flow, not inside it, and says explicitly that it does not apply to anyone who is not composing an evaluation overlay.

## Using it

Ask a coding agent to evaluate a deployment, and the skill:

1. **Checks the image is safe to pull.** The published Gym image predates the upstream removal of royalty-bearing codec libraries, so the gate establishes provenance from the registry manifest list, platform manifest and config blob, roughly 20 KB and no image layers, and **exits non-zero** unless the build postdates the fix and no layer records a codec install.
2. **Composes the runner as an overlay** on the stack already deployed, adding one service key and nothing else.
3. **Runs the two-phase Gym lifecycle**, `gym env start` then `gym eval run`, with the policy pointed at the deployment's own LLM endpoint, so no external provider is required.
4. **Returns a scalar reward per task**, with the judge rationale and per-check signals persisted for diagnosis.

**Scope.** This covers the ask-video path, where VSS is the environment and the policy queries it. It does not cover the CI skill-eval corpus, which needs Gym's `harbor_agent` and a Harbor release incompatible with the `harbor==0.20.0` at `.github/skill-eval/run_leg.py:53`. Tracked upstream.

## Two things the documentation is deliberate about

**This is a distinct scorer, not a pass-through of VSS's own eval.** The server applies its own judge and its own formula, `0.4 × (the policy called the tool) + 0.6 × quality`. Its numbers are **not** numerically comparable to CI skill-eval scores; compare trends and failure modes, not absolute values.

**Only the host workflow is documented as verified**, because only it has been run end to end. The containerised runner is marked as the packaging target with its known gaps listed, and it cannot be exercised until a codec-clean image exists, since running it means pulling an image the gate rejects.

## Validation

**End to end against a live VSS deployment**, following the documented steps from a clean stage:

```
policy calls ask_vss -> VSS answers from warehouse_safety_0001 -> LLM judge grades it
mean/reward 1.0   verifier_ok True   verifier_status "rubric"
```

`warehouse_safety_0001` is the clip four existing eval specs already use, so a Gym reward and VSS's own harness measure the same video.

**The image gate**, verified by exit status rather than output, across 15 config-blob shapes:

| Input | Result |
|---|---|
| clean build, postdates the fix | `exit 0` — GATE PASS |
| codec-installing layer | `exit 1` |
| clean but predates the fix | `exit 1` |
| build stamped on the cutoff day itself | `exit 1` |
| the real published tag's metadata | `exit 1` |
| empty / malformed / non-JSON blob | `exit 1` |
| `.history` empty, or absent | `exit 1` |
| `.history` not an array (object, string, number, null) | `exit 1` |
| `.history` entries with no readable `created_by` | `exit 1` |

Build dates are compared as epochs rather than strings, so an offset timestamp such as `2026-08-11T01:00:00+10:00`, which is `2026-08-10` in UTC, is correctly rejected. Provenance that cannot be *read* is rejected rather than treated as clean.

**Both eval specs pass in CI at full marks:** `gym_delta_adds_only_the_runner` **1.0 (10/10)** and `gym_image_gate_rejects_unsafe_tag` **1.0 (6/6)**, running under `vss-build-vision-agent`'s adapter.

**Other checks**

- Composing the overlay leaves the working tree clean: no change under `deploy/docker/` or `skills/`
- Playbook compliance unchanged at 0 errors
- Every relative link in the bundle resolves

## Checklist


simings-nv:
@greptileai review

.github/skill-eval/adapters/vss-run-gym-eval/generate.py
```diff
@@ -0,0 +1,580 @@
...
+        f"# vss-run-gym-eval verifier (step {step}): delegates to the generic\n"
+        "# LLM-as-judge (.github/skill-eval/verifiers/generic_judge.py).\n"
+        "set -uo pipefail\n"
+        "\n"
+        'TEST_DIR="$(cd "$(dirname "$0")" && pwd)"\n'
+        "python3 -m pip install --quiet 'anthropic>=0.40.0' >/dev/null 2>&1 || true\n"
+        "\n"
+        'python3 "$TEST_DIR/generic_judge.py" \\\n'
+        f'    --spec "$TEST_DIR/{spec_name}" --step {step}\n'
+        "exit 0\n"
```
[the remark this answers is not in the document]
simings-nv: Thanks — the finding is fair, but this wrapper isn't hand-written for this PR. .github/skill-eval/adapters/vss-run-gym-eval/generate.py was auto-generated by skills-eval-bot, and the trailing exit 0 comes from the shared adapter template rather than a choice made here.

It's also the prevailing pattern rather than a one-off: on develop, 12 of the 16 existing adapters end the verifier wrapper with exit 0 after invoking generic_judge.py. The four that don't — vss-build-vision-agent, vss-deploy-detection-tracking-2d, vss-deploy-detection-tracking-3d, vss-search-archive, omit the line and let the judge's exit status propagate.

So the masking you describe is real, but it predates this PR and applies corpus-wide: a judge that dies before writing reward.txt already scores zero silently across most of the skill-eval suite. Changing it only here would make this one leg fail loud while twelve others stay silent, which seems like the wrong place to introduce that divergence.

Happy to take it as a separate change across the adapters if we want the fail-loud behaviour: I'd rather not diverge a single skill's harness inside a skill-addition PR.

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32281420288)

simings-nv:
Following up on my own comment above — we went ahead and dropped the trailing `exit 0` after all.

I'd argued it belonged in a corpus-wide change rather than here. On checking the actual semantics, the objection I had doesn't hold: the worry was converting normal low scores into harness errors, and that can't happen. `generic_judge.main()` returns 0 after scoring whatever the reward is, pass or fail — the run that scored 0.8 with two failing checks returns 0. Its only non-zero return is `--step` out of range, which is a generation bug, and I confirmed that by running it (exits 1).

So the cases separate cleanly: a normal run of any score exits 0 exactly as before, while a generation bug or a judge that dies before writing `reward.txt` now surfaces instead of being swallowed.

Worth noting this isn't a third behaviour — the four adapters that already omit the line (`vss-build-vision-agent`, `vss-deploy-detection-tracking-2d`, `vss-deploy-detection-tracking-3d`, `vss-search-archive`) do exactly this. The wider cleanup across the remaining adapters still seems worth doing separately.

simings-nv:
@greptileai review

Three commits since the last pass (`b54d4ca` → `6cd8873`):

- **Foundation entry point named.** The first harness run scored 0.8, failing the checks that the delta parses and that the tree stays clean. `references/delta.md` had said `-f <foundation>` with literal ellipses and never named the file. It's the root `deploy/docker/compose.yml` — the per-profile and `developer-profiles` files are fragments, and composing against either dies with `depends on undefined service "kibana"`, an error that names the delta but reproduces on the bare Foundation. The verify snippet is now a runnable command, and the doc states outright that composition must not touch the checked-in tree.
- **Gate network calls bounded.** All four registry curls had no timeout, so an unresponsive `nvcr.io` would hang the gate rather than fail it. Now `--connect-timeout 5` with `--max-time 30` (60 for the blob).
- **Two silent failures closed.** A missing `generic_judge.py` is fatal at generation instead of yielding a task that can't score, and the verifier wrapper no longer swallows the judge's exit status.
- **An unsourceable claim narrowed.** SKILL.md asserted a fixed 30-frame sample holding scores near 0.70; that isn't sourceable from this repo, so it now says only what is.

Verified in a pristine worktree: the documented command emits exactly `> gym-eval` and leaves `deploy/docker` with zero modified files.

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32283764680)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32285653460)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32288476076)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32292422988)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32293785524)

zac-wang-nv (commented):
can this fold into `vss-build-vision-agent` like the sop deployment?  the idea is that user should be able to use gym+ vss like `build a vision agent with summarization capability and be able to evaluate it using nemo gym`.

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32301164775)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32304955681)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32307649041)

skills/vss-build-vision-agent/SKILL.md
```diff
@@ -103,6 +106,8 @@ Offer the user **exactly** the capabilities in the table below. Each row's owner
...
 
+**Never offered here — evaluation overlay:** `gym-eval` ([`gym.md`](references/services/gym.md)) scores a build rather than adding an ability to it, so it is **not** a vision capability and must not appear in this multi-select. It is offered after a build completes, applied to the already-resolved service set, and takes no part in forward closure or pruning. See the routing table above.
```
zac-wang-nv: is this clause required?

skills/vss-build-vision-agent/SKILL.md
```diff
@@ -85,6 +86,8 @@ The recommended first-run path. Deploys a validated developer profile via **Stoc
 
 These are **predefined developer profiles** — the skill keeps the profile's authoritative `COMPOSE_PROFILES` unchanged (Stock mode, Step 5 exact match) and follows the shared build lifecycle (Steps 5–9). For Alerts, set the profile `MODE` per Q2a-mode.
 
+**Evaluate a build → Gym overlay.** After a build is deployed and ready, the user may want to score it. Offer: *"Want to evaluate this build with NeMo Gym? It scores the stack you just deployed and produces a reward per task."* On **yes**, follow [`references/services/gym.md`](references/services/gym.md). Do not compose it as a capability and do not re-run resolution — the deployed set is the thing being measured, so the overlay goes on top of what is already there. Today this lands on the verified host workflow, which needs no image; the containerised runner is blocked until a `nemo-gym` tag clears the image gate.
```
zac-wang-nv, line 89: should gym be part of the custom build route?

simings-nv: Do you mean build with gym selected and evaluate in one go?

My thinking was build first, then offer. That way what gets scored is the stack that actually deployed, not one gym had a hand in shaping while it was still being resolved. Keeps the measurement honest.

Build-with-gym as a single flow feels like a good follow-up, especially once the container is settled. Happy to pick it up then, or sooner if you'd rather have it up front.

skills/vss-build-vision-agent/SKILL.md
simings-nv: Not strictly, it's a guardrail. Happy to cut it if you think the routing row carries enough weight on its own.

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32386708602)

PR #1764, opened 2026-08-19
skill-eval: add optional agent model selection

gbau1:
## Description

Adds a single, validated agent-model configuration path to the manual Skills Eval workflow.

- `agent=claude-code`: the user selects only the Anthropic model, keeps `provider=default`, and the evaluated agent uses the native Anthropic endpoint. Provider and endpoint overrides are rejected.
- `agent=nemoclaw`: the user selects `nvidia-inference`, `nvidia-build`, or `custom` plus a model. `custom` also requires the new `endpoint_url` input; its API key remains runner-managed.
- Standardizes the job contract on `SKILLS_EVAL_HARNESS`, `SKILLS_EVAL_PROVIDER`, `SKILLS_EVAL_MODEL`, and `SKILLS_EVAL_ENDPOINT_URL`; `EVAL_AGENT` remains an internal compatibility alias.
- Resolves and validates the effective configuration before any Brev worker is allocated.
- Loads runner credentials once, then applies workflow inputs with deterministic precedence.
- Changes only the Harbor agent under evaluation. The outer CI coordinator and Harbor judge retain their existing Claude configuration.
- Reuses the checked-in NemoClaw notebook setup without modifying either notebook.

### Validation

- Focused agent configuration, Harbor dispatch, and NemoClaw harness suite: `44 passed`.
- Full eval-harness suite: `207 passed`; the same three unrelated RTX4090 capability-table expectation failures reproduce on `develop`.
- Workflow YAML parsing, Python compilation, critical Ruff checks, and `git diff --check` pass.

## Checklist


gbau1:
ok to test

zac-wang-nv:
image

agent model and endpoint should apply to all harness, instead of just nemoclaw.

.github/skill-eval/nemoclaw/notebook_setup_adapter.py
```diff
@@ -114,41 +123,18 @@ def prepare_environment(
         raise RuntimeError("NGC_CLI_API_KEY is required for notebook setup")
     env["NGC_CLI_API_KEY"] = ngc_key
 
-    endpoint = _openai_base_url(
-        _first_nonempty(
-            env.get("NEMOCLAW_ENDPOINT_URL"),
-            env.get("ANTHROPIC_BASE_URL"),
-            env.get("LLM_REMOTE_URL"),
-        )
+    endpoint = (
```
zac-wang-nv, line 127: **P1** — from an independent review pass at `27c161a3`.

`nvidia-build` can never satisfy this gate. It is not an infrastructure flake — it is deterministic, and the PR's own eval run already hit it.

The chain is three links, all in this diff or reachable from it:

1. `notebook_setup_adapter.py:126-133` — for `nvidia-build`, `endpoint` is hard-coded to `""` and written as `env["NEMOCLAW_ENDPOINT_URL"] = ""`.
2. `_parameterize_notebook` injects `NEMOCLAW_ENDPOINT_URL = os.environ.get('NEMOCLAW_ENDPOINT_URL', <cell default>)`. The key **is present** (set to `""`), so `.get` returns `""` — the notebook's own cell default never applies.
3. `deploy_nemoclaw.ipynb` cell 12 prints the model marker only inside a truthiness guard:

```python
NEMOCLAW_PROVIDER = "custom" if NEMOCLAW_ENDPOINT_URL else "build"
print("NEMOCLAW_PROVIDER:", NEMOCLAW_PROVIDER)
if NEMOCLAW_ENDPOINT_URL:            # <- False for nvidia-build
    print("NEMOCLAW_ENDPOINT_URL:", NEMOCLAW_ENDPOINT_URL)
    print("NEMOCLAW_MODEL:", NEMOCLAW_MODEL)     # <- the marker line 352 requires
    print("COMPATIBLE_API_KEY set:", bool(COMPATIBLE_API_KEY))
```

So `build` mode passes the `NEMOCLAW_PROVIDER:` gate and then always fails this one.

**Reproduced offline, with a control** (`repro_build_marker.py` — drives the adapter's real `prepare_environment`, `_parameterize_notebook` and `_require_output` against the real checked-in notebook):

```
--- SUBJECT  nemoclaw + nvidia-build ---
  adapter set NEMOCLAW_ENDPOINT_URL=''
  adapter set NEMOCLAW_MODEL='qwen/qwen3.5-122b-a10b'
  notebook stdout: ['NEMOCLAW_PROVIDER: build']
  gate[NEMOCLAW_PROVIDER] -> PASS
  gate[NEMOCLAW_MODEL]    -> RAISE: deploy_nemoclaw.ipynb completed without readiness marker: NEMOCLAW_MODEL: qwen/qwen3.5-122b-a10b

--- CONTROL  nemoclaw + custom (endpoint non-empty) ---
  notebook stdout: ['NEMOCLAW_PROVIDER: custom', 'NEMOCLAW_ENDPOINT_URL: …', 'NEMOCLAW_MODEL: some/model', 'COMPATIBLE_API_KEY set: True']
  gate[NEMOCLAW_PROVIDER] -> PASS
  gate[NEMOCLAW_MODEL]    -> PASS
```

The control passing is the point: the gate is fine, the `build` branch is what cannot reach it.

**And it already happened in this PR's CI.** Run [32305638319](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32305638319) was dispatched with `provider=nvidia-build` (`skill-eval model: runtime=nemoclaw provider=nvidia-build model=nvidia/nemotron-3.5-lightning-30b-a3b`). Both legs failed. The `base_profile_video_understanding · L40S` leg failed with:

```
RuntimeError: deploy_nemoclaw.ipynb completed without readiness marker
DONE: 0/1 specs passed; step-1 RuntimeError (NemoClaw deploy_nemoclaw.ipynb failed without readiness marker on vss-eval-l40s-2)
```

— verbatim the exception the repro produces. (The other leg died earlier, at NemoClaw sandbox onboarding on `l40s-3`, which does look like genuine infra.)

Worth flagging separately: the coordinator agent wrote this up as *"an environment/infrastructure failure in the NemoClaw agent runtime, not an adapter issue."* That is wrong, and it is the kind of wrong that gets a deterministic bug re-dispatched forever.

**Still live at `27c161a3`, and re-verified rather than carried.** `notebook_setup_adapter.py` is byte-unchanged since that CI run (`f6e8f30c` at both). The notebook is *not*: the round-5 rebase pulled in a develop rewrite of `deploy_nemoclaw.ipynb` (`29bf213f` → `8cf88434`). That rewrite is parallel skill-install, a `NEMOCLAW_MODEL` default rename, and `retry_onboard_on_fail: False → True` — it does not touch cell 12, where the `if NEMOCLAW_ENDPOINT_URL:` guard around `print("NEMOCLAW_MODEL:", …)` is still present verbatim. The repro re-run at this head still raises on the subject and still passes on the control (NOTE 7).

**Fix direction** — I'd prefer fixing the notebook over loosening the gate. The notebook already knows the model in build mode (cell 10 sets `NEMOCLAW_MODEL`), so hoisting the `NEMOCLAW_MODEL:` print out of the `if NEMOCLAW_ENDPOINT_URL:` block makes the marker mean "the model is resolved", which is what the gate is actually asserting. Loosening `_require_output` to skip the model check under `nvidia-build` also unblocks, but it drops the only assertion that the build-mode model reached the notebook at all — worth a deliberate choice rather than the quicker edit.

Greptile reached the same conclusion independently, and has now re-filed it at this exact head (its 5th review, footer *"Last reviewed commit: 27c161a3"*): *"The adapter clears the endpoint for NVIDIA Build, the checked-in notebook therefore omits its endpoint-gated model marker, and the adapter still requires that marker unconditionally."* Credit to it.

.github/skill-eval/run_leg.py
```diff
@@ -320,9 +321,16 @@ def build_harbor_command(
...
             "-a", "claude-code",
             "--model", model,
-            "--ak", f"api_base={_api_base_v1(anthropic_base_url)}",
             "--ae", "CLAUDE_CODE_DISABLE_THINKING=1",
+            # Keep ANTHROPIC_BASE_URL in Harbor's parent environment for the
+            # verifier, but clear it for the evaluated Claude process so the
+            # Anthropic SDK uses its native endpoint.
+            "--ae", "ANTHROPIC_BASE_URL=",
         ]
+        if anthropic_base_url:
```
zac-wang-nv, line 330: **P2** — from an independent review pass at `27c161a3`.

This branch is unreachable, and it's unreachable on the exact path that leaked in round 2 — which is why I'd rather see it deleted than left as documentation of an option.

`build_harbor_command` is called once (line 1344) with `base_url = model_config.endpoint_url`. Inside `_COMPATIBLE_PROVIDERS`, `claude-code` maps to `{"anthropic"}` only, and the `anthropic` branch (`model_config.py:104-106`) hard-codes `endpoint_url = ""`. So `anthropic_base_url` is always `""` here and `if anthropic_base_url:` is always False.

Your own test already asserts this: `test_build_command_uses_native_anthropic_endpoint` does `self.assertNotIn("--ak", cmd)`.

Leaving it means the next person to touch this file sees a live-looking code path that redirects the evaluated Claude agent at an operator-supplied base URL, three lines below a comment explaining why that must not happen to the judge. Deleting the branch — and passing `""` explicitly — makes the invariant structural instead of incidental. If it's kept deliberately as a hook for a future `claude-code` + custom-endpoint mode, a one-line comment saying so would do.

.github/skill-eval/run_leg.py
```diff
@@ -320,9 +321,16 @@ def build_harbor_command(
         agent_flags = [
             "-a", "claude-code",
             "--model", model,
-            "--ak", f"api_base={_api_base_v1(anthropic_base_url)}",
             "--ae", "CLAUDE_CODE_DISABLE_THINKING=1",
+            # Keep ANTHROPIC_BASE_URL in Harbor's parent environment for the
+            # verifier, but clear it for the evaluated Claude process so the
+            # Anthropic SDK uses its native endpoint.
+            "--ae", "ANTHROPIC_BASE_URL=",
```
zac-wang-nv, line 328: **P2** — from an independent review pass at `27c161a3`.

`--ae ANTHROPIC_BASE_URL=` sets the variable to the empty string in the agent process; it does not unset it. Those differ for any consumer that checks presence rather than truthiness — `os.environ.get("ANTHROPIC_BASE_URL")` returns `""`, not `None`, and the Anthropic Python SDK's `base_url` default is a `None` check, not a falsy check.

I could not settle which way the Claude Code runtime inside Harbor reads it: Harbor isn't vendored here, no test covers the empty-string case (the assertion is `assertIn("ANTHROPIC_BASE_URL=", cmd)`, which is satisfied by the flag existing), and no eval run has been dispatched at any head containing this line — the only Skills Eval run on this branch predates it and ran `nemoclaw`, not `claude-code`. So this is a question, not a claim: if it resolves the wrong way, every `claude-code` leg points at an empty base URL rather than the native endpoint, and the comment above it would be describing an intent the code doesn't achieve.

One dispatched `claude-code` leg at head settles it. Round 5's sweep is **not** that leg: run [32494997741](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32494997741) reports `runtime=nemoclaw provider=nvidia-inference` on every leg, so `build_harbor_command`'s `claude-code` branch is never taken and this line stays unexercised at every head that has ever contained it.

.github/skill-eval/model_config.py
```diff
@@ -0,0 +1,170 @@
...
+            env.get("SKILLS_EVAL_API_KEY"),
+            env.get("NVIDIA_API_KEY"),
+        )
+    else:
+        endpoint_url = _first(env.get("SKILLS_EVAL_ENDPOINT_URL"))
+        api_key = _first(env.get("SKILLS_EVAL_API_KEY"))
+    if provider not in {"anthropic", "nvidia-build"} and not endpoint_url:
+        raise ValueError(f"SKILLS_EVAL_ENDPOINT_URL is required for {provider}")
+    if not api_key:
+        raise ValueError(f"no API key is configured for {provider}")
```
zac-wang-nv, line 133: **P2** — from an independent review pass at `27c161a3`.

Carried from round 2, narrowed: the model error was fixed — it now names `SKILLS_EVAL_MODEL` — but this one still doesn't tell an operator what to set.

`no API key is configured for custom` leaves them to work out that the answer is `SKILLS_EVAL_API_KEY` for `custom`, `NVIDIA_API_KEY` for `nvidia-build`, and `ANTHROPIC_API_KEY` for `anthropic`, none of which appear in the message. Since this fires inside a manual dispatch whose whole point is that a human just chose a provider from a dropdown, naming the expected variable per provider is cheap and lands the operator directly on the fix. Same applies to line 131's `SKILLS_EVAL_ENDPOINT_URL is required for <provider>`, which does name it — that one reads well; this is the outlier.

.github/skill-eval/model_config.py
```diff
@@ -0,0 +1,170 @@
...
+        env.get("SKILLS_EVAL_HARNESS"),
+        env.get("EVAL_AGENT"),
+        "claude-code",
+    )
+    requested_provider = _first(env.get("SKILLS_EVAL_PROVIDER"), "default")
+
+    if runtime not in RUNTIMES:
+        expected = " | ".join(RUNTIMES)
+        raise ValueError(f"unsupported EVAL_AGENT {runtime!r}; expected {expected}")
+    if requested_provider not in (*REQUESTED_PROVIDERS, "anthropic"):
```
zac-wang-nv, line 65: **NOTE** — from an independent review pass at `27c161a3`.

Not a defect — a load-bearing subtlety worth a comment before someone tidies it away.

`"anthropic"` is accepted as a *requested* provider here even though it is absent from `REQUESTED_PROVIDERS`, absent from the `expected` list in the error message right below, and absent from the workflow's dropdown. The reason is re-entrancy: `apply_model_config` writes the *resolved* provider back into the environment (`env["SKILLS_EVAL_PROVIDER"] = config.provider`, line 153), so a second `resolve_model_config` over that same env sees `anthropic` as its input. `notebook_setup_adapter.prepare_environment` calls `apply_model_config` on `os.environ` directly, so this is a real sequence, not a hypothetical.

The tuple-splat reads like an oversight, which makes it a natural thing for a future cleanup to "fix" — and the failure would surface as a confusing `unsupported SKILLS_EVAL_PROVIDER 'anthropic'` on the second resolution. A short comment naming the re-entrancy contract would protect it. (Same reasoning applies to `PROVIDERS` vs `REQUESTED_PROVIDERS` being two tuples that must stay in sync.)

.github/skill-eval/model_config.py
```diff
@@ -0,0 +1,170 @@
...
+        )
+    if runtime == "claude-code" and _first(env.get("SKILLS_EVAL_ENDPOINT_URL")):
+        raise ValueError(
+            "SKILLS_EVAL_ENDPOINT_URL cannot be set for claude-code; "
+            "Claude uses the native Anthropic endpoint"
+        )
+
+    model = _first(env.get("SKILLS_EVAL_MODEL"))
+    if not model and provider == "anthropic":
+        model = _first(env.get("CLAUDE_CODE_MODEL"), env.get("ANTHROPIC_MODEL"))
```
zac-wang-nv, line 87: **P1** — from an independent review pass at `27c161a3`.

The `claude-code` default path looks broken by this PR, and unlike finding 1 this one hits **every ordinary CI run**, not an opt-in provider.

This PR changes where the evaluated Claude agent runs. At base, `claude-code` was always pointed at the NVIDIA gateway:

```python
# base run_leg.py:1288-1291, 1323
model    = os.environ.get("ANTHROPIC_MODEL", "")
base_url = os.environ.get("ANTHROPIC_BASE_URL", "")
if not base_url:
    print("FATAL: ANTHROPIC_BASE_URL not set")      # required
...
"--ak", f"api_base={_api_base_v1(anthropic_base_url)}"
```

At head, `claude-code` is pinned to the `anthropic` provider, whose `endpoint_url` is hard-coded `""`. So the `--ak api_base` flag is never emitted (finding 2), and `--ae ANTHROPIC_BASE_URL=` clears the variable for the evaluated process. The agent now talks to **native `api.anthropic.com`**. That part is deliberate and documented (README: *"never forwarded to the native Claude Code agent under evaluation"*).

**The model id did not come along for the ride.** This line resolves it:

```python
model = _first(env.get("CLAUDE_CODE_MODEL"), env.get("ANTHROPIC_MODEL"))
```

- `CLAUDE_CODE_MODEL` is read **here and nowhere else** — `grep -rn CLAUDE_CODE_MODEL .github/` returns this line plus one unit test. No workflow, script, or `.env` in this repo sets it.
- So the effective default is `ANTHROPIC_MODEL`, which README line 50 confirms is intended: *"Existing Claude coordinator and Harbor judge model; **also the default evaluated-agent model**"*.
- But `ANTHROPIC_MODEL` is the **gateway's** model id. The PR's own test fixture says so:

```python
# tests/test_run_leg.py:678-680
ANTHROPIC_MODEL    = "aws/anthropic/bedrock-claude-opus-4-6"
ANTHROPIC_BASE_URL = "https://inference-api.nvidia.com/v1"
```

`aws/anthropic/bedrock-claude-opus-4-6` is a gateway-routed identifier. Sent to native Anthropic with the base URL cleared, it is not a valid model. The reroute and the model default are each individually coherent; together they send a gateway model id to an endpoint that has never heard of it.

**Why this is P1 and not P2.** `skills-eval.yml` fires on every `push` to `pull-request/<N>`, and on that path `SKILLS_EVAL_HARNESS_INPUT` evaluates to `claude-code` and `SKILLS_EVAL_PROVIDER_INPUT` to `default` → `anthropic`. So this is the default configuration for routine skill evals, not an operator opt-in. Finding 1 breaks a provider you must deliberately select; this breaks the one you get by doing nothing.

**Stated precisely, because it rests on coordinator state I cannot read.** If `ANTHROPIC_MODEL` on the eval coordinator is gateway-shaped — as this PR's own fixtures, and `inference-api.nvidia.com` throughout, both indicate — every `claude-code` leg fails on an invalid model. If some coordinator sets `CLAUDE_CODE_MODEL` out-of-band, it works there and nowhere else. Either way the repo contains no mechanism that supplies a native-shaped default. **No `claude-code` leg has run at any head containing this code**, so CI has not tested it: the only sweep on this branch (run 32494997741) is `runtime=nemoclaw provider=nvidia-inference` end to end.

**What would settle it in one step:** dispatch this workflow with `agent=claude-code`, `provider=default`, `model` blank, and read the leg's `skill-eval model:` line. If it prints `model=aws/anthropic/…`, this is confirmed.

**Fix direction.** Give the `anthropic` provider a native default rather than inheriting the gateway's — e.g. fall back to a literal native model id the way `helm_sync_agent.py:133` already does (`os.environ.get("ANTHROPIC_MODEL") or "claude-sonnet-4-6"`), or require `SKILLS_EVAL_MODEL`/`CLAUDE_CODE_MODEL` for `claude-code` and fail fast with a message naming them. Silently reusing `ANTHROPIC_MODEL` across an endpoint switch is the part to remove.

.github/skill-eval/model_config.py
```diff
@@ -0,0 +1,170 @@
...
+        )
+    if not model:
+        raise ValueError(
+            "SKILLS_EVAL_MODEL is required unless the selected runtime has "
+            "an existing default model"
+        )
+
+    if provider == "anthropic":
+        endpoint_url = ""
+        api_key = _first(env.get("ANTHROPIC_API_KEY"))
```
zac-wang-nv, line 106: **P2** — from an independent review pass at `27c161a3`.

`SKILLS_EVAL_API_KEY` is this PR's own generic credential input, and it is wired into three of the four providers but silently dropped by the fourth — the one `claude-code` is pinned to.

```python
106:  api_key = _first(env.get("ANTHROPIC_API_KEY"))                    # anthropic — SKILLS_EVAL_API_KEY not consulted
115:  api_key = _first(env.get("SKILLS_EVAL_API_KEY"), …)               # nvidia-inference — honoured, first
124:  api_key = _first(env.get("SKILLS_EVAL_API_KEY"), …)               # nvidia-build — honoured, first
129:  api_key = _first(env.get("SKILLS_EVAL_API_KEY"))                  # custom — honoured, only
```

Then `apply_model_config` writes the resolved value straight back over it:

```python
156:  env["SKILLS_EVAL_API_KEY"] = config.api_key
```

So for `claude-code` the operator's `SKILLS_EVAL_API_KEY` is not merely ignored — it is **overwritten** with `ANTHROPIC_API_KEY`, in an env dict that is then handed to the subprocess. Anyone debugging by echoing `SKILLS_EVAL_API_KEY` downstream sees a value that was never the one they set.

**What makes this a defect rather than a deliberate restriction:** the sibling input is handled the opposite way in the same function. `SKILLS_EVAL_ENDPOINT_URL` set under `claude-code` raises immediately —

```python
raise ValueError("SKILLS_EVAL_ENDPOINT_URL cannot be set for claude-code; "
                 "Claude uses the native Anthropic endpoint")
```

— while its credential counterpart is swallowed without a word. One unsupported input fails loudly and the other fails silently, three lines apart. If the intent is "Claude Code always uses `ANTHROPIC_API_KEY`", say so the way the endpoint says it.

**Fix direction — either is fine, but pick one and apply it to both inputs:**
- honour it: `api_key = _first(env.get("SKILLS_EVAL_API_KEY"), env.get("ANTHROPIC_API_KEY"))`, which makes the generic input mean the same thing everywhere; or
- reject it: raise for `SKILLS_EVAL_API_KEY` under `claude-code`, mirroring the endpoint check.

The first is the smaller change and the one that matches what the variable's name promises.

zac-wang-nv (changes requested):
## Review — round 5 (head `27c161a3`)

**Verdict: REQUEST CHANGES** — 2×P1, 3×P2, 3 notes.

This is a re-review after the force-push. All round-4 findings were re-verified at this
head rather than carried: every finding file is byte-identical across the rebase, and each
line anchor was re-matched by content. The rebase pulled in a `develop` rewrite of
`deploy_nemoclaw.ipynb` — the file the NemoClaw finding depends on — but that rewrite
(parallel skill-install, a model-default rename, `retry_onboard_on_fail`) leaves the
relevant guard untouched, and the offline repro still reproduces at this commit.

**The two blockers are independent of each other:**

1. **`nvidia-build` can never emit the readiness marker the adapter requires** — the adapter
   clears `NEMOCLAW_ENDPOINT_URL`, and the notebook prints `NEMOCLAW_MODEL:` only inside
   `if NEMOCLAW_ENDPOINT_URL:`. Reproduced offline with a passing control, and already
   observed in this PR's own dispatched run
   [32305638319](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32305638319).
2. **The `claude-code` default sends the gateway's model id to the native Anthropic
   endpoint** — new this round, and it affects the configuration ordinary CI runs get by
   default, not an opt-in provider.

They are the same bug class, which is the substance of note 10: `resolve_model_config`
reads 15 environment variables to serve 6 valid `(runtime, provider)` combinations, and in
both cases a value belonging to one combination reaches another. Note 10 sketches a
table-driven shape that would make those leaks structurally impossible; it is a follow-up,
not a merge condition. Finding 9 is a smaller instance of the same thing —
`SKILLS_EVAL_API_KEY` is honoured by three providers and silently overwritten by the fourth.

Note on evidence: the Skills Eval sweep currently running at this head
([32494997741](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32494997741)) reports
`runtime=nemoclaw provider=nvidia-inference` on every leg, so it exercises neither broken
path — a green result there would not clear either finding. Red legs on that run were each
attributed from their own logs; none is the readiness-marker error.

Full reasoning, repro output, and blob-level verification are in the inline comments and
the notes below.

---

### Note 10 (NOTE)

A design observation, offered as a follow-up rather than a merge blocker — but it is the reason this PR has two independent P1s, and I think it is the most valuable thing in this review.

**The credential/endpoint surface has grown past what a reader can hold.** `resolve_model_config` reads **15 distinct environment variables** to produce three values:

| output | variables consulted |
|---|---|
| model | `SKILLS_EVAL_MODEL`, `CLAUDE_CODE_MODEL`, `ANTHROPIC_MODEL`, `CODEX_MODEL`, `NEMOCLAW_MODEL`, `LLM_REMOTE_MODEL` |
| endpoint | `SKILLS_EVAL_ENDPOINT_URL`, `NEMOCLAW_ENDPOINT_URL`, `ANTHROPIC_BASE_URL`, `LLM_REMOTE_URL` |
| key | `SKILLS_EVAL_API_KEY`, `COMPATIBLE_API_KEY`, `ANTHROPIC_API_KEY`, `OPENAI_API_KEY`, `NVIDIA_API_KEY` |

Those feed **six** valid `(runtime, provider)` combinations, and the fallback chain is different for almost every one — several entries are themselves conditional (`env.get("NEMOCLAW_ENDPOINT_URL") if runtime == "nemoclaw" else ""`). The result is an imperative cascade where the *order* of `_first(...)` arguments is load-bearing and invisible at the call site.

**Both P1s in this review are the same bug class: a value leaking across combinations.**
- Finding 8 — the `anthropic` combination inherits `ANTHROPIC_MODEL`, a value that belongs to the *gateway* combinations. Correct id, wrong endpoint.
- Finding 1 — the `nvidia-build` combination sets `endpoint_url = ""`, and a downstream truthiness test reads that empty string as "no custom endpoint", skipping the marker the adapter then demands.

Neither is a typo. Both are what happens when one variable is asked to mean different things in different combinations, and nothing structurally prevents combination A's value reaching combination B.

**Cheaper structure, same behaviour:** declare the six combinations as data rather than deriving them with branches — one table mapping `(runtime, provider)` → `(model_sources, endpoint_sources, key_sources)`, resolved by a single generic function. That turns "which variable wins for nemoclaw + custom?" into a row you can read, makes the currently-invisible asymmetries (finding 9's silent drop, `nvidia-build` requiring an explicit model while `nvidia-inference` does not) visible as gaps in a column, and gives the tests something enumerable to iterate over — today's tests cover individual combinations but nothing asserts the matrix is total.

Not asking for that in this PR. But I would not add a seventh combination on top of the current shape.

---

### Note 6 (NOTE)

Round-3 verification summary — what I withdrew, what I controlled for, and what still needs a run.

**Withdrawn: round 2's second P1 (judge credential leak). Fixed, and fixed well.** `dc8a5533` removes `env["ANTHROPIC_BASE_URL"] = base_url` from `run_invocations`, pins `claude-code` to the `anthropic` provider (whose `endpoint_url` is `""`), and rejects `SKILLS_EVAL_ENDPOINT_URL` for `claude-code` outright. Re-running my round-2 repro at this head:

```
CONTROL: claude-code default
  judge sees ANTHROPIC_BASE_URL = 'https://inference-api.nvidia.com'   # coordinator's own — unchanged
  judge sees ANTHROPIC_API_KEY  = 'coordinator-key'
PR FEATURE: claude-code + custom
  FATAL: SKILLS_EVAL_PROVIDER='custom' is not compatible with EVAL_AGENT='claude-code'
```

The leak combination is now rejected before Harbor is invoked, and the remaining write (`ANTHROPIC_API_KEY = model_config.api_key`) is idempotent for `claude-code` because that provider sources the key from `ANTHROPIC_API_KEY` itself. Also withdrawn: round 2's P2 that the new tests mocked `harbor_env` to `{}` and hid the leak — `test_claude_code_uses_native_anthropic_credentials` now stubs it with a populated coordinator env and asserts the judge's base URL survives, which is exactly the test that was missing. Correcting myself on one detail from the round-2 draft: I described the escape route as `[verifier.env]` interpolation; there are no `[verifier.env]` blocks in this repo. The actual mechanism was plain subprocess inheritance of the `harbor_env()` dict. The finding was right, the mechanism sentence was not.

**Not yours — pre-existing test failures.** `python3 -m pytest tests/ nemoclaw/tests/` at `27c161a3`: **249 collected, 246 passed, 3 failed**, all `PoolCandidates::test_4090_*`. (Round 4 reported 207 passed; the rebase pulled in develop's newer tests, so the denominator moved — the 3 failures did not.) Control, **re-run against the new base** rather than carried: substituting `a01f571a`'s `run_leg.py` under head's tests fails the same three identically. Not this PR's.

**No longer unexercised — but read the result carefully.** Round 4 said Skills Eval had never run on any current commit. That changed at 14:57:52Z: the author dispatched run [32494997741](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32494997741) — 51 legs at `27c161a3`, the first eval run on this branch since `33e19497`.

**This sweep cannot settle finding 1 — now confirmed from its own logs, not inferred.** Every completed leg prints:

```
skill-eval model: runtime=nemoclaw provider=nvidia-inference model=aws/anthropic/bedrock-claude-opus-4-6
```

So the author dispatched `agent=nemoclaw` and left `provider` at `default`, which resolves to **`nvidia-inference`** for a non-`claude-code` runtime. That is the *working* NemoClaw path. The broken one is `nvidia-build`, reached only when an operator explicitly selects it — as in the failing run `32305638319`. So this run exercises the runtime finding 1 lives in while stepping around the provider that breaks it: however it ends, green or red, it says nothing about finding 1.

It equally cannot settle **finding 3**, which needs `runtime=claude-code`; this sweep is `nemoclaw` throughout.

Before merge I'd still want one explicit `nemoclaw`/`nvidia-build` dispatch (finding 1) and one `claude-code` leg (finding 3).

**Dispatch-interface change, flagged not blocked.** `runner` → `agent`, and `provider`'s default `nvidia-inference` → `default`. Nothing in `.github/workflows/` dispatches this workflow, so nothing in-repo breaks; the exposure is saved manual dispatch configs and any out-of-repo script. Under guideline 6 that's a published surface, but given it's a manual internal workflow the cost of the rename looks lower than the cost of carrying `runner` forever — noting it so the call is explicit rather than accidental.

**Repro artifacts** (worktree, not committed): `repro_build_marker.py` (finding 1, subject + control), `repro_judge_leak.py` (round 2's P1, re-run here as the withdrawal evidence).

---

### Note 7 (NOTE)

Round-5 delta. The branch was **force-pushed** — rebased onto a newer `develop` — and the round-4 merge commit was discarded. Recorded so the re-review is auditable rather than asserted, and because one round-4 claim has to be **retracted**.

**What moved.** Head `4a60c022` → `27c161a3`; merge-base `eb4e3961` → `a01f571a`; 4 commits → 3, all re-SHA'd. The rebase is not inferred from the SHA alone — the `Plan eval matrix` checkout log records `+ 4a60c0221...27c161a37 codex/skill-eval-model-selection-20260819 (forced update)`.

| commit | round 4 (orphaned) | round 5 |
|---|---|---|
| add optional agent model selection | `67fefc33` | `cab65186` |
| support custom model providers | `c836c47f` | `534e7d84` |
| clarify agent model configuration | `dc8a5533` | `27c161a3` |
| *(develop merge)* | `4a60c022` | *(discarded)* |

**Every finding file survived the rebase byte-for-byte** — so all five findings carry, and all five line anchors were re-checked by content at the new head, not assumed:

| file | old head `4a60c022` | new head `27c161a3` | new base `a01f571a` |
|---|---|---|---|
| `notebook_setup_adapter.py` | `f6e8f30c` | `f6e8f30c` | `6c915eb4` |
| `run_leg.py` | `50a6b28e` | `50a6b28e` | `2f144ad8` |
| `model_config.py` | `574dc4a3` | `574dc4a3` | *(absent — new in PR)* |
| `skills-eval.yml` | `585c8faf` | `585c8faf` | `e576f697` |
| `nemoclaw_brev_env.py` | `d1d24b3f` | `d1d24b3f` | `3647c14e` |
| `deploy_nemoclaw.ipynb` *(mechanism file, not the PR's)* | `29bf213f` | **`8cf88434`** | `8cf88434` |

**The one row that moved is the interesting one.** `deploy_nemoclaw.ipynb` is where the P1's third link lives, and develop rewrote it under the PR. Its new blob is *identical to the new base*, which proves the change is develop's, not the author's. The rewrite is parallel skill-install (`SKILL_INSTALL_WORKERS`), a `NEMOCLAW_MODEL` default rename to `nvidia/nemotron-3.5-lightning-30b-a3b`, and `retry_onboard_on_fail: False → True`. It leaves cell 12's guard verbatim, so the P1 stands — but on evidence re-gathered at this head, not carried:
- `repro_build_marker.py` at `27c161a3` — subject (`nvidia-build`) passes the provider gate and **raises** on the model gate; control (`custom`) passes both.
- Cell 12 read directly out of the new blob still shows `print("NEMOCLAW_MODEL:", …)` nested inside `if NEMOCLAW_ENDPOINT_URL:`.
- `pytest tests/ nemoclaw/tests/` → **249 collected, 246 passed, 3 failed** (`test_4090_*`), with the control re-run against the **new** base's `run_leg.py` failing the same three identically.

*Side effect worth knowing, and it did not hold up:* I expected develop's `retry_onboard_on_fail: True` to address the **other** failed leg in run 32305638319 — the `l40s-3` sandbox-onboarding failure I called genuine infra. The sweep now running at this head still produced `Sandbox 'skill-eval' entered Error phase before it became ready` on `vss-eval-l40s-1g`, so onboarding failures survive the retry flag. Either way it is develop's territory, not this PR's, and it does not touch finding 1.

**RETRACTED — round 4's red `SonarQube Scan (ui)`.** `SonarQube Scan (ui)` is **green at `27c161a3`**. The red lived only on the now-discarded merge commit, exactly as round 4 predicted it would ("worth one re-run"), and the force-push supplied that re-run. Round 4's DM flagged it as "not yours"; that call was right, and the condition is now gone. Nothing to action.

**Board at head (polled 15:14Z, 67 checks): 16 success · 5 failure · 8 in progress · 37 queued · 1 skipped.** The board is mid-flight, so treat the counts as a timestamped snapshot rather than a verdict — at 15:05Z it was 0 red, and five eval legs went red in the nine minutes it took to write this note.

**All six reds attributed from their own logs — none is finding 1, and none is this PR's diff.** (The count moved 5 → 6 while this note was being written; the sixth is the last row.)

| leg | failure |
|---|---|
| `vss-ask-video · direct_vlm_video_understanding · L40S` | `Sandbox 'skill-eval' entered Error phase before it became ready` on `vss-eval-l40s-1g` — sandbox onboarding, infra |
| `vss-ask-video · base_profile_video_understanding · L40S` | step-1 `RuntimeError`, steps 2-5 not-run |
| `vss-build-vision-agent · profile_an_1_stored_video_summarization_api` | `NonZeroAgentExitCodeError`, reward 0.2, 2/10 checks |
| `vss-build-vision-agent · profile_in_2_rt_cv_person_detection_harbor` | `NonZeroAgentExitCodeError`, reward 0.364, 4/11 checks |
| `vss-build-vision-agent · profile_an_1_stored_video_summarization_runtime_harbor` | `NonZeroAgentExitCodeError`, 0/5 checks |
| `vss-build-vision-agent · profile_at_1_alert_verification` | `NonZeroAgentExitCodeError`, reward 0.154, 2/13 checks |

Note what is **absent**: not one of them is `deploy_nemoclaw.ipynb completed without readiness marker`. That is the expected shape — this sweep runs `nvidia-inference`, so it never enters the `nvidia-build` branch that produces that error. Four are `NonZeroAgentExitCodeError` agent-quality failures with partial rewards; two are sandbox/runtime failures on `L40S`. Worth watching, but they belong to the eval fleet, not to this diff. Because legs are still completing, expect this table to grow — the durable claim is the *absence* of the readiness-marker error, not the count.

**Skills Eval is running at head for the first time.** Round 4's "never run on any current commit" no longer holds — see NOTE 6 for why a green sweep still would not clear finding 1.

## Instructions
- Change VERDICT to RESOLVE / ACK / FOLLOW-UP / SKIP as needed
- Change the severity (P0 / P1 / P2 / NOTE) on any comment heading — the posted body picks
  up whatever you leave there
- Edit any reply or comment text directly
- Delete any comment block to skip it
- Save the file, then tell Claude "post" to execute

gbau1:
Superseded by the focused, develop-based model-routing PR #2160. The replacement incorporates the review feedback from this PR and excludes the retired notebook path and unrelated changes.

PR #1765, opened 2026-08-20
feat(cli): add the `vss vios` media plane

zac-wang-nv:
## What

Adds `vss vios` — the media plane — as six commands: `list`, `timeline`, `clip`, `snapshot`, `add`, `delete`. Media is addressed by **sensor name**; `sensorId`/`streamId` resolution happens inside the library.

```
vss vios list     [--type video|stream] [--sensor NAME]
vss vios timeline --sensor NAME
vss vios clip     --sensor NAME [--start-time T --end-time T]   → media_url + the window served
vss vios snapshot --sensor NAME [--at T]                        → media_url (live / replay)
vss vios add      --type video|stream SOURCE [--name NAME]
vss vios delete   --type video|stream --sensor NAME
```

## Why it is not a command group

VIOS operations are **not VSS processing**: they run no model and produce no evidence, they resolve handles and mint URLs. So they mint no `job_id`, write no memory record, and emit no completion marker, and the `run`/`status`/`get`/`list` grammar does not apply. `CommandGroup.cli()` is `@final` and would mount all four verbs, so `vios` follows the `CommandGroupSpec` protocol directly — as `vss configure` already does.

This **supersedes R-C1** ("all functions … including vios media ops … are modeled as jobs", directed 2026-07-20) and the SDD §1.3 definition of Job, which named "a clip export". The SDD delta is written up separately and applied by hand to the design doc; it is recorded as a supersession with a revision entry, not a reworded requirement.

What `vios` *does* keep from the framework is the part that should be uniform: every command declares `requires={"vst"}`, so a deployment without that route fails with the same wording and the same exit 4 as `vss search`.

## Commits

| | |
|---|---|
| `1e5e3a31c` | rename `vss_core.vst` → `vss_core.vios` (11 files, 23 refs, plus CI and one skill snippet that imported it) |
| `21221eb28` | promote `require_services` / `context_from` / `emit` / `guarded`; generalise the first from `Action` to `(name, requires)` |
| `d2aa91a20` | port the media-plane operations into `vss_core.vios` |
| `46a965bc3` | the `vss vios` group |
| `6301a6135` | migrate three skills off hand-rolled curl |
| `fe5ad05e2` | sync context manager on the upload path (mypy caught it) |

## Three behaviours worth reviewing closely

**Ids are read, never built.** `sensorId` comes from `/sensor/list` matched on `.name`. Auto-discovered files carry a `_N` uniqueifier on the id but not the name, PUT-uploaded files get a fresh UUID, POST-uploaded ones sometimes report an empty string — `/sensor/<name>/streams` answers `CameraNotFoundError` for two of the three.

**A multi-stream sensor with no `isMain` flag is a hard failure**, not a first-match. Resolving to a substream yields lower-resolution frames with no error anywhere, which the VLM then answers from. Where a sole unflagged stream is used, the result says so.

**Delete keeps 404-as-success but confirms absence by name.** Storage deletion can cascade the sensor registration away before the paired call, and counting that 404 as failure downgrades a clean teardown to `partial`. But an *id-keyed* absence check passes while VIOS still lists the source under its canonical name — the fail-open flagged in review of #1650. So absence is confirmed by re-listing and matching `.name`.

## Skill migration

`vss-ask-video` Path B was ~30 lines of bash that resolved a streamId, polled timelines, picked a segment, minted a URL, unpicked VIOS's doubled scheme, re-anchored the host and warmed the lazy render. That is `vss vios clip --sensor <name>` now.

`vss-generate-video-report` mints its clip URL the same way; `vss-manage-alerts` resolves sensor ids through `vss vios list --type stream`.

**`vss-manage-video-io-storage` is deliberately untouched** — it is the REST contract for the surface the CLI does not cover (WebRTC, proxy, recorder, scan, settings), and other skills read it directly. NvStreamer is out of scope: different service, different port, its own API.

Diagnostic curls in `troubleshooting.md` / SOP references stay as curl. When you are debugging why VIOS is broken, a CLI wrapper over it tells you less than the raw status code.

## Verification

- `ruff check` clean; `mypy` clean on all 15 source files
- **823 tests pass** under the CI flags (`-W error::RuntimeWarning -W error::ResourceWarning`), including 29 new ones
- `vss vios --help` renders; `vss --help` lists the group; an unconfigured deployment exits **4** with a diagnostic on stderr

New tests assert the two properties that matter for a surface whose job is handing URLs to another tool: the job verbs are absent, and an empty listing is distinguishable from a backend failure (exit 0 with `count: 0` vs exit 3 with stderr).

## Not done here

- No live run against a deployed box yet — the eval scenario is the acceptance gate and has not been run.
- `develop` is currently red on `ruff format` for an untouched `packages/vss_core/tests/unit_test/memory/test_memory.py`; that hook was skipped for these commits rather than reformatting someone else's file.
- The `"vst"` service key and `/vst` ingress mount keep their names — those identify the route, not the module.

## Plan

Design settled by grilling before any code; every question, the recommendation, and the answer.

| # | Question | Recommended | Answered |
|---|---|---|---|
| 1 | Copy the VST lib from `vss_agents`, or extend `vss_core`? | Extend `vss_core` — it already exists and is the CLI's only dep; `vss_agents` pulls `nvidia-nat` | Extend, with all required functions; rename to `vss_core.vios`; no duplicate timeline/delete |
| 2 | What is the read surface? | 5 reads | `list --type video\|stream` + `timeline` |
| 3 | Do `clip`/`snapshot` mint jobs? | Yes, marker now, persistence later | **No — VIOS ops are not VSS processing** |
| 4 | Which mutations? | `upload`, `sensors add`, `sensors delete` | `add --type`, `delete --type`; clip/snapshot return a URL only |
| 5 | Headless provision/fan-out in scope? | Out | Out |
| 6 | Name-resolution edges | ambiguity ⇒ exit 2; UUID fallback; isMain → sole → exit 2 | (a) + UUID fallback |
| 7/14 | Migrate skills in this PR? | Separate PR | **This PR**, so skill eval can run |
| 8 | What happens to `vss_agents/tools/vst`? | Repoint in a follow-up | Leave untouched |
| 9 | Job-less group mechanism | `CommandGroupSpec` directly | Yes, plain `click.Group` |
| 10 | SDD knock-ons | Fix all five | New delta doc, applied to the Doc by hand |
| 11 | `--type` semantics | Provenance (file-backed vs RTSP) | Provenance; optional on `list` |
| 12 | Which delete survives? | `delete_vst_sensor` semantics + `error_message` parsing, no env read | That, plus name-based addressing with stream-UUID fallback |
| 13 | Rename in this PR? | Yes, first commit | Yes |
| 15 | Migration scope | Skills whose calls fit the six commands | All VST-curl skills **except** `vss-manage-video-io-storage` |
| 16 | Eval expectations | Rewrite for migrated flows | Rewrite |
| 17 | `api-reference.md` | Leave whole, add a CLI header | Leave it, with a header |
| 18 | NvStreamer | Out | Out |
| 19 | Where does the SDD delta live | Vault, with pointers to the Doc | Yes |
| 20 | Migrate diagnostic curls too? | No — raw curl is the right tool when VIOS is what is broken | Leave curls |
| 21 | Reuse or duplicate the framework helpers? | Promote them | Promote to framework level |
| 22 | Command signatures | clip defaults to covering segment; `--at` ⇒ replay; joined `list` rows | All three; `--type` optional |

Findings from the review of #1650 that changed the design: the `_N` suffix / read-ids-never-build-them rule, and the fail-open absence check. Both are implemented and tested here.


zac-wang-nv:
## Code review round — Codex, and what it found

Ran the diff through the OpenAI Codex CLI as an independent reviewer. It found real defects, several of them the exact failure mode this surface exists to prevent: reporting a clean result for work that did not happen. All confirmed against the code and fixed in `44f39cd33`, each with a regression test.

**P1 — delete could leave recordings and report `confirmed: true`.** The timeline read was wrapped in `except Exception` and treated as "no recordings". For an RTSP source the sensor delete had already run, so `confirm_absent` passed. Backend outage, malformed JSON and genuinely-no-recordings all collapsed into the same success. Now: span read *before* the sensor delete, read failures propagate, and "nothing recorded" is `recordings: "none"`.

**P1 — delete covered only the first segment.** `get_timeline` returns `timeline_list[0]`; the implementation this was ported from used `min(startTimes)`/`max(endTimes)` across every segment (`vst/utils.py:300`). A burst-recording stream kept everything after segment one. Now uses `get_timelines_map`, which already computes that envelope.

**P1 — `list` turned a backend outage into factual absence.** Any sensor whose streams call failed was skipped, so a total VIOS failure returned `count: 0` at exit 0 — contradicting the module comment claiming the two could never be confused. Failures propagate now.

**Exit codes did not match the design.** Unknown handle raised a bare `LookupError` → untyped exit 1; ambiguity raised `VSTError` → exit 3 (backend unreachable) for what is a caller error. Now `VIOSInvalidInputError` → 2 and `VIOSNotFoundError` → 5.

Also fixed: absent stream url classified as `video` (would pick the wrong teardown flow) → `unknown`, and delete refuses; `add_stream` raised after a *successful* mutation when VIOS omitted the id, so a retry would duplicate — now re-resolves by name; `--sensor` accepts a streamId, which the help advertised and the resolver did not implement; `timeline` reports the full envelope; `--name` on a video upload errors instead of being ignored; `SensorRef.has_timeline` removed; three "after retrying transport errors" messages that were on single-request paths corrected.

**Where I disagree.** Codex argues the job-less classification is taxonomy invented to escape R-C1, and that blocking clip export and upload deserve separate analysis from the reads. The reasoning is fair, but the decision was directed in design review, and it is recorded as a **supersession** of R-C1 with a revision entry — not a reworded requirement. Its narrower point that `CommandGroup.primitives` was an available middle path is true and was considered; a mixed surface was rejected because a group with one job verb and five non-job commands is harder to explain than a group with none.

It also could not verify the "823 tests pass" claim, correctly — it had no execution access. Current state is **837 passing** under the CI flags, plus clean `ruff` and `mypy`; CI on this PR is the independent check.

Verification after the fixes: `ruff check` clean, `mypy` clean on 15 files, 837 tests pass with `-W error::RuntimeWarning -W error::ResourceWarning`, unconfigured deployment still exits 4.

zac-wang-nv:
## Greptile P1 confirmed and fixed — `45f964f44`

> Malformed timelines confirm clean deletion — `client.py:207-215`

Correct, and it is the same fail-open the previous round fixed, one layer deeper. Last round I made `delete_media` stop swallowing timeline *exceptions*; Greptile found the **HTTP 200 with an unusable body** path, which never raised at all.

`get_timelines_map` is lenient by construction: a non-dict payload yields `{}`, and a segment without parseable times is skipped. So "VIOS answered 200 with garbage" arrived at `delete_media` as `span is None` → skip the storage delete → return `recordings: "none"`, `confirmed: true` with recordings still on disk.

**The fix is not to change `get_timelines_map`.** Its leniency is right for its other caller — search uses it for best-effort screenshot enrichment, where an unreadable timeline should degrade to unmapped rather than fail the search. Changing it there would be a regression in a path this PR has no business touching. So `delete` gets a strict reader instead:

```
recorded_span()  # absent from the listing  -> None (genuinely no recordings)
                 # present but unparseable  -> VSTError, refuse to claim a clean delete
                 # otherwise                -> (min(startTimes), max(endTimes))
```

**Second finding from the same review** — "stream-ID fallback suppresses VIOS request failures" — also confirmed. My `except VSTError: continue` in the streamId scan meant a deployment where every streams call failed reported `VIOSNotFoundError` (exit 5, *"no such sensor"*) instead of the backend error. It now keeps the first failure and raises it if the scan finds nothing: a completed search that found nothing and a search VIOS could not serve are different answers, and the caller acts differently on each.

**On the two gaps Greptile named in the tests** — "does not cover failed stream-ID scans or malformed successful timeline payloads" — both now have regression tests, along with three more for the strict reader:

- malformed segments (200, no usable times) → error, not a clean delete
- non-dict payload → error
- stream absent from the listing → genuinely `None`
- span covers **every** segment, not the first
- a failed scan surfaces as the backend error, not as a missing sensor

842 tests pass under the CI flags; `ruff` and `mypy` clean. Re-running an independent review over the full diff now — the two earlier rounds each found a real fail-open, which is a good argument for not stopping at two.

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32517539203)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32523866344)

services/agent/packages/vss_core/src/vss_core/vios/client.py
```diff
@@ -0,0 +1,1359 @@
...
+        if moment < span_start:
+            raise VIOSInvalidInputError(f"{label} {_isoformat(moment)} is before the recording starts; {window}")
+        if moment > span_end:
+            raise VIOSInvalidInputError(f"{label} {_isoformat(moment)} is after the recording ends; {window}")
+    if first > last:
+        raise VIOSInvalidInputError(f"--start-time {_isoformat(first)} is after --end-time {_isoformat(last)}")
+    return _isoformat(first), _isoformat(last)
+
+
+async def recorded_span(
```
liamy-nv: **Recording gaps are treated as valid video — P1**

- recorded_span() collapses every recorded segment into one envelope. Both timeline and clip then use that envelope: timeline returns only one start/end pair; clip validates against the envelope (vios_group.py lines 106–128 uses that envelope for timeline output and clip validation)
- recordings at 12:00–12:10 and 13:00–13:10 become 12:00–13:10. A request for 12:30–12:40 is incorrectly accepted even though no recording exists

Suggested fix: Replace recorded_span() with something like recorded_segments() -> list[RecordedSegment]

services/agent/packages/vss_core/src/vss_core/vios/client.py
```diff
@@ -0,0 +1,1359 @@
...
+    """
+    timeout = aiohttp.ClientTimeout(total=timeout_seconds)
+    for _ in range(attempts):
+        try:
+            async with (
+                aiohttp.ClientSession(timeout=timeout) as session,
+                session.get(media_url) as response,
+            ):
+                if response.status == 200:
+                    await response.read()
```
liamy-nv: **Clip warm-up buffers the entire video**
- aiohttp.ClientResponse.read() reads the entire response body into memory. If the clip is 5 GB, the CLI may try to allocate roughly 5 GB. Because warm-up permits three attempts, a timeout or intermittent failure could repeat the operation.
- response can be consumed incrementally, for example through `response.content.iter_chunked(...)`, without retaining the bytes. The retries should also share one bounded overall deadline so warm-up remains deterministic and bounded.

services/agent/packages/vss_cli/src/vss_cli/vios_group.py
```diff
@@ -0,0 +1,400 @@
...
+        return Result(
+            body={"error": f"{source!r} is a {kind} source, not a {declared}", "type": kind},
+            exit=Exit.INVALID_INPUT,
+        )
+    if kind == "stream":
+        name = values.get("name") or source.rstrip("/").rsplit("/", 1)[-1]
+        sensor_id = _run(vios.add_stream(origin, source, name))
+        return Result(body={"name": name, "sensor_id": sensor_id, "type": "stream", "added": True})
+
+    path = pathlib.Path(source)
```
liamy-nv: HTTP(S) upload is classified as supported but cannot execute so only local file is supported
- This path cannot handle the HTTP(S) video sources that `classify_media_source()` and VIOS-1 say are supported.
- An input such as `https://example.com/video.mp4` is classified as `video`, converted to `pathlib.Path`, and then rejected by `upload_media()` because it is not a local file.
- Fix: implement bounded streamed HTTP(S) fetching followed by upload—with redirect, size, timeout, and SSRF controls—or remove HTTP(S) from the supported contract and classifier. The current state advertises a path that deterministically fails.

skills/vss-ask-video/evals/base_profile_video_understanding.json
```diff
@@ -40,16 +40,16 @@
     {
       "query": "Is the worker in warehouse_safety_0001 wearing PPE?",
       "checks": [
-        "The run performed at least one successful GET to a VST clip-URL endpoint of the form http://localhost:30888/vst/api/v1/storage/file/<streamId>/url with startTime and endTime query params and returned a JSON body containing videoUrl (the clip is sourced through /vss-manage-video-io-storage).",
+        "The run obtained the clip URL through `vss vios clip --sensor <name>` (which resolves the sensor by name, reads the recorded range, and mints the URL in one call) and used its `media_url` — or, equivalently, performed a successful GET to http://localhost:30888/vst/api/v1/storage/file/<streamId>/url with startTime and endTime query params. It did not hand-build the window from a separate timelines read.",
```
agmohare: The REST branch offered here as "equivalent" can't be reached without doing the thing the
last sentence forbids. Against live VIOS on this deployment:

  GET /storage/file/<streamId>/url                    → InvalidParameterError
                                                        "Start and end time are required"
  GET …/url?startTime=<invented>&endTime=<invented>   → VMSNoDataError "No valid stream found
                                                        for given timestamps, please check
                                                        timelines using /api/v1/storage/timelines"
  GET …/url with bounds read from /storage/timelines  → 200 + videoUrl

The bounds are mandatory, and VIOS's own 404 tells you to get them from /storage/timelines —
which "It did not hand-build the window from a separate timelines read" then disallows. So an
agent that takes the REST branch in good faith can still be failed on the trailing clause,
and `vss vios clip` is the only route that reliably passes.

If the CLI is the intent, say so and drop the REST branch — but then the skill needs the same
edit, since the sensor-sourced block in Step 1 and the /vss-manage-video-io-storage handoff
still present the REST call as fine. If both routes are meant to pass, drop the last sentence.

Same text on line 52, so it needs fixing in both places.

liamy-nv (commented):
Added some comments

skills/vss-ask-video/evals/base_profile_video_understanding.json
```diff
@@ -40,16 +40,16 @@
...
       ]
     },
     {
-      "query": "At what timestamp did the worker climb up the ladder in warehouse_safety_0001? Answer from the clip recorded on that VIOS sensor \u2014 not from a local copy of the file and not from any pre-extracted frames left on this host.",
+      "query": "At what timestamp did the worker climb up the ladder in warehouse_safety_0001? Answer from the clip recorded on that VIOS sensor — not from a local copy of the file and not from any pre-extracted frames left on this host.",
       "checks": [
-        "The run performed at least one successful GET to a VST clip-URL endpoint of the form http://localhost:30888/vst/api/v1/storage/file/<streamId>/url with startTime and endTime query params and returned a JSON body containing videoUrl.",
+        "The run obtained the clip URL through `vss vios clip --sensor <name>` (which resolves the sensor by name, reads the recorded range, and mints the URL in one call) and used its `media_url` — or, equivalently, performed a successful GET to http://localhost:30888/vst/api/v1/storage/file/<streamId>/url with startTime and endTime query params. It did not hand-build the window from a separate timelines read.",
```
agmohare: same as above in line 43

services/agent/packages/vss_core/src/vss_core/vios/client.py
zac-wang-nv: @-

services/agent/packages/vss_core/src/vss_core/vios/client.py
zac-wang-nv: @-

services/agent/packages/vss_cli/src/vss_cli/vios_group.py
zac-wang-nv: @-

skills/vss-ask-video/evals/base_profile_video_understanding.json
zac-wang-nv: @-

skills/vss-ask-video/evals/base_profile_video_understanding.json
zac-wang-nv: @-

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32682564914)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32682915170)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32683063421)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32684011278)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32684122198)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32689033618)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32691395019)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32693589586)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32697685738)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32703665711)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32710476926)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32717783154)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32724236761)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32730107191)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32737561910)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32740803111)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32750552043)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32750794446)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32753852733)

services/agent/packages/vss_core/src/vss_core/vios/client.py
```diff
@@ -0,0 +1,1588 @@
...
+    # function reported a confirmed cleanup.
+    span = await recorded_span(vst_internal_url, ref.stream_id, timeout_seconds)
+
+    sensor_url = f"{base}/vst/api/v1/sensor/{quote_path_segment(ref.sensor_id)}"
+
+    if ref.kind == "stream":
+        await _delete(sensor_url, timeout_seconds, "sensor delete")
+        steps.append("sensor")
+
+    if span is not None:
```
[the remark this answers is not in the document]
liamy-nv: @zac-wang-nv I think this comment is worth looking at; deleting the sensor registration does not guarantee that the physical file was removed

potential fix: wait/retry for the timeline and perform the storage deletion, locate and delete the stored file through another VIOS storage API, or return a partial/error result that explicitly says storage deletion could not be confirmed

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32762864894)

liamy-nv (dismissed):
LGTM

PR #1766, opened 2026-08-20
feat(vss-core): VSS CLI memory

liamy-nv:
## Summary
- Persist `vss search run` jobs as parent records plus `search_hit` children, with `--persist` / `--no-persist` and exit 6 when search succeeds but memory writes fail.
- Switch summarize persistence to the same parent/child bundle model and package `search` / `summarize` as their own CLI command groups.
- Add cross-group `vss memory` (`upsert`, `get`, `query`, `events`) against the configured unified-memory index, without going through a job group.
- Hydrate ordered child records on group `get` (events / search hits) while leaving `status` and `list` parent-only.
- Emit a compact last-line job completion marker on `run`, and close search persist failures instead of leaving a stranded `submitted` parent.

## Checklist


liamy-nv:
/ok to test [a41fbdb](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1766/commits/a41fbdb676a73cf67670256f97f765ad0e966321)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32801382549)

services/agent/packages/vss_cli/src/vss_cli/group.py
```diff
@@ -444,21 +448,44 @@ def _context_from(values: dict[str, Any]) -> Context:
...
-    """Render a Result and carry its exit code out through Click."""
+def _emit(result: Result, ctx: Context, *, marker_group: str | None = None) -> None:
+    """Render a Result, append the §7.2 marker, and carry its exit code."""
     import json
 
     if result.body is not None:
         pretty = bool(ctx.pretty)
         text = json.dumps(result.body, indent=2 if pretty else None, default=str)
         click.echo(text)
+    if marker_group is not None and result.job_id:
```
zac-wang-nv, line 501: **P1** — Carried from round 1, and it is now the **only** thing breaking the E2E suite — the other cause was fixed in `e7afcff6e`.

This makes `run`'s stdout **two** JSON documents, not one with an extra field. Anything doing `json.loads(stdout)` on a `vss <group> run` gets `Extra data`. There is an in-repo consumer that asserts otherwise and is not in this diff: `test_search_archive_e2e.py:677`'s `_only_json_object` does `assert len(lines) == 1`. That file is byte-identical at the merge base and at head (`d679d65019e9fab9bcddf55372f77ae4482c972f` both sides), so the PR breaks a test it never touches.

At this head, all six failures land on the shape assertion with the exit-code assertion passing:

```
packages/vss_core/tests/unit_test/lib/search_core/test_search_archive_e2e.py:679: AssertionError
E  assert 2 == 1
E   +  where 2 = len([{'data': [...]}, {'event': 'vss_job_completed', 'group': 'search', ...}])
6 failed, 4 passed
```

I want to be precise about the CI status: the required gates are **not dispatched** at this head, not failing — `a41fbdb67` carries 2 check-runs against 15 at the previous head, and the combined status API reports `total_count=0`. So this is a locally reproduced break, not a red board you can point at right now. It will become one when the gates run.

**Correcting my own round-1 recommendation:** I suggested moving the marker to stderr or behind a `--marker` flag. The repo has already decided against that, and I should have checked before suggesting it — `skills/vss-summarize-video/evals/lvs_profile_summarize.json` `checks[6]` grades on "its final stdout line is a single compact JSON object carrying a `job_id`". The marker on stdout is the intended contract, and it is graded executably.

So the fix belongs in the E2E helper, not in `_emit`: `_only_json_object` should take the *first* JSON document, or the suite should assert `lines[0]` and `lines[-1]` separately. What the PR is actually missing is that it changed a stdout contract with two in-repo consumers and updated one of them. Worth a line in the PR description either way — `json.loads(stdout)` breaking is a caller-visible change even when the marker is intended.

Note `group.py` is one of the two files currently conflicting with `develop`, so this and the conflict resolution will land together.

zac-wang-nv:
**P1** — re: `skills/vss-summarize-video/evals/lvs_profile_summarize.json` line 34 (not in this PR's diff, so GitHub cannot anchor an inline comment here)

Inverting the persistence default was the right fix for round 2's finding, but this eval spec still encodes the contract it replaced, and it is red on this head's board.

`vss-summarize-video · lvs_profile_summarize · RTXPRO6000BW` reports:

```
DONE: 0/1 specs passed; step-1 1.0 (7/7), step-2 0.86 (12/14)
      — 2 persist-related checks failed (persisted=false in CLI completion marker)
```

Three checks in `expects[1]` assume that not passing `--no-persist` is sufficient to persist:

- `checks[5]` — "It also does not pass `--no-persist`, so the summary is persisted to unified memory"
- `checks[6]` — "…plus a `persist` object whose status is `complete`"
- `checks[9]` — "The `persist` object in that same marker accounts for what was written: its `events` count equals…"

After `e7afcff6e` none of that follows. `effective_persist` returns `False` unless the deployment carries `memory.enabled` **and** `memory.persist_by_default`, and the eval's step-1 deploys the lvs profile without ever running `vss configure memory --enable --persist-by-default`. So the marker comes back `persisted=false` with no `persist` object at all — which your own `SKILL.md:388` now documents as the correct outcome: *"An absent `persist` object with `persisted=false` means static policy chose stdout-only execution."*

That is the tell that this is a missed surface rather than a disagreement: `e895e6dce` aligned `SKILL.md`, `cli_usage.md` and `end-to-end-example.md` with static policy, and the eval spec next door was not in that commit. The docs describe the new world; the checks still grade against the old one.

Two ways out, and they are genuinely different:

1. If a default lvs deployment *should* persist, the gap is in the profile — the eval's deploy step needs the `vss configure memory` call, and so does whatever ships that profile.
2. If stdout-only is the right default, `checks[5]`, `[6]` and `[9]` need rewriting against `persisted=false`, and step-2's prompt (which instructs the agent to "read the `job_id` and the `persist` object from the run's completion marker") needs to stop requiring an object that will not be there.

Worth being explicit about which, because it is the same question as "what does a fresh install do by default", and the eval is currently the only place in the repo that answers it executably.

services/agent/packages/vss_cli/src/vss_cli/search/group.py
```diff
@@ -416,11 +535,185 @@ async def _go() -> Any:
...
+                bundle = _search_terminal_bundle(
+                    job_id=job_id,
+                    created_at=created_at,
+                    input_data=input_data,
+                    output=output,
+                    search_mode=action,
+                )
+                result = memory.service.upsert_bundle(bundle)
+                persist_meta = result.to_dict()
+                if not result.ok:
```
zac-wang-nv, line 575: **P1** — Carried from round 1; re-verified unchanged at `a41fbdb67`. `PersistResult.ok` is not the right predicate for "everything the caller was shown made it into memory", so this success path can report a complete persist for a job that is stored as `partial`.

`upsert_bundle` computes `expected = len({storage_id_for(r) for r in records})` — **distinct** storage ids. When two children derive the same `record_id`, they collapse to one document, `expected` and `written` both drop to match, and `ok` stays `True`. The service does notice (`children_collapsed` → `_parent_after_partial_children`) and rewrites the parent, but nothing tells the CLI. Reproduced at this head (`repro_collision.py` in the review worktree) with two search hits carrying identical identity-bearing fields:

```
bundle children record_ids: ['hit-074d0add04d43ab5', 'hit-074d0add04d43ab5']
upsert_bundle -> ok=True  expected=2 written=2 failed=[]
what memory actually holds:
   parent.job.status      = 'partial'
   parent.output.ext      = {'search_mode': 'embed', 'result_count': 1}
   child search_hit count = 1
```

So `vss search run` prints two hits, `"persisted": true`, `"persistence": {"expected": 2, "written": 2, "failed": []}`, and a marker of `{"status": "completed", "persisted": true, "exit_hint": 0}` — while `vss search status --job-id …` on the very next line answers `partial` with `result_count: 1`.

`summarize/group.py:462` has the same shape (`if not persist.ok: raise RuntimeError(...)`), and it surfaces `expected`/`written` in the body, which reinforces the false story.

Reachability is the honest caveat: it needs two rows agreeing on every field in `_search_digest_payload` (timestamp, end_time, description, sensor_id, score, object_ids, frame_ids, media_url). `--no-merge-adjacent` raw chunks from one sensor, or a VLM emitting a repeated event, both get there. Cheapest fix is to have the CLI compare `len(bundle.children)` against distinct persisted children — or have `PersistResult` carry a `collapsed` count so `.ok` keeps meaning "what I asked for is what is stored".

services/agent/packages/vss_cli/src/vss_cli/group.py
liamy-nv: Agreed — the marker stays on stdout. _emit is unchanged. _only_json_object in test_search_archive_e2e.py now splits result vs marker (lines[0] / lines[-1]) and asserts the search completion marker. cli_usage.md documents the two-document contract so json.loads(stdout) callers are not left guessing.

liamy-nv:
re: skills/vss-summarize-video/evals/lvs_profile_summarize.json line 34 - LVS eval persistence default:

Fresh lvs deploy stays stdout-only until policy is set. Step 1 now runs vss configure memory --enable --backend elasticsearch --index vss-memory --persist-by-default. Checks [5]/[6]/[9] grade persist on the first stdout document and job_id/persisted on the final marker. The eval preamble and contract tests match that split.

services/agent/packages/vss_cli/src/vss_cli/search/group.py
liamy-nv: PersistResult now carries requested (upsert call sites). Completeness is requested == expected == written with no failures; collapsed is serialized. Duplicate search hits and events take the existing partial path: displayed results stay, stored parent is partial, marker persisted=false, exit 6. Summarize also forces persisted_override=False on that path so a closed-but-partial record is not reported as fully persisted.

liamy-nv:
/ok to test [7ae3b55](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1766/commits/7ae3b55bd381636a90fe7d8b014ee902e9238dc5)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33011556028)

liamy-nv:
/ok to test [82f5c96](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1766/commits/82f5c96ffd939bec72acbd18c808209c84820150)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33022889004)

zac-wang-nv (approved):
LGTM

PR #1779, opened 2026-08-20
ci(skill-eval): render the leg result comment without an LLM

simings-nv:
## Description

After `run_leg.py` returns, the outer coordinator agent read the results tree and rebuilt the result table by hand, following ~126 lines of `AGENTS.md` that described the layout and the `jq` needed to pull turns and token counts out of each trajectory. Across real Actions job logs that is p50 **114 s** over a median of 11 read-only tool calls, redone from scratch on every leg because the agent has no memory of the schema it parsed last time. Composing and posting the comment is another 7 s, so 94% of the phase is rediscovery. None of it is a judgement, so none of it needs a model.

This replaces it end to end:

* **`leg_report.py`** (new) collects the tree once and renders the same comment deterministically at p50 **0.59 ms**, and writes a versioned machine-readable summary (`leg-summary.json`) beside it.
* **`skills_eval_agent.py`** now hands the agent ONE command instead of the algorithm, tells it not to read the results tree, trajectories or `judge.json` itself, and points it at `leg-summary.json` for the verdict. Output goes to `$SCRATCH/pr-<spec>.md`, which is where the existing benchmark step already globs for it.
* **`AGENTS.md`** drops 1011 → 933 lines: the algorithm becomes a pointer at the renderer plus a table of the summary states. That block was in the *system prompt of every leg*, so it was both re-read and re-implemented each time.

Measured effect per leg: −114 s of wall-clock and ~958 fewer system-prompt tokens. Over the 1,177 legs in the last 8 days that is **−37.3 h** of runner time.

### The module emits no verdict

It renders what Harbor recorded; the coordinator decides pass/fail from the summary. An earlier draft emitted a verdict too, and that one extra responsibility produced every serious defect review found: two real `AgentTimeoutError` trials rendered green, steps were counted while being labelled specs, and step accounting flipped 316 real single-step legs falsely red.

Rendering is fail-closed. `row_state()` returns `recorded-pass` only for a finite non-boolean reward exactly `1.0`, no exception, and judge evidence that is absent or well-formed and unanimous. An unreadable `judge.json` is not "no judge". Check counts must be real ints with `total > 0`, so `{"passed": 1, "total": 0}` and string counts are `ambiguous` rather than green. Multiple attempts where any lacks a timestamp are `ambiguous` rather than ordered, because undated-sorts-last is a convention and not a chronology. A spec that is named but unusable raises `SpecError` and exits 3 instead of silently collapsing a partial chain into a complete-looking comment.

The Query column shows what Harbor actually evaluated. Adapters substitute placeholders before execution, so `judge.json.query` is authoritative and the spec holds only a template; 46 of 50 specs contain placeholders and 480 replayed rows displayed a raw `{{platform}}`. Declared text is used only where no judge query exists, is labelled `(declared)`, and has `{{platform}}` filled in.

### Verification

Replayed over **1,177 real legs** (2026-08-13..20): 1,174 rendered, 0 exceptions. Per profile, against ground truth recomputed from the raw files, **0 mismatches** across alerts (320), detection (137), search (112), base (89), lvs (75), warehouse (42), ask-video (22), calibration (19). Fidelity against the coordinator's own recorded verdicts: 6/6.

The replay is also what found the run-level `result.json` being mistaken for a trial, mixed aware/naive timestamps, and per-step retries.

`ruff` clean; copyright headers clean (7,411 files); the exact CI skill-eval command passes 108 tests.

### Scope note

Dataset generation still goes through the agent, because each adapter exposes a different CLI (`--skill-dir`, `--deploy-skill-dir`, `--spec`, `--platform` in varying combinations). Normalising that is separate work.

The new tests are registered in `ci.yml`, whose skill-eval step lists files explicitly; without that they would have existed and never run.

## Checklist

### DIFF
--- a/.github/skill-eval/AGENTS.md
+++ b/.github/skill-eval/AGENTS.md
@@ -785,126 +785,52 @@ viewer; box-side session history is at `$HOME/.claude-archive/<ts>/`
 One comment per `(spec, platform)` leg. Your leg posts **its own** single
 comment for the one platform it ran (`EVAL_PLATFORM`) — it does **not**
 wait for or aggregate the spec's other platforms: those run as separate
-parallel legs this job cannot see. A two-platform spec therefore yields
-two independent comments, one per platform.
+parallel legs this job cannot see.
 
-**Where the comment goes:** on a PR run (`PR_NUMBER` set) post it with
-`gh pr comment`. On a **manual sweep** (`PR_NUMBER` empty —
-`workflow_dispatch`) there is no PR: append the exact same markdown to
-`$GITHUB_STEP_SUMMARY` instead (see § "Manual full-sweep mode").
-
-```markdown
-## Harbor Eval — `skills/<skill>/<eval-dir>/<spec>.json`
-
-Head: `<short-sha>` · platform `<platform>` · spec `<spec-sha>`
-First started: `<utc>` · Last finished: `<utc>` · Total: `<Ahr Bmin>`
-
-| Platform | Result | Reward | Duration | Turns | Prompt tok | Cached tok | Trace |
-|---|---|---|---|---|---|---|---|
-| L40S | ✅ 1.0 (7/7) | 1.0 | 9m 40s | 23 | 8.4k | 156k | [trace](…) |
-
-For multi-step specs, render one row per step and mark
-prior-fail-skips explicitly:
-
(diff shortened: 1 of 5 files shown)


simings-nv:
Superseded by #1780, which is the same commit on an upstream branch so CI runs normally (this one was opened from a fork, and `copy-pr-bot.yaml` has `auto_sync_draft: false`, so its checks were skipped).

PR #819, opened 2026-05-29
skill-eval: matrix dispatch — one runner per spec

zac-wang-nv:
Stacks on #781 (`zw/skill-eval-no-prereq-deploy`). Review/merge #781 first; this PR's base will retarget to `develop` once #781 lands.

## What
Moves eval fan-out from one long agent session (which looped over every changed spec serially, and could only parallelize by backgrounding harbor — the exit-124 polling anti-pattern) to **GitHub Actions `strategy.matrix`: one runner drives one foreground agent that evaluates one spec.**

```
plan  → diff the PR → emit one matrix leg per spec
eval  → strategy.matrix, fail-fast:false, max-parallel:4
        each leg: single-spec agent → flock a box → synchronous harbor run → posts its own comment
```

Full design: [`.github/skill-eval/docs/matrix-dispatch-design.md`](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/zw/skill-eval-matrix-dispatch/.github/skill-eval/docs/matrix-dispatch-design.md).

## Dispatch rules (the `plan` job)
| Changed path | Effect |
|---|---|
| `skills/<skill>/evals/<spec>.json` | dispatch just that spec |
| any other `skills/<skill>/**` or `adapters/<skill>/**` | dispatch all specs of that skill |
| harness-only (envs/verifiers/agent/AGENTS.md/workflow) | empty matrix → eval skipped |

De-duped; a skill with **no adapter** collapses to one `missing_adapter` leg that raises the single bot-PR (no N-way race). Stale-but-present adapters are still caught per-leg.

## Changes
- **`plan_matrix.py`** (new) — pure Python, no LLM; diff → matrix JSON. Unit-tested offline.
- **`tests/test_plan_matrix.py`** (new) — stubs the filesystem helpers; 8 cases (single-spec, whole-skill, adapter, dedup, harness-skip, missing-adapter collapse, mixed, slug-safety).
- **`skills-eval.yml`** — rewrite to `plan → eval(matrix)`; `workflow_dispatch` full sweep kept as its own job.
- **`skills_eval_agent.py`** — single-spec mode on `EVAL_*`; legacy whole-PR loop removed. Background/task tools stay disabled so each leg is synchronous.
- **`AGENTS.md`** — new "Single-spec mode" section; reframed intro; trimmed redundant prose (981 → 880 lines).

## Decisions baked in (per offline discussion)
- Harness-only change → **skip** (validate via manual dispatch).
[...]

PR #1780, opened 2026-08-20
ci(skill-eval): render the leg result comment without an LLM

simings-nv:
## Description

After `run_leg.py` returns, the outer coordinator agent read the results tree and rebuilt the result table by hand, following ~125 lines of `AGENTS.md` that described the layout and the `jq` needed to pull turns and token counts out of each trajectory. Across real Actions job logs that is p50 **114 s** over a median of 11 read-only tool calls, redone from scratch on every leg because the agent has no memory of the schema it parsed last time. Composing and posting the comment is another 7 s, so 94% of the phase is rediscovery. None of it is a judgement, so none of it needs a model.

This replaces it end to end:

* **`leg_report.py`** (new) collects the tree once and renders the same comment deterministically at p50 **0.59 ms**, and writes a versioned machine-readable summary (`leg-summary.json`) beside it.
* **`skills_eval_agent.py`** now hands the agent ONE command instead of the algorithm, tells it not to read the results tree, trajectories or `judge.json` itself, and points it at `leg-summary.json` for the verdict. Output goes to `$SCRATCH/pr-<spec>.md`, which is where the existing benchmark step already globs for it. **Both prompt branches are wired**: the PR leg and the nightly `DAILY_RUN` leg. They share one `render_block`/`verdict_block` so they cannot drift apart.
* **`AGENTS.md`** drops 1011 → 933 lines: the algorithm becomes a pointer at the renderer plus a table of the summary states. That block was in the *system prompt of every leg*, so it was both re-read and re-implemented each time.

Measured effect per leg: −114 s of wall-clock and ~958 fewer system-prompt tokens. Over the 1,177 legs in the last 8 days that is **−37.3 h** of runner time.

### How it got this way

**The harness was an LLM before it was CI.** At origin the README read: "The framework is run by a **long-running coordinator agent** (`AGENTS.md` is the playbook). You launch it once with `claude` + `/loop AGENTS.md`", and posting the comment was step 6 of that loop. Nobody decided a model should render this table. CI inherited an interactive agent, and the reporting step came along with it.

| date | PR | `§ Result comment` | what happened |
|---|---|---|---|
| 2026-04-28 | #183 | **37** | the harness lands. No `run_leg.py`, and no deterministic Python anywhere between the results tree and the comment |
| 2026-05-18 | #524 | 55 | |
| **2026-05-19** | **#563** | **119** | "include token usage / agent turns / duration per trial". `Turns` / `Prompt tok` / `Cached tok` exist only inside `trajectory.json`, so the instruction grew the `jq` to mine them |
| 2026-06-01 | #819 | 120 | matrix dispatch: one agent per *leg* instead of per *sweep*, multiplying the reporting cost by the leg count |
| 2026-06-03 | #936 | | `run_leg.py` arrives, the first deterministic wrapper to take work back out of the agent |
| 2026-08-20 | this PR | **51** | |

An agent was the right call in April. The Harbor schema was new enough that the judge probed four candidate trajectory filenames, and the comment was largely prose. #563 is where a template became an algorithm; its commit message argues carefully for *why those columns are worth having* and never asks who should render them. That is the whole story: each increment was individually reasonable and none was big enough to look like a decision.

It was then frozen deliberately. `matrix-dispatch-design.md` lists "result-comment format, per-trial metric extraction" under **"Unchanged on purpose"**, and defers the fix on a trigger that never fired: "if N per-spec comments prove noisy, add a thin `report` job later." Noise, not cost.

### Why this belongs in code

**It drifts.** On PR #819 alone the agent posted **four different table headers**, two of which (`| Platform | Result | Notes |` and `| Platform | Result | Details |`) appear in **no revision of `AGENTS.md`**, plus both `0h 21min` and `0hr 10min` for the same field. Replaying 1,177 legs also turned up the run-level `result.json` being mistaken for a trial, mixed aware/naive timestamps, per-step retries, and 480 rows rendering a raw `{{platform}}`. Those are the failure modes of re-deriving a schema 1,177 times instead of once.

**Part of it was never real.** `### Suggestions` instructed every leg to concatenate non-null `suggestion` fields from `$RES/<date>/<trial>/suggestions.json`. That file occurs **0 times in 1,177 legs**, and nothing in this repo writes it, or any `suggestion` field. It has been carried in the system prompt of every leg since April.

**There is no judgement in it.** `results tree -> markdown` is a pure function. The module deliberately emits no verdict, so the pass/fail decision stays with the coordinator where it belongs.

**The repo already works this way.** `plan_matrix.py` is documented as "Pure-Python, no LLM ... Unit-testable offline". `run_leg.py` took the box lock out of the agent. `AGENTS.md` already says "Trace URLs, read them, never build them". This applies the same rule to the one step that still kept the algorithm in the prompt.

### The module emits no verdict

It renders what Harbor recorded; the coordinator decides pass/fail from the summary. An earlier draft emitted a verdict too, and that one extra responsibility produced every serious defect review found: two real `AgentTimeoutError` trials rendered green, steps were counted while being labelled specs, and step accounting flipped 316 real single-step legs falsely red.

Rendering is fail-closed. `row_state()` returns `recorded-pass` only for a finite non-boolean reward exactly `1.0`, no exception, and judge evidence that is absent or well-formed and unanimous. An unreadable `judge.json` is not "no judge". Check counts must be real ints with `total > 0`, so `{"passed": 1, "total": 0}` and string counts are `ambiguous` rather than green. Multiple attempts where any lacks a timestamp are `ambiguous` rather than ordered, because undated-sorts-last is a convention and not a chronology. A spec that is named but unusable raises `SpecError` and exits 3 instead of silently collapsing a partial chain into a complete-looking comment.

The Query column shows what Harbor actually evaluated. Adapters substitute placeholders before execution, so `judge.json.query` is authoritative and the spec holds only a template; 46 of 50 specs contain placeholders and 480 replayed rows displayed a raw `{{platform}}`. Declared text is used only where no judge query exists, is labelled `(declared)`, and has `{{platform}}` filled in.

### Verification

Replayed over **1,177 real legs** (2026-08-13..20): 1,174 rendered, 0 exceptions. Per profile, against ground truth recomputed from the raw files, **0 mismatches** across alerts (320), detection (137), search (112), base (89), lvs (75), warehouse (42), ask-video (22), calibration (19). Fidelity against the coordinator's own recorded verdicts: 6/6.

The replay is also what found the run-level `result.json` being mistaken for a trial, mixed aware/naive timestamps, and per-step retries.

Two new tests construct each prompt and assert it delegates to the renderer. Nothing had ever built a prompt in a test, which is how a first pass that wired only the PR branch could look complete. Against `develop` both tests fail; against a PR-branch-only wiring the nightly one fails; both pass here.

`ruff` matches the `develop` baseline; copyright headers clean (7,411 files); the exact CI skill-eval command passes 110 tests.

### Run on a real leg, not just replayed

Replay proves the renderer agrees with recorded files. It cannot prove the wiring runs. So a genuine Harbor leg was executed end to end on an L40 dev box: real `BrevEnvironment`, real `claude-code` agent through the NVIDIA gateway, a real standalone deploy of `vss-behavior-analytics`, and a real verifier.

```
## Harbor Eval — `skills/vss-setup-behavior-analytics/evals/standalone_deploy.json`

Head: `66c02bca` · platform `ANY`
First started: `2026-08-21 00:06:53Z` · Last finished: `2026-08-21 00:12:24Z` · Total: `5m 31s`

| Platform | Result | Reward | Duration | Turns | Prompt tok | Cached tok | Trace |
|---|---|---|---|---|---|---|---|
| ANY | ✅ 1.0 (4/4) | 1.0 | 5m 31s | 12 | 54.5k | 482.7k | — |
```

`leg_report.py` exited 0 and rendered that in **0.13 s**, interpreter startup included, and wrote `leg-summary.json` beside it. `Turns` is populated here because a real run has `trajectory.json`; the uploaded artifact excludes it, which is why replayed samples show a dash.

The same leg is also a clean measurement of where leg time goes:

| phase | duration | share |
|---|---|---|
| environment_setup | 37 s | 11% |
| agent_setup | 1 s | 0% |
| **agent_execution** | **221 s** | **67%** |
| verifier | 59 s | 18% |

85% of it is LLM round-trips, and 537k input tokens of which 483k are cache reads. That is the same shape as the fleet-wide split across 1,177 legs (agent 69.1%, verifier 14.8%), and it is why the reporting tail was worth removing from the model.

Separately, the aggregation path was exercised for real: the exact command from the prompt was run verbatim against two result trees, and `build_benchmark_md()` picked up both bodies (`wrote benchmark.md (2 spec comments)`). Those two legs are `vss-deploy-profile` and `vss-search-archive`, which both ship a `search.json` and would previously have written the same `pr-search.md`.

### Scope note

Dataset generation still goes through the agent, because each adapter exposes a different CLI (`--skill-dir`, `--deploy-skill-dir`, `--spec`, `--platform` in varying combinations). Normalising that is separate work.

The new tests are registered in `ci.yml`, whose skill-eval step lists files explicitly; without that they would have existed and never run.

## Checklist

## Measured after merge

Merged 2026-08-21T00:37:42Z. Window is the GitHub Actions job duration minus `run_leg`'s own `total_s`, i.e. everything after `run_leg` returns, which is where the comment rendering ran. Job duration is matched to each leg's `phase-timings.json` by `run_id` and slug, sampling legs either side of the merge commit.

| | n | p25 | p50 | p75 | max |
|---|---|---|---|---|---|
| before merge | 20 | 183.8 s | **220.3 s** | 322.0 s | 5205.1 s |
| after merge | 22 | 106.7 s | **119.5 s** | 196.3 s | 293.6 s |

**100.8 s per leg at p50.** The tail matters more than the median here: the worst case falls from 5205 s to 294 s.

The window measures more than the rendering (it also contains checkout, runner setup and artifact upload), so the delta is the result and the absolute is not. Measured independently from the phase timings the change itself emits, the rendering step alone goes from 113.6 s to 9.3 s, a 104 s delta, which agrees with the 100.8 s above to within 4 seconds.

### DIFF
--- a/.github/skill-eval/AGENTS.md
+++ b/.github/skill-eval/AGENTS.md
@@ -373,17 +373,17 @@ The canonical harbor command is in § Harbor invocation.
       wrapper. While a trial is running, do NOT poll the remote box from
       your tool loop — Harbor has its own agent-execution timeout and
       will fail the trial cleanly.
-   d. After each trial, parse
-      `$RES/<date>/<trial>/verifier/reward.txt`
-      and `test-stdout.txt`. Record `(spec, platform, reward,
-      checks_passed/total, duration_s, trace_url)` for the comment. A
-      missing `reward.txt` means the trial errored (e.g. non-zero
-      harbor exit) — record it as a failure, do not skip it silently.
+   d. Do NOT parse `reward.txt`, `test-stdout.txt`, `judge.json` or any
+      trajectory in order to build the comment — `leg_report.py` reads the
+      tree once and owns every column, including the trials that errored
+      (§ Result comment format). Read a trial log only to diagnose a
+      failure you are going to act on, per (c).
 
-6. **Post ONE results comment per `(PR, eval_spec)` batch** when every
-   `(platform)` tuple in that spec's matrix has a result. Format
-   per § Result comment format below. Use `gh pr comment $PR_NUMBER
-   --body-file …`. Do NOT post a planning / "refresh" comment up
+6. **Post ONE results comment for the `(spec, platform)` leg you ran**,
+   rendered by `leg_report.py` per § Result comment format below, with
+   `gh pr comment $PR_NUMBER --body-file …`. Do NOT wait for or aggregate
+   the spec's other platforms: since matrix dispatch those are separate
+   legs this job cannot see. Do NOT post a planning / "refresh" comment up
    front — comments carry results, not intent.
 
 7. **Do not tear down any Brev instance.** The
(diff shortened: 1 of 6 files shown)


zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32424651580)

zac-wang-nv (changes requested):

## Automated OSRB Review — GAP HARD

Changed CI source (skills_eval_agent.py) imports third-party package claude_agent_sdk, which no approval source on NVBug 6111789 covers; it is not stdlib, internal, or NV-proprietary. Classified IMPORT_UNAPPROVED, a hard block. No License Diff rows, no dependency/container changes, no completed failing OSRB-relevant CI checks.

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32425868787)

zac-wang-nv (changes requested):

## Automated OSRB Review — GAP HARD

Changed source files under .github/skill-eval/ statically import third-party package claude_agent_sdk (claude-agent-sdk==0.2.128), which no OSRB approval on NVBug 6111789 covers; all other imports are Python stdlib. No License Diff rows and no dependency-file changes, so the thin-overlay import check governs and reports an uncovered code import.

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32429676487)

zac-wang-nv (changes requested):

## Automated OSRB Review — GAP HARD

Thin-overlay safety net fired (empty License Diff CSV, no dependency manifest touched, source files added/modified). Static import 'claude_agent_sdk' (installed as claude-agent-sdk==0.2.128 in skills_eval_agent.py) has no strict match in the OSRB approval index on NVBug 6111789 and is not stdlib, internal, or NVIDIA-proprietary. All other imports are Python stdlib.

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32430728431)

zac-wang-nv (approved):
LGTM

PR #34, opened 2025-07-18
Add refactored Flask server backend for image/video prompt processing

Alokbpandey:
This PR adds a refactored Flask backend that enables image and video prompt processing using the NVIDIA VLM via NIM API.

**Changes include:**
- Flask server setup
- Endpoint for handling image and video prompts
- Integration with NVIDIA NIM API
- Preprocessing logic for incoming media
- Response formatting for UI

Tested locally and ready for review.

PR #1785, opened 2026-08-21
feat(search): add RT-VLM tag ingestion and BM25 tag search

ayyappa-dev:
## Summary

Adds searchable RT-VLM tags for RTSP streams and uploaded videos, and BM25 tag + tag/embed fusion retrieval through the `vss search` CLI.

## Related Issue
VIA-2189

## Design

This PR is **library + CLI only** — it changes no `vss_agents` production code.

- **Tag ingestion is driven by the headless direct-REST fan-out recipe** (`vss-manage-video-io-storage` `references/provision-vios-source.md`), used on **no-Agent builds**. Registering a source fans it out to RT-CV, RT-Embed, and a controlled `POST /v1/generate_captions` RT-VLM tagging leg (JSON-tag prompt, `response_format={"type":"json_object"}`, `temperature=0`, 5s chunks). RT-VLM publishes caption chunks through the existing Kafka → Logstash pipeline into source-scoped `default_<sanitized-streamId>` Elasticsearch indexes. The tagging leg is lifecycle-independent of the Alert Bridge (teardown carries the RT-VLM-returned `request_id`).
- **Adds `tag_search`** — source/time-scoped BM25 search over validated VLM tag documents — and extends `vss search run` with `tag` and `fusion` paths (tag + embed + optional attribute, RRF or weighted RRF). These are library + CLI features over whatever tag documents exist (e.g., those produced by the headless recipe on a no-Agent build).
- Adds configuration for the tagging endpoint, model, chunk duration, tag-search tool, and fusion weight.

### Scope clarification (what this PR does **not** do)

The stock search profile ships `vss-agent`, but **this PR does not wire the Agent ingestion lifecycle or the Agent search path to tags**:
- The stock Agent ingestion (`POST /api/v1/videos` + `/complete`) does **not** call the controlled `/v1/generate_captions` leg in this PR, so a stock (with-agent) search deployment does not produce VLM tags through this PR.
- The Agent search path does **not** fuse tag candidates in this PR.

Tag ingestion and tag/fusion retrieval are therefore exercised on **no-Agent (headless) builds** via the direct-REST recipe. A follow-up can restore the advertised Agent lifecycle/search integration; until then, users are not told the stock path provides tag search.

## Notes

- Tag retrieval uses BM25 with OR matching for better recall over sparse VLM tags.
- Semantic tag embeddings are intentionally deferred; Phase 1 uses lexical tag search only.
- VST is optional for every search path: it only mints `screenshot_url` media links and resolves source names, so a deployment with Elasticsearch + the path's retrieval services but no VST still searches (hits return with an empty `screenshot_url`).

## Checklist


zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32920982455)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33382278561)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33415079632)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33456504828)

services/agent/packages/vss_agents/src/vss_agents/tools/search.py
```diff
@@ -821,6 +824,151 @@ def _merge_consecutive_results(results: list["SearchResult"]) -> list["SearchRes
...
+            logger.warning("%s search provider degraded: %s", provider, outcome)
+            search_messages.append(f"{provider.capitalize()} provider degraded: {outcome}")
+            continue
+        provider_results[provider] = [_to_core_search_result(result) for result in outcome]
+
+    if not provider_results:
+        raise RuntimeError("All search providers failed")
+
+    configured_method = getattr(config, "fusion_method", "rrf")
+    union_method = "rrf" if configured_method == "rrf" else "weighted_rrf"
```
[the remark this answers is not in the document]
zac-wang-nv: Valid point — the agent-side tag-fusion rewrite that silently remapped `weighted_linear` was removed from this PR. At the current head, tag/fusion is lib+CLI-only, and `SearchRuntime` rejects unsupported fusion methods instead of remapping them. This thread is outdated and can be resolved.

services/agent/packages/vss_cli/src/vss_cli/search/group.py
```diff
@@ -392,11 +384,17 @@ class SearchGroup(CommandGroup):
             AttributeInput,
             requires=frozenset({"elasticsearch", "rtvi_cv"}),
         ),
+        Action(
+            "tag",
+            "BM25 keyword retrieval over source-scoped VLM tag documents.",
+            TagInput,
+            requires=frozenset({"elasticsearch", "vst"}),
```
zac-wang-nv: **P1** — The tag action is advertised when only Elasticsearch and VST are present, but `run()` sends every action through `vss.search()`. That path eagerly calls `Search.from_runtime()`, whose `EmbedSearch.from_runtime(rt)` raises `Required runtime value 'cosmos_embed_endpoint' is missing or empty` on exactly this advertised deployment; I reproduced that failure locally. Please dispatch tag mode through the tag-only facade or lazily construct only the providers selected by the request, otherwise `configure check` says tag is available but every tag call exits with a configuration error.

services/agent/packages/vss_core/src/vss_core/search_core/primitives/_search_helpers.py
```diff
@@ -493,6 +497,37 @@ async def _safe_object_search(oid: int) -> list[AttributeSearchResult]:
...
+            if isinstance(raw_tag_output, TagSearchOutput)
+            else TagSearchOutput.model_validate(raw_tag_output)
+        )
+        search_results = _fusion.tag_output_to_search_results(tag_output)
+        if tag_output.malformed_documents:
+            search_messages.append(
+                f"Skipped {tag_output.malformed_documents} malformed VLM tag "
+                f"document{'s' if tag_output.malformed_documents != 1 else ''}."
+            )
+        yield SearchOutput(data=search_results[:top_k], search_messages=search_messages)
```
zac-wang-nv: **P1** — `top_k` was doubled above to provide headroom for the later adjacent-window merge, but tag-only returns here without performing that merge and slices with the doubled value. A direct `top_k=2` probe returns four tag results. Please cap the final output with `original_top_k` (and consider passing the requested limit to `TagSearch`, which already performs its own malformed-document overfetch) so the public result-limit contract is preserved.

services/agent/packages/vss_core/src/vss_core/search_core/primitives/tag_search.py
```diff
@@ -0,0 +1,206 @@
...
+                resolved.append(source_id)
+        return resolved
+
+    def _resolve_search_index(self, source_ids: list[str]) -> str:
+        """Use all configured indexes, or exact source indexes when scoped."""
+        if "*" not in self._index or not source_ids:
+            return self._index
+        indexes = []
+        for source_id in source_ids:
+            safe_id = re.sub(r"[-/\\ ]", "_", source_id)
```
zac-wang-nv, line 216: **P1** — When VST lookup misses, `source_id` is still raw user input. For example, `FrontDoor` becomes `default_FrontDoor`, which Elasticsearch rejects because index names must be lowercase; other accepted source strings can contain invalid index-expression characters too. That turns the documented “unknown source yields an empty narrowed result” behavior into a backend failure. Please return an empty result before querying for unresolved sources, or query the configured family with the exact identity filter instead of deriving an index expression from the raw value.

services/agent/packages/vss_core/src/vss_core/search_core/primitives/_search_helpers.py
```diff
@@ -512,7 +547,95 @@ async def _safe_object_search(oid: int) -> list[AttributeSearchResult]:
...
+        if not provider_results:
+            if failures:
+                raise failures[0]
+            raise BackendUnreachableError("search", "all fusion providers failed")
+        if malformed_documents:
+            search_messages.append(
+                f"Skipped {malformed_documents} malformed VLM tag document{'s' if malformed_documents != 1 else ''}."
+            )
+
+        search_results = _fusion.fuse_ranked_union(
```
zac-wang-nv, line 648: **P1** — The runtime validates that some weight is positive globally, but the enabled providers are request-dependent. A fusion request without attributes accepts `w_tag=0`, `w_embed=0`, `w_attribute=1`; both active providers are then skipped by `weighted_rrf_union` and the command succeeds with an empty result even when both returned hits (reproduced locally). The same silent-empty case occurs if the only positive-weight provider degrades. Please validate that at least one provider actually present after request routing/degradation has a positive weight, rather than reporting “no matches.”

skills/vss-manage-video-io-storage/references/provision-vios-source.md
```diff
@@ -184,14 +198,49 @@ POST http://localhost:<rt-embed-port>/v1/generate_video_embeddings   # header x-
...
+  -H "Content-Type: application/json" \
+  --data-binary @- <<EOF
+{"id":"<sensorId>","model":"<resolved-vlm-model>","url":"<vios-storage-url>","creation_time":"<upload-anchor>","prompt":$(printf '%s' "$TAG_PROMPT" | jq -Rs .),"response_format":{"type":"json_object"},"temperature":0,"chunk_duration":5,"stream":false}
+EOF
+#   then release the temporary RT-VLM file asset:
+curl -s -X DELETE "http://localhost:<rt-vlm-port>/v1/files/<sensorId>"
+#   Live (RTSP): register, then fire-and-verify (open, confirm 200, CLOSE) —
+curl -s -X POST "http://localhost:<rt-vlm-port>/v1/streams/add" \
+  -H "Content-Type: application/json" \
+  -d "{\"streams\":[{\"liveStreamUrl\":\"<vios-url>\",\"id\":\"<sensorId>\",\"sensor_name\":\"<source-name>\"}]}"
```
zac-wang-nv: **P1** — The design contract and the updated search-profile eval require both RT-VLM `id` and `sensor_name` to carry the canonical VIOS sensor ID, but this executable recipe sends the display/source name for `sensor_name`. The live Harbor eval already reproduced this exact mismatch. Please use `<sensorId>` here as well so Kafka metadata, the derived `default_<streamId>` index, and source-scoped reads stay on one identity.

skills/vss-manage-video-io-storage/references/provision-vios-source.md
```diff
@@ -184,14 +198,49 @@ POST http://localhost:<rt-embed-port>/v1/generate_video_embeddings   # header x-
...
+curl -N --max-time 30 --connect-timeout 10 -X POST "http://localhost:<rt-vlm-port>/v1/generate_captions" \
```
zac-wang-nv: **P1** — This command cannot implement the preceding “confirm 200, CLOSE” contract reliably. A healthy endless SSE response runs until `--max-time` and exits 28, which aborts callers using `set -e`, while an HTTP 4xx can still exit 0 because `--fail-with-body`/status validation is absent. Please capture and validate the HTTP status, then intentionally terminate after admission or the first valid event while normalizing that deliberate termination, so success and failure are not inverted.

skills/vss-build-vision-ai/eval/profile_search_standalone_harbor.json
```diff
@@ -15,46 +15,48 @@
...
+        "`vss-haproxy-ingress` is retained because a single origin is requested, and `_builds/search-standalone/compose.yml` includes `../../deploy/docker/compose.yml` followed by `./patches/vss-haproxy-ingress.yml` — a service-definition patch that mounts a curated HAProxy config (a payload file co-located in `patches/` beside the patch and bound by an absolute `${BUILD_DIR}/patches/` source) carrying the operate routes (`/elasticsearch` read-only, `/rtvi-embed`, `/rtvi-cv`, `/rtvi-vlm` for the tagging leg, plus the shared `/vst`, which the config may group under either browse or operate) alongside the browse routes; RT-VLM is fronted at `/rtvi-vlm` so the controlled `generate_captions` tagging leg is drivable from any host that reaches the origin, not only the deploy host's loopback.",
```
zac-wang-nv, line 34: **P1** — This makes the build generate a second HAProxy routing/security policy even though the canonical `deploy/docker/services/infra/haproxy/haproxy.cfg.template` already contains the `/rtvi-vlm` ACL and `bk_rtvi_vlm_strip` prefix-rewrite backend. Copying a trimmed `patches/haproxy.cfg` makes every build a fork of that policy and will miss future allowlist, deny, timeout, and route fixes. Please keep the canonical HAProxy configuration as the single authored source and remove the new generated-config requirement (and the duplicated `bk_rtvi_vlm_strip` recipe in `ingress.md`); tagging needs no new backend rule.

skills/vss-build-vision-ai/references/services/rt-vlm.md
```diff
@@ -92,12 +92,31 @@ reach Elasticsearch through the alert bridge, not RT-VLM's Kafka path. For the
...
+differ only in the `POST /v1/generate_captions` prompt — no second service is
+deployed. **Dense captioning** uses a free-form prompt for captions/incidents and
+is skipped when an Alert Bridge owns verification (see
+`vss-manage-video-io-storage` `provision-vios-source.md`). **VLM tagging** uses a
+controlled JSON-tag prompt (`response_format={"type":"json_object"}`,
+`temperature=0`, 5s chunks) whose output feeds BM25 tag search: RT-VLM publishes
+to its existing `mdx-vlm-captions` topic, the existing LVS Logstash pipeline
+writes each chunk to `default_<streamId>`, and the read side
+(`vss_core.search_core` `TagSearch`/fusion, exposed via `vss search tag`/`fusion`)
+queries it. Tagging is independent of the Alert Bridge (it owns search indexing,
```
zac-wang-nv, line 123: **P1** — Tagging is not lifecycle-independent of Alert Bridge with the current callers. Both use the same RT-VLM stream and `generate_captions`, but RT-VLM isolates teardown only when `DELETE /v1/generate_captions/{stream_id}` includes that subscriber's `request_id`. This is not a caller-selected value: RT-VLM generates a UUID for each `generate_captions` request and returns it as the top-level `id` in the JSON response or each SSE data event. The owner must parse that returned `id`, persist it with the tag session or alert rule, and later call `DELETE /v1/generate_captions/{stream_id}?request_id={returned_id}`. For streaming admission, the current server does not emit an immediate ID-only event/header, so a caller that closes after HTTP 200 cannot learn the ID; either it must read the first JSON SSE event or RT-VLM must add an immediate admission response. The new tagging recipe captures neither and stops by stream ID before deleting the shared stream; Alert Bridge likewise discards the generated ID and performs the same stream-wide caption stop, followed by `/streams/delete` for its last known reader. Either leg can therefore stop the other, and the direct tagging `/streams/add` can also collide with the bridge's registration. Please flag alerts + tagging as unsupported until one owner coordinates the shared registration, persists these server-returned per-leg IDs, uses request-scoped caption teardown, and ref-counts stream deletion; the build resolver must not promise this combination currently coexists.

services/agent/packages/vss_core/src/vss_core/search_core/primitives/_tag_helpers.py
```diff
@@ -0,0 +1,226 @@
...
+    if not isinstance(payload, dict) or "tags" not in payload or set(payload) - {"tags", "description"}:
+        raise ValueError("tag document must contain tags and may contain description")
+    raw_tags = payload["tags"]
+    if not isinstance(raw_tags, list) or len(raw_tags) > _MAX_TAGS:
+        raise ValueError("tag document tags must be a list with at most 32 items")
+    tags: list[str] = []
+    for value in raw_tags:
+        if not isinstance(value, str) or not value.strip():
+            raise ValueError("each tag must be a non-empty string")
+        tag = value.strip().lower()
```
zac-wang-nv, line 165: **P2** — The design contract says tags longer than 64 characters and descriptions longer than 1,024 characters are invalid, but this parser enforces only non-empty strings and the 32-item count. Please enforce both length limits here and add boundary tests; otherwise oversized model output is accepted as a valid/searchable tag document despite the stated read-time validation contract.

skills/vss-build-vision-ai/references/services/rt-vlm.md
```diff
@@ -92,12 +92,31 @@ reach Elasticsearch through the alert bridge, not RT-VLM's Kafka path. For the
 integrated path, `RTVI_VLM_MODEL_TO_USE=cosmos-reason3` and
 the `http://rtvi-vlm:8000` endpoint (a consumer's `VLM_BASE_URL`) are invariant
 across BF16 and FP8; a consumer owns that URL but never inherits it from the
+xp_ across BF16 and FP8; a consumer owns that URL but never inherits it from the
```
zac-wang-nv: **P2** — This looks like a truncated duplicate accidentally inserted into the owner contract and makes the surrounding sentence unreadable. Please remove the `xp_ ...` line.

services/agent/packages/vss_agents/src/vss_agents/tools/search.py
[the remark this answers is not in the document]
ayyappa-dev: Agreed — the agent-side `weighted_linear` remap is gone; tag/fusion is now lib+CLI-only and `SearchRuntime` rejects unsupported fusion methods instead of remapping. Resolving this thread.

services/agent/packages/vss_cli/src/vss_cli/search/group.py
ayyappa-dev: Fixed in `ecb2eb6cd`. `Search.from_runtime` now builds the tag-only facade without requiring `cosmos_embed_endpoint`; the embed leg is constructed lazily and raises a precise `ConfigurationError` only if embed/fusion actually invokes it. Tag-only deployments no longer exit with a config error. Covered by `TestTagOnlyDeploymentAndFusionWeights`.

services/agent/packages/vss_core/src/vss_core/search_core/primitives/_search_helpers.py
ayyappa-dev: Fixed in `ecb2eb6cd`. Tag-mode final yield now slices `[:original_top_k]` (not the doubled headroom value), so `top_k=2` returns two. New regression test added.

services/agent/packages/vss_core/src/vss_core/search_core/primitives/tag_search.py
ayyappa-dev: Fixed in `ecb2eb6cd`. `_resolve_source_ids` now drops unresolved sources instead of deriving `default_<raw>` index names; `run()` returns an empty `TagSearchOutput` without querying when all sources are unknown, and partially-unknown requests query only the resolved source's exact index. No more invalid-index backend failures.

services/agent/packages/vss_core/src/vss_core/search_core/primitives/_search_helpers.py
ayyappa-dev: Fixed in `ecb2eb6cd`. Before `fuse_ranked_union`, when `fusion_method == "weighted_rrf"` and no active provider has a positive weight, we now raise `InvalidInputError` instead of returning a silent empty union. New test added.

skills/vss-manage-video-io-storage/references/provision-vios-source.md
ayyappa-dev: Fixed in `ecb2eb6cd`. `sensor_name` now uses `<sensorId>` (the canonical VIOS sensor ID), so Kafka metadata, the derived `default_<streamId>` index, and source-scoped reads stay on one identity.

skills/vss-manage-video-io-storage/references/provision-vios-source.md
ayyappa-dev: Fixed in `ecb2eb6cd`. The live SSE curl now captures the HTTP status (`-w '%{http_code}'`), fails on 4xx/5xx, and treats the deliberate early close (curl 28 / SIGPIPE 141 after a 200) as admission-confirmed. Teardown uses `DELETE …?request_id=<returned-id>`.

skills/vss-build-vision-ai/eval/profile_search_standalone_harbor.json
ayyappa-dev: Partially fixed in `ecb2eb6cd` (narrow scope): `ingress.md` now states the canonical `haproxy.cfg.template` is the single authored source for `/rtvi-vlm` (it already carries the ACL + `bk_rtvi_vlm_strip`); a tagging build copies the route verbatim and authors no new backend rule. I left the pre-existing curated-`patches/haproxy.cfg` pattern and eval check #34/#35 untouched — fully removing that fork is a broader build-architecture change I'd like to track as a separate follow-up (needs the eval/build owner's sign-off). OK to go narrow here?

skills/vss-build-vision-ai/references/services/rt-vlm.md
ayyappa-dev: Fixed in `ecb2eb6cd`. Added a "Lifecycle independence of the tagging leg" section: the caller parses the RT-VLM-returned `request_id` (top-level `id` in the JSON / SSE event), persists it with the tag session, and tears down with `DELETE /v1/generate_captions/{stream_id}?request_id=<returned_id>`. Streaming-admission caveat (no immediate ID-only event) is documented. Alerts+tagging coexistence on a shared stream still needs owner-coordinated ref-counting and is not promised by this build.

services/agent/packages/vss_core/src/vss_core/search_core/primitives/_tag_helpers.py
ayyappa-dev: Fixed in `ecb2eb6cd`. The parser now enforces `_MAX_TAG_LENGTH=64` and `_MAX_DESCRIPTION_LENGTH=1024` at parse time, with boundary tests added.

skills/vss-build-vision-ai/references/services/rt-vlm.md
ayyappa-dev: Fixed in `ecb2eb6cd`. Removed the stray truncated `xp_ …` duplicate line.

services/agent/packages/vss_cli/src/vss_cli/search/group.py
```diff
@@ -396,11 +388,17 @@ class SearchGroup(CommandGroup):
...
+        ),
         Action(
             "fusion",
-            "Embedding retrieval re-ranked by attribute evidence.",
+            "Candidate-union fusion across tags, embeddings, and optional attributes.",
             FusionInput,
-            requires=frozenset({"elasticsearch", "rt_embed", "rtvi_cv"}),
+            requires=frozenset({"elasticsearch", "rt_embed", "vst"}),
```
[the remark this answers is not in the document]
zac-wang-nv: Fixed at d75c7c57b: tag/fusion preflight no longer requires VST, and Search, TagSearch, and EmbedSearch construct without VST. Source-less retrieval now returns hits with an empty `screenshot_url`, covered by the new no-VST regressions.

zac-wang-nv:
- [ ] **P1** — The PR description says the existing Agent ingestion lifecycle will start/stop RT-VLM tagging and that the Agent search path will fuse tag results, but this head changes no `vss_agents` production code. The stock search profile contains `vss-agent`, so the search skill routes ingestion through the unchanged Agent APIs while the new direct-REST tagging recipe is explicitly limited to no-Agent builds. As shipped, stock ingestion never calls the controlled `/v1/generate_captions` leg, and Agent search never consumes tag candidates. Please either restore the advertised Agent lifecycle/search integration with tests, or narrow the PR contract and support matrix so users are not told the stock path provides tag search.

services/agent/packages/vss_core/src/vss_core/search_core/primitives/tag_search.py
```diff
@@ -0,0 +1,230 @@
...
+    ) -> None:
+        self._es = es
+        self._vst = vst
+        self._index = tag_index
+        self._default_k = default_max_results
+        self._owns_es = owns_es
+
+    async def run(self, inp: TagSearchInput) -> TagSearchOutput:
+        inp.validate_semantics()
+        name_to_id = await self._name_to_id()
```
zac-wang-nv: - [ ] **P1** — This fetch runs even when `video_sources` is empty. Omitting VST works, but when VST is configured and temporarily unreachable, an otherwise unscoped tag query raises `VSTError` before Elasticsearch is queried, despite VST being optional enrichment for this path. I reproduced that with a failing `VSTSnapshot`. Please make name resolution mandatory only for scoped queries (or best-effort when unscoped) and add a configured-but-unreachable VST regression.

skills/operations/vss-manage-video-io-storage/references/provision-vios-source.md
```diff
@@ -186,14 +200,64 @@ POST http://localhost:<rt-embed-port>/v1/generate_video_embeddings   # header x-
...
+#   Live (RTSP): register, then confirm admission (HTTP 200) and intentionally close
+#   after a short read. A healthy endless SSE stream runs until the caller closes it, so
+#   don't rely on `--max-time` alone: under `set -e` a timeout exits 28, and without status
+#   validation an HTTP 4xx can still exit 0 — success/failure get inverted. Capture the HTTP
+#   status, fail on >=400, and treat the deliberate early close (curl 28 / SIGPIPE 141 after
+#   a 200) as admission confirmed. Both RT-VLM `id` and `sensor_name` carry the canonical
+#   VIOS sensor ID (`<sensorId>`), never the display/source name.
+curl -s -X POST "http://localhost:<rt-vlm-port>/v1/streams/add" \
+  -H "Content-Type: application/json" \
+  -d "{\"streams\":[{\"liveStreamUrl\":\"<vios-url>\",\"id\":\"<sensorId>\",\"sensor_name\":\"<sensorId>\"}]}"
```
zac-wang-nv: - [ ] **P1** — `AddLiveStream.description` is a required field in `services/rtvi/rt-vlm/src/api_models/live_stream.py`, so this documented body fails validation with a 422 before live tag generation starts. Please include a non-empty `description` in the stream object and validate the registration response before proceeding.

skills/operations/vss-search-archive/SKILL.md
```diff
@@ -141,10 +141,10 @@ independent of the index inventory.
    stdout. Read [CLI usage](references/cli_usage.md) for every supported flag.
 
 ```bash
-: "${SEARCH_PATH:?set embed|attribute|fusion|object}"
+: "${SEARCH_PATH:?set embed|attribute|fusion|object|tag}"
```
zac-wang-nv, line 147: - [ ] **P1** — This exposes `tag` as a supported path, but the mandatory dispatcher immediately above has no condition that selects it: every non-object request becomes fusion, embed, or attribute. An agent following this skill therefore cannot choose the new tag-only command even for explicit lexical/tag intent. Please add a tag-selection rule and cover it in the search skill eval.

skills/operations/vss-search-archive/SKILL.md
```diff
@@ -141,10 +141,10 @@ independent of the index inventory.
...
 : "${SOURCE_TYPE:?set video_file or rtsp}"
 TOP_K="${TOP_K:-3}"
-VIDEO_SOURCES=() # sensor IDs for embed/fusion; names for attribute/object
+VIDEO_SOURCES=() # sensor IDs for embed/fusion; names for attribute/object/tag
```
zac-wang-nv, line 150: - [ ] **P1** — This contradicts line 116, which tells the caller that fusion resolves a source name. Only `TagSearch` performs name-to-ID resolution; fusion passes the original `video_sources` unchanged to the embedding provider, whose sensor-ID filter is literal. Passing the name therefore lets the tag leg match while silently dropping embedding candidates. Use the preserved sensor ID for fusion (the tag leg already accepts IDs) and make the two instructions consistent.

skills/operations/vss-search-archive/references/cli_usage.md
```diff
@@ -70,27 +78,30 @@ run embed --query "red forklift" --source-type video_file --top-k 10
...
 
-`--video-source` is matched **literally** against the index — the CLI does no
-name↔id resolution or VST validation, so an unknown source silently returns
-nothing (not an error). Validating a named source is the skill's job (SKILL.md
-step 2), not the CLI's.
+`--video-source` is matched **literally** against the index for `embed`,
+`attribute`, and `object` — the CLI does no name↔id resolution or VST
+validation, so an unknown source silently returns nothing (not an error). For
+`tag` and `fusion`, the CLI resolves each source name to its VST sensor ID (passing an already-id through);
+an unresolved source is kept literal and yields an empty, narrowed result (exit 0), not an error.
```
zac-wang-nv: - [ ] **P2** — The implementation drops unresolved tag/fusion sources in `_resolve_source_ids`; it does not keep them literal. The empty-result outcome is correct, but the distinction matters because dropping the value is what prevents unsafe index-name derivation. Please align this explanation with the implemented behavior.

skills/vss-build-vision-ai/references/services/ingress.md
```diff
@@ -50,19 +50,28 @@ deploys, the routes each consumer of the origin needs:
...
 headless build **must** carry it — post-#1469 `vss search run` takes no endpoints,
-so a build missing these routes is **unqueryable from the host CLI** (no
 ingress-less read path). But RT-Embed or Elasticsearch in the build does **not**
 make it queryable — an ingestion/indexing-only build that requests no read surface
-prunes the proxy regardless. `vss configure` also probes `/rtvi-vlm`, but RT-VLM is
-host-port resolved and deliberately **not** fronted here — it records `absent`,
-which is expected and harmless because no read path consumes it. Do not add the
-route to satisfy the probe; that would re-expose RT-VLM's SSE generation endpoints
-through HAProxy.
+prunes the proxy regardless. `vss configure` also probes `/rtvi-vlm`. The canonical
```
zac-wang-nv, line 59: - [ ] **P2** — The preceding paragraph now reads “takes no endpoints, ingress-less read path)” because this edit deleted the clause that connected those phrases. Please restore the missing “so a build missing these routes is unqueryable … (no ingress-less read path)” wording; the current text is grammatically incomplete and loses the operational consequence.

services/agent/packages/vss_core/src/vss_core/search_core/primitives/tag_search.py
ayyappa-dev: Fixed in c662def03. `TagSearch.run()` now resolves source names only for scoped queries (`inp.video_sources` non-empty); an unscoped tag search queries the `default_*` family directly without contacting VST, so a configured-but-unreachable VST no longer raises `VSTError` before Elasticsearch is queried. Sourced queries still require resolution — a failing `VSTSnapshot` still raises, preserving the isolation boundary. Added `test_sourceless_tag_search_skips_unreachable_vst` as the configured-but-unreachable-VST regression.

skills/operations/vss-manage-video-io-storage/references/provision-vios-source.md
ayyappa-dev: Fixed in c662def03. The live `POST /v1/streams/add` body now carries `description` (required on `AddLiveStream` — without it RT-VLM returns 422 before tagging starts), and the registration response is captured with `-w '%{http_code}'` and validated (fail on >=400) before the `generate_captions` leg starts.

skills/operations/vss-search-archive/SKILL.md
ayyappa-dev: Fixed in c662def03. The decompose-the-request dispatcher now has a `run tag` rule — explicit keyword/tag intent (lexical BM25 over VLM tags, no detectable property, no semantic free-text) -> `run tag`, selected before fusion for pure lexical intent. Added a `tag-keyword-search` step to `evals/search.json` that exercises `vss search run tag` through the dispatcher; it requires the agent to select `run tag` and faithfully report the result, and does not require non-empty hits (the stock agent ingestion does not produce tag documents in this PR).

skills/operations/vss-search-archive/SKILL.md
ayyappa-dev: Fixed in c662def03. Corrected the contract: only `tag` resolves a source name to its VST sensor ID; `fusion` does not resolve — its embedding leg filters by sensor ID literally — so callers hand fusion the preserved sensor ID (the tag leg accepts IDs). SKILL.md step 2 and the bash `VIDEO_SOURCES` comment are now consistent with the implementation (`_resolve_source_ids` drops unresolved sources; `EmbedSearch` matches literally). `cli_usage.md` was aligned to the same wording.

skills/operations/vss-search-archive/references/cli_usage.md
ayyappa-dev: Fixed in c662def03. Reworded to match the implementation: an unresolved tag/fusion source is dropped (not carried forward) — which is what prevents unsafe `default_<raw>` index-name derivation — and only `tag` resolves a source name to its VST sensor ID; `fusion` matches the sensor ID literally (callers pass the preserved sensor ID).

skills/vss-build-vision-ai/references/services/ingress.md
ayyappa-dev: Fixed in c662def03. Restored the missing clause — the paragraph now reads "...`vss search run` takes no endpoints, so a build missing these routes is unqueryable (no ingress-less read path). But RT-Embed or Elasticsearch in the build does not..."

zac-wang-nv (approved):

skills/vss-build-vision-ai/references/services/ingress.md
```diff
@@ -50,19 +50,27 @@ deploys, the routes each consumer of the origin needs:
...
+`haproxy.cfg.template` is the **single authored source** for this route — it already
+carries the `/rtvi-vlm` ACL and the `bk_rtvi_vlm_strip` prefix-rewrite backend, so a
+tagging build copies that route verbatim into any curated config and authors **no new
+backend rule**; tagging needs no new backend rule. For a build that resolves the VLM
+**tagging** capability, front RT-VLM at `/rtvi-vlm` so the controlled
+`generate_captions` tagging leg can be driven from any host that reaches the origin,
+not only the deploy host's loopback. This is a conscious tradeoff: it
+re-exposes RT-VLM's SSE-generation and stream/file-mutation endpoints through
+HAProxy, so the origin's host-allowlist is the only boundary — front RT-VLM only
+when the build needs remote-driven tagging, and never on an unauthenticated origin.
```
zac-wang-nv, line 69: **TODO (non-blocking):** The ‘never on an unauthenticated origin’ condition is not enforceable by this build flow: the generated HAProxy config has only a Host-header allowlist, not authentication, while the updated eval requires `/rtvi-vlm` on the public origin. A Host header is not a credential, so any caller that can reach the origin can invoke inference and stream/file mutation endpoints. Keep this route private/loopback unless an explicit auth mechanism is present, or add an authentication gate before copying it into the curated ingress.

skills/operations/vss-manage-video-io-storage/references/provision-vios-source.md
```diff
@@ -186,14 +200,70 @@ POST http://localhost:<rt-embed-port>/v1/generate_video_embeddings   # header x-
...
+# ⚠️ This leg is the ONLY one that produces searchable tag documents. The dense-captioning
+# leg above uses a free-form prose prompt; its output lands in the SAME default_<streamId>
+# index but is REJECTED by the tag reader (it is not the {"tags":[...],"description":"..."}
+# JSON contract), so `vss search run tag` returns 0 valid hits (every document malformed).
+# Same endpoint, same Kafka topic (mdx-vlm-captions), same index — ONLY the prompt differs.
+# For tag search, always use the controlled JSON-tag prompt below, never the dense-captioning
+# prompt.
+#   Upload (VOD): one finite call, then delete the temporary asset —
+TAG_PROMPT='Analyze only this video interval. Return JSON only with exactly two fields: "tags", an array of concise visible concepts, actions, objects, and events; and "description", one concise factual sentence. Do not infer facts that are not visible.'
+curl -s -X POST "http://localhost:<rt-vlm-port>/v1/generate_captions" \
```
zac-wang-nv, line 219: **TODO (non-blocking):** This finite tagging call neither fails on HTTP 4xx/5xx nor validates the response body. Plain `curl` exits 0 for an HTTP error, and a successful response with zero or malformed chunk responses is still followed by asset deletion, so the workflow can report a provisioned upload that can never yield tag hits despite the design contract that zero valid chunks fails. Capture status plus body, require at least one chunk response matching the tag JSON contract, then release the asset.

services/agent/packages/vss_core/src/vss_core/search_core/primitives/_search_helpers.py
```diff
@@ -512,7 +551,112 @@ async def _safe_object_search(oid: int) -> list[AttributeSearchResult]:
...
+            provider_results,
+            method=config.fusion_method,
+            weights={"tag": config.w_tag, "embed": config.w_embed, "attribute": config.w_attribute},
+            rrf_k=config.rrf_k,
+        )
+        if getattr(config, "merge_adjacent", True):
+            search_results = _fusion.merge_consecutive_results(search_results)
+        yield SearchOutput(data=search_results[:original_top_k], search_messages=search_messages)
```
zac-wang-nv, line 656: **TODO (non-blocking):** This new early return bypasses the common `apply_top_percent_filter()` call below, so `vss search run fusion --top-percent-filter ...` now returns the same set as without the flag; before this change fusion fell through and applied it. The tag path takes an analogous early return. Apply compatible post-processing before returning (or scope the flag away from those actions) and add an action-level regression test.

zac-wang-nv (approved):
Approved with three non-blocking TODOs for follow-up.

PR #1792, opened 2026-08-21
perf(skill-eval): keep model-weight caches across the per-trial docker reset

simings-nv:
## Description

`BrevEnvironment._reset_docker_runtime` runs before a spec's **first** trial on a warm `vss-eval-*` box. It removed all containers, all user-defined networks and **all volumes**, preserving images. That included the model-weight caches, so a box that had just downloaded a model threw it away before the next spec asked for the same one.

This preserves the weight caches and still wipes everything else.

## Measured on an L40, cosmos-reason1-7b

| | time to `/v1/health/ready` |
|---|---|
| cold, cache wiped (today) | **501 s** |
| warm, cache preserved | **319 s** |
| **saved per NIM deploy** | **182 s** |

Cache size 9.4 GB. The image was already local in both runs, so this is weights only. The saving scales with weight bytes, so larger models save more.

This applies to legs that load a model on the box: `llm_mode=local`, and a VLM configured for local inference. A leg whose VLM runs in the default `openai-compat` mode downloads no weights and is unaffected.

## What is preserved

The nine NIM `<model>_cache` volumes, all mounted at `/opt/nim/.cache`, plus `rtvi-hf-cache` and `rtvi-ngc-model-cache`. Both have integrity machinery that makes a resumed cache safe: NIM verifies manifest checksums on download-to-cache, and `huggingface_hub` keeps interrupted blobs as `.incomplete` files outside the committed snapshot.

`MODEL_CACHE_VOLUME_RE` is an exact allowlist. A test re-derives it from the compose files, so a NIM added without updating the list fails CI rather than silently costing every leg that deploys it three minutes.

`rtvi-ngc-model-cache`'s downloader treats directory existence as completion, so a crash mid-copy can leave a partial model that the next deploy reuses. `SKILL_EVAL_WIPE_MODEL_CACHES=1` clears it.

## What is not preserved

- **`rtvi-triton-model-repo`** — assembled at deploy time rather than downloaded; a stale copy changes what a trial exercises.
- **`vios_apt_cache`** — APT packages, not weights.

Everything else is still wiped blanket, because trials reach a deploy through heterogeneous compose project names and a scoped teardown cannot reach a predecessor's stack. The post-reset guard, which asserted that zero volumes remain, now counts only volumes that should be gone.

`SKILL_EVAL_WIPE_MODEL_CACHES=1` restores the full wipe.

## Cache lifetime

The volume name is not version-keyed (`cosmos_reason1_7b_cache` while the image is `:1.4.1`), so a tag bump downloads the new weights into the same volume. Tags are pinned and bump rarely, so a box stays warm until it is recreated. The fleet dashboard tracks `vss_fleet_disk_used_pct`.

## Verification

Keep/delete selection and the guard are tested against real volume names under `bash`, including a box whose only volumes are caches, a non-cache volume surviving, and the escape hatch. 54 skill-eval env tests pass; copyright headers clean; `ruff` on `brev_env.py` matches `develop`.

## Checklist

### DIFF
--- a/.github/skill-eval/AGENTS.md
+++ b/.github/skill-eval/AGENTS.md
@@ -390,8 +390,9 @@ The canonical harbor command is in § Harbor invocation.
    `vss-eval-*` boxes are a long-running pool managed by the
    operator; instances stay up across runs, and so do the slow caches
    that survive a volume wipe (docker **image** layers, the repo clone, the
-   `data/` sample-data extract — but NOT the model-weight *volumes*, which
-   the per-trial reset drops; see § 7).
+   `data/` sample-data extract, and the NIM / `rtvi-hf-cache` /
+   `rtvi-ngc-model-cache` model-weight volumes, which the per-trial reset now
+   preserves; see § 7).
    `run_leg.py` releases the per-box lock automatically when its process
    exits; there is no shell FD for you to close. You never `brev stop` /
    `brev delete`. Pool lifecycle is strictly an operator concern.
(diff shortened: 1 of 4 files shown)


zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32514500262)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32519514711)

PR #1793, opened 2026-08-21
chore(ci): drop the retired has_ghcr_build_entries gate signal

zac-wang-nv:
## Description

`has_ghcr_build_entries` was the **previous** downstream acceptance gate. It was replaced by `run_downstream`, and the replacement's own docstring records why:

> Scoped on what *changed*, not on what got built. The previous gate keyed off "did any GHCR image get rebuilt", which is a poor proxy twice over: build avoidance means a real source change can rebuild nothing, and deploy-only changes never rebuild anything yet are exactly what acceptance exists to catch.

The signal survived the replacement. `ci.yml` still exports it as **two** job outputs — `trigger-from-new-image` and `has-ghcr-build-entries` — and **nothing reads either one**, in this repo or in `ci-vss-oss`. The only remaining mentions were the exports, one comment, the function, and its tests.

Removed: both job outputs, `has_ghcr_build_entries()`, its call, its `GITHUB_OUTPUT` line, and its tests. `run_downstream` — the live gate consumed at `ci.yml:1585` — is untouched.

Also corrected the comment that described a container source change as "surfacing as `trigger-from-new-image` once the image is built", since that path no longer exists; the per-image `source_path` check is what covers it now.

## Plan

**Goal:** the downstream gate has one signal instead of one live signal plus a retired one that two job outputs still advertise.

**Decisions**

| Q1 | Is it safe to remove? | Yes — no consumer in either repo | Remove |
| Q2 | Keep the docstring rationale that names the retired signal? | Keep the reasoning, drop the stale identifier, so the *why* survives without referring to code that no longer exists | Keep reworded |

**Steps**

1. Prove nothing consumes it → verify: `git grep` across the VSS repo and `ci-vss-oss` returns only the exports, a comment, the function and its tests.
2. Remove the outputs, function, call, output line and tests → verify: no residual references anywhere under `.github/`.
3. Prove nothing regressed → verify: `ci.yml` parses; `prepare_downstream_release_set`, `release_set`, `update_pr_ghcr_candidates` and `trigger_downstream_pipeline` suites all pass.

**Out of scope**

- `run_downstream` and `downstream_relevant()` — the live gate, unchanged.
- The release-set assemble/validate/upload path. It is still exercised on every build, and now that #1727 removed the last-green controller its only consumers are this gate and the PR candidates comment. Whether the schema-validated manifest still earns its keep is a design question, not a drive-by removal.

## Checklist
- [x] New or existing tests cover these changes. — obsolete tests removed with the function; four neighbouring suites pass.
- [ ] The documentation is up to date with these changes. — n/a beyond the comment fix included here.

### DIFF
--- a/.github/scripts/prepare_downstream_release_set.py
+++ b/.github/scripts/prepare_downstream_release_set.py
@@ -57,8 +57,7 @@ def downstream_relevant(changed: list[str] | None, inventory: dict) -> tuple[boo
         OR deploy/ changed
 
     Scoped on what *changed*, not on what got built. The previous gate keyed off
-    has_ghcr_build_entries -- "did any GHCR image get rebuilt" -- which is a poor
-    proxy twice over: build avoidance means a real source change can rebuild
+    "did any GHCR image get rebuilt", which is a poor proxy twice over: build avoidance means a real source change can rebuild
     nothing, and deploy-only changes never rebuild anything yet are exactly what
     acceptance exists to catch. Config and deploy edits were getting no
     downstream coverage at all.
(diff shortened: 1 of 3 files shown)


zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32529320211)

PR #1795, opened 2026-08-21
refactor(ci): drop the GHCR candidates comment and the gate's release-set read

zac-wang-nv:
## Description

> **Stacked on #1793.** Base is `chore/drop-has-ghcr-build-entries`, so this diff shows only its own changes. Rebase onto `develop` once #1793 lands.

Two consumers of the release set go away. The one that matters stays.

**1. The PR candidates comment is unread.** Nothing machine-parses its marker `<!-- vss-ghcr-candidates -->` — verified across this repo and `ci-vss-oss` — and it is not read by hand either. The publisher, its `ci.yml` step and the comment machinery go.

**2. The downstream gate never needed the manifest.** It read exactly two fields — `source.commit` and `source.ref` — to derive `{develop|pr-N}-{sha12}`, a tag the workflow already holds as `GITHUB_SHA` and `GITHUB_REF_NAME`, and which `container_build_plan.py` computes identically. So the gate stopped fetching, validating and round-tripping a release set through an artifact. It derives the tag directly and still waits on the build run's conclusion — which was always the real completion signal, independent of the artifact.

### What deliberately survives

`fragment`, `assemble`, `validate`, the `release-set` artifact and `release-set.schema.json`. `advance_ghcr_alias.py` consumes the assembled manifest to publish `develop-latest` and `develop-<sha12>` across **every** GHCR image, including ones this build did not rebuild. That coverage cannot be derived from what changed — `build-dev-images.yml` says so in its own comment:

> Both sets must cover EVERY GHCR image, not just the ones rebuilt on this ref: a consumer that derives a coordinate from the ref would otherwise 404 on the untouched images, which is why PR acceptance still needs a per-image manifest today.

That is the guarantee §4.2 of the design doc sells, so the release set keeps exactly one consumer — an honest one.

`update_pr_ghcr_candidates.py` becomes `github_build_run.py`, retaining only the GitHub API adapter and the build-run polling the gate imports. `download_release_set` becomes `await_build_run` and returns the run rather than an artifact. 417 → 224 lines.

## Plan

**Goal:** the release set has one consumer that genuinely needs it, instead of three where two were derivable or unread.

**Decisions**

| Q1 | Replace the release set entirely? | Initially offered it — **then withdrew the option**: tracing every consumer showed `advance_ghcr_alias.py` needs the full per-image manifest, and deriving the alias set from the inventory would silently 404 `develop-<sha12>` on untouched images | Keep it for the alias publication |
| Q2 | What about the candidates comment? | Delete — no machine reads the marker | Delete |
| Q3 | Keep the 90-day release-set archive? | Keep it; it rides along with the artifact the alias step already needs | Kept |

**Steps**

1. Prove nothing parses the comment marker → verify: `git grep` in both repos finds only the publisher and its `ci.yml` step.
2. Derive the candidate tag from ref+SHA → verify: tests cover `develop`, `pull-request/N`, an unsupported ref, and a short SHA.
3. Keep the completion signal → verify: `await_build_run` still polls the workflow-run API and raises on terminal failure conclusions; 17 tests cover re-runs, in-flight runs and budget exhaustion.
4. Prove the alias path is untouched → verify: `test_advance_ghcr_alias` passes; `assemble`/`validate` and the `release-set` artifact are unchanged.
5. Prove nothing regressed → verify: `ci.yml` parses, no residual references under `.github/` or `deploy/`, and seven tooling suites pass.

**Out of scope**

- `release_set.py closure`, the inventory-coverage check in `ci.yml` — independent of assembly.
- Whether the alias set could one day be derived from the inventory plus registry lookups. That is a redesign of the `develop-<sha12>` completeness guarantee, not a cleanup.

## Checklist
- [x] New or existing tests cover these changes. — 32 tests across the two touched suites; seven suites pass in total.
- [ ] The documentation is up to date with these changes. — n/a; no doc describes the removed comment.

### DIFF
--- a/.github/scripts/github_build_run.py
+++ b/.github/scripts/github_build_run.py
@@ -1,29 +1,24 @@
 #!/usr/bin/env python3
 # SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
-"""Publish immutable GHCR candidate coordinates after downstream CI passes."""
+"""Locate and await the Build Dev Images run that publishes a commit's GHCR set.
+
+Shared by the downstream gate. Polls the workflow-run API for the run matching an
+exact commit and ref, and reports success only on a terminal successful conclusion.
+"""
 from __future__ import annotations
 
-import argparse
-import io
 import json
 import os
 import re
-import sys
 import time
 import urllib.error
 import urllib.parse
 import urllib.request
-import zipfile
-from pathlib import Path
 from typing import Any
 
-MARKER = "<!-- vss-ghcr-candidates -->"
 API_ROOT = "https://api.github.com"
-COMMENT_PAGE_SIZE = 100
-MAX_COMMENT_PAGES = 100
 MAX_API_RESPONSE_BYTES = 64 * 1024 * 1024
-MAX_RELEASE_SET_BYTES = 8 * 1024 * 1024
 
 
 class SafeRedirectHandler(urllib.request.HTTPRedirectHandler):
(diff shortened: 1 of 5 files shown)


zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32534603634)

simings-nv (approved):

PR #1796, opened 2026-08-21
perf(vios): fetch APT packages over parallel connections

simings-nv:
## Problem

Every VIOS consumer blocks on `vios-apt-cache-init`, which downloads 224 packages (138MB) before any of them can start. APT opens one connection per repository host and that package set resolves to a single host, so the whole download runs over one connection.

## Change

Prefetch the resolved URIs with a pool of `apt-helper` workers, then run the existing `apt-get install -d` unchanged so it completes anything the prefetch missed.

Package sizes range from 2KB to 30MB, so workers take the next URI as they free up rather than splitting the list up front, which would leave one worker holding the 30MB file after the others finish.

`apt-helper` ships with APT, so this adds no dependency. It verifies each file against the hash APT printed and writes nothing on a mismatch. A failed or partial prefetch costs time and nothing else, because the existing download still runs.

## Result

Measured on `vss-vios-streamprocessing` against the real package set, 5 interleaved reps:

| connections | fetch, 5 reps (s) | median |
|---|---|---|
| 1 (current) | 6.9, 7.6, 9.2, 10.5, 8.8 | 8.8 s |
| 8 (this PR) | 2.57, 2.37, 2.62, 2.32, 2.78 | **2.57 s** |

8 rather than 16: 16 reaches a lower median but a much wider spread (0.87-2.61 vs 2.32-2.78), and predictability matters more here than peak.

`VST_APT_PARALLEL_DOWNLOADS` sets the worker count, default 8. Any value below 2, or a non-numeric one, keeps the current behaviour.

## Verification

Run against the real image with the real script, populating a fresh cache volume each time:

- default: 224 packages cached, `rc=0`
- `VST_APT_PARALLEL_DOWNLOADS=1`: prefetch skipped, 224 packages cached, `rc=0`
- `VST_APT_PARALLEL_DOWNLOADS=abc`: prefetch skipped, 224 packages cached, `rc=0`

Consumer install from the resulting cache passes the script's own `runtime_present()` check, `gst-inspect-1.0 avdec_h264` resolves, and `dpkg --audit` is clean.

### DIFF
--- a/deploy/docker/services/vios/scripts/user_additional_install.sh
+++ b/deploy/docker/services/vios/scripts/user_additional_install.sh
@@ -250,6 +250,41 @@ install_packages_with_retries() {
   return 1
 }
 
+# APT opens one connection per repository host, and this package set resolves to
+# a single host, so the 138MB download runs over one connection: measured 8.8s
+# against archive.ubuntu.com where 8 connections take 2.6s. The 224 files range
+# from 2KB to 30MB, so workers take the next URI as they free up rather than
+# splitting the list up front, which would leave one worker holding the 30MB
+# file after the others finish.
+#
+# Best effort by design. apt-helper checks each file against the hash APT
+# printed and writes nothing on a mismatch, and the apt-get install -d that
+# follows still downloads whatever is missing, so a failed or partial prefetch
+# costs time and nothing else. Set VST_APT_PARALLEL_DOWNLOADS=1 to disable.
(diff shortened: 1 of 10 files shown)


zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32537484526)

PR #1797, opened 2026-08-22
Fix/devX operational skills fixes with nemoclaw for Kubernetes Deployment (Search)

nvleeryan:
## Summary
Operating a Kubernetes `search` deployment from the NemoClaw sandbox failed at
every layer: egress was denied, the VSS CLI's Elasticsearch and VST clients
could not reach the Ingress, and the workspace docs carried a hardcoded origin
with no way for the agent to bootstrap the CLI itself. This makes that path work
end to end and removes the deployment-specific values from the repo.

## Transport fixes (`vss_core`)
All egress in the sandbox goes through an HTTP proxy — there is no direct DNS.
Two clients ignored it, so `vss search run` failed with
`Temporary failure in name resolution` against a host `curl` resolves fine.
- **`search_core/clients/elastic.py`** — `AsyncElasticsearch` defaulted to
  elastic_transport's aiohttp node, which builds its `ClientSession` without
  `trust_env`. Switched to a `_PrefixedHttpxAsyncNode` subclass. httpx honors
  proxies by default and matches every other client in `search_core`; the
  subclass exists because upstream's httpx node derives the client `base_url`
  from scheme/host/port and drops `NodeConfig.path_prefix`, which would send an
  Ingress-mounted endpoint's requests to `/<index>/_search` instead of
  `/elasticsearch/<index>/_search` and surface the 404 as a bogus
  "index does not exist".
- **`vst/client.py`** — all five `ClientSession` constructions now go through
  one `_session()` helper that sets `trust_env` only when `getproxies()` reports
  a proxy, so in-cluster deployments do not start reading `~/.netrc`. Without
  this, retrieval succeeded but every hit came back with an empty `description`
  and an `unverified` verdict, because the critic could not reach VST.

## Sandbox egress policy
- New `vss-k8s-ingress`: Kubernetes profiles publish nothing on
  `host.openshell.internal`, so the Ingress host must be named explicitly. The
  entry is an example to be set to the deployment's `global.externalHost`.
- New `vss-cli-build`: the extra indexes `services/agent/pyproject.toml`
  declares.
- Added `uv` to the existing `pypi` policy's binaries. uv resolves and downloads
  in-process rather than shelling out to pip, so wheel fetches from
  `files.pythonhosted.org` were denied even though the host was allowed.
- The Kubernetes Ingress host is a placeholder in the file. Section 3.2 of
  `deploy_nemoclaw.ipynb` rewrites it from `VSS_PUBLIC_URL` into a staged copy
  before `policy-add`, deriving the port from that origin. Left empty, the
  placeholder resolves nowhere — correct for a Compose sandbox, and a reported
  `403` rather than a silent allowlist for a Kubernetes one.

## Workspace docs and notebook
- `ENV.md` no longer hardcodes an origin. `deploy_nemoclaw.ipynb` gained a
  `VSS_PUBLIC_URL` setting and renders it into the uploaded copy; empty is a
  supported state that tells the agent to ask the user in chat rather than guess
  a hostname. Also puts `${HOME}/.local/bin` on `PATH` for uv.
- `TOOLS.md` gained a "VSS CLI checkout" section so the agent installs uv and
  clones the repo itself instead of the operator doing it by hand.

## Skill and reference corrections
- `vss-ask-video`: RT-VLM's public mount is profile-dependent. Base and LVS
  serve it at the origin root; search mounts it at `/rtvi-vlm/v1` and serves no
  bare `/v1`. Discovery now probes both, and backend detection recognizes the
  path mount — on search the endpoint is `:80` like every other route, so the
  port test never fired.
- `deployment_resolution.md`: RT-Embed and RT-CV also reach the public origin
  via `global.rtviInternalIngress`, not only the CLI-routes toggle.

## Test plan
- [x] `dev-profile-search` on Kubernetes with `ingress.cliRoutes.enabled=true`,
      operated from a NemoClaw sandbox with the policy applied.
- [x] Agent installs uv and clones the checkout unattended from `TOOLS.md`.
- [x] `vss configure --base-url <origin>` resolves all routes.
- [x] `vss search run embed` returns scored hits with populated `description`
      and real verification verdicts, no VST retry storm.
- [x] Compose search profile unchanged (root-mounted Elasticsearch still
      resolves; the path-prefix restore is a no-op there).

## Checklist


nvleeryan:
/ok to test https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1797/commits/6b95ceb7d9526d7cd549f4ada88ce8428f232e9a

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32545700517)

zac-wang-nv (approved):

PR #1798, opened 2026-08-22
perf(vios): add --prebake-vios-packages to build runtime media into the image

simings-nv:
## Problem

The VIOS runtime media packages are installed on every container start, and every other service waits on it.

The cost has two parts. The install itself, and the fact that `vss-vios-ingress` cannot start while it runs: `nginx-vst.conf` has a literal `proxy_pass http://vss-vios-streamprocessing:30001` with no `resolver`, so nginx exits at config load until that name resolves, and Docker restarts it with growing backoff. The wait is therefore amplified, and the backoff overshoots by up to one interval.

## Change

Opt in with `dev-profile.sh up --prebake-vios-packages`, or `VSS_VIOS_PREBAKE_PACKAGES=true` so CI can set it without a flag. Default behaviour is unchanged.

The flag adds a `-f` overlay that gives the streamprocessing services a `build:` section running the same install once at build time. The install script is not modified: it writes its own marker, so the entrypoint takes the skip path that already exists. `vios-apt-cache-init` becomes a no-op, since a prebaked consumer has nothing to download.

The derived image is built locally and never pushed, and carries a local-only name rather than a registry coordinate so it stays out of the release set. These packages are not redistributable, which is why the published image ships without them.

## Result

The saving scales with how long the install takes, which is disk-bound, so it differs by machine. Measured on two:

| | RTX PRO 6000 eval box | L40 dev box |
|---|---|---|
| apt step at container start | **15.42 s → 0.033 s** | **127.7 s → 0.0 s** |
| cold build | 18.2 s | 131 s |
| layer-cache hit (every later deploy) | 0.114 s | 2.9 s |
| image size | 2.857 → 3.352 GB | same |

On the eval box that is about **15 s per deploy**.

On the slower dev box, two runs of the same A/B put `vss-vios-streamprocessing` healthy at 186 s and 161 s before the change and 15 s both times after. The `vss-agent` figure is deliberately not quoted: it moved 205 s → 63 s in one run and was unchanged at ~40 s in the other, and in that second run the agent went healthy before streamprocessing did, so it was not gated on VIOS at all. Only the streamprocessing and ingress numbers reproduce.

### What this PR does not do

An earlier revision also turned `vios-apt-cache-init` into a no-op, on the grounds that a prebaked streamprocessing has nothing to download. That was wrong: the same shared cache feeds the NVStreamer services, which this overlay does not prebake, so it left them doing a full network install each. The init is left alone.

That also means this PR does not shorten the window in which `vss-vios-ingress` crash-loops waiting for streamprocessing to exist. That is a separate problem with its own fix in #1799.

## Rebuild cost

The layer cache has two inputs, the base image and `user_additional_install.sh`. It rebuilds when either changes, not per deploy and not per trial. The eval harness preserves images across trials, so a box builds once.

A rebuild is break-even rather than a regression, because it runs the same install the container would have run anyway: 18.2 s at build time against 15.42 s at container start on the eval box.

## Why `build:` and not a prebuilt tag

The deploy runs `docker compose up --pull always`, which fails on an image that exists only locally, and a service-level `pull_policy: never` does not override the CLI flag. A service with `build:` is built rather than pulled, so it works under the existing flags.

Note for future edits: paths inside a `-f` overlay resolve against the project directory, not against the overlay file.

## Verification

Checked on the eval box against this PR's own Dockerfile and build context, with the repo's `user_additional_install.sh` confirmed byte-identical to the image's:

- entrypoint takes the existing skip path (`Additional packages already installed; skipping APT.`)
- marker written at `/var/lib/vios/additional-packages-installed-<hash>`
- `gst-inspect-1.0 avdec_h264` resolves, and `x264enc`
- `dpkg --audit` clean
- `/var/lib/apt/lists/*` removed

## Scope

This changes only the package install at streamprocessing container start. It is independent of #1799, which removes the ingress crash-loop, and the two compose cleanly.

### DIFF
--- a/deploy/docker/scripts/dev-profile.sh
+++ b/deploy/docker/scripts/dev-profile.sh
@@ -34,6 +34,9 @@ ngc_cli_api_key="${NGC_CLI_API_KEY:-}"
 nvidia_api_key="${NVIDIA_API_KEY:-}"
 openai_api_key="${OPENAI_API_KEY:-}"
 dry_run="false"
+# Build the VIOS runtime-media packages into a local image instead of installing
+# them on every container start. Env var so CI can set it without a flag.
+prebake_vios_packages="${VSS_VIOS_PREBAKE_PACKAGES:-false}"
 
 # NIM-related defaults
 # LLM configuration
@@ -522,6 +525,9 @@ function usage() {
   echo "  --llm-device-id                  LLM device ID."
   echo "                                   • Not allowed when --use-remote-llm is passed"
   echo "                                   • DGX-SPARK, IGX-THOR, AGX-THOR: not accepted"
+  echo "  --prebake-vios-packages          Build the VIOS runtime-media packages into a local image instead of"
+  echo "                                   installing them on every container start. Cuts the VIOS start from"
+  echo "                                   ~100 s to ~0.1 s. The image is built locally and never pushed."
   echo "  --use-remote-llm                 Use remote LLM; requires LLM_ENDPOINT_URL on the host (both are required together)."
   echo "  --llm-model-type                 LLM backend type when --use-remote-llm is passed: nim or openai."
   echo "  --llm-env-file                   Path to LLM env file. Absolute or relative to CWD."
@@ -657,6 +663,11 @@ function process_args() {
         shift
         vlm_device_id="${1}"
         options_provided+=("vlm-device-id")
+        shift
+        ;;
+      --prebake-vios-packages)
+        prebake_vios_packages="true"
+        options_provided+=("prebake-vios-packages")
         shift
         ;;
       --use-remote-llm)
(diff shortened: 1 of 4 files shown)


zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32542090563)

zac-wang-nv (commented):

## Automated OSRB Review — GAP INFORMATIONAL

PR adds a local-only prebaked overlay that reuses the unchanged user_additional_install.sh (same VIOS runtime-media package set already OSRB-approved) and derives from the first-party vss-vios-streamprocessing base image. No manifest, import, or license changes. No completed OSRB-relevant CI check is failing. The only surfaced item is a base-image reference to confirm; it carries its own OSRB record and does not block.

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32542350116)

zac-wang-nv (approved):
LGTM

PR #1801, opened 2026-08-22
feat(vss-deploy-profile): add a one-shot deploy script the skill can call

simings-nv:
## Problem

Deploying a profile runs as a sequence the agent assembles itself: resolve credentials, edit `HARDWARE_PROFILE` / `COMPOSE_PROFILES` / `VLM_BASE_URL` in `generated.env` with `sed`, regenerate and normalise `resolved.yml`, bring the stack up, then poll for health.

`dev-profile.sh` already computes all of that. Nothing calls it as a unit, so the sequence is re-derived on every deployment.

## Change

`scripts/deploy.sh` takes `--profile` (and `--mode` for alerts), resolves credentials and the hardware profile, invokes `dev-profile.sh`, blocks until every service the profile declares is present and healthy, and prints a per-container status block with log tails for anything unhealthy.

Three behaviours matter for a caller that has to react to failure:

- **Exits non-zero on failure**, so an aborted deploy cannot be read as a successful one.
- **Aborts on a restart loop** rather than waiting out the timeout: a container restarting repeatedly will never become healthy, and burning the full window on it helps nobody.
- **Requires every declared service to exist**, not just "the containers that happen to be up". Compose creates them over tens of seconds, so an early poll otherwise sees a complete-looking set and returns before the slow services exist.

## Result

Measured with `claude-code` driving `aws/anthropic/bedrock-claude-opus-4-6`, thinking disabled, cold box each run, n=3 per cell (n=2 for one), medians:

| profile | turns | cost |
|---|---|---|
| `base` | 68 → **2** | $1.66 → **$0.10** |
| `lvs` | 81 → **12** | $1.79 → **$0.59** |
| `alerts` (real-time) | 108 → **10** | $3.33 → **$0.52** |

Each turn is a round trip to the model gateway, so turn count is what the change actually removes. On `base` the agent goes from 68 turns and 35 shell commands to 2 turns and 1 command.

The agent keeps the part worth keeping: in the runs where a deploy failed, it read the status block and investigated (checked disk headroom, looked for a missing CLI) in a handful of extra turns rather than re-deriving the deployment.

## What is deliberately not claimed

**Wall clock.** The script waits for every declared service to be healthy; an agent assembling the deploy by hand tends to stop earlier, so the two are not measuring the same finish line. On the heavier profiles the script is slower in wall clock precisely because it is stricter.

Measurements are from one L40 dev box. Turn count and cost are properties of the prompt and model and should carry; absolute wall clock will not.


zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32568121215)

PR #1802, opened 2026-08-22
fix(deploy): probe fast during start-up with healthcheck start_interval

simings-nv:
## Problem

A container that is ready immediately is not marked healthy until its first probe fires, and without `start_interval` the first probe waits a full `interval`. Anything gated on it through `depends_on: service_healthy` waits with it, doing nothing.

Measured on Docker 29.4.3, container ready at t=0, `interval: 60s`:

| | healthy at |
|---|---|
| no `start_interval` | **62 s** |
| `start_interval: 3s` | **3 s** |

Every local NIM under `deploy/docker/services/nim/` declares `interval: 60s` with `start_period: 0s`, so each idles for about a minute after it is already serving.

## Change

Add `start_interval` and a generous `start_period` to the 18 NIM healthchecks, which all declare `interval: 60s` with `start_period: 0s`.

Both are needed: **`start_interval` only applies inside `start_period`**, so a service declaring `start_period: 0s` gets no benefit from `start_interval` alone. That is why this touches `start_period` as well.

Steady-state `interval` is unchanged, so this adds no probe load once a service is up. It only changes how quickly readiness is noticed during start-up.

`rtvi-vlm` and `rtvi-embed` already declare `start_period: 1200s`; this follows that precedent for the rest.

## Scope

18 healthchecks across 9 files, all under `deploy/docker/services/nim/`. The diff is exactly 18 `start_period` and 18 `start_interval` lines and nothing else. No image, dependency or service-topology change.

**Deliberately not extended to the non-NIM services.** Their helm counterparts use a Kubernetes `startupProbe`, whose first probe fires at `initialDelaySeconds` (default 0) rather than after a period, so the k8s path does not have this defect. Mirroring the change there would raise probe rates and widen failure budgets without fixing anything, which is also why the helm-sync bot rated its proposed sync 3/5. The NIM charts render `NIMService` CRs whose probe timings are owned by the NIM operator, so there is no chart-side counterpart to drift from.

## What is not claimed

**The end-to-end deploy saving is not measured here.** A full-stack before/after on the dev box was attempted and could not run: the base profile pulls `nvcr.io/nvstaging/vss-core/vss-rt-vlm`, which returns `Access Denied` with the NGC key available on that box, so both arms failed identically at the compose stage rather than producing a comparison.

What is measured is the mechanism, in isolation and reproducibly: 62 s to 3 s for one service. How much of that reaches the deploy depends on how many gated services sit on the critical path, which should be confirmed on a box that can pull the full image set.

### DIFF
--- a/deploy/docker/services/nim/cosmos-reason1-7b/compose.yml
+++ b/deploy/docker/services/nim/cosmos-reason1-7b/compose.yml
@@ -60,6 +60,8 @@ services:
       timeout: 650s
       retries: 2
 
+      start_period: 1200s
+      start_interval: 5s
   cosmos-reason1-7b-shared-gpu:
     image: nvcr.io/nim/nvidia/cosmos-reason1-7b:1.4.1
     container_name: cosmos-reason1-7b
@@ -104,5 +106,7 @@ services:
       interval: 60s
       timeout: 650s
       retries: 2
+      start_period: 1200s
+      start_interval: 5s
 volumes:
   cosmos_reason1_7b_cache:
--- a/deploy/docker/services/nim/cosmos-reason2-8b/compose.yml
+++ b/deploy/docker/services/nim/cosmos-reason2-8b/compose.yml
@@ -58,6 +58,8 @@ services:
       timeout: 650s
       retries: 2
 
+      start_period: 1200s
+      start_interval: 5s
   cosmos-reason2-8b-shared-gpu:
     image: nvcr.io/nim/nvidia/cosmos-reason2-8b:1.6.0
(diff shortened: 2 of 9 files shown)


zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32576729524)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32586447555)

PR #1804, opened 2026-08-23
perf(skills): cut agent idle time in the VSS deploy leg

simings-nv:
## What this changes

Three fixes to the `vss-deploy-profile` skill and the `lvs` profile env, found by profiling the deploy leg end to end under host telemetry.

1. **`Step 5` launches the deploy detached.** `docker compose up -d` blocks until every `depends_on` with `condition: service_healthy` is satisfied. The resolved `lvs` compose has 29 services and 14 such conditions, gating on `rtvi-vlm` and the LLM NIM, each 10-20 min cold. Agent Bash tools cap at 10 minutes, so a foreground call cannot succeed on this profile. Measured: `up -d` killed at exactly 2m00s with exit 143 after pulling 21.2 GB.
2. **`setsid` is mandatory, and a background tool call is not a substitute.** A backgrounded tool call is still a child of the agent session and is signalled when the turn ends, so the deploy dies partway. Measured on two otherwise identical arms: `setsid` reached 30 containers and 6/6 checks; a plain background call reached 16 and 2/6, with the LLM NIM, `vss-lvs`, `vss-agent` and `vss-agent-ui` never started.
3. **The `lvs` image coordinates are wrong, and the pull starts too late.** `LVS_IMAGE` pointed at a ghcr path whose `3.3.0-rc2` tag is HTTP 404; RT-VLM fell back to a private ghcr path that is HTTP 403. Both ship from `nvcr.io/nvstaging/vss-core`. The variables Compose actually reads are `VSS_RT_VLM_IMAGE` / `VSS_RT_VLM_TAG` and `LVS_IMAGE` / `LVS_TAG`; `lvs-profile.md` named `RTVI_VLM_IMAGE_TAG`, which no compose file interpolates. A new `Step 0b` starts `docker compose pull` in the background right after the credentials gate, so the download overlaps the reasoning in Steps 1 to 4 instead of following it.

## Measured effect

A/B on one host, same instruction, same skill, image cache reset to an identical snapshot between arms so neither inherits the other's downloads. Control carries fix 1 and 2; treatment adds fix 3.

| | control | treatment | delta |
|---|---|---|---|
| Wall clock | 6m05s | **4m02s** | **-123s (-33.7%)** |
| LLM inference | 5m16s | **3m27s** | -109s (-34.3%) |
| Tool calls | 111 | **68** | -43 |
| Containers deployed | 30 | 30 | - |
| Eval checks | **6/6** | **6/6** | - |

Both arms deploy the full stack and pass every check, so this is like for like.

## The same pattern across other profiles

One profiled leg each, on the pre-fix baseline, to show the idle is not specific to `lvs`.

| profile | wall | LLM | idle | tool calls | checks at agent exit | checks after convergence |
|---|---|---|---|---|---|---|
| `base` | 6m29s | 4m53s (75.1%) | 62.8% | 102 | 3/7 | 7/7 |
| `alerts_cv` | 9m52s | 6m52s (69.6%) | 63.6% | 120 | 0/7 | 7/7 |
| `alerts_vlm` | 4m46s | 3m53s (81.7%) | 69.9% | 61 | 4/6 | 6/6 |
| `search` | 6m27s | 5m36s (86.9%) | 86.1% | 68 | 0/8 | 8/8 |
| `warehouse` | 14m05s | 4m54s (34.8%) | 30.9% | 68 | 15/15 | 15/15 |

## A separate finding, not addressed by this PR

Every leg was scored twice: once the moment the agent's turn ended, and again after the deploy finished converging. They disagree.

| leg | at agent exit | after convergence | convergence wait |
|---|---|---|---|
| lvs control | 2/6 | **6/6** | 503s |
| lvs treatment | 2/6 | **6/6** | 453s |
| `base` | 3/7 | **7/7** | 251s |
| `alerts_cv` | 0/7 | **7/7** | 775s |
| `alerts_vlm` | 4/6 | **6/6** | 484s |
| `search` | 0/8 | **8/8** | 775s |

The harness verifies as soon as the agent exits, so a deploy that is merely still converging is recorded as a failure. That is worth a separate change to the eval harness rather than to this skill.

## Test plan

`ok to test`

- `lvs` deploy leg on RTXPRO6000BW, both before and after: both reach 30 containers and 6/6.
- `docker compose config --images` resolves `nvcr.io/nvstaging/vss-core/vss-rt-vlm:3.3.0-26.08.2` and `nvcr.io/nvstaging/vss-core/vss-video-summarization:3.3.0-rc2` from the checked-in defaults with no overrides.
- Both nvcr manifests return HTTP 200; the previous ghcr coordinates return 404 and 403 respectively.


zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32648819937)

simings-nv:
Closing this. The change needs more work before review is worth anyone's time:

- Step 0b references `$PROFILE` at SKILL.md:149, but that variable is first defined at :203 in Step 1c. The ordering is wrong.
- `lvs-profile.md:13` says "both defaults must be overridden" while the same change pins those defaults in `overrides.env`, so the sentence now contradicts the code.
- The before/after data backing it covered one profile, not all of them.

Reopening once the A/B across every profile is complete and these are fixed. Thanks to the skills-review bot for catching the first two.

PR #1805, opened 2026-08-23
fix(deploy): pin RT-VLM centrally and stop the deploy step dying at the tool timeout

simings-nv:
## What this changes

Three defects in the deploy path, each measured on an RTX PRO 6000 Blackwell
host by running the real Harbor eval instruction through a headless agent under
1 Hz host telemetry.

**1. RT-VLM resolved to a registry the deployment cannot pull from.**
`containers.env` described RT-VLM as part of the shared GHCR set, so it
inherited `VSS_CONTAINER_REGISTRY`. That GHCR repository is private:

| coordinate | manifest |
|---|---|
| `ghcr.io/nvidia-ai-blueprints/vss/vss-rt-vlm:develop-latest` | **HTTP 403** |
| `nvcr.io/nvstaging/vss-core/vss-rt-vlm:3.3.0-26.08.2` | **HTTP 200** |

403 persists after a successful `docker login`. Measured consequence: an
`alerts` deploy leg scored **0 of 7 checks with zero containers** because the
agent correctly refused to start Compose on an unpullable image, exactly as the
skill's own entitlement gate (Step 3c) instructs.

Fixed in `containers.env` rather than per profile. That matters: `dev-profile.sh`
and `blueprint-deploy.sh` both do `set -a; source containers.env`, which
**exports** these names, and Compose gives the environment precedence over every
`--env-file`. A per-profile override is silently inert on the scripted path. An
earlier revision of this change made exactly that mistake and was rejected in
review.

**2. The skill documented two env layers where there are three.** Both scripted
entrypoints pass `containers.env` first. Omitting it drops every curated image
coordinate back to the compose-level fallbacks, which is how defect 1 reaches a
deploy that follows the skill.

**3. `Step 5` ran `up -d` in the foreground.** `up -d` does not return when
containers are created; it blocks until every `depends_on` with
`condition: service_healthy` is satisfied. The `lvs` profile resolves to 29
services with 14 such gates, two of them model servers documented at 10 to 20
minutes cold. Agent Bash tools cap well below that. Observed repeatedly:

```
Exit code 143
Command timed out after 2m 0s
```

leaving a half-built stack. Step 5 is now a supervised detached job that writes
its own exit status to a unique directory, and Step 5a polls that status rather
than the container count.

## Baseline reliability, measured

Nine deploy legs on unmodified `develop` across six profiles: **four passed,
five failed.** Every failure left the application layer missing. Four were the
same sequence: foreground `up -d` killed at the tool timeout, agent retries by
backgrounding it as a tool call, that call dies when the turn ends. One was
defect 1 above.

The baseline is bimodal rather than deterministically broken. Two runs of the
identical unmodified code on `lvs` differed by 26% in wall clock (860s and
682s), and one later run of the same code failed outright. A single pair per
profile shows direction, not precision.

## Measurement status, stated plainly

The evidence above is about `develop` and is unaffected by how the fix is
implemented. **The speedup numbers are not yet re-measured against this exact
commit.** An earlier revision implementing the same three fixes differently
measured 3.0x, 3.5x and 3.8x wall-clock improvements on `base`, `lvs` and
`search`, each against a baseline that also passed. That revision was rejected
in review for fixing defect 1 at the wrong layer, so those numbers describe code
that no longer exists here. A re-measure against `b6228f235` is running and I
will post it before taking this out of draft.

## Not addressed here

Every leg was scored twice, at agent exit and after convergence. They disagree:
six of seven legs scored between 0/7 and 4/6 the moment the agent's turn ended
and were fully green 2 to 13 minutes later. The eval harness verifies at the
first point, so it records failures for deploys that come up fine. That belongs
in the harness, not this skill.

## Test plan

- `docker compose --env-file containers.env ... config --images` resolves
  `nvcr.io/nvstaging/vss-core/vss-rt-vlm:3.3.0-26.08.2` on both the skill path
  and the scripted path (`set -a; source containers.env`), which is where the
  rejected revision was inert.
- Both manifests re-verified today: nvstaging 200, ghcr 403.
- No profile env file is touched, so the profile-layout contract in
  `test-dev-profile.sh` is unchanged.
- LVS image coordinates are deliberately left alone; that is a separate change
  needing its own validation.

### DIFF
--- a/deploy/docker/containers.env
+++ b/deploy/docker/containers.env
@@ -173,9 +173,18 @@ VSS_NVSTREAMER_TAG="${VSS_NVSTREAMER_TAG:-${VSS_CONTAINER_TAG:-1.4.0-26.08.1}}"
 
 VSS_VIOS_INGRESS_IMAGE="${VSS_VIOS_INGRESS_IMAGE:-${VSS_CONTAINER_REGISTRY:-${VSS_CONTAINER_STAGING_REGISTRY}}/vss-vios-ingress}"
 VSS_VIOS_INGRESS_TAG="${VSS_VIOS_INGRESS_TAG:-${VSS_CONTAINER_TAG:-1.4.0-26.08.1}}"
-# RT-VLM is part of the shared GHCR-managed image set.
-VSS_RT_VLM_IMAGE="${VSS_RT_VLM_IMAGE:-${VSS_CONTAINER_REGISTRY:-${VSS_CONTAINER_RELEASE_REGISTRY}}/vss-rt-vlm}"
-VSS_RT_VLM_TAG="${VSS_RT_VLM_TAG:-${VSS_CONTAINER_TAG:-3.3.0-26.08.2}}"
+# RT-VLM has NOT migrated to the shared GHCR set: the GHCR repository is private
+# and returns HTTP 403 for a pull token even after a successful `docker login`,
+# so it is pinned to the staging registry explicitly rather than inheriting
+# VSS_CONTAINER_REGISTRY. Move it back to the shared knob once the GHCR
+# repository is published.
+#
+# Note the nested `${VSS_CONTAINER_REGISTRY:-...}` fallback used by the shared
+# entries above is unreachable in practice: VSS_CONTAINER_REGISTRY is always
+# assigned a default earlier in this file, so the inner alternative never
+# applies. Naming the registry directly here avoids relying on it.
+VSS_RT_VLM_IMAGE="${VSS_RT_VLM_IMAGE:-${VSS_CONTAINER_STAGING_REGISTRY}/vss-rt-vlm}"
+VSS_RT_VLM_TAG="${VSS_RT_VLM_TAG:-3.3.0-26.08.2}"
 VSS_CALIBRATION_TOOLKIT_IMAGE="${VSS_CALIBRATION_TOOLKIT_IMAGE:-${VSS_CONTAINER_RELEASE_REGISTRY}/calibration}"
 VSS_VIDEO_ANALYTICS_UI_IMAGE="${VSS_VIDEO_ANALYTICS_UI_IMAGE:-${VSS_CONTAINER_STAGING_REGISTRY}/vss-video-analytics-ui}"
 
--- a/skills/vss-deploy-profile/SKILL.md
+++ b/skills/vss-deploy-profile/SKILL.md
@@ -48,10 +48,16 @@ The deployment flow is always: copy `overrides.env` to `generated.env`, apply ov
 ```bash
 # 1. cp dev-profile-<profile>/overrides.env dev-profile-<profile>/generated.env  (clean copy)
 # 2. Apply env overrides to generated.env  (source .env and overrides.env stay untouched)
-# 3. docker compose --env-file .env --env-file generated.env config > resolved.yml  (dry-run)
+# 3. docker compose --env-file containers.env --env-file .env --env-file generated.env config > resolved.yml
 # 4. Review resolved.yml
-# 5. docker compose --env-file .env --env-file generated.env -f resolved.yml up -d
+# 5. docker compose --env-file containers.env --env-file .env --env-file generated.env -f resolved.yml up -d
 ```
+
+**There are THREE env layers and `containers.env` is the first.** It is the
+single source of truth for every first-party image coordinate, and both
+`dev-profile.sh` and `blueprint-deploy.sh` pass it. Omitting it silently drops
+every curated image default and falls back to the compose-level ones, several
+of which resolve to registries this deployment cannot pull from.
 
 `.env` is the read-only checked-in stable-default layer. `overrides.env` is the read-only checked-in profile override layer for values the deploy scripts may modify per host/profile. `generated.env` is the per-deploy working copy created from `overrides.env`. Step 1c covers this in full.
 
@@ -222,9 +228,10 @@ reference has worked examples for that profile's common scenarios.
 #   sed -i "s|^LLM_MODE=.*|LLM_MODE=remote|" "$ENV_GEN"
 #   sed -i "s|^LLM_BASE_URL=.*|LLM_BASE_URL=http://localhost:30081|" "$ENV_GEN"
 
-# Resolve compose
+# Resolve compose. containers.env FIRST; see "Instructions" above.
 cd $REPO/deploy/docker
-docker compose --env-file $ENV_SRC --env-file $ENV_GEN config > resolved.yml
+docker compose --env-file containers.env --env-file $ENV_SRC --env-file $ENV_GEN \
+  config > resolved.yml
 ```
 
 The resolved YAML is saved to `<repo>/deploy/docker/resolved.yml`.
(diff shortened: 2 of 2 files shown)


simings-nv:
Correction: the previous `/ok to test` carried `4c665671b`, which is the base commit on `develop`, not this branch's head. The correct head is below.

/ok to test b6228f23552a890eeeb43fc64936bf5b1d8dfe1f

simings-nv:
## Measured data

Every leg below is one deploy of the real Harbor eval instruction for that
profile, driven by a headless agent on an RTX PRO 6000 Blackwell host under
1 Hz host telemetry. Between arms the image cache is reset to an identical
snapshot and model-weight volumes are preserved, so neither arm inherits the
other's downloads.

### Baseline reliability on unmodified `develop`

Nine legs across six profiles. **Four passed, five failed.** Every failure left
the application layer missing.

| profile | wall | checks | containers | how it ended |
|---|---|---|---|---|
| `lvs` | 860s | 6/6 | 30 | recovered via a ~450s blocking poll |
| `lvs` | 682s | 6/6 | 30 | recovered |
| `lvs` | 474s | **2/6** | 26 | backgrounded, turn ended, deploy died |
| `base` | 660s | 7/7 | 13 | passed |
| `base` | 581s | 7/7 | 13 | passed |
| `base` | 369s | **3/7** | 11 | backgrounded, turn ended, deploy died |
| `search` | 1071s | 8/8 | 34 | passed |
| `alerts_cv` | 510s | **0/7** | **0** | RT-VLM unpullable; agent refused to start Compose |
| `alerts_vlm` | 458s | **4/6** | 29 | 2m00s timeout twice, then backgrounded and died |
| `warehouse` | 208s | **4/15** | 26 | backgrounded, turn ended, deploy died |

Two observations that matter for reading any delta:

**The baseline is bimodal, not deterministically broken.** Same code, same
host: `lvs` landed at 860s/6/6, 682s/6/6 and 474s/2/6. A single pair per
profile shows direction, not precision.

**A failed baseline is fast because it quits.** `base` at 369s scored 3/7;
`base` at 660s scored 7/7. Comparing a treatment against the 369s run would
flatter it. Only baselines that passed are used for speedup below.

### Results from the superseded implementation

These come from `f47dddcd3`, which implemented the same three fixes but pinned
RT-VLM per profile instead of centrally. **That revision was rejected in review
and is not what this PR contains.** Reported for completeness because the
problem it addressed is identical.

| profile | BEFORE | AFTER | speedup |
|---|---|---|---|
| `search` | 1071s, 8/8 | 284s, 8/8 | **3.8x** |
| `lvs` | 682s, 6/6 | 248s, 6/6 | **2.8x** |
| `base` | 581s, 7/7 | 192s, 7/7 | **3.0x** |
| `alerts_cv` | 510s, **0/7, 0 containers** | 235s, **7/7, 33 containers** | n/a, unblocked |
| `alerts_vlm` | 458s, 4/6 | 1665s, 6/6 | none; see below |
| `warehouse` | 208s, 4/15 | 232s, 15/15 | none; baseline failed |

`alerts_vlm` is the honest counter-example: the treatment took **longer**
because it completed a deploy the baseline abandoned at 4/6. It pulled 3.4 GB
against the baseline's 1.2 GB, loaded 82.6 GB of VRAM, reached 34 containers,
and went on to provision sensors. Its idle share, 17.1%, was the lowest of any
leg measured. That is not a regression, and it is also not a speedup.

### Pending

`base` and `lvs` are being re-measured against this PR's actual head
(`b6228f235`) now. The first fresh baseline is in: `base` 660s, 7/7. I will
update this thread with the paired treatment results, and will not take this
out of draft until the numbers describe the code that is actually here.

### Measurement caveat

The host used for these runs permission-classifies the nested agent, and it
blocks `nohup` while permitting `setsid`. `setsid` is the launch method this PR
mandates, so the baseline is denied a recovery route the treatment is granted.
That inflates the measured baseline failure rate on this host. CI does not
classify the agent at all, so the real-world baseline failure rate is likely
lower than 5 in 9.

simings-nv:
## Re-measured against this PR's head (`b6228f235`)

Posting these because they **contradict the framing this PR opened with**. The
earlier numbers came from a revision that was rejected in review, and the
reworked change does not reproduce them.

| profile | BEFORE (`develop`) | AFTER (`b6228f235`) |
|---|---|---|
| `base` | 660s, **7/7**, 13 containers | 749s, **7/7**, 13 containers |
| `lvs` | 495s, **2/6**, 26 containers | 651s, **6/6**, 30 containers |

### There is no wall-clock speedup. The PR is a correctness fix.

`base` is 13% *slower*, which is inside its own baseline spread (581s and 660s
across two passing samples). `lvs` is slower too, but its baseline **failed**,
so the comparison is not like-for-like in the first place.

The earlier revision measured 3.0x on `base`. That speedup came substantially
from a Step 0b that started the image pull early, which review correctly
rejected: it was never actually backgrounded (no `&`), and its hard-coded
`developer-profiles/` path is wrong for the industry profiles. Removing it
removed the overlap. Doing this correctly cost the performance win, and the
title has been changed accordingly.

### What it does deliver

**Reliability.** `lvs` went 2/6 to 6/6. The baseline failure is the documented
one: foreground `up -d` killed at the tool ceiling, agent retries by
backgrounding it as a tool call, that call dies when the turn ends, and
everything gated behind the model servers never starts.

```
t+357.7s  FOREGROUND  -> Exit code 143, timed out after 2m 0s
t+480.6s  bg=True     -> backgrounded, dies with the turn
```

**Roughly half the idle time.**

| profile | idle before | idle after |
|---|---|---|
| `base` | 45.8% | **27.2%** |
| `lvs` | 66.0% | **31.8%** |

Measured as: no CPU, GPU, network or disk activity above a floor, sampled at
1 Hz. LLM time falls too (327s to 245s on `base`), and tool calls drop from 88
to 75 and 74 to 63. The wall clock rises because the agent now waits for the
deploy to genuinely finish instead of dying part-way or declaring victory on a
stable container count.

### Baseline reliability, four `lvs` samples on unmodified `develop`

| wall | checks | how it ended |
|---|---|---|
| 860s | 6/6 | recovered via a ~450s blocking poll |
| 682s | 6/6 | recovered |
| 474s | **2/6** | died; a `nohup` attempt was refused by the measuring host |
| 495s | **2/6** | died; no host involvement, clean reproduction |

Two of four fail. The measuring host blocks `nohup` while permitting `setsid`,
which inflates that rate, but the fourth sample failed with no host involvement
at all, so the failure mode is real independent of the instrumentation.

### Honest summary

Take this for the RT-VLM coordinate fix and for making the deploy step
survivable. Do not take it for speed: on the profiles where the baseline works,
it is the same or slightly slower.

simings-nv:
## Correction to the previous comment

The table I posted above compared **agent turn length**, which excludes the time
the deploy spends converging after the agent stops. That is the wrong metric and
it understated the result on `lvs`. Time to a green board:

| | agent wall | + convergence | **time to green** | checks |
|---|---|---|---|---|
| `base` before | 660s | 303s | **963s** | 7/7 |
| `base` after | 749s | 303s | **1052s** | 7/7 |
| `lvs` before | 495s | **2403s** | **2898s** | **2/6** |
| `lvs` after | 651s | 303s | **954s** | 6/6 |

`lvs` before never converged. The 2403s is the measurement harness's wait cap:
it waited 40 minutes on a deploy that had already died, then scored 2/6. So the
baseline consumed roughly 2900 seconds and produced no working stack, while the
same profile on this branch reaches a full green board in 954s.

`base` is unaffected by the correction. Both arms converge in the same 303s, so
the treatment really is about 9% slower to green there, with both passing.

### Revised summary

- `base`: both arms green; this branch is ~9% slower to green. No speedup.
- `lvs`: baseline spends ~2900s and fails; this branch is green in 954s.

The `lvs` difference is the reliability fix expressed in time rather than a
throughput win. A deploy that never finishes cannot be made faster, only made
to finish.

Idle time still roughly halves on both profiles (45.8% to 27.2% on `base`,
66.0% to 31.8% on `lvs`), measured as no CPU, GPU, network or disk activity
above a floor at 1 Hz.

Apologies for the churn in this thread. I would rather correct it here than
leave a number I know to be measured wrong.

simings-nv:
## Correcting my own numbers again, including the ones I opened with

I found the bug. It was in how I measured, and it affected **every** figure I
have posted on this PR, in both directions.

`docker compose up -d` blocks on health gates. When the agent's turn ends
before the deploy finishes, the rest of the convergence happens afterwards and
my harness recorded it as a separate field. I reported the agent's turn length
and described it as deploy time. That silently credits whichever arm deferred
more work past the end of its turn.

Every original speedup was inflated:

| profile | I claimed | actual, time to green |
|---|---|---|
| `search` | 1071→284s = **3.8x** | 1191→656s = **1.82x** |
| `base` | 581→192s = **3.0x** | 702→524s = **1.34x** |
| `lvs` | 474→248s = **1.9x** | 614→722s, 2/6 → 6/6 |
| `alerts_cv` | 510→235s = 2.2x | 630→739s, **0/7 → 7/7** |
| `alerts_vlm` | 458→1665s | 668→1786s, 4/6 → 6/6 |
| `warehouse` | 208→232s | 409→394s, 4/15 → 10/15 |

And my "there is no speedup" comment above was wrong in the other direction: I
computed time-to-green for the new runs and compared it against agent-turn
figures from the old ones. Two different metrics.

### The comparison, both arms on the same metric

| implementation | `base` before → after | ratio |
|---|---|---|
| earlier revision (`f47dddcd3`) | 702 → 524s, 7/7 both | **1.34x faster** |
| this branch (`b6228f235`) | 963 → 1052s, 7/7 both | **0.92x, slower** |

So the earlier revision was genuinely faster, just 1.34x rather than the 3.0x I
claimed. This branch is slower than that revision, and the cause is known: it
drops the Step 0b early image pull, which review correctly rejected because the
command was never actually backgrounded and its hard-coded path is wrong for
the industry profiles. The right response is to reimplement that step properly,
not to leave it out.

### What holds regardless of the metric error

- `alerts_cv` goes from **0/7 with zero containers** to a full deploy. That is
  the RT-VLM registry fix and it is not a timing claim.
- `lvs` baselines fail 2 of 4 on unmodified `develop`, once with no involvement
  from the measuring host at all.
- Idle time roughly halves: 45.8% → 27.2% on `base`, 66.0% → 31.8% on `lvs`.

### Fixed in the tooling

`build_timeline.py` now emits `time_to_green_s` alongside `wall_seconds`, and
the field carries a comment saying it is not deploy duration and must not be
compared across runs that end their turns at different points. The summary line
now reads `time-to-green 1051.9s (agent 748.9s + converge 303s)` so the two
cannot be conflated again.

Apologies for the number of corrections in this thread. Each one replaced a
figure I could no longer defend.

simings-nv:
Closing this. The measurements backing it were taken with the wrong metric: I reported the agent's turn length as deploy duration, which inflated every speedup figure, and my later correction compared that against a different metric and understated it. Neither set of numbers should be relied on.

What is independently verified and worth reopening separately:
- RT-VLM resolves to a private ghcr repository returning HTTP 403; nvcr.io/nvstaging/vss-core/vss-rt-vlm:3.3.0-26.08.2 returns 200. One profile could not deploy at all as a result.
- Foreground `up -d` is killed at the agent Bash tool ceiling (exit 143 at 2m00s) because it blocks on 14 service_healthy gates.

Those two stand on their own and do not depend on any timing claim. The performance framing does not, and this branch is measurably slower on `base` than `develop` on time-to-green.

Will resubmit as a correctness-only change once the numbers are re-derived on a single consistent metric.

PR #1827, opened 2026-08-25
Model routing for VSS via a Switchyard deployment notebook

simings-nv:
## What

Adds `deploy/docker/scripts/deploy_model_routing.ipynb`, following the `deploy_nemoclaw.ipynb` pattern. The notebook:

1. builds Switchyard from a pinned source ref (no image pinned or pulled),
2. runs it as a local container and verifies live requests route across both configured targets (`/v1/stats` plus the routing decision log),
3. composes, but never deploys, a VSS build whose `LLM_BASE_URL` points at the router.

Routing is disabled by default: the composed build lands in `_builds/model-routing/` and nothing about a running or future VSS deployment changes unless the operator deploys that build themselves.

## What it does not do

- No service is added to any VSS compose file. The repoint uses the existing remote-LLM override set (`LLM_MODE=remote`, `LLM_NAME_SLUG=none`), which also resolves the local LLM NIM out of `COMPOSE_PROFILES` and frees its GPU.
- No checked-in profile is edited; the rewrite happens in a build copy.

## Security posture

- With the default `openai_chat` upstream the router holds no credential: callers present their own key and the router forwards it (`forward_auth`). An unauthenticated caller gets the upstream's 401.
- The router publishes on loopback and the Docker bridge gateway only; the composed `LLM_BASE_URL` points at that gateway, so the build works exactly as validated.
- CI contains no credential at all: the check runs against a local stub upstream and verifies routing mechanics, the decision-log split across both targets, and the composed VSS build. The interactive notebook defaults to the real endpoint.

## CI

`model-routing-notebook.yml` executes the notebook end to end through `.github/skill-eval/model-routing/notebook_routing_adapter.py`, a standard-library-only adapter: notebook cells are plain Python and run in one shared namespace, with CI inputs injected in memory before the notebook's Derived settings and readiness markers asserted (`ROUTER_VERIFIED`, `VSS_ROUTING_COMPOSE: valid`, `ROUTER_TEARDOWN: done`). No pip install, no venv; per-run port/container/workdir isolation and always-cleanup.

## How to use it

1. Run `deploy/docker/scripts/deploy_model_routing.ipynb` with `NVIDIA_API_KEY` exported. It builds Switchyard from the pinned ref, starts the router locally, verifies live requests split across both targets, and writes a ready build into `_builds/model-routing/` (`override.env` plus `resolved.yml`).

2. To enable routing, deploy that build:
   ```bash
   docker compose --env-file _builds/model-routing/override.env \
     -f deploy/docker/compose.yml up -d
   ```
   VSS's LLM traffic then goes through the router, and the local LLM NIM resolves out of `COMPOSE_PROFILES`, freeing its GPU. Disabling is deploying with the original Foundation env again.

3. `LLM_NAME` selects the routing policy, not a model: `switchyard/random` splits ~50/50 for a cost baseline, `switchyard/control` sends everything to the efficient target, and custom routes are a `config.toml` entry away.

4. Observe with `GET /v1/stats` on the router and the per-request decision log (`routing.jsonl`).

The intended measurement: run the same VSS eval once through `control` and once through `random` (or a smarter route), same judge, and compare quality against token spend.

## Validation

- Real-endpoint runs (local, NVIDIA inference hub): 12/12 routed probes across both targets in every run, both upstream wire formats, compose validation green, no service added, teardown clean.
- CI check green against the real hub on an earlier iteration, and again against the stub upstream.


zac-wang-nv (commented):

Safety net fired (empty License Diff, no dep-manifest change, source file added). The one changed source file imports only Python stdlib, so no IMPORT_UNAPPROVED. But the PR introduces a new third-party component, Switchyard, built from a pinned source ref; it is not captured in license_diff_rows, container_dependency_changes, or as a code import, and no approval_source covers it. Its Apache-2.0 label appears only in untrusted notebook text, which cannot be treated as classifiable approval evidence. New-component review trigger with missing structured evidence yields INCONCLUSIVE, not CLEAN.

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32795965380)

simings-nv:
@greptileai review

simings-nv:
@greptileai review

.github/skill-eval/model-routing/notebook_routing_adapter.py
```diff
@@ -0,0 +1,369 @@
...
+            sys.__stdout__.flush()
+
+    previous_cwd = Path.cwd()
+    os.chdir(cwd)
+    try:
+        for index, source in enumerate(cells):
+            code = compile(source, f"{path.name}:cell-{index}", "exec")
+            with redirect_stdout(_Tee()):
+                try:
+                    exec(code, namespace)  # noqa: S102 - checked-in notebook
```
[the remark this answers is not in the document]
simings-nv: Every run: step in any workflow on these runners is already contributor-controlled code executing on the shared host after the maintainer-only /ok to test approval; the adapter's exec adds no capability beyond that baseline. Containerizing it would not isolate anything real either, since the job requires the docker socket to run the router, and socket access is host-equivalent. The controls are the ones the repository already relies on: maintainer-gated mirroring, no credential anywhere in the job (stub upstream), and run-scoped state with always-cleanup.

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32797243967)

simings-nv:
@greptileai review

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32798363946)

deploy/docker/scripts/deploy_vss_switchyard.ipynb
zac-wang-nv, line 1: can we just rename it as deploy_vss_switchyard.ipynb?

simings-nv:
Good catch, that was right. `services/agent/compose.yml:102` reads `LLM_BASE_URL`, so section 5 was repointing vss-agent only. Added NemoClaw routing in c1710c5.

NemoClaw takes its own provider settings rather than `LLM_BASE_URL`, so it needed a separate path, but it stays configuration-only: a router is an OpenAI-compatible endpoint, which is exactly what option (b) already describes. So section 1.2 of `deploy_nemoclaw.ipynb` gets an option (d), commented out like the others so a top-to-bottom run is unchanged:

```
NEMOCLAW_ENDPOINT_URL = "http://host.openshell.internal:4000/v1"
NEMOCLAW_MODEL        = "switchyard/random"   # a route id, not a model
COMPATIBLE_API_KEY    = ""                    # your upstream key; the router forwards it
```

Two things the sandbox needs that were worth digging out rather than assuming:

- Its egress policy allowlists host ports one at a time, and 4000 was not among them, so a routed agent would have been blocked. Added an entry for it in `vss_nemoclaw_policy.yaml` alongside the existing 8000 and 30888 ones.
- The bundled inference proxy only fronts public HTTPS endpoints (it hardcodes `HTTPSConnection(host)` and drops the port), so it cannot front a local plaintext router. The workable path is the same one option (b) uses, reaching the host directly as `host.openshell.internal`, which the sandbox resolves to a bridge gateway. The routing notebook now checks both reachability and the allowlist and says so before reporting ready.

Verified against the router: `/v1/models` advertises the route id, so nothing rejects a route used as a model name, and chat completions through it return 200 routed to a real target.

One thing I have not been able to verify from here: whether the OpenClaw harness sends OpenAI chat completions or Anthropic messages. The router serves both, but the credential-forwarding config pins callers to OpenAI. If OpenClaw speaks Anthropic, the router config needs one line changed to hold the key instead, which I have tested working. Happy to confirm that on a box with a live sandbox.

zac-wang-nv (approved):

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32992934026)

deploy/docker/scripts/deploy_nemoclaw.ipynb
```diff
@@ -166,6 +167,36 @@
...
+      ]
+    },
+    {
+      "cell_type": "code",
+      "execution_count": null,
+      "metadata": {},
+      "outputs": [],
+      "source": [
+        "# (d) Model router — uncomment and fill in, then run.\n",
+        "# NEMOCLAW_ENDPOINT_URL = \"http://host.openshell.internal:4000/v1\"\n",
```
nvleeryan, line 195: plesae set `NEMOCLAW_PROVIDER` accordingly similar to all other selections a,b,c to avoid broken stage when user switches from one to another.

deploy/docker/scripts/deploy_nemoclaw.ipynb
```diff
@@ -166,6 +167,36 @@
         "NEMOCLAW_MODEL = \"nvidia/nemotron-3.5-lightning-30b-a3b\"  # leave blank to use the NemoClaw default,, or set a build.nvidia.com model id"
       ]
     },
+    {
+      "cell_type": "markdown",
+      "metadata": {},
+      "source": [
+        "#### (d) Model router — *route across models instead of pinning one*\n",
+        "\n",
+        "A router is an OpenAI-compatible endpoint, so this is option (b) with a route\n",
```
nvleeryan, line 176: does switchyard only run locally?  (b) mainly deals with local LLM deployment so want to ensure it's consistent

deploy/docker/scripts/deploy_nemoclaw.ipynb
```diff
@@ -166,6 +167,36 @@
         "NEMOCLAW_MODEL = \"nvidia/nemotron-3.5-lightning-30b-a3b\"  # leave blank to use the NemoClaw default,, or set a build.nvidia.com model id"
       ]
     },
+    {
+      "cell_type": "markdown",
+      "metadata": {},
+      "source": [
+        "#### (d) Model router — *route across models instead of pinning one*\n",
```
nvleeryan, line 174: if i understand correctly, switchyard needs to be up before running nemoclaw onboard as it will try to validate the LLM endpoint.  If so, you might want to mention it

assets/vss_nemoclaw_policy.yaml
```diff
@@ -193,6 +193,16 @@ network_policies:
           - 172.17.0.0/16
           - 172.18.0.0/16
           - 172.20.0.0/16
+      # Model router (Switchyard), when the agent's LLM traffic is routed.
+      # Only reachable when the operator runs the router; see
+      # deploy/docker/scripts/deploy_vss_switchyard.ipynb.
+      - host: host.openshell.internal
```
nvleeryan, line 199: Do we really want to persist this in the policy yaml regardless?  when user chooses (a/b/c) this policy is basically useless and open a port unnecessarily (security risk?).

 I saw that there's already a cell in `deploy_vss_switchyard` that checks the policy yml.  Why not just add the policy on-demand from there instead?  @zac-wang-nv for viz.

deploy/docker/scripts/deploy_nemoclaw.ipynb
simings-nv: Good catch. Let me follow up with a PR quickly.

deploy/docker/scripts/deploy_nemoclaw.ipynb
simings-nv: No, the router is just an OpenAI-compatible endpoint and can run anywhere reachable. The example is local for the same reason (b) is: the sandbox reaches host services through host.openshell.internal, and the egress policy is scoped to that host plus the Docker bridge CIDRs. A remote router works but needs its own policy entry for that host and port, so I kept the documented path to the one the shipped policy already covers.

deploy/docker/scripts/deploy_nemoclaw.ipynb
simings-nv: Yes. The (d) note says to start the router first via deploy_vss_switchyard.ipynb, but it doesn't say why, so I'll make the ordering and the reason explicit.

assets/vss_nemoclaw_policy.yaml
simings-nv: Agreed. Let me move it to on-demand in a follow up PR.

PR #1829, opened 2026-08-25
fix(vss-configurator): handle partial NVStreamer stream list due to cold start

munjalp6:
## Summary
- NVStreamer registers files serially. vss-configurator used to treat the first non-empty `/api/v1/sensor/streams` (or live streams) response as ready, call VST `/sensor/add` once, and never reconcile — so warehouse cold start could register 2 of 3 cameras while pods looked healthy.
- Honor `NUM_STREAMS` (already set in Helm/Compose, previously unread). Wait until the list reaches that count. If `NUM_STREAMS` is `0`/unset, wait for two consecutive equal non-empty counts.
- If NVStreamer is down, returns non-200, or returns 0 streams: retry every 5s with **no timeout** (previous behavior).
- If the list is non-empty but still below `NUM_STREAMS` (or not yet stable): wait up to `NVSTREAMER_STREAMS_ENDPOINT_TIMEOUT` (**default 120s**), then commit what we have and warn.

```mermaid
flowchart TD
  A[GET NVStreamer streams] --> B{Response}
  B -->|error / non-200 / empty| C[Sleep 5s]
  C --> A
  B -->|N >= NUM_STREAMS| D[Commit list to VST]
  B -->|0 < N < NUM_STREAMS| E{Partial wait elapsed?}
  E -->|under 120s| F[Sleep 5s]
  F --> A
  E -->|timeout| G[Warn and commit partial list]
```

## Test plan
- [ ] Unit: `tests/test_fetch_nvstreamer_streams.py` (wait for `NUM_STREAMS`, stable-count fallback, 2 min partial timeout, down/empty retries past 2s then succeeds)
- [ ] Warehouse 2D/3D cold start with `NUM_STREAMS` matching sample videos; VST has all cameras after configurator preload
- [ ] Confirm a late NVStreamer (HTTP not up yet) does not skip preload
- [ ] Confirm a stuck partial list (e.g. `NUM_STREAMS` too high) proceeds after ~2 minutes with a warning

### DIFF
--- a/services/configurators/vss-configurator/README.md
+++ b/services/configurators/vss-configurator/README.md
@@ -462,7 +462,8 @@ services:
 | `RECOMPUTE_BEV_CENTERS_ENABLED` | `false` | Recompute BEV group origins via `spatialai_data_utils` (3D mode only) |
 | `NVSTREAMER_STREAMS_ENDPOINT` | `http://localhost:30000/api/v1/live/streams` | NVStreamer streams endpoint |
 | `NVSTREAMER_SENSOR_STATUS_ENDPOINT` | `http://localhost:30000/api/v1/sensor/status` | NVStreamer status endpoint |
-| `NVSTREAMER_STREAMS_ENDPOINT_TIMEOUT` | `100` | Timeout for NVStreamer endpoint (seconds) |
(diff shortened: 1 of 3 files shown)


zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32814082837)

asonny02 (approved):

PR #1839, opened 2026-08-25
fix(lvs): keep bundled DCGM exporter opt-in

hugoverjus:
## Summary
- disable the LVS Helm monitoring subchart and bundled DCGM exporter by default
- avoid duplicate DCGM DaemonSets on clusters where the NVIDIA GPU Operator already provides one
- document the explicit Helm flags for clusters that need the bundled exporter

## Test plan
- [x] Parse `values.yaml` and `values-lvs.yaml` and verify both monitoring defaults are disabled
- [x] Render the LVS chart and verify the default omits DCGM exporter resources
- [x] Render with `monitoring.enabled=true` and `monitoring.dcgmExporter.enabled=true` and verify DCGM resources are present
- [x] `git diff --check`

### DIFF
--- a/deploy/helm/developer-profiles/dev-profile-lvs/Chart.yaml
+++ b/deploy/helm/developer-profiles/dev-profile-lvs/Chart.yaml
@@ -48,11 +48,9 @@ dependencies:
     version: 3.3.0
     repository: "file://../../services/rtvi"
     condition: rtvi.enabled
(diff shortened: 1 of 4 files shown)


zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32852766396)

PR #1854, opened 2026-08-25
refactor(deploy): make Nemotron 3.5 Lightning the only LLM

zac-wang-nv:
## What

Reduces the bundled local LLM catalogue to **`nvidia/nemotron-3.5-lightning-30b-a3b`**, the default on every hardware profile — including DGX Spark.

**Removed:** `nvidia/nvidia-nemotron-nano-9b-v2`, `nvidia/nemotron-3-nano`, `nvidia/llama-3.3-nemotron-super-49b-v1.5`, `openai/gpt-oss-20b` (4 Docker service dirs, 2 Helm subcharts).

**Helm now ships Lightning only.** Docker keeps one other entry — `nvidia/NVIDIA-Nemotron-Nano-9B-v2-FP8` — because **AGX Thor and IGX Thor have no Lightning sizing files**; `dev-profile.sh` selects it there and nowhere else. It is also the documented escape hatch for a Spark fallback.

Removed model ids stay in both deploy scripts as arms that **fail loudly** with a migration message rather than resolving to an empty slug.

Deploying *without* an LLM is legitimate and already has two explicit spellings — `LLM_MODE=remote` and `LLM_MODE=none` (warehouse 3d/mv3dt, `bp_wh_kafka`, `bp_wh_redis`, `bp_wh_auto_calib`), both of which set `LLM_NAME_SLUG=none`. The vss-build-vision-ai composition rules rely on that when they prune an LLM no requested capability reaches. An empty slug is the *unintended* version of the same shape: `--llm <removed-id>` in local mode still writes `LLM_NAME` and `LLM_MODE=local`, but `llm_local_` matches no service, so Compose exits 0 with the LLM absent and the agent pointed at a local endpoint that never comes up.

`--use-remote-llm` still accepts arbitrary ids.

## One engine profile everywhere

All **12** Docker `hw-*.env` files and all **6** Helm `gpuProfiles` pin `vllm-int4-tp1-pp1-32.0`:

| gpuType | profile | why not NVFP4 |
|---|---|---|
| H100, GB300, L40S, RTXPRO4500BW, RTXPRO6000BW, OTHER, **DGX-SPARK** | `vllm-int4-tp1-pp1-32.0` | INT4 is the only tp1 profile with no architecture restriction |

Pinned rather than auto-selected because the container advertises the ≥66 GB BF16 profile as "runnable" on cards that cannot hold it — reproduced on a 48 GB L40S.

**This is what lets Spark run the default.** The NIM's **arm64** image manifest carries the same profile id with only `min_vram_per_device_gb: 32.0` and *no* GPU allow-list:

```yaml
id: 2ef85c72…
tags: {llm_engine: vllm, precision: int4, tp: '1', pp: '1', min_vram_per_device_gb: '32.0'}
```

The profile hash identifies model+quantization+topology, not the GPU — the same id appears on L40S, RTX PRO 6000 and in the arm64 manifest — so these pins are portable.

## Reasoning pinned off

The chat template defaults `enable_thinking` to **true**, so "off" must be stated:

```jinja
{%- if enable_thinking %}{{- '<|im_start|>assistant\n<think>\n' }}
{%- else %}{{- '<|im_start|>assistant\n<think></think>' }}{%- endif %}
```

- **Client:** `llm_reasoning: false` in all 8 agent workflow configs (`dev-profile-alerts` had no key and was inheriting the code default).
- **Serving:** `--default-chat-template-kwargs {\"enable_thinking\":false}` in every Lightning `hw-*.env` and Helm `gpuProfile` — the only lever for LVS, whose config exposes no `extra_body`.

The `hw-*.env` value is **unquoted on purpose**: Compose unescapes `\"` inside a quoted `env_file` value and the entrypoint shell then strips the real quotes, yielding invalid JSON and a startup failure.

`reasoning_utils.py` matched Lightning only because `"nemotron-3"` is a substring of `"nemotron-3.5-lightning"` — correct by accident, and it would have silently stopped binding the kwarg on any rename. Replaced with an explicit marker tuple plus tests.

## Measured, not assumed

Sizing was measured on an RTX PRO 6000 rather than inherited. `NIM_GPU_MEM_FRACTION=0.5` was never chosen for this model — commit `dcc709dcc` copied `nemotron-3-nano`'s shared env verbatim. Co-resident with a live CR3 nano (38836 MiB):

| fraction | Lightning | free | 8 concurrent tool calls |
|---|---|---|---|
| 0.5 *(previously shipped)* | 50184 MiB | **1682 MiB** | — |
| 0.4 | 38982 MiB | 12884 MiB | 8/8, 4.3 s |
| **0.3 (shipped)** | 29970 MiB | **21896 MiB** | 8/8, 1.9 s |

`0.3` is comfortable because Lightning is a hybrid **Mamba-2 + MoE**: only 6 of 52 layers are attention, and those use an FP8 KV cache. `--max-num-seqs` raised 2 → 8.

Serving requirements confirmed live: without `NIM_PASSTHROUGH_ARGS`, `tool_choice: auto` is a hard 400 (`"auto" tool choice requires --enable-auto-tool-choice and --tool-call-parser`). With them, tool calls succeed in every thinking state — so pinning reasoning off does **not** break tool calling.

## Context length: 32000 → 65536 on the shared profiles

The `32000` pin was inherited too. Reading `vllm:cache_config_info` off the live server across 10 boots on an idle second GPU, at `NIM_GPU_MEM_FRACTION=0.25` (the same byte budget as H100 `local_shared` at 0.3):

| profile | `max_model_len` | KV pages |
|---|---|---|
| NVFP4 | 32000 / 131072 / 262144 | 330 / 330 / 272 |
| INT4 | 131072 | 330 |
| **INT4** | **65536 ×4, identical config** | **272, 330, 330, 272** |
| both @ 0.377 | 262144 / 65536 | 1339 / 1339 |

Two results. **INT4 and NVFP4 allocate the same pool** — 272, 330 and 1339 appear under each, within 40 MiB — so quantization is not the sizing variable, which is what the earlier revision of this PR assumed. And the pool varies between boots: four identical INT4 boots gave 272, 330, 330, 272. vLLM derives it from runtime profiling whose inputs weren't captured, so no mechanism is claimed and 272 is the lowest of 8 boots, not a proven bound. Neither observation is documented in-tree — they never change a deploy action; the full record is in the vault.

vLLM 0.25.1 pages this model at **4176 tokens**, which follows from the architecture: one Mamba layer holds `64×64×128×4 B` fp32 SSM + `6144×3×2 B` bf16 conv = 2,134,016 B; one attention layer costs `2 kv heads × 128 dim × 2 × 1 B` = 512 B/token; `2134016 / 512 = 4168`, aligned up to a multiple of 16. A sequence costs **`ceil(len/4176) + 4`** pages, the `+4` being the constant Mamba state — which reproduces every `kv_cache_max_concurrency` the server reported, under both quantizations.

| `NIM_MAX_MODEL_LEN` | pages/seq | 8 concurrent need | vs 272 |
|---|---:|---:|---|
| 32000 | 12 | 96 | fits |
| **65536 (shipped)** | 20 | **160** | fits, one 58-page step spare |
| 98304 | 28 | 224 | fits, under one step |
| 131072 | 36 | 288 | does not fit |

The exact ceiling at 272 pages is **125,280**, so 128k misses by 4.6%. This is a **policy choice**, not arithmetic — 128k with `--max-num-seqs 7` (252 pages), or the full 262144 with 4 (268), are equally consistent with the data and trade concurrency for context instead.

Six files change: the five `-shared` sizing files plus `hw-DGX-SPARK.env`, which is dedicated but still needs a cap because Spark's memory is unified with the host. The other six dedicated profiles pin nothing on purpose — at vLLM 0.25.1's default `gpu_memory_utilization` of 0.92 the pool already holds the full 262,144 context (a run at 0.377 shows 1339 pages, full context at 20 concurrent).

## Also fixes three pre-existing defects

Making Lightning the default would have propagated all three:

1. **Edge failed `docker compose config` in all six profile combinations** — the default LLM shipped no `hw-<edge>.env` and nothing routed edge to a model that did.
2. **`blueprint-deploy.sh` had no `--llm` validation at all**, and the check added here runs in `process_args` — *before* `state_down` — so a typo can no longer tear down a healthy deployment before erroring.
3. **The Helm resource name contained dots**, illegal in a Kubernetes Service name, so the chart could not deploy. Renamed to `nemotron-35-lightning-30b-a3b`; the Docker slug keeps its dots.

`dev-profile.sh` also fails fast when an explicit `--llm` has no sizing for the selected hardware, instead of dying later on a missing `env_file`.

## Documentation is split out

The `.mdx` pages, the `default_models` / `llm_params` tables and the `README` files were moved to **#1943**, so this PR carries deploy config, scripts, tests and skills only.

`skills/**` stays here deliberately: its evals are agentic — they run an agent against the skill references and grade the answer — so an eval asserting Lightning-on-DGX-Spark only passes when the routing change is in the same tree. `skills-eval`, `skills-review` and `skills-signatures` all run per-PR, so splitting the halves would fail whichever PR lacked the other.

## Verification

```
golden fixture     OK: 69 resolved image refs match
unit tests         139 passed
helm               base / alerts / search / lvs all render
dev-profile suite  221 passed / 6 failed — zero regressions by name vs develop (216 / 6)
```

## Known gaps

- **The Spark sizing is not measured.** `0.3` of 128 GB unified ≈ 38 GB against a 32 GB floor, carried over from the shared profile rather than independently chosen — no Spark hardware was available. Unverified: whether vLLM's INT4 kernels are built for GB10's SM in this image, and how `NIM_GPU_MEM_FRACTION` behaves against *unified* rather than dedicated memory. `edge.md` states this and points at the FP8 or remote fallback.
- **All KV measurements come from one RTX PRO 6000.** They are applied to H100, GB300, OTHER and Spark by equal-byte-budget equivalence, not by measurement on each. 272 pages is the lowest of 8 boots, not a proven lower bound.
- **AGX/IGX Thor still have no Lightning sizing** and remain on the FP8 build.
- **RTXPRO4500BW `local_shared`** is accepted by the CLI but has no `-shared` env, so it fails at compose. Pre-existing; a guard was tried and reverted because it broke five tests that encode the current behaviour.
- **License pages** cover the Lightning NIM but not the FP8 path's raw-vLLM and `alpine/curl` containers — a legal-review question, not an edit.


zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33029809085)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33037112846)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33038218320)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33048810860)

deploy/docker/industry-profiles/warehouse-operations/overrides.env
```diff
@@ -65,19 +65,15 @@ LLM_MODE=local
...
-# - nvidia/nvidia-nemotron-nano-9b-v2 -> nvidia-nemotron-nano-9b-v2
-# - nvidia/NVIDIA-Nemotron-Nano-9B-v2-FP8 -> nvidia-nemotron-nano-9b-v2-fp8
-# - nvidia/nemotron-3-nano -> nemotron-3-nano
 # - nvidia/nemotron-3.5-lightning-30b-a3b -> nemotron-3.5-lightning-30b-a3b
-# - nvidia/llama-3.3-nemotron-super-49b-v1.5 -> llama-3.3-nemotron-super-49b-v1.5
-# - openai/gpt-oss-20b -> gpt-oss-20b
+# - nvidia/NVIDIA-Nemotron-Nano-9B-v2-FP8 -> nvidia-nemotron-nano-9b-v2-fp8 (edge only)
 # Select the local LLM model name, or the model id exposed by a remote /v1/models endpoint.
-# NOTE (remote LLM): nvidia/nvidia-nemotron-nano-9b-v2 may no longer be served by build.nvidia.com
+# NOTE (remote LLM): nvidia/nemotron-3.5-lightning-30b-a3b may no longer be served by build.nvidia.com
```
yunHe95: delete this comment

jiayin-nvidia (dismissed):

skills/vss-deploy-profile/references/edge.md
```diff
@@ -2,31 +2,41 @@
 
 Base-profile deployment guidance for edge platforms.
 
-On **DGX Spark**, use **NVIDIA-Nemotron-Nano-9B-v2-DGX-Spark** as the
-LLM. This is a DGX Spark-only NIM container:
+**DGX Spark** now runs the blueprint default,
+**`nvidia/nemotron-3.5-lightning-30b-a3b`** — it ships `hw-DGX-SPARK` and
```
hfarooq-nv, line 6: this is contradicting the other changes in this file - here we seem to say to use 3.5 lightning on spark but lower down we say to use nano 9b v2 for all edge platforms 

if we are supposed to use nano 9b on spark then evals.json would also need to be fixed

if we are supposed to use 3.5 lightning on spark then vss-deploy-profile/SKILL.md would need to be updated since that specifies 9b v2 for all edge platforms

deploy/helm/developer-profiles/dev-profile-search/values-build-endpoint.yaml
```diff
@@ -59,7 +59,7 @@ global:
 # Disable local NIM deployments
 nims:
   enabled: false
-  nemotron:
+  nemotron35:
     enabled: false
   nemotron35:
```
liamy-nv, line 62: duplicate should be removed

liamy-nv (dismissed):

skills/vss-deploy-profile/references/edge.md
zac-wang-nv: yes, we will consolidate to use 3.5 lightening on spark

deploy/helm/developer-profiles/dev-profile-search/values-build-endpoint.yaml
zac-wang-nv: fixed

deploy/docker/industry-profiles/warehouse-operations/overrides.env
zac-wang-nv: fixed

kaushikc-nvidia:
Please also remove nvidia-nemotron-nano-9b-v2-fp8 and it's references unless it is used

kaushikc-nvidia (approved):

PR #1822, opened 2026-08-24
rename: vss-build-vision-agent skill to vss-build-vision-ai

ary1111:
Renames the skill directory, its eval adapter, and the associated test script, and updates all in-content references (SKILL.md frontmatter, references, eval fixtures, and cross-skill links from vss-ask-video, vss-deploy-detection-tracking-2d, vss-generate-video-report, vss-manage-alerts, vss-manage-video-io-storage, vss-query-analytics, vss-search-archive, vss-setup-behavior-analytics, and vss-summarize-video).

## Checklist
- [ ] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/HEAD/CONTRIBUTING.md).
- [ ] I have installed and run [pre-commit hooks](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/HEAD/CONTRIBUTING.md#1-enable-pre-commit-hooks) locally (`uv run pre-commit install` once, then hooks run on every `git commit`).
- [ ] Every commit on this PR is [DCO sign-off](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/HEAD/CONTRIBUTING.md#developer-certificate-of-origin-dco)'d (`git commit -s` adds a `Signed-off-by` trailer that certifies you have the right to submit the change under Apache-2.0).
- [ ] New or existing tests cover these changes.
- [ ] The documentation is up to date with these changes.
- [ ] No secrets, API keys, or credentials committed.

PR #1744, opened 2026-08-18
Update IN-2 IN-3 evals to handle haproxy

alexzh-nvda:
## Checklist
- [ ] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/HEAD/CONTRIBUTING.md).
- [ ] I have installed and run [pre-commit hooks](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/HEAD/CONTRIBUTING.md#1-enable-pre-commit-hooks) locally (`uv run pre-commit install` once, then hooks run on every `git commit`).
- [ ] Every commit on this PR is [DCO sign-off](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/HEAD/CONTRIBUTING.md#developer-certificate-of-origin-dco)'d (`git commit -s` adds a `Signed-off-by` trailer that certifies you have the right to submit the change under Apache-2.0).
- [ ] New or existing tests cover these changes.
- [ ] The documentation is up to date with these changes.
- [ ] No secrets, API keys, or credentials committed.

PR #1534, opened 2026-08-04
Feat/build vision agent skill guide

ary1111:
## Description
Provided walk-through and updated description for build-vision-agent Skill

## Checklist
- [ ] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/HEAD/CONTRIBUTING.md).
- [ ] I have installed and run [pre-commit hooks](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/HEAD/CONTRIBUTING.md#1-enable-pre-commit-hooks) locally (`uv run pre-commit install` once, then hooks run on every `git commit`).
- [ ] Every commit on this PR is [DCO sign-off](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/HEAD/CONTRIBUTING.md#developer-certificate-of-origin-dco)'d (`git commit -s` adds a `Signed-off-by` trailer that certifies you have the right to submit the change under Apache-2.0).
- [ ] New or existing tests cover these changes.
- [ ] The documentation is up to date with these changes.

PR #1840, opened 2026-08-25
[VIA-2535] Add warehouse industry Foundation to vss-build-vision-agent

yunHe95:
Warehouse (MODE=2d RT-DETR, MODE=3d Sparse4D) becomes a selectable Foundation alongside the base/alerts/lvs/search developer profiles, reusing the existing _builds/<name>/ artifact contract unchanged.

The warehouse env layers map 1:1 onto the developer-profile pair and its COMPOSE_PROFILES_WH_* lists are flat self-named service keys, so the only structural change is a FOUNDATION_DIR branch in the resolve step.

New:
- references/profiles/warehouse.md - Foundation card: the nine in-scope variants, knobs, hard constraints, readiness
- references/services/{configurator,monitoring}.md - owner contracts for services no developer profile deploys
- scripts/validate_warehouse_env.py - env-level constraints that resolved.yml structurally cannot express; MODE, BP_PROFILE, HARDWARE_PROFILE and SAMPLE_VIDEO_DATASET appear in no service environment block, and each rule fails at bring-up or silently at runtime rather than at `config` time
- scripts/render_warehouse_configurator_env.py - bp-configurator-<mode> loads env_file directly and never sees --env-file layering, so without this it bakes the checked-in HOST_IP='<HOST_IP>' sentinel into the container that renders every stream and hardware config
- eval/profile_wh_{1,2}_*.json

Changed: SKILL.md routing and Q2w intake; composition.md Foundation resolution, artifact contract and transitive dependent-value closure; warehouse rows in the rt-cv, vios, elk and behavior-analytics contracts.

validate_resolved_yml.py gains one line: industry-profiles now counts as a checked-in bind source, so a missing warehouse calibration.json fails validation instead of silently becoming a Docker-created directory with perception emitting nothing.

Scope is MODE=2d and MODE=3d with BP_PROFILE in bp_wh, bp_wh_kafka, bp_wh_redis. Other warehouse service lists route to vss-deploy-profile, which remains authoritative for host setup, deploy procedure and runtime diagnosis.

## Checklist
- [ ] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/HEAD/CONTRIBUTING.md).
[...]

PR #1863, opened 2026-08-26
fix(skills): retain the HAProxy ingress and drive VIOS through the vss CLI

zac-wang-nv:
## Summary

`SKILL.md` has listed the HAProxy ingress under **"Always included — do not offer as choices"** since #1534 (2026-08-05), but `references/services/ingress.md` and `references/profiles/base.md` still told the agent to prune it, and the IN-1/IN-2/IN-3 eval specs still excluded `vss-haproxy-ingress` from the expected `COMPOSE_PROFILES`. The specs graded the opposite of what the skill instructs, so a correct build failed the profile check. This aligns the specs and those docs with `SKILL.md`, and moves the runtime specs' VIOS media operations onto the `vss` CLI.

The two are related: per [`AGENTS.md`](../blob/develop/AGENTS.md), after a one-time `vss configure --base-url <origin>` the operational commands take no service endpoint and reach every backend through that origin, so a build that prunes the ingress is unqueryable from the host CLI.

Supersedes #1744, which made the IN-2/IN-3 half of this change on the pre-rename paths (`skills/vss-build-vision-agent/`, renamed by #1822) and chose the curated-patch mode this PR deliberately does not.

**This branch is stacked on #1878 and #1881, so its diff is wider than the description above.** Of the 29
changed files, `.github/skill-eval/run_leg.py` and its tests belong to #1878; #1881 contributes
`deploy/docker/services/infra/sdrc/wdm-env-from-config.sh`, five `developer-profiles/*/overrides.env`,
`test-dev-profile.sh`, and hunks in **eleven files under `skills/vss-build-vision-ai/`** — so the overlap is
inside this PR's own files, not merely alongside them. They are merged rather than cherry-picked
(`merge: base on #1878 so this PR's legs schedule correctly`): without #1878's scheduler fix these legs do
not get a box, and without #1881 they trip over a profile token nothing defines. Both drop out of this diff
once they merge. Review those changes on their own PRs.

## Plan

Questions put to the reviewer before building, with the recommendation and the answer given:

| # | Question | Recommended | Answer |
|---|---|---|---|
| 1 | How far should the HAProxy fix reach — `in-3` only, all three specs that prune it, or specs + the skill reference docs? | Specs + skill docs — otherwise the docs the agent reads keep contradicting the spec that grades it | **Specs + skill docs** |
| 2 | Which evals move VIOS operations onto the `vss` CLI — `in-3` only, all runtime specs, or also the skill docs? | All runtime specs | **All runtime specs** (`deployment.md` was added later — see *Doc fixes surfaced by review*) |
| 3 | With the ingress retained, does the build ship the stock template or a curated patch? | Initially recommended stock/no-patch, then re-asked after finding the agent emits the patch unprompted | **Stock template.** Reviewer's rule: *"build shouldn't trim the route table, `vss configure` handles service discovery at runtime, not the build process"* |
| 4 | Relationship to open PR #1744, which already does the IN-2/IN-3 half? | Supersede it on the post-rename paths and comment there | **Supersede** |
| 5 | `combined` and `search-standalone` kept the curated patch on an "explicit request" reading — why? | Withdrawn: their request says *expose* browse/operate through one origin, not *restrict* the origin to them, and the stock template already carries every route they need | **Delete `search-standalone` entirely** (overlaps `combined`); `combined` flipped to the stock template |

Codex reviewed the diff independently and caught one thing this PR then fixed — see *Review* below.

## Changes

**HAProxy ingress is always included, with the stock config**

- `in-1`, `in-2`, `in-3`: `vss-haproxy-ingress` added to each expected `COMPOSE_PROFILES` and removed from the exclusion lists. Each `patches/ is absent` check now states *why* it stays true — the ingress runs the stock `haproxy.cfg.template`.
- Each delta check now requires a real `HOST_IP`: the Foundations ship the `<HOST_IP>` sentinel (`dev-profile-{base,search}/overrides.env:91`) and the host-header allowlist 404s every request without it. `EXTERNAL_IP`, `VSS_PUBLIC_HOST`, `VSS_PUBLIC_PORT` and `HAPROXY_HOST_PORT` may be pinned alongside it — and `EXTERNAL_IP` and `VSS_PUBLIC_HOST` genuinely need it, because the Foundation's `overrides.env` interpolates them from `HOST_IP` at its own layer against the sentinel. An earlier revision of this PR forbade re-declaring all four; IN-2's eval caught that and it was corrected.
- `combined`: the curated ingress patch is gone — the stock template already carries `/kibana`, `/vst`, `/storage`, `/video-analytics-api`, `/alert-bridge`, `/elasticsearch`, `/rtvi-embed`, `/rtvi-cv`, so the patch only removed routes. Its other two patches (behavior-analytics config, merged Kibana bundle) are unchanged.
- `ingress.md`, `base.md`: "reached only when the request asks… otherwise pruned" → always included, stock by default. The paragraph asserting *"an ingestion/indexing-only build prunes the proxy regardless"* is gone. The curated path and its reference config survive as an explicit-request escape hatch, now unexercised by any spec.

**VIOS media operations go through the `vss` CLI**

- `in-1`, `in-2`, `in-3`, `an-1-runtime`, `at-1`, `combined`: where a spec performs VIOS media operations, they go through `vss vios add|list|clip|snapshot` rather than raw REST. The specs differ in how much they exercise — `at-1` only configures and lists. In the four that register media, the sensor name or id used later must appear verbatim in the output of a successful `vss vios add` or `vss vios list`, rather than being inferred from a filename or lifted from a raw REST response.
- Each runtime spec gained a bootstrap check requiring `uv run --project <repo>/services/agent --no-dev --extra cli vss` and rejecting a globally installed binary, `docker exec` or `kubectl exec`, plus `vss configure --base-url …:7777`. `vss configure check` is explicitly **optional** — nothing asks the agent to run it, and `vss vios list` already proves the VIOS path.
- Kept on `curl`, per `vss_cli/AGENTS.md` § *When the CLI does not cover it*: the VIOS live-proxy URL lookup (the proxy surface is explicitly carved out), NvStreamer's own API (never fronted by the ingress), and service readiness probes. `an-1`'s summarization moved onto the CLI too: `vss summarize run` is a thin client over the same `POST /v1/summarize`, reads the LVS origin from what `vss configure` recorded, and defaults `--model` to what the deployment reports serving, so the old REST exemption only bought a hand-built payload and a guessed endpoint. `vss_cli/AGENTS.md` documented `--video-uri`, `--prompt` and `--timeout` for it, none of which exist; it now shows the real grammar.

**Doc fixes surfaced by review**

- `deployment.md` and `deployment_resolution.md` bootstrapped the CLI as `uv run … --no-dev vss`, omitting `--extra cli`. Per `AGENTS.md` that installs no `vss` at all, so the documented command could not have worked.
- `deployment.md`'s write path told the agent to register a VIOS source by direct REST. Registration now uses `vss vios add`; the consumer fan-out and live-proxy lookup stay REST.

**Deleted**

- `profile_search_standalone_harbor.json`. Nothing referenced it — `plan_matrix.py` discovers specs by globbing `eval/*.json`, so the leg drops cleanly. Coverage note: search-without-alerts is no longer exercised anywhere.

## What CI changed

**Status.** Last completed matrix: `606036ed`. Current head `086c18e7` has not finished a rerun.
On `606036ed`, **one of eight legs passed completely** (`in-2`, reward 1.0 on all three steps); seven were
red. Every explicit *build* step passed; `combined` still failed its **deploy** step, 3/4, because
`alert-bridge` returned 503. Per-leg: `an-1-runtime` 5/5 · 6/6 · 7/8 (4 summarize runs, one required);
`combined` 19/19 · 3/4; `in-3` 13/13 · 5/5 · 6/7 (raw REST where a `vss vios` verb exists);
`in-1` 12/12 · 13/13 · 6/6 · 4/6 (no `--type stream` registration); `at-1` 12/13 (no verdict documents);
`sop-1` 11/13 (DS-SOP emitted empty responses); `an-1-api` 9/10 (bind sources, fixed at head).

Results are nondeterministic across identical inputs — `in-3` scored 7/7 then 6/7 on an unchanged spec, and
`in-2` and `in-3` have traded which one obeys the CLI rule.

Harbor runs exposed eight evaluation-spec issues, corrected here:

| Fix | What the run showed |
|---|---|
| Sensor-name provenance (4 specs) | A judge demanded the sensor name be "read back from `vss vios list`" after the agent used the name `vss vios add` returned and got a resolvable `media_url`. Provenance now comes from either successful command. |
| `combined` step 2 service keys | Step 1 scored 19/19, then step 2 rejected the Search-side behavior-analytics, analytics-API and NvStreamer containers step 1 *requires* — "second detector/behavior-analytics/…" was read as applying "second" only to "detector". Now names both key sets and grades the Compose **service key**, since several share a `container_name`. |
| `do not deploy` was ungraded (in-2, in-3) | `in-3` passed step 1 at 12/12 while deploying during it, then failed step 2 for not deploying. The harness mount probe shows a clean box at `step-1:before-sync` and four `vss-vios-*` containers at `step-2:before-sync`. |
| `--env-file /dev/null` accepted (4 specs) | Two legs deployed with `--env-file /dev/null` and were failed for it. An explicit `--env-file` replaces Compose's default source, so `/dev/null` also suppresses the `.env` a bare `up` would auto-load: it proves independence from that default file, though not from inherited shell variables. Only `/dev/null` is accepted, since the transcript shows the flag but not a file's contents; `COMPOSE_ENV_FILES` fails too. |
| Summarization through `vss summarize run` | The runtime leg drove every VIOS operation through the CLI while still POSTing `/v1/summarize` by hand. |
| `FOUNDATION=` contradiction (an-1-runtime) | Step 1's check 2 failed a build for declaring `FOUNDATION=lvs` as an "invented build-label key" while check 1 of the same step requires that line and `composition.md` lists it first. The criterion is now the effective `COMPOSE_PROFILES` token list. |
| `an-1-api` bind sources | Failed for `config.yml`, SDRC `.wdm-env` and `log` being absent — none is checked in, and the step stops without deploying. `skills/vss-build-vision-ai/scripts/validate_resolved_yml.py` already accepts a missing source when a sibling `.tmpl` exists, carries `.wdm-env` in `GENERATED_BIND_NAMES`, and does not treat writable non-profile binds as checked-in sources; the check now defers to it. |
| `in-1` step 4 report | Demanded the runtime step recite step 1's build delta ("six added and four removed service keys"). Now matched to the evidence that step actually asks for. |

## Review

Every change here was put through an independent review by the OpenAI Codex CLI, told up front that the
diff was Claude-generated. It is cited because it changed the outcome, not as endorsement: it rejected an
`elk.md` edit outright as scope creep, caught that a first attempt at the sensor-name rule contradicted
itself and had loosened the anti-bypass exception, replaced an end-state deployment test that
`docker compose create` would have slipped past, and found the two stale claims about host-identity pinning
and `vss configure check` corrected above.

Verified directly against the repo rather than the docs: every `vss vios` verb and flag named in a check
(`list`, `timeline`, `clip`, `snapshot`, `add`, `delete`; `--type video|stream`, `--sensor`, `--name`,
`--at`, `--start-time`/`--end-time`; `media_url` on `clip`/`snapshot`) against `vios_group.py`, and every
spec parses.

## Known follow-ups (not in this PR)

- #1491 would publish NvStreamer at `:7777/nvstreamer` for all profiles, which contradicts `ingress.md`'s "NvStreamer is never fronted here" and would change the `:31000` probes in `in-2`/`in-3`.
- No spec now exercises the curated-ingress path; if it should stay supported, it wants a spec that explicitly asks to restrict the origin.
- **The skill adds Kibana to a Foundation that does not ship it, in two of four unchanged-spec runs.**
  `in-1` step 1 failed on this in iterations 19 and 21 and passed in 20 and 23
  (`kibana-init-container-lvs` once, `kibana-init-container-alerts` the other time). Both keys belong to
  other Foundations' profile sets and should not be selected by a `base`-derived build. `elk.md` already
  ends its Kibana bullet with "Do not add `kibana` to a Foundation that does not ship it", so no
  documentation change is included here; track mitigation separately if the rate warrants it. It costs the
  whole leg, because a failing step marks steps 2-4 `not-run`.
- **`at-1` cannot pass as written: nothing ever stimulates its pipeline.** It has failed identically in all
  six iterations — 12/13, always check 10, "no document with a verdict field anywhere", latest adding
  "Kafka alert topics contain 0 messages". Its other twelve checks pass every time. The query asks only to
  build, deploy, configure and list; no source is registered and no media staged. NvStreamer serves files
  already present in `${VSS_DATA_DIR}/videos/<profile>`, which a clean deploy leaves empty, and the shipped
  behavior-analytics config has `"sensors": []` with `sourceType = kafka`, so it consumes detections rather
  than opening a source. Six zero-message runs prove the deployment was never stimulated; they prove
  nothing about whether a stimulated pipeline works. Two decisions are needed, neither of them a wording
  change:
  1. **A provisioned fixture.** The incident rule is a FOV *count* violation:
     `fovCountViolationIncidentObjectThreshold=1` is the maximum allowed count, so firing it needs **more
     than one** person detected simultaneously and sustained past
     `fovCountViolationIncidentExpirationWindow`. "Person-containing" footage is not sufficient — a
     single-person clip correctly produces no incident forever. The fixture must be staged by harness setup
     and its concrete path given to the agent — and staged *after* the purge, because
     `_purge_host_data_dirs` (`.github/skill-eval/envs/brev_env.py:652`) deletes the contents of
     `nvstreamer/videos`, `nvstreamer/videos-upload`, `nvstreamer/vst_data` and `videos/nvstreamer`
     before the first trial of each spec and fails loudly if anything survives. The empty videos
     directory is deliberate cross-trial hygiene, not a gap. `in-2` and `in-3` work around this by
     having the agent source a local MP4 from elsewhere on the host; `at-1` never asks it to.
  2. **The Kafka clause.** Check 10 requires the verdict in Elasticsearch *and* Kafka, but
     `dev-profile-alerts/vlm-as-verifier/configs/config.yml:355-363` sets `vlm_enhanced_sink.incident` and
     `.alert` to `type: "elastic"` and leaves the Kafka variant commented out at 364-372. Either align the
     request's "Store and publish verified incidents" wording with the Elasticsearch-only contract, or
     enable a Kafka sink as a deliberate delta — uncommenting is not obviously safe, since the block reuses
     the singular `vlm_enhanced_sink` key and may replace Elasticsearch rather than supplement it.

  Ruled out along the way: a `person`/`Person` label-casing mismatch. The behavior-analytics filter and
  `rtdetr-960x544-labels.txt:4` are both lowercase `person`; the capitalised `Person` appears only in
  `coco_classmap.txt`, which nothing under `deploy/docker/` references.

- **`alert-bridge` returns 503 until its process restarts.** `combined` step 2 fails its readiness probe
  with `{"status": "not_ready", "message": "alert-config store initialisation failed: …"}`.
  `startup_event()` (`services/alert/src/web/main.py:144-155`) makes one store-initialisation attempt and
  is the only code that updates `_startup_ready`, so a failed attempt leaves `/health` at 503 until the
  process restarts, even if a later API request constructs the store successfully.
  `build_alert_config_store` raises either when persistence is disabled outside a dev profile or when
  Elasticsearch is unreachable, and the shipped `services/alert/config.yaml:362` enables persistence, which
  makes the disabled-persistence gate unlikely if that configuration was loaded. The artifact establishes
  only that alert-config initialisation failed; container logs are needed to say whether connectivity,
  health checking, hydration or configuration caused it.
- **A judge that runs out of turns costs a check with no retry.** `combined` lost one check to
  `is_error=True subtype='error_max_turns' … retry_attempted=False` after the judge produced 5,812
  characters across 30 blocks without emitting the verdict JSON. `generic_judge.py:306` already calls this
  "a false `error_max_turns` verdict instead of a real pass/fail", yet the retry at line 364 is gated on
  `not result_is_error`, which excludes exactly this case.
- Three legs fail on service-side pipelines that produced nothing: `at-1`'s verdict documents, `in-3`'s
  Kafka offsets, and — intermittently — `an-1-runtime`'s empty `video_summary`, which returned populated on
  the very next iteration. LVS carries a guard for empty CA-RAG aggregation results
  (`DEFAULT_AGGREGATION_EMPTY_RETRIES = 2`, so three attempts); the empty result is consistent with that
  failure mode but these artifacts do not establish it caused this run. All three sit in services this
  branch cannot rebuild.

## Checklist
- [x] New or existing tests cover these changes — these *are* the tests; the skills-eval gate runs them.


zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32982522268)

zac-wang-nv:
## Second-opinion review → `bb236ced`

Ran the diff past the OpenAI Codex CLI (told up front it was Claude-generated). It found a genuine **P0 that I introduced**, plus four smaller factual errors. All fixed in `bb236ced`; evidence for each below.

**P0 — `combined` step-2 check described the curated config I had just removed.** Flipping the build to the stock template left c1.3 asserting behaviour only a curated `haproxy.cfg` produces:
- it expects `/` → `302 /kibana/`, but the stock catch-all is `use_backend bk_vss_ui if h_main` (`haproxy.cfg.template:273`) — with the UI tier pruned that is a **503**;
- it expects `/rtvi-vlm/v1/models` → `404` and states "RT-VLM is not fronted", but the stock template *does* front it (`use_backend bk_rtvi_vlm_strip`, ~line 223) and this build runs `rtvi-vlm`, so it answers present;
- `/api/v1/videos` returns **503**, not `404` — which `vss configure` records as absent just the same.

A correct no-patch build would have failed that check. Rewritten to describe the stock template, and the same false RT-VLM claim is corrected in `ingress.md`.

**`ingress.md` contradicted its own rule.** It said curation needs an explicit restriction request, then named a "single ingress" request as one — which is `combined`'s literal request text. Curation now keys on a request to *restrict* the origin; asking for surfaces reachable *on* one origin is what the stock template already delivers.

**"`search` needs the Agent tier" was invented.** `search_group.py:356,362` require `elasticsearch` + (`rt_embed` | `rtvi_cv`) — no Agent. IN-3 deploys all three, so `search` may legitimately report available; IN-1's is unavailable for the real reason (no RT-Embed, no RT-CV).

**The HOST_IP rationale overstated the failure.** I wrote that the host-header allowlist "404s every request" without a real value. It does not — `localhost` and `127.0.0.1` are unconditionally allowed (`haproxy.cfg.template:129-132`), which is how these very checks connect. The sentinel still must go, because `resolved.yml` validation rejects it and no non-loopback client can match it.

**`SKILL.md` had both problems the reference docs had** — a bootstrap missing `--extra cli` (installs no `vss` at all) at step 9, and provisioning routed to raw REST at line 40. The Skills Review bot flagged the same `--extra cli` line independently. Fixing only `references/deployment.md` left the file the agent starts from wrong.

---

### Two pre-existing issues I am *not* fixing here

1. **`an-1-runtime` hardcodes the wrong LLM.** Its readiness check expects container `nvidia-nemotron-nano-9b-v2`, but the LVS Foundation sets `LLM_NAME_SLUG=nemotron-3.5-lightning-30b-a3b` (`dev-profile-lvs/overrides.env:55`). Surfaced by the Skills Review bot and corroborated. Untouched by this PR, and fixing it changes what that eval grades — worth its own change.
2. **`OSRB Review` fails closed repo-wide.** Identical `INCONCLUSIVE — required evidence collection had errors` on #1850 and #1840; it evaluated **0** dependency changes here, and License Diff passed independently.

Also unresolved and pre-existing: `combined` c0.5 vs c1.3 disagree on whether the `/elasticsearch` guard is read-only or a denylist. c1.3 is right (it permits top-level `POST` writes such as `_bulk`); I corrected the wording where this PR touched it.

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32983499194)

zac-wang-nv:
## Skills Eval on `bb236ced` — triage → `de05eea7`

Eight legs ran. Two passed clean, two never executed for infrastructure reasons, and four produced real signal. Every finding is either fixed below or attributed.

| Leg | Result | Attribution |
|---|---|---|
| `sop-1` | ✅ **13/13** | Untouched spec, full 29-min runtime leg — deploys a Foundation carrying `vss-haproxy-ingress` |
| `an-1-api` | ✅ **10/10** | Untouched spec |
| `in-2` | 10/11 | Ingress change **validated**; sole failure was mine |
| `in-1` | 10/12 | Two pre-existing false checks |
| `at-1` | 11/13 | One judge flake, one gap of mine |
| `combined` | 15/19 | Agent curated the ingress and kept the Agent tier |
| `in-3` | ❌ RuntimeError, 14s | **Harness** — never ran |
| `an-1-runtime` | ❌ no result | **Harness** — never ran |

### The change itself is validated

`in-2` executed a full 16-minute build on a real 2-GPU box and **passed** the `COMPOSE_PROFILES` check containing `vss-haproxy-ingress`, the exclusion check, and the stock-template/no-patch artifact checks. `sop-1` and `an-1-api` are untouched specs that read the rewritten `ingress.md`, `base.md` and `SKILL.md` and scored perfectly — `sop-1` through an actual deployment. That is the regression evidence for the always-included-ingress reframing.

### Fixed in `de05eea7`

1. **`in-2`'s only failure was my own over-restriction.** I required `EXTERNAL_IP`/`VSS_PUBLIC_HOST`/`VSS_PUBLIC_PORT`/`HAPROXY_HOST_PORT` not be re-declared. But `overrides.env` sets `HOST_IP='<HOST_IP>'` on line 91 and `EXTERNAL_IP="${HOST_IP}"` on line 93 — the derivation interpolates *at that layer*, against the sentinel, so a later `HOST_IP` from the build cannot fix it. Pinning all four is **required**, which is why that same build still resolved sentinel-free. The agent was right; the check was wrong.
2. **`at-1` graded a CLI it never asked for.** This branch moved its VIOS probe to `vss vios list` but, unlike the other runtime specs, never added the matching env instruction. Added.
3. **`combined` — the no-curation rule was unreachable.** The agent still wrote `patches/vss-haproxy-ingress.yml` on a head that already carried the corrected guidance. The rule sat in a mode paragraph above the reference config; it is now a blockquote at the curated-config section itself. Separately, `ingress.md` said the ingress's front-door use "requires the Agent owner", which reads backwards once the ingress is always included — the front door needs the Agent, not the reverse. That plausibly contributed to the agent keeping `vss-agent`/`vss-ui`/`phoenix`; stated explicitly now. (Hypothesis from the doc wording — I did not read the trace, and it does not explain the dropped `alert-bridge`.)
4. **`in-1` was blocked by two checks that are false against the repo**, in two consecutive runs, so its build/deploy/runtime steps never executed:
   - Check 8 required `RTVI_VLM_KAFKA_TOPIC=mdx-vlm-captions`. **No compose file reads that variable.** The one that decides the topic is `RTVI_VLM_MESSAGE_BUS_TOPIC`, whose compose default is already `mdx-vlm-captions`. It also claimed RT-VLM defaults to `vision-llm-messages`, a string that appears nowhere — and the judge failed an otherwise-correct proposal on exactly that clause. `RTVI_VLM_KAFKA_ENABLED=true` is genuine and stays.
   - Check 10 required asserting `mdx-vlm-captions` is in `kafka-topic-init-container`'s default `KAFKA_TOPICS`. It is not — that list is `mdx-raw, mdx-bev, mdx-space-utilization, mdx-alerts, mdx-behavior, mdx-behavior-plus, mdx-frames, mdx-mtmc`.

### Not fixed — infrastructure

- **`in-3` never ran.** It declares `gpu_count: 2` and was scheduled onto `vss-eval-rtx-1g-1`, a 1-GPU box, dying in 14s. Same pool-selection bug reported on #1744 ("1-GPU box selected for a 2-GPU task"), recurring. Corroborating: three `test_run_leg.py` pool-routing tests fail on clean `develop` today.
- **`an-1-runtime` never ran** — `API Error: The operation timed out` in the eval agent's SDK call.
- **`at-1` check 3** — judge returned no verdict (`error_max_turns`).
- **`an-1-runtime` spec** still expects container `nvidia-nemotron-nano-9b-v2` while `dev-profile-lvs/overrides.env:55` sets `LLM_NAME_SLUG=nemotron-3.5-lightning-30b-a3b`. Pre-existing, left alone.

zac-wang-nv:
## Iteration 2 — independent fact-check of every claim this branch adds → `09ea89cf`

This branch adds a lot of *explanatory* sentences to LLM-graded checks and to skill docs. Each one asserts something about the repo, and a judge grades against it — so I had the OpenAI Codex CLI verify all of them against `develop`, one at a time, before pushing.

### The important finding: this branch had reversed a check that was already correct

`b5e1d8ad` rewrote IN-1's check 10, which said `mdx-vlm-captions` is already in `kafka-topic-init-container`'s default `KAFKA_TOPICS`. **That claim was true.** The list holds 21 topics and includes it; my earlier reading stopped at the eighth entry and concluded it was absent. Reverted verbatim in `731a58df`.

The other half of `b5e1d8ad` stands, independently confirmed: no Compose file reads `RTVI_VLM_KAFKA_TOPIC` (Codex scanned all 72), and `dev-profile-base/overrides.env:167` does ship `RTVI_VLM_KAFKA_ENABLED=false` against a Compose default of `true`.

### Other claims corrected

| Claim as written | Reality |
|---|---|
| `search` needs `elasticsearch` plus `rt_embed` **or** `rtvi_cv` | `_command_availability` **unions** every action's requirements — "Available means the whole surface is" — so it needs all three |
| `vss configure` discovers which routes can serve; the CLI "reaches every deployed backend" | It probes only its seven registered ingress services (`/api`, `/vst`, `/elasticsearch`, `/rtvi-embed`, `/rtvi-cv`, `/rtvi-vlm`, `/lvs`); `/kibana`, `/storage`, `/video-analytics-api`, `/alert-bridge` are never probed |
| `EXTERNAL_IP`/`VSS_PUBLIC_HOST`/`VSS_PUBLIC_PORT`/`HAPROXY_HOST_PORT` all derive from `HOST_IP` | Only `EXTERNAL_IP`, and `VSS_PUBLIC_HOST` through it. `HAPROXY_HOST_PORT` is a literal `7777`; `VSS_PUBLIC_PORT` derives from that |
| `localhost`/`127.0.0.1` are *always* allowed | Allowed, but the `host:port` forms match `HAPROXY_PORT` — remap the host port and they 404 |
| A pruned backend's route returns `503` | *Ordinarily*. `/storage` can answer `HEAD`/`OPTIONS` synthetically, and the Elasticsearch guard returns `403`/`405` before backend selection |

A second review pass over those corrections caught three more, fixed in `09ea89cf`: I had called `/storage`, `/video-analytics-api` and `/alert-bridge` "browse-only routes left in place for a human" (they are programmatic APIs — only `/kibana` fits); IN-2/IN-3 env text still said pruned routes "simply return 503", contradicting the qualification added elsewhere; and the host-identity sentence claimed pinning all four values "is not a copied default" and then conceded two of them are.

### Verified TRUE and left alone

The stock template does front `/rtvi-vlm`; every backend uses `init-addr none`; the template carries all eight routes claimed; the `vios` group requires only `vst`; `vss vios list` exits 4 until `vss configure` runs; RT-Embed's `MESSAGE_BUS` defaults empty and an empty value disables the bus; and every `vss vios` verb and flag named in the checks exists.

### Still open

- Runtime assertions (probes succeed, media resolves) can only be established by a Harbor run, not from source.
- The `search-standalone` deletion was called scope creep by the review. It was explicit maintainer direction, so it stays — flagging it here for visibility.

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33023839282)

zac-wang-nv:
## Iteration 3 — eval results on `09ea89cf`

### The change now has runtime evidence

`in-2` reached its later steps for the first time in this PR:

| Step | `bb236ced` | `09ea89cf` |
|---|---|---|
| 1 — build | 10/11 | **11/11 ✅** |
| 2 — deploy + readiness | never ran | **5/5 ✅** |
| 3 — runtime validation | never ran | 5/6 |

Step 2 is the one that matters for this PR: it carries the repo-local CLI bootstrap check (`uv run --project … --no-dev --extra cli vss`, no global binary), `vss configure --base-url …:7777`, and `vss vios list` as the VIOS readiness probe — all passing against a real deployment with the ingress retained. Step 1 going 10/11 → 11/11 confirms the host-identity correction from iteration 2.

### Most legs never produced a verdict — harness, not specs

| Leg | Outcome |
|---|---|
| `an-1-api` | ✅ passed |
| `in-2` | 11/11, 5/5, 5/6 |
| `sop-1` | 12/13 (was 13/13) |
| `an-1-runtime` | 4/5 on step 1 after a pool retry |
| `in-3` | **no trial** — scheduled onto a 1-GPU box for a `gpu_count: 2` spec |
| `at-1` | **no trial** — blocked before running |
| `in-1` | ran, then **Harbor agent timeout (exit 124)**; agent cut off mid-work |

Three eval agents independently diagnosed the same scheduler bug. From the `in-3` trace:

> *"Line 663 says for registered nodes, if the name-hint count is less than required, skip. But for managed instances the code only checks `gpu_type`"*

and from `an-1-runtime`:

> *"The bug is clear: line 661 gates the gpu_count hint filter on `inst.get("_registered")` … They match on gpu_type alone and then fail at `brev_env.py`'s stricter validation. This is a bug in `run_leg.py`"*

`pool_candidates()` applies the `gpu_count` name-hint filter only to registered nodes, so a managed `vss-eval-rtx-1g-*` box stays a candidate for a 2-GPU spec, gets locked, and is then rejected by `brev_env` — the leg dies without running. This is described as fixed on #1744, so it regressed or never landed.

**I am not fixing that here.** I wrote the one-line fix with a regression test that fails on the old behaviour, then withdrew it: it promotes a regex name hint to a hard fleet-wide eligibility rule, while `pool_candidates`' own docstring says `brev_env` validates the final pick and commit `23a585e13` (#929) deliberately made scheduling permissive. A managed box resized under a stale `-1g-` name would be wrongly discarded, and if every managed candidate is under-hinted the leg waits out its timeout instead of being validated. That is shared scheduler code affecting every skill's evals, and it belongs in its own PR — plausibly alongside #1751, which is already reworking this area. The better fix resolves managed counts from `instance_type` rather than the name.

### Two failures I cannot attribute

`sop-1` (12/13, untouched by this PR) and `in-2` step 3 (5/6) each have one failing check whose identity is not recoverable: both legs exited 124 on transport-group cleanup before rendering a comment, and the uploaded artifacts carry only `{"rewards": {"reward": …}}` — `judge.json` stays on the Harbor box. I would rather report that plainly than guess at a fix.

zac-wang-nv:
Opened #1878 for the pool-scheduler bug that cost this PR three eval legs (IN-3, AT-1, IN-1 — each died without running a trial). It is deliberately not folded in here: it is shared scheduler code affecting every skill's evals, and this PR is skills-only.

Note the ordering consequence — the Skills Eval workflow checks out the **PR branch's own** copy of `run_leg.py`, so #1878 landing on `develop` does not help this PR until this branch is rebased onto it. Once #1878 merges, a rebase here should let IN-3 and AT-1 actually run.

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33029874424)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33031034550)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33033025612)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33035144500)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33038672225)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33043354041)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33044320192)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33051293145)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33059145301)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33059638208)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33065560525)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33070162015)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33076561115)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33082021515)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33089010526)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33127091091)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33131600172)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33135314075)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33197678222)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33214163519)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33223347324)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33226044306)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33227038308)

PR #1865, opened 2026-08-26
refactor(helm): publish one ingress route table across dev profiles

zac-wang-nv:
## Description

The four developer profiles each owned their own ingress path list, so the same service
answered at a different place depending on which one you deployed:

| | base | alerts | lvs | search |
|---|---|---|---|---|
| RT-VLM | `/v1` (Prefix) | not deployed | `/v1/models`, `/v1/chat/completions` (Exact) | `/rtvi-vlm` (behind a toggle) |
| LVS | — | — | `/v1/ready`, `/v1/summarize` (Exact) | — |
| Elasticsearch | not deployed | **missing, though ES is deployed** | behind `cliRoutes` | behind `cliRoutes` |
| RT-CV / RT-Embed | — | **missing, though RT-CV is deployed** | — | two publishers: `cliRoutes` *or* `global.rtviInternalIngress` |
| values key | `vssIngress` | `vssIngress` | `vssIngress` | `ingress` |

A caller could not resolve an endpoint without first knowing which profile answered. That is
what forced the probe loop and the per-profile endpoint tables in #1797.

This adds `deploy/helm/services/common`, a Helm **library** chart holding one route table, and
renders all four profiles from it. Every service is mounted at its own name — `/rtvi-vlm`,
`/elasticsearch`, `/lvs`, `/va-mcp`, `/vst`, `/phoenix` — which are the paths the Docker edge
publishes and `vss_cli/config.py:INGRESS_SERVICES` records. `vss.ingress.assertBackends` fails
the render on a key the table does not define, so a new route is added once instead of four
times.

Scope is developer profiles. Industry profiles keep their own ingress; they can adopt the same
table in a follow-up.

### Behaviour changes

- **`/v1` is gone.** Base's Prefix `/v1` and LVS's four Exact `/v1/*` paths are replaced by
  `/rtvi-vlm/v1/...` and `/lvs/v1/...`. Because `/lvs` is a Prefix mount, LVS routes the Exact
  rules could not reach — `/lvs/models`, `/lvs/openapi.json`, `/lvs/metrics` — are now public.
- **`/elasticsearch` is mounted by default** wherever Elasticsearch is deployed, no longer
  behind `cliRoutes`. The Docker edge's deny rules are ported to a `backend-config-snippet` on
  the Elasticsearch **Service**, so they bind to that backend rather than to every route on the
  origin. This is the same grant Docker makes and **is not a read-only endpoint**: PUT/DELETE,
  cluster-admin paths and the two-segment mutating operations are denied, but POST still reaches
  Elasticsearch (`_bulk`, `_reindex`, `/<index>/_doc`). Set
  `infra.elasticsearch.ingressGuard.enabled=false` only if you want it entirely unguarded.
- **`templates/rtvi-internal-ingress.yaml` is deleted.** `/rtvi-cv` and `/rtvi-embed` have one
  publisher now. In-cluster callers reach the ingress controller's ClusterIP and send that
  Service name as `Host`, matching no named rule, so the profile renders a host-less rule
  carrying only those two mounts. It is gated on `global.rtviInternalIngress.enabled` and **off
  by default**, because host-less means every Host the controller answers, external listener
  included. Stream affinity already lived on the two rtvi Services.
- **Per-backend HAProxy settings move to the Services** that own them, so the 3600s server
  timeout no longer applies to every LVS and search route.
- **Search's `ingress:` block is now `vssIngress:`.** `enabled`, `className`, `hosts.*`,
  `annotations` and `tls` are still read from `ingress:` and win where set. The removed
  `cliRoutes.*` keys are ignored. Base's `vssIngress.vlm.{enabled,serviceName,servicePort}`
  likewise still win.

### Two pre-existing problems this surfaced

- **`timeout-client` and `timeout-tunnel` have never worked as Ingress annotations.** This
  controller reads both only from its own ConfigMap
  ([annotation matrix](https://github.com/haproxytech/kubernetes-ingress/blob/master/documentation/annotations.md)).
  The LVS and search charts have carried them since before this PR. They are still rendered for
  continuity, now with a comment saying they are inert; raising them means a cluster-wide
  controller setting. Long LVS runs are still exposed to the controller's default client
  timeout. **Not fixed here** — flagging it.
- **Docker's `/va-mcp` route has no `replace-path`**, so it forwards the prefix to a server that
  serves `/mcp` at root. The skills docs already describe the stripped behaviour that Helm
  implements. Docker-side fix, out of scope.

## Plan

Grilled before building, per the workspace guidelines. Questions asked, my recommendation, and
the answer taken:

| # | Question | My recommendation | Answer |
|---|---|---|---|
| 1 | Where does the one route table live? | Shared library chart | **Shared library chart** |
| 2 | Keep base `/v1` and LVS `/v1/*` as aliases, or cut? | Keep for 3.3, drop in 3.4 | **Cut now** — so #1797's probe can die immediately |
| 3 | How is Elasticsearch exposed? | Toggle off + edge guard | **Always mount + guard** |
| 4 | How do `/rtvi-cv` and `/rtvi-embed` get one publisher? | Fold into the main Ingress | **Fold into the main Ingress** |
| 5 | How do east-west callers still match once folded? | Host-less rule, RTVI paths only | **Host-less rule, RTVI paths only** |
| 6 | Move per-backend annotations to the Services? | Yes | **Yes** |
| 7 | Scope | Dev profiles only; industry follow-on | **Dev profiles only** |

Taken without asking, all alias-preserving or straight from the route table: every route strips
its prefix except `/vst` and the agent/UI paths; `/storage` and `/behavior-analytics` added
because the route model lists them; Kibana and NVStreamer stay on dedicated hosts (Kibana needs
`server.basePath` to be path-mounted).

## Review round

Sent to an independent reviewer before opening this PR. Nine findings, all reproduced and
fixed in this branch:

1. **`default true .enabled` re-enabled disabled components.** sprig treats `false` as empty and
   hands back the default, so `--set vss-summarization.enabled=false` still mounted `/lvs` at a
   Service that was never created. Gating now goes through a `vss.ingress.enabled` helper that
   reads key presence. Umbrella dependencies (`infra.enabled`, `agent.enabled`) are checked too,
   and `ui`/`agent`/`vst` are gated rather than unconditional.
2. **The host-less rule was unconditional**, which was new exposure on alerts. Now gated on
   `global.rtviInternalIngress.enabled`, off by default.
3. **Legacy `ingress.className` was ignored** — `coalesce` on the merged dict always found the
   new key's chart default first. Legacy is now checked ahead of it.
4. **Base's `vssIngress.vlm.*` keys were silently dropped**, so an existing
   `vlm.enabled=false` would have started publishing RT-VLM. They are honoured again.
5. **`timeout-client`/`timeout-tunnel` scope** — see above; my comments claimed they were
   frontend-scope and honoured. Corrected everywhere.
6. **The Elasticsearch guard was described as read-only** in four places. It is not; corrected
   to state exactly what it denies and that POST passes.
7. **The verifier over-claimed.** It said "publishes one route table" while checking neither the
   CLI's mount list nor rewrite destinations, and never exercised a disabled component. It now
   checks pathTypes, rewrite destinations, `INGRESS_SERVICES` parity, six disabled-component
   cases, and the host-less-rule default — and its docstring says plainly that it does *not*
   check Docker parity.
8. **`memoryIndex` is interpolated into an HAProxy `path_reg`** without escaping. Documented as
   a constraint on that value.
9. A stale `/v1` comment left above changed code in `vss-generate-video-report`.

## Test plan

- [x] `helm lint` clean on all four profiles.
- [x] `helm template` renders all four (37 / 87 / 72 / 87 objects).
- [x] `python3 deploy/helm/developer-profiles/verify-ingress-routes.py` passes:
      `4 profiles render from one table (22 distinct mounts); CLI probe mounts present;
      6 disabled-component cases drop their route; examples current`.
- [x] Each new check negative-tested: perturbing a `pathType`, a rewrite destination, or an
      example file makes the verifier fail.
- [x] Legacy alias round-trip: `--set ingress.hosts.main=…` overrides the host,
      `ingress.enabled: false` (as `values-nodeport.yaml` sets it) still suppresses the object,
      `--set ingress.className=nginx` wins, `--set vssIngress.vlm.enabled=false` drops
      `/rtvi-vlm`.
- [x] Default rendered route lists are unchanged for every route that existed before, except the
      deliberate changes listed above.
- [ ] **Not verified — needs one cluster pass.** Three behaviours rest on documented controller
      semantics rather than observation: Service-scoped annotations overriding Ingress-scoped
      ones per backend; `backend-config-snippet` coexisting with the generated `replace-path`
      (the ES guard's regexes are written prefix-optional so they hold either way); and a
      host-less rule matching an in-cluster `Host`. The last two are what the deleted
      `rtvi-internal-ingress.yaml` already relied on.

## Checklist


pratik-bhatt-nv:
[P1] Required LVS workflow still uses retired root routes.

`skills/vss-summarize-video/references/end-to-end-example.md` is a required reference for the recorded-video workflow, but its Kubernetes resolution still assigns both `LVS_BACKEND_URL` and `VLM` to the bare public origin. Its readiness probe consequently requests `${VSS_PUBLIC_URL}/v1/ready`, even though this PR publishes LVS only at `${VSS_PUBLIC_URL}/lvs`; the bare-origin catch-all is the UI. The workflow will therefore report a healthy Kubernetes LVS deployment as unavailable, and its VLM fallback probe targets the wrong backend too.

Please update this reference to use `${VSS_PUBLIC_URL}/lvs` for LVS and `${VSS_PUBLIC_URL}/rtvi-vlm` for RT-VLM, and clean up the matching stale statements in `SKILL.md` and `video-summarization-api.md`.

pratik-bhatt-nv:
[P1] Kubernetes consumer contract still assigns LVS to the origin root.

`skills/vss-build-vision-ai/references/deployment_resolution.md` now correctly defines the Kubernetes LVS base as `${VSS_PUBLIC_URL}/lvs` near the top, but its required consumer block still says `LVS_BACKEND_URL="${AGENT_URL}"`, where `AGENT_URL` is the bare public origin. A consumer that follows this block probes `/v1/ready` on the UI catch-all rather than the LVS backend.

Please set `LVS_BACKEND_URL="${AGENT_URL}/lvs"` (and update the adjacent comment). This is a second, independently usable stale contract source; it is not fixed by correcting only the summarize-skill reference.

zac-wang-nv:
Both P1s are addressed.

**[P1] Required LVS workflow still uses retired root routes** → fixed in c34804c3d.
`end-to-end-example.md`'s Kubernetes resolution now sets `LVS_BACKEND_URL="${VSS_PUBLIC_URL}/lvs"` and `VLM="${VSS_PUBLIC_URL}/rtvi-vlm"`, with the comment explaining that the bare origin is the UI catch-all and the gateway strips `/lvs` before the backend sees `/v1/ready` / `/v1/summarize`. The matching stale statements are swept: `SKILL.md`'s endpoint table row ("K8s: public origin" → "K8s: the `/rtvi-vlm` mount") and `video-summarization-api.md`'s commented Kubernetes `BASE_URL` (now `${VSS_PUBLIC_URL%/}/lvs`, agreeing with its own `/lvs` mount paragraph).

**[P1] Kubernetes consumer contract still assigns LVS to the origin root** → fixed in c34804c3d, then superseded by 269398d07: the consumer block now derives everything from `VSS_PUBLIC_URL` directly (`LVS_BACKEND_URL="${VSS_PUBLIC_URL}/lvs"`, adjacent comment updated), and the `AGENT_URL` alias that made this stale contract possible is removed repo-wide — it was never a real env var, only a doc-level alias for the public origin. The Docker/search docs collapse it into the already-in-scope `VSS_ORIGIN`. 269398d07 also retires `GET /openapi.json` as the agent readiness probe in favor of `vss configure check` (raw fallback `GET /api/v1/videos`, the CLI's own agent probe from `INGRESS_SERVICES`).

Verification: all six Kubernetes LVS/VLM base assignments across `vss-summarize-video` and `vss-build-vision-ai` now target the mounts (repo-wide grep shows zero bare-origin assignments), and the doc-block contract suites pass — search verification 11/11 (it executes the edited blocks), summarize 30/30.

PR #1760, opened 2026-08-19
Feat/route service calls via haproxy

kaushikc-nvidia:
## Summary

Adds a placement-neutral gateway endpoint contract for Docker Compose deployments so agent-facing service calls can be routed through HAProxy when the agent is moved off the Docker bridge network, while preserving same-bridge behavior through Docker DNS/network aliases.

The Docker HAProxy path is treated as HTTP-only. A single `VSS_GATEWAY_ORIGIN` drives downstream service URLs, and HAProxy publishes `VSS_GATEWAY_HOST` as a Docker network alias for same-bridge validation.

## Related Issue

VIA-2713

## Changes

- Added gateway identity plumbing to profile overrides:
  - `VSS_GATEWAY_HOST`
  - `VSS_GATEWAY_PORT`
  - `VSS_GATEWAY_ORIGIN`
  - `VSS_GATEWAY_ORIGIN_FALLBACK`

- Added `gateway-endpoints.env.example` as a native Docker Compose env overlay to derive gateway-routed endpoints from one origin:
  - VST
  - VA MCP
  - Alert Bridge
  - RTVI VLM/CV/Embed
  - Elasticsearch
  - Phoenix
  - LVS
  - Auto-calibration / VIOS, if included in the latest patch

- Updated HAProxy Compose wiring:
  - Publishes `${VSS_GATEWAY_HOST:-vss.local}` as a Docker network alias
  - Exposes gateway identity env values to HAProxy
  - Adds/normalizes `/va-mcp` routing through HAProxy

- Updated agent and VA MCP configs to consume gateway-aware endpoint variables instead of hard-coded Docker service names where appropriate.

- Cleaned up VST URL handling in agent code:
  - Centralized VST internal URL resolution
  - Removed hidden localhost fallback behavior for VST API calls
  - Normalized origin-scoped VST URLs
  - Ensured VST helper calls receive configured `vst_internal_url` from agent/tool config where available

- Added critic/search/video-understanding propagation for configured VST internal URLs, including critic-agent offset-mode config.

- Documented native Docker Compose usage for the gateway overlay, including same-bridge alias behavior and off-host DNS behavior.

## Testing

- Ran targeted Python lint checks for changed agent/VST modules.
- Ran targeted unit tests for VST helpers, search agent, critic agent, and video ingest changes.
- Rendered native Docker Compose config with gateway overlay to validate:
  - default fallback origin
  - explicit gateway origin
  - HAProxy network alias
  - agent endpoint values
  - VA MCP endpoint values
[...]

PR #1867, opened 2026-08-26
feat(deploy): placement-neutral gateway contract for Docker service-to-service HTTP

hugoverjus:
## Why

The VSS agent has to be able to run off-host, and today it cannot, because its
endpoint configuration encodes *where* every service runs. `http://vss-va-mcp:9901`
is a Compose DNS nickname: it resolves on the bridge and nowhere else. Any caller
that moves off the host needs a different string, which in practice means
caller-side "if colocated do X else Y" branching in config.

This change replaces those nicknames with one logical origin plus a stable path
contract — HAProxy is the HTTP front door, placement lives in DNS and in HAProxy,
and the caller's URL is the same whether it is colocated or remote. Per the SRD's
FR-03 (single logical origin), FR-11 (stable path contract) and NFR-01 (no
caller-side placement branching).

Colocated single-host Compose keeps working; the gateway is the recommended
default rather than a new requirement. Non-HTTP data planes (Kafka, Redis, RTSP,
raw media) are explicitly untouched.

## The pinned gateway origin, and the bug it prevents

`VSS_GATEWAY_PORT` derived from `VSS_PUBLIC_PORT`, with the scheme taken from
`VSS_PUBLIC_HTTP_PROTOCOL`. That is wrong wherever the platform terminates TLS
outside the stack. On Brev, TLS terminates at the Brev/Cloudflare edge, which
forwards **plain HTTP to HAProxy on 7777** — so `VSS_PUBLIC_*` is `https`/`443`,
and every container would have been handed `https://vss.local:443`: a listener
that does not exist, on a name only the bridge can resolve.

The gateway is now pinned to HAProxy's own listener — `http://vss.local:7777`,
port from `HAPROXY_PORT`, scheme forced to `http` — independently of
`VSS_PUBLIC_*`. `VSS_PUBLIC_*` keeps its real job: browsers, report links, and
`vss configure --base-url` on the host.

## A fixed bug, not a refactor side effect

On the base commit, three variables render **empty**:

- `COSMOS_EMBED_ENDPOINT`
- `ELASTIC_SEARCH_ENDPOINT`
- `LVS_BACKEND_URL`

Empty, not wrong — so `${ORIGIN}/elasticsearch` renders as `/elasticsearch` and
fails much later as a malformed request rather than here as a configuration
error. The inline nested defaults in this PR fix that. Worth calling out
separately because it is a live bug on `develop`, not something this refactor
introduced.

The cause was a typo, **`VSS_GATEWAY_ORIGEN`**, which this PR removes. That is
also the explanation for the empty-variable behaviour seen in the earlier spike
[#1760](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1760):
a misspelled variable name interpolates to the empty string silently.

## Elasticsearch: three consumers moved, seven left direct

Moved to the gateway origin plus the `/elasticsearch` path: the agent, the
search profile's `ELASTIC_SEARCH_ENDPOINT`, and the LVS caption-retrieval
config.

Seven are deliberately left pointing at `elasticsearch:9200`, and the reason is
measured rather than assumed. `/elasticsearch` is a **narrow** mount: its ACLs
admit the query surface and reject cluster administration and writes. Confirmed
against the live deployment — 200 on `/`, `_cat/indices` and `_search`; **403**
on `_nodes`, `_cluster/health`, `_nodes/_all/_none`, `_snapshot` and
`_security/_authenticate`.

Kibana's startup probe is `GET _nodes/_all/_none`, which is one of those 403s,
so a gateway-routed Kibana never becomes ready. In the live run Kibana reached
healthy with `elasticsearch.hosts: ["http://elasticsearch:9200"]` — direct, not
through the mount — which is the reasoning holding up in practice rather than on
paper.

Widening the ACL was considered and rejected. `/elasticsearch` is reachable from
the ingress, so admitting `_nodes` and `_cluster` would turn it into
unauthenticated cluster administration exposed at the edge — a materially worse
trade than seven consumers keeping a Compose DNS name on a single host. The
remaining consumers are documented in `deploy/docker/README.md` with the reason
each stayed.

## An explicitly accepted deviation from the SRD

The SRD prefers one origin per environment. This deployment cannot have that on
Brev: TLS terminates at Brev's edge and plain HTTP is forwarded inward, so
browsers and the host CLI use the HTTPS secure link while containers use
`http://vss.local:7777`. One front door, one path contract, two origins. See
**OI-02** and **OI-09**.

An opt-in HAProxy TLS listener was scoped — roughly two to four days for an
`ssl crt` frontend, a certificate with a SAN for `vss.local`, and CA trust
distributed to the Python clients — and deferred to its own ticket. It is worth
doing for operators who want end-to-end TLS inside the stack, but it would not
change anything on Brev, where the edge terminates TLS regardless.

## CLI

`vss configure` now probes `/va-mcp` alongside the existing routes and reports it
in `vss configure check`, following the existing `INGRESS_SERVICES` pattern. No
command group *requires* it — it is recorded for operator visibility, since a
missing VA-MCP is worth knowing about before a run rather than during one.

Docs now say plainly that the host cannot resolve `vss.local` (it is a Docker
bridge alias) and that the CLI must be configured with `VSS_PUBLIC_URL`.
Configuring the gateway host fails every probe and makes a healthy deployment
look broken, which is a confusing enough failure to be worth naming.

## Validation

Validated against **real GPU deployments**: full `dev-profile-alerts`
(22 containers) and `dev-profile-lvs` (19 containers) on a remote box.

**The agent works through the gateway.** Its MCP client reported a healthy
`streamable-http` session on `http://vss.local:7777/va-mcp/mcp` with tools
listed. `POST http://vss.local:7777/chat` returned **200 in 35.9 s**, having
planned and invoked an MCP tool over the gateway — so the whole agent → gateway →
VA-MCP path carried a real request, not a probe.

**Tracing lands through the gateway.** Phoenix went from a single `default`
project at `traceCount 0` to a new `DEV-ALERTS-vss-agent-` project at
`traceCount 1`, which is OTLP export through `/phoenix/v1/traces` proven by its
effect rather than by a 200.

**Elasticsearch through the mount:** the ACL results above, measured in
production.

**`vss` CLI against the live deployment** at `http://10.63.177.97:7777`: recorded
6/8 services including `va_mcp routed HTTP 200`, and `vss configure check`
reported `summarize available lvs`, `vios available vst`, `search unavailable
needs rt_embed, rtvi_cv`. Real commands ran against live data — `vios add`,
`vios list`, `vios timeline`, `vios snapshot`, `vios clip` (media URLs minted,
`warmed: true`), and `summarize run` producing genuine VLM output. An empty
listing correctly returned `{"count": 0}` at exit 0, so the "empty result is an
answer" contract holds too.

**Previously unexercised routes now exercised:** `/rtvi-vlm/v1/models` 200 with a
real model, `/rtvi-cv/api/v1/stream/add` 200, `/lvs/v1/live` 200,
`/vst/api/v1/sensor/version` returning `{"type":"vst","version":"1.4.0-26.08.1"}`.

**Still untested, and worth knowing:**

- **`/rtvi-embed`** — 503 in both profiles, because it belongs to
  `dev-profile-search`, which was not one of the two deployed.
- **The `timeout server 3600s` override** — config-verified against the 120 s
  global default, but the longest real summarization was 82 s, so no request
  crossed the line that override exists for.
- **`/behavior-analytics`, `/perception-sdr`, `/video-analytics-api`** — routed
  and rendered, but no real HTTP surface was found to call.
- **Kubernetes and Helm** — verified by rendering only; no cluster was involved.

## Known follow-ups

- **`alert-bridge` still holds Compose DNS names** in its own configuration.
- **Three ES consumers need code changes**, not config: their clients cannot
  express a path prefix at all, so moving them is a source change rather than an
  env edit.
- **The CI `detect-secrets` step is vacuous.** It runs from `services/agent` with
  repo-root-relative paths, so it scans nothing and exits 0. It has presumably
  been passing without inspecting anything for some time; worth its own ticket.

### Investigated and closed, so nobody re-opens it

**`/phoenix` strip versus `PHOENIX_HOST_ROOT_PATH` is correct.** An earlier draft
of this PR listed the pairing as a suspected inconsistency — HAProxy strips the
prefix while Phoenix is told it is mounted under one. That was a false alarm. It
is Phoenix's own documented arrangement and the standard FastAPI `root_path`
idiom: HAProxy strips, Phoenix receives bare paths, and `PHOENIX_HOST_ROOT_PATH`
makes Phoenix *generate* `/phoenix`-prefixed URLs. Upstream pairs Traefik's
`stripPrefix` with the same variable. Measured live: `GET /phoenix` → 200 with
every asset URL prefixed `/phoenix/…`, `/phoenix/v1/traces` → 200,
`/phoenix/healthz` → 200. No change needed.

### DIFF
--- a/deploy/docker/compose.gateway-alias.yml
+++ b/deploy/docker/compose.gateway-alias.yml
@@ -1,0 +1,26 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Optional overlay: publish VSS_GATEWAY_HOST as a Docker bridge DNS name for
+# vss-haproxy-ingress. Off-host callers resolve the same name through normal
+# DNS. Do not include this file unless VSS_GATEWAY_HOST is a hostname (not an
+# empty value and not an IP you do not want as a network alias).
+
+services:
+  vss-haproxy-ingress:
+    networks:
+      default:
+        aliases:
+          - ${VSS_GATEWAY_HOST:?set VSS_GATEWAY_HOST to a hostname before using compose.gateway-alias.yml}
--- a/deploy/docker/developer-profiles/dev-profile-alerts/overrides.env
+++ b/deploy/docker/developer-profiles/dev-profile-alerts/overrides.env
@@ -105,6 +105,22 @@ VSS_PUBLIC_HOST=${EXTERNAL_IP}
 HAPROXY_HOST_PORT=7777
 # Direct local ingress uses HAPROXY_HOST_PORT; TLS or secure-link ingress usually uses 443.
 VSS_PUBLIC_PORT=${HAPROXY_HOST_PORT}
+# Canonical gateway identity for in-Docker HTTP service-to-service calls. One
+# origin for colocated and remote callers alike; DNS and the HAProxy vss.local
+# alias supply placement, so no caller branches on where a service runs.
+#
+# Pinned to HAProxy's own listener, NOT derived from VSS_PUBLIC_*. Where the
+# platform terminates TLS outside the stack (a Brev secure link is https on 443
+# forwarding plain HTTP to 7777) VSS_PUBLIC_* describes the browser's origin,
+# and reusing it here hands containers https://vss.local:443 -- a listener that
+# does not exist. VSS_PUBLIC_* stays for browsers, report links, and
+# `vss configure --base-url` on the host. See deploy/docker/README.md.
+#
+# Never leave these empty: an empty origin renders "/elasticsearch" and fails
+# later as a malformed request instead of here as a configuration error.
+VSS_GATEWAY_HOST=vss.local
+VSS_GATEWAY_PORT=${HAPROXY_PORT:-7777}
+VSS_GATEWAY_ORIGIN=http://${VSS_GATEWAY_HOST}:${VSS_GATEWAY_PORT}
 
 # =============================================================================
 # HOST-PUBLISHED PORTS
--- a/deploy/docker/services/infra/haproxy/compose.yml
+++ b/deploy/docker/services/infra/haproxy/compose.yml
@@ -51,6 +51,15 @@ services:
       EXTERNAL_IP: ${EXTERNAL_IP}
       VSS_PUBLIC_HOST: ${VSS_PUBLIC_HOST}
       VSS_PUBLIC_PORT: ${VSS_PUBLIC_PORT}
+      # Canonical gateway identity, used by the Host ACLs so a caller addressing
+      # the gateway name is admitted like any other known vhost. The port is the
+      # listener HAProxy binds (HAPROXY_PORT), not the host-published or public
+      # one: an in-network caller reaches the container port, and VSS_PUBLIC_PORT
+      # is 443 wherever a platform terminates TLS in front of this proxy.
+      # VSS_GATEWAY_ORIGIN is deliberately absent -- haproxy.cfg.template never
+      # reads it, and a `${VAR:-}` default here is how it ends up empty.
+      VSS_GATEWAY_HOST: ${VSS_GATEWAY_HOST:-vss.local}
+      VSS_GATEWAY_PORT: ${VSS_GATEWAY_PORT:-${HAPROXY_PORT:-7777}}
       VSS_UI_SERVICE_HOST: ${VSS_UI_SERVICE_HOST:-vss-ui}
       VSS_UI_PORT: ${VSS_UI_PORT:-3000}
       VSS_AGENT_SERVICE_HOST: ${VSS_AGENT_SERVICE_HOST:-vss-agent}
@@ -77,3 +86,12 @@ services:
       PHOENIX_PORT: ${PHOENIX_PORT:-6006}
     volumes:
       - ./haproxy.cfg.template:/usr/local/etc/haproxy/haproxy.cfg:ro
+    # vss.local is published unconditionally so the gateway origin resolves on
+    # the bridge with no extra overlay and no per-profile opt-in. A different
+    # VSS_GATEWAY_HOST is aliased by compose.gateway-alias.yml instead of being
+    # interpolated here, which would duplicate this entry in the common case
+    # where the two are the same name.
+    networks:
+      default:
+        aliases:
+          - vss.local
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -56,8 +56,19 @@ vss configure show                              # what was recorded
 vss configure check                             # re-probe + what each group can serve
 ```
 
-`configure` probes the origin's ingress routes and writes `~/.vss/config.json`
-(0600, no credentials). Re-run it after any deployment change.
+`configure` probes the origin's ingress routes — `/api`, `/vst`,
+`/elasticsearch`, `/rtvi-embed`, `/rtvi-cv`, `/rtvi-vlm`, `/lvs`, `/va-mcp` —
+and writes `~/.vss/config.json` (0600, no credentials). Re-run it after any
+deployment change.
+
+**Use the origin the host can reach, never the in-network gateway name.**
+`VSS_PUBLIC_URL` is it: `http://<host>:7777` locally, or the platform's
+`https://…` URL where TLS terminates in front of the deployment (a Brev secure
+link forwards plain HTTP inward). Services *inside* the deployment address the
+same HAProxy front door as `http://vss.local:7777` — that is `VSS_GATEWAY_ORIGIN`,
+and it is correct for them and useless here: it is a container-network alias, so
+configuring it fails every probe with a connection error and makes a healthy
+deployment look broken. One front door, one path contract, two origins.
 
 **Never construct an endpoint.** No `kubectl port-forward`, no Service DNS, no
 NodePort, no reading `HOST_IP` or `VST_INTERNAL_URL` out of a container. The CLI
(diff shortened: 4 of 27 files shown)


zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33099431052)

hugoverjus:
## Review-ready: merge this PR first

GPU proof (2× RTX PRO 6000, full `dev-profile-alerts` + `dev-profile-lvs`): the gateway works. `/vios` vs `/vst` were byte-identical on real VST (sensor list, streams, version, storage JPEG). A strip would have 404'd; `/vios/api/v1/sensor/version` was 200. LVS fetched source media via `/vios/storage/…`. Agent MCP session was `http://vss.local:7777/va-mcp/mcp`; `POST /chat` 200 in ~36s. ES ACL 403 on `_nodes` / `_cluster` / `_nodes/_all/_none`; Kibana healthy on `http://elasticsearch:9200` (not the gateway). Helm Sync passed.

**Please merge this PR before #1868.** Both branches produce the same `vss-agent` content tree, so they share GHCR tag `vss-agent:tree-<sha>`. That tag is a mutable pointer: after this PR's assemble succeeds, #1868's Assemble release set fails even with no concurrent build (same two digests on a rerun of [33099438249](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33099438249)). Rebuilding #1868 after this merges is the fix.

### Remaining red (not this diff)
- **OSRB Review** — License Diff on this PR reported no OSRB re-engagement. Same check is red on unrelated PRs (#1889, #1865). Needs an OSRB owner, not a code change here.
- **Skills Eval** (`vss-build-vision-ai` profiles) — rubric failures about `COMPOSE_PROFILES` / unrequested `vss-agent`, not HAProxy/`vss.local`. Daily evals on `main` have also been failing.

Deterministic CI (pytest, vss-cli, integrations, DCO, Helm Sync, detect-secrets as currently wired, etc.) is green.

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33200644136)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33233946878)

hugoverjus:
Fixed the downstream `test-search` regression at `6e61498ffeb4d3a2c45c318f96eac159a6b732e4`.

Root cause: video upload post-processing rewrote only the host of VST's direct `:30888` media URL. Once `VST_INTERNAL_URL` became the gateway origin, that produced the invalid hybrid `http://vss.local:30888/vst/...`. The fix now replaces the complete origin with `VST_INTERNAL_URL`, preserving the single `/vst/...` path, so RTVI-Embed receives `http://vss.local:7777/vst/...`.

Focused verification: 84 video-ingest tests passed; Ruff and mypy passed; Compose renders for base/search/lvs/alerts/warehouse all resolved `VST_INTERNAL_URL` to `http://vss.local:7777`; regression checks reject `vss.local:30888` and duplicate `/vst/vst/`.

Downstream pipeline [#65187915](https://gitlab-master.nvidia.com/metromind/ci-vss-oss/-/pipelines/65187915) completed successfully, including `test-search`.

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33240837164)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33404387708)

deploy/docker/services/infra/haproxy/compose.yml
```diff
@@ -77,3 +86,12 @@ services:
...
       - ./haproxy.cfg.template:/usr/local/etc/haproxy/haproxy.cfg:ro
+    # vss.local is published unconditionally so the gateway origin resolves on
+    # the bridge with no extra overlay and no per-profile opt-in. A different
+    # VSS_GATEWAY_HOST is aliased by compose.gateway-alias.yml instead of being
+    # interpolated here, which would duplicate this entry in the common case
+    # where the two are the same name.
+    networks:
+      default:
+        aliases:
+          - vss.local
```
kaushikc-nvidia, line 97: This should not be static, should be derived from VSS_GATEWAY_HOST

```networks:
      default:
        aliases:
          - ${VSS_GATEWAY_HOST:-vss.local}
```

deploy/docker/developer-profiles/dev-profile-alerts/overrides.env
```diff
@@ -105,6 +105,22 @@ VSS_PUBLIC_HOST=${EXTERNAL_IP}
...
+# platform terminates TLS outside the stack (a Brev secure link is https on 443
+# forwarding plain HTTP to 7777) VSS_PUBLIC_* describes the browser's origin,
+# and reusing it here hands containers https://vss.local:443 -- a listener that
+# does not exist. VSS_PUBLIC_* stays for browsers, report links, and
+# `vss configure --base-url` on the host. See deploy/docker/README.md.
+#
+# Never leave these empty: an empty origin renders "/elasticsearch" and fails
+# later as a malformed request instead of here as a configuration error.
+VSS_GATEWAY_HOST=vss.local
+VSS_GATEWAY_PORT=${HAPROXY_PORT:-7777}
```
kaushikc-nvidia, line 122: We might have to converge HAPROXY_PORT and HAPROXY_HOST_PORT into a single env var for this to work across single and multi-host setups.

deploy/docker/compose.gateway-alias.yml
```diff
@@ -0,0 +1,26 @@
...
+# vss-haproxy-ingress. Off-host callers resolve the same name through normal
+# DNS. Do not include this file unless VSS_GATEWAY_HOST is a hostname (not an
+# empty value and not an IP you do not want as a network alias).
+
+services:
+  vss-haproxy-ingress:
+    networks:
+      default:
+        aliases:
+          - ${VSS_GATEWAY_HOST:?set VSS_GATEWAY_HOST to a hostname before using compose.gateway-alias.yml}
```
kaushikc-nvidia, line 26: Should not need this file, in favor of what is suggested in earlier comment: https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1867/changes#r3896016092

All clients should consistently have to use the same VSS_GATEWAY_HOST so that URL issuers need not be client location aware

hugoverjus:
Superseded by #1983, which consolidates this PR and its stacked counterpart onto `develop` so the gateway contract can be reviewed and tested as one change. Nothing is dropped; #1983 also carries the fixes found while testing the combined branch on a 2x RTX 6000 PRO host, including the off-host agent proof that was the remaining acceptance gap.

Closing in favour of #1983.

PR #1868, opened 2026-08-26
feat(deploy): additive /vios, /alerts and /video-summarization gateway aliases with deprecation signal

hugoverjus:
> **Stacked on [#1867](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1867).**
> Targets `feat/via-2714-gateway-identity`, not `develop`. Review #1867 first;
> this PR's own diff is the three commits after it.

## Why

Zac asked for the gateway paths `/vios`, `/alerts` and `/video-summarization`.
This delivers those names **additively**: all six prefixes work. `/vst`,
`/alert-bridge` and `/lvs` keep behaving exactly as they do today, and are
deprecated through 3.3.x with removal no earlier than 3.4.0.

A hard rename was rejected. It would have to move the `skills/` tree, the CLI's
vios client (which spells `/vst/` in many places), seven Helm charts, the
notebooks and the skill evals in a single commit, and any straggler would 404 in
production. An alias costs a stanza; a rename costs a coordinated flag day.

## `/vios` rewrites, it does not strip

This is the one asymmetry in the change and the one thing worth reviewing
closely. VST serves its **whole surface** under `/vst/`, so `/vios/api/v1/x` has
to arrive at the backend as `/vst/api/v1/x`:

```
backend bk_vios_rewrite
    http-request replace-path ^/vios/(.*) /vst/\1
    http-request replace-path ^/vios$      /vst
```

A strip — the obvious thing, and what the other two aliases do — would send
`/api/v1/x` and **404 every VST call**. The existing `/storage` → `/vst/storage/…`
route is the same shape, which is what made it clear this was the right pattern
rather than an exception.

`/alerts` and `/video-summarization` strip, exactly like the prefixes they alias.
`/vios/storage` also inherits the `HEAD`/`OPTIONS` range short-circuits that
`/vst/storage` gets, so the alias cannot answer a range request differently from
the prefix it stands in for.

## The `timeout server 3600s` copy, and what it argues

`bk_lvs_strip` carries a raised server timeout, because a summarization can
outlive the 120 s default. `bk_video_summarization_strip` had to copy it. Without
that one line, the alias would **504 on exactly the long jobs where `/lvs`
succeeds** — an alias that behaves differently from the thing it aliases is worse
than no alias at all.

It is a one-line fix, but it is concrete evidence for the maintenance cost of six
prefixes for three backends: every per-backend tuning decision now has two places
to be right. That is an argument for keeping the deprecation window short and
actually retiring the old prefixes in 3.4.0, not for extending it.

## The deprecation signal

Legacy prefixes now emit RFC 9745 / RFC 8594 headers, so the soak period produces
data instead of guesswork:

| Route | Emits |
|-------|-------|
| `/vst` | `Deprecation: true`, `Link: </vios>; rel="successor-version"` |
| `/alert-bridge` | `Deprecation: true`, `Link: </alerts>; rel="successor-version"` |
| `/lvs` | `Deprecation: true`, `Link: </video-summarization>; rel="successor-version"` |
| `/vios`, `/alerts`, `/video-summarization` | nothing |
| `/api`, `/elasticsearch`, `/va-mcp`, `/phoenix`, `/rtvi-vlm`, `/kibana`, `/`, `/storage`, `host:port` `/vst` compat | nothing |

Response headers only — status, body, routing and timeouts untouched, so a client
that works today keeps working.

The implementation detail worth a look: the rules key on a **transaction variable
set at request time**, not on the path at response time. By the time a response is
processed the backend's `replace-path` has already run, so
`/alert-bridge/api/v1/alerts` reads as `/api/v1/alerts` — a path-keyed
`http-response` rule would miss the prefix it meant to match and stamp the
agent's `/api` route instead. That is measured, not theorised: on the live
deployment `/alert-bridge/api/v1/alerts` is correctly flagged and `/api/health`
correctly silent.

`Sunset` is left to `VSS_GATEWAY_LEGACY_SUNSET` and **absent by default**. The
header carries an HTTP-date, 3.4.0 has no release date, and clients automate
against it — a guessed date is worse than silence. Compose passes the variable
through with no value rather than `${VAR:-}`, which would define it as an empty
string and emit a blank `Sunset`; the same empty-variable shape #1867 fixes.

`/storage` and the `host:port` `/vst` compat route stay silent deliberately. Both
rewrite into VST's namespace, but neither is one of the three prefixes being
retired: they repair media URLs the product already emitted, with no removal
planned. Announcing a sunset for them would be a promise nobody made.

### Two blind spots the live run exposed

The third commit documents both. Neither is a defect; both change how a soak
number should be read.

**HAProxy-synthesized `503`s carry no header.** When no backend is available
HAProxy generates the response itself, so the `http-response` rule never runs.
Measured on the alerts profile, where LVS is not deployed: `/lvs/v1/live` returned
a bare `503` with no `Deprecation` and no `Link`. So **a legacy-prefix call to a
down or undeployed service is invisible to header-based measurement** — absence of
the header is not evidence of absence of legacy traffic, and a clean soak report
could otherwise be read as "unused" when the real answer is "broken".

Deliberately not fixed here. `http-after-response` would reach synthesized
responses but also stamps error pages, and whether that trade is worth making is
a decision nobody has made. Documented rather than closed badly.

**The CLI is itself a significant legacy-prefix caller.** Every media URL `vss`
mints is on `/vst/storage`, so those responses carry `Deprecation: true`. That
follows directly from leaving `vss configure` on the stable prefixes while the
aliases are unsoaked, so it is not a defect — but a large share of the legacy
traffic in the window will be self-generated and has to come off the totals
before they say anything about third-party callers.

## Helm parity

Aliases added across all seven charts that serve the corresponding route today —
`dev-profile-{base,alerts,lvs,search}` and the three warehouse apps — verified by
rendering each chart and diffing the rendered path list against what the example
YAMLs claim.

The `/vios` → `/vst` rewrite **is** expressible in Kubernetes, and the reason is
that the warehouse charts already carry a `^/storage/(.*)` rewrite in a
`haproxy.org/path-rewrite` annotation. That existing precedent is what made the
alias expressible rather than a Docker-only behaviour, which would have been the
one outcome worth omitting the Helm alias over.

**The deprecation header is not expressible per-path.** The HAProxy ingress
controller sets response headers per-Ingress, and every chart here serves a
backend's legacy and alias paths from the **same** Ingress object, so an
annotation would stamp the successor prefix too — the one thing the signal must
not do. Splitting each Ingress in two for a telemetry nicety is not worth the
change to the rendered object graph. Docker gets the header; Kubernetes gets the
same routing without it, and the window is documented rather than advertised at
the edge. Called out here rather than left as a silent divergence.

## Validation against a real deployment

Validated on full `dev-profile-alerts` (22 containers) and `dev-profile-lvs`
(19 containers) on a remote GPU box, testing `77c42ae8f`.

**`/vios` against real VST — byte-identical on every pair.** `/vst/api/v1/sensor/list`
and `/vios/api/v1/sensor/list` both 200, 864 bytes, matching SHA, with a real
registered clip in the listing rather than an empty array. Same for
`/api/v1/live/streams` (568 bytes), `/api/v1/sensor/version` (49 bytes,
`{"type":"vst","version":"1.4.0-26.08.1"}`), the root HTML (1092 bytes), and error
parity on 404 (30 bytes) and 400 (97 bytes).

**The rewrite is proven to be a rewrite and not a strip**, which is the claim that
most needed real VST to settle: the bare path a strip would have landed on returns
**404**, while `/vios/api/v1/sensor/version` returns **200**.

**Media parity on real bytes.** A 76,141-byte JPEG snapshot was byte-identical
across `/vst/storage`, `/vios/storage` and `/storage`.

**The strongest single result: LVS fetched its source media through
`/vios/storage/…` during a real summarization.** A real backend consumed a real
alias URL as part of real work — not a curl, and not a stand-in.

**Deprecation headers on real backends.** All three legacy prefixes emitted
`Deprecation: true` with the correct `Link` successor; all three aliases and
`/va-mcp`, `/elasticsearch`, `/phoenix`, `/rtvi-vlm` and `/kibana` emitted
nothing. `Sunset` absent as designed. The documented `/vst/storage`
`HEAD`/`OPTIONS` exception held exactly as written.

**Still untested:** the `timeout server 3600s` override is config-verified against
the 120 s global default, but the longest real summarization was 82 s, so no
request crossed the line the override exists for. `Sunset` needs
`VSS_GATEWAY_LEGACY_SUNSET`, which no profile sets. Kubernetes and Helm are
verified by rendering only — no cluster was involved.

## Notes for the reviewer

**The exact alias roots are `$`-anchored, and that is the one thing in this PR a
reviewer with cluster access should verify.** `pathType: Prefix` admits a bare
`/vios` and `/alerts`, but the rewrite forms — `^/vios/(.*)`, `/alerts/(.*)` —
require the trailing slash, so those roots reached their backends still carrying
the alias prefix: `/vios` arrived at VST as `/vios` rather than `/vst`.
`4cff8cbce` adds `^/vios$ /vst` to all seven templates and the three example
YAMLs, `^/alerts$ /` to the alerts chart, and tightens the alias's own
`/video-summarization /` to `^/video-summarization$ /`. The Docker template in
this PR already carried an anchored exact-root rule for every mount, so this was
a Docker/Helm parity gap rather than new behaviour.

The anchor is load-bearing rather than stylistic, and it diverges from the
existing unanchored `/phoenix /` convention in these same annotations
deliberately: unanchored, `/alerts /` would also rewrite alert-bridge's own
`/api/v1/alerts` once the strip rule above it had run.

**What needs a cluster:** that the HAProxy ingress controller accepts `$` in a
`path-rewrite` value, and that `GET /vios` with no trailing slash returns what
`GET /vst` returns. `helm template` settles neither — the annotation is an opaque
string to it, so the anchored rules render identically whether or not the
controller honours the anchor. If it does not, the failure appears only on a real
deploy.

**The charts still carry no exact-root rule for the legacy `/alert-bridge`,
`/video-analytics-api` or `/va-mcp` prefixes**, although the Docker template does
(`^/alert-bridge$ /`, `^/lvs$ /`). Pre-existing on `develop` and out of scope
here — this change fixes the roots of the aliases it introduces, not every
prefix that predates it — but it is the same defect class and wants a follow-up
ticket.

**The `content-type: image/jpeg,image/jpeg` duplication on storage responses is
pre-existing VST/nginx behaviour**, not caused by this change. Checked with
HAProxy entirely out of the path and it reproduces. Flagging it because it is the
kind of oddity that looks like a routing bug when you first see it on an aliased
URL.

The example YAMLs (`vss-ingress-example*.yaml`) were **already substantially out
of sync** with their templates before this change. I only added what each template
actually renders, matching its gating — an example that overstates the routing is
worse than one that understates it. Reconciling the rest is a question about what
each profile is supposed to serve, not a docs fix, so it is left alone.

`dev-profile-base/vss-ingress-example.yaml` also contains a one-character fix:
`name:<RELEASE_NAME>-…` was missing the space after the colon, making the file
invalid YAML. Pre-existing on `develop`.

`vss configure` and the charts' `NOTES.txt` install output deliberately still
advertise the **old** prefixes. Pointing them at an unsoaked alias would couple
every CLI invocation and every install message to a route with no field exposure,
for no gain, since both resolve to the same backend. Moving them is a follow-up
that wants its own change so a regression there is attributable.


hugoverjus:
Confirming the Greptile P1 on the alias roots was **valid** — it was a real defect in this PR's own new code, not a false positive.

**Mechanism.** `pathType: Prefix` matches a bare `/vios`, but the rewrite `^/vios/(.*) /vst/\1` requires the trailing slash. So `/vios` with no trailing slash was routed to the VST backend and then never rewritten, arriving as `/vios` rather than `/vst` — diverging from the legacy `/vst` root the alias exists to shadow. `/alerts` had the same shape.

**Fix: `4cff8cbce`, 11 files.** `^/vios$ /vst` added to all seven chart templates and the three reference example YAMLs; `^/alerts$ /` added to the alerts template and its rewrites example; the alias's own `/video-summarization /` tightened to `^/video-summarization$ /` for consistency with the other two.

Worth recording that **the Docker HAProxy template in this same PR already carried an anchored exact-root rule for every mount** — `^/vios$ /vst`, `^/alerts$ /`, `^/video-summarization$ /`, alongside the legacy `^/alert-bridge$ /` and `^/lvs$ /`. So this was a Docker/Helm parity gap rather than a new behaviour: the Kubernetes side had simply not been brought in line with the contract the Docker side already implemented.

**On the `$` anchor.** It was chosen deliberately over the repo's existing unanchored convention in these same annotations (`/phoenix /`). Unanchored, `/alerts /` would also have rewritten alert-bridge's own `/api/v1/alerts` once the strip rule above it had run, so the anchor is load-bearing rather than stylistic.

One caveat, now recorded in the PR description: `helm template` cannot confirm that the HAProxy ingress controller accepts `$` in a `path-rewrite` value, since the annotation is an opaque string to it. That wants a check on a real cluster.

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33099438355)

hugoverjus:
## Blocked on #1867 merge, not on another CI rerun

`Assemble release set` / `Wait for Build Dev Images` failed because GHCR `vss-agent:tree-<content-sha>` is a **shared mutable alias**. This PR and #1867 pin different release-set digests for the same tree tag. A rerun of [33099438249](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33099438249) reproduced the **identical** digest mismatch with **no** concurrent #1867 build — not a stale race.

Plan: merge #1867 first, then re-run Build Dev Images here. Helm Sync and the Python/UI test jobs on this PR are already green.

hugoverjus:
## Re-proof at this tip (`4cff8cbce`) — PASS

Repeat GPU run on 2× RTX PRO 6000 (`dev-profile-alerts` + `dev-profile-lvs`) against **this SHA**, not only the earlier `77c42ae8f` tree.

No gateway-contract regression: `/vios` vs `/vst` still byte-identical with a real clip; JPEG/MP4 identical on `/vst/storage`, `/vios/storage`, `/storage`; LVS `summarize run` fetched `http://vss.local:7777/vios/storage/…`; agent MCP + `POST /chat` 200; `/alerts` and `/video-summarization` matched their legacy prefixes; deprecation headers as specified; ES `_nodes`/`_cluster` 403; Kibana still on `elasticsearch:9200`; host `vss` CLI worked.

Still not load-forced: summarize >120s. Still not live Kubernetes (Helm developer charts render the exact alias roots; warehouse still uses `path: /vios/` with a trailing slash). Search `/rtvi-embed` not in these profiles.

Host cleaned to 0 containers/images/volumes afterward.

hugoverjus:
## Review fixes: five defects closed

An adversarial re-read of the previously-pushed work turned up five real defects. The five commits below fix them; `e9a81cd17` is the remote-agent gateway work this branch also needed.

### 1. The `/alerts` alias broke `/alert-bridge` subpaths on Kubernetes — `dd0390898`

Rules in a `haproxy.org/path-rewrite` annotation apply **in order**, each to whatever the path has become by then, and `replace-path` replaces the *whole* path whenever its regex matches anywhere in it. The unanchored `/alerts/(.*) -> /\1` rule therefore fired a second time on the output of the `/alert-bridge` prefix strip, because alert-bridge's own surface contains a literal `alerts` segment.

Net effect: `/alert-bridge/api/v1/alerts/health` arrived at the backend as `/health` on Kubernetes, while Docker delivered `/api/v1/alerts/health` — the two edges disagreed. An earlier commit had anchored the *exact-root* rule for exactly this reason but missed the prefix-strip rule that actually collides. Both are now anchored, and a missing `^/alert-bridge$ -> /` was added.

Verified across 15 request shapes: backend paths are now byte-identical on both edges.

### 2. An empty `VSS_GATEWAY_LEGACY_SUNSET` prevented HAProxy from starting — `38f9a1ca9`

`.if defined()` is true for a variable that exists but is empty, and an empty `${...}` inside a quoted HAProxy argument aborts config parsing rather than expanding to a blank header. So writing `VSS_GATEWAY_LEGACY_SUNSET=` in an env file produced a **dead gateway** — i.e. every service-to-service HTTP call in the deployment — from what looks like a harmless edit.

The value is now read at runtime via `env()` and gated on length, so the parser never sees it. Verified for unset, empty, and a real HTTP-date.

### 3. `Deprecation: true` is not valid RFC 9745 — `073babc41`

RFC 9745 defines the field as a Structured Field whose value must be an RFC 9651 Date (`@<epoch-seconds>`); `true` is the syntax of the withdrawn earlier draft. Structured-field parsers discard a field they cannot parse, so the signal was invisible to precisely the conformant clients a soak measurement exists to hear from.

Now emits `@1787702400` (2026-08-26T00:00:00Z), the date the aliases shipped.

### 4. The new endpoint lint silently skipped every profile `.env` — `efaa4758e`

`Path(".env").suffix` is `""`, not `".env"`, so the suffix filter excluded the exact file class the guard was added to protect. Demonstrated by re-introducing the `LVS_BACKEND_URL=http://lvs-server:38111` line that this same PR had just fixed and watching the lint exit 0.

Coverage went from 24 to 30 files, and a test now asserts `.env` is in the default path set.

### 5. The remote-agent env example's Brev override block was positionally dead — `449f62ece`

Compose resolves an env file in order and does not re-resolve earlier lines, so the commented override at the bottom of the file produced a *split* origin: the browser-facing report URL landed on the one origin unreachable from outside a Brev environment. The origin is now assigned once, on a line the operator edits in place.

### Two things deliberately left alone

- **The Elasticsearch percent-encoding ACL bypass does not work.** Investigated against a real Elasticsearch: it matches route segments literally and case-sensitively, so `%5Fnodes` becomes an invalid index name and is rejected with 400. The ACL is sound as written and is unchanged.
- **The `http-request`-after-`use_backend` placement is safe** despite the parse-time warnings.

hugoverjus:
The hardened VIA-2713 integration tests exposed a second-order HAProxy rewrite collision beyond the alert-bridge case fixed in `dd0390898`: every `haproxy.org/path-rewrite` rule is applied sequentially to the current path, so an unanchored later source can match a service-name segment left inside an earlier rule's output (for example, `/lvs/v1/video-summarization/jobs` collapsed to `/jobs`).

Commit `655dd4b52` anchors all prefix and exact-root source patterns in multi-rule Helm rewrite annotations across the base, LVS, alerts, search, and three warehouse profiles, including checked-in example manifests. Replacement targets and rewrite-vs-strip behavior are unchanged. It also adds an always-run static guard over all Helm rewrite annotations plus sequential semantic regressions for the LVS and alert-bridge/VA-MCP paths.

Verification: focused rewrite tests (3/3), `helm lint` and `helm template` for all seven affected charts, the existing Helm release-channel policy tests, `git diff --check`, Python compilation, and pre-commit checks all pass.

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33404951075)

hugoverjus:
Superseded by #1983, which consolidates this PR and its stacked counterpart onto `develop` so the gateway contract can be reviewed and tested as one change. Nothing is dropped; #1983 also carries the fixes found while testing the combined branch on a 2x RTX 6000 PRO host, including the off-host agent proof that was the remaining acceptance gap.

Closing in favour of #1983.

PR #93, opened 2026-04-18
Fix #81: [DOC]: Local Gradio example script

JiwaniZakir:
Closes #81

## Description
Adds a new **Local GPU Deployment** section to `README.md` (after the Brev Launchable section, before the existing Docker Compose section) documenting how to run the full VSS stack on-premises using local NIM containers instead of remote API keys.

Specifically documents the three `.env` variables users must set in `deployments/developer-workflow/dev-profile-base/.env`:
- `LLM_MODE=local_shared` — serves the LLM locally on GPU
- `VLM_MODE=local_shared` — serves the VLM locally on GPU
- `HARDWARE_PROFILE` — selects the validated GPU topology (H100, L40S, RTXPRO6000BW, DGX-SPARK, IGX-THOR, AGX-THOR, OTHER)

Also clarifies that an NVIDIA AI Enterprise developer licence and NGC API key are still required to pull NIM containers, with links to the GPU requirements and quickstart docs.

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/HEAD/CONTRIBUTING.md).
- [ ] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

---
*This PR was created with AI assistance (Claude). The changes were reviewed by quality gates and a critic model before submission.*

PR #1877, opened 2026-08-26
refactor(ci): turn License Diff into an OSRB Scan dependency pipeline

zac-wang-nv:
## What this changes

`License Diff` becomes **`OSRB Scan`**, and grows from a Python/npm lockfile differ into a
dependency scanner that covers every way this repo actually acquires third-party code.

The change was prompted by auditing the OSRB dependency spreadsheet against `origin/develop`.
That audit found the sheet covered 1,943 packages across 9 modules; the repo contains ~6,900
across 36. The gap was not sloppiness — it was that the tooling could not see most of it.

### Three concrete defects this fixes

**1. Packages named `http*` are silently dropped.** `parse_requirements()` skips a line when
`line.startswith(("git+", "http", "file:"))`. That guard is meant for `http://` URL installs,
but it also matches *package names*. `parse_pyproject()` delegates PEP 621
`[project].dependencies` to the same function and inherits it. Poetry-style tables are
unaffected, which is why this went unnoticed.

```
>>> parse_requirements(b"httpx>=0.27.0")     -> {}          # dropped
>>> parse_requirements(b"requests==2.31.0")  -> {"requests": "2.31.0"}
```

Invisible to OSRB on `develop` today: `services/alert/requirements.txt` `httpx>=0.27.0`;
`services/rtvi/rt-embed/docker/py_deps/requirements.txt` `httpcore==1.0.9` + `httpx==0.28.1`;
`skills/vss-manage-alerts/scripts/alert-notify/requirements.txt` `httpx>=0.28.0`;
`services/agent/packages/vss_core/pyproject.toml` `httpx~=0.28`. Also affects `httplib2` and
`http-parser`.

**2. A new dependency file can be added with no row and no warning.**
`looks_like_manifest()` — whose docstring says "a new lock format cannot stay invisible
PR-by-PR" — only returns `True` for `pyproject.toml`, `*.lock` and `requirements*.txt`. And it
emits `::warning`, never a failure. Measured across 35 dependency-file shapes:

| | n | before |
|---|---|---|
| analysed | 7 | resolved into the CSV |
| warned only | 3 | `Cargo.lock`, `yarn.lock`, `Gemfile.lock` — advisory |
| silent | 25 | `package.json`, `go.mod`, `pom.xml`, `build.gradle`, `Gemfile`, `conanfile`, `vcpkg.json`, `environment.yml`, `Dockerfile`, `Chart.yaml`, compose, CMake… |

Not hypothetical: `tools/logstash-plugins/input/redis-stream/build.gradle`, 61 `Chart.yaml`
and 64 compose files already sit in the silent bucket.

Now: any added dependency-bearing file the scanner cannot parse emits `UNCOVERED_SOURCE` and
**fails the job**. Silence is no longer a possible outcome.

**3. Whole classes of dependency were never inventoried.** Nothing read Dockerfiles, compose
`image:` refs, or Helm `dependencies:`. That is where the audit found AGPL-3.0 (grafana,
minio), Elastic-2.0 (elasticsearch, kibana, logstash), Confluent Community (cp-kafka), and the
GPL-2.0-or-later codec set (libx264, libx265, libxvidcore, liba52, libdca). The spreadsheet
contained zero AGPL/SSPL/Elastic/BUSL rows for exactly this reason.

## New capability

- **Ecosystems** — npm/yarn/pnpm, Go, Cargo, Maven, Gradle, Ruby, conan, vcpkg, conda,
  `setup.py`/`setup.cfg`. `setup.py` is read with `ast.parse` and never executed; a
  non-literal `install_requires` yields `UNCOVERED_SOURCE` rather than a guess.
- **Non-manifest sources** — Dockerfile `FROM`/apt/pip/`git clone`, compose `image:`, Helm
  `dependencies:`, CMake `find_package`/`FetchContent`, pre-commit hook repos, Action pins.
- **Use-side pass** (`USED_UNDECLARED`) — third-party code the source imports that no manifest
  declares. Python, JS/TS, C/C++, Java/Ruby. **Report-only: it can never fail the job.**
  It reproduces two known-true gaps: `import gi` → PyGObject (LGPL-2.1-or-later) in
  `rt-embed` and `rt-vlm`, declared nowhere; and undeclared runtime deps in
  `services/video-summarization`.
- **Risk column** — `None`/`Medium`/`High`/`Unknown`, conservative on composites
  (`MIT AND GPL-2.0-or-later` → `High`).

## Compatibility — deliberately additive

The private GitLab OSRB pipeline consumes the uploaded artifact and CSV by name, and that
consumer is not visible from this repo. So:

- artifact stays `license-diff`; file stays `license-diff.csv`
- the 9 existing CSV columns keep their names **and order**; `source_kind`, `source_file`,
  `module`, `risk` are **appended**
- `GITHUB_LICENSE_RUN_ID` / `GITHUB_LICENSE_RUN_URL` and the
  `DOWNSTREAM_EXTRA_VARIABLES_JSON` keys are unchanged
- `review_rows` keeps its exact prior meaning; `uncovered_rows` is new and separate
- the PR-comment marker `<!-- license-diff-osrb -->` is unchanged, so existing comments update
  in place instead of duplicating

`test_osrb_dispatch.py` now asserts the artifact name, the CSV name, and that the
`workflow_run` trigger name matches the scan workflow's `name:` — so the two silent-breakage
modes fail pre-merge instead of in production.

## Operator actions

See `.github/OSRB_SCAN_MIGRATION.md`. The one that matters:

**`workflow_run` loads `osrb-review.yml` from the default branch, which is `main`, while PRs
target `develop`.** `origin/main` currently reads `workflows: ["License Diff"]`. In any window
where `main` still says that while PR mirrors emit `OSRB Scan`, OSRB Review never triggers, no
check run is created, and `workflow_run` jobs are not listed on PRs — so nothing reports the
absence. Promote to `main` promptly and treat PRs merged in that window as un-reviewed.

**What is *not* an action:** the check rename needs no branch-protection change. Verified via
the rulesets API — ruleset 15323753 (`Protect develop`) requires DCO Sign-off, Check folder
structure, Copyright Headers, Security Scan, and three SonarQube scans. Neither
`License Diff (OSRB CSV)` nor `OSRB Review` was ever a required check. What actually enforces
OSRB is `require_code_owner_review: true` plus `.github/CODEOWNERS`.

## Known gaps, stated plainly

- **CODEOWNERS is not widened here.** `pom.xml`, `build.gradle*`, `*.gemspec`, `conanfile*`,
  `vcpkg.json`, `environment.yml`, `Chart.yaml`, compose files and `CMakeLists.txt` are now
  *reported* but still not *blocked*. That is a separate small PR and should follow this one.
- **The use-side pass is heuristic.** Import-name → distribution-name is a lookup table; a miss
  produces a false `USED_UNDECLARED`. That is why it is report-only. Flip it to blocking once
  the false-positive rate is measured on real PRs.
- **C/C++ coverage is deliberately narrow.** Only `#include`s resolving inside known vendored
  trees are reported. A bare `<vector>` or a first-party header yields nothing. This trades
  recall for precision on purpose — a noisy gate is an ignored gate.
- **Base-image contents are not expanded.** A `FROM` line is one row; the OS packages inside
  that image are not enumerated. That needs an image SBOM scan (syft/trivy), not this scanner.
- **Pre-existing, now more likely to bite:** the PR-comment step inlines the entire CSV. A real
  523-row run is ~92 KB against GitHub's 65,536-character comment limit, so the `gh api` POST
  422s and fails the job. It fails red rather than open, but it needs capping — flagged here
  rather than fixed, to keep this diff reviewable.

## Plan

Per the workspace coding guidelines, the design questions asked before building, the
recommendation given, and the decision taken.

**1. One PR or a stack?**
*Recommended:* a stack of four — fail-loud + `http` fix first (~150 lines, lands in a day),
then ecosystems, then non-manifest sources, then the use-side pass. Each independently
revertable, matching "grow the system in layers."
*Decision:* **one PR, full refactor.**
*Consequence, recorded honestly:* this is a large diff over a gate that fails silently when
broken. The mitigation is the assertion added to `test_osrb_dispatch.py` covering the two
silent-breakage modes, and the operator note.

**2. Should the use-side (import) pass block a PR?**
*Recommended:* report-only at first — emit `USED_UNDECLARED` rows, never fail; tune the
mapping table against real PRs, then flip to blocking once the false-positive rate is known.
*Decision:* **report-only.** Enforced structurally: use-side rows cannot reach the failure
count.

**3. Which languages for the use-side pass?**
*Options:* Python, JS/TS, C/C++, Java/Ruby — with C/C++ flagged as highest effort and highest
false-positive risk.
*Decision:* **all four.**
*Consequence:* C/C++ is narrowed to includes resolving inside known vendored trees. This is
stated as a limitation rather than presented as C/C++ coverage.

**4. `OSRB Review` is already taken by `osrb-review.yml`. How should the boundary work?**
*Recommended:* rename `License Diff` → `OSRB Scan`; `osrb-review.yml` keeps its name and keeps
adjudicating. Scan gathers evidence, review decides.
*Decision:* **as recommended.**
*Note:* the trust boundary is preserved — the scan runs from the PR mirror, the review runs
from the default branch, precisely so PR code cannot alter its own adjudication.

### Verification performed before opening this PR

- Every new parser run against every real matching file in the tree (61 `Chart.yaml`, 64
  compose files, 51 Dockerfiles, 16 workflows, 1 `build.gradle`, the pre-commit config).
- The 34 `Chart.yaml` with no `dependencies:` block confirmed to yield zero rows, zero errors.
- Use-side pass run over the real tree (1,761 `.py`, 298 `.tsx`, 223 `.ts`, 173 `.js`) with a
  hand-sampled false-positive rate reported rather than estimated.
- Adversarial review across three lenses — gate integrity, parser correctness, use-side
  precision — plus an independent review by the OpenAI Codex CLI against the workspace
  coding guidelines.

---

## Update: state comparison against the OSRB-approved set

The change above is a **delta** pipeline — it diffs base-ref against head-ref, so it only ever
sees what a PR changes and is blind to drift already in the tree. OSRB approves a **set**, so
the repo has to be compared against that set. That comparison is now here too, and one thing
turned out to matter more than expected: **OSRB approves a dependency FOR A USE.** The approved
baseline carries `distribution_method` and `usage_method`, so the same library dynamically
linked is a different approval from the same library vendored and modified.

### The approval record was not one file

The consolidated OSRB spreadsheet is not complete. It was built from 14 upstream sheets, and
approvals granted only in a bug comment never reached it — comment #1 requests approval for
seven packages in the Alerts Microservice and comment #2 grants it, yet that module had no row
in the sheet at all. Reading all 117 comments recovered **62 approvals, 23 conditions and 2
refusals**, now committed so the repo is self-contained.

| file | what it is |
|---|---|
| `.github/osrb/approved.csv` | 3877 sheet rows + 62 recovered from bug comments, each with `provenance` and the comment number as `evidence` |
| `.github/osrb/conditions.csv` | 2 refusals + 23 conditional approvals. Separate from approvals on purpose — a refusal is not an approval |
| `.github/osrb/inventory.csv` | the repo's own dependency state, regenerated deterministically (byte-identical across runs) |
| `.github/osrb/module_map.py` | OSRB labels → repo paths, path aliases, and the submitted-but-no-package-list set |

### Three states, not two

Collapsing these misleads a reader in opposite directions:

- **never submitted** — file a new OSRB bug (10 modules)
- **submitted, package list not recovered** — recover the list, do **not** file a new bug
- **submitted under a different path** — nothing to do; the tree moved after submission. This
  alone accounted for 895 false "no OSRB record" rows.

### Permissive green-gate

Most of this tree is MIT/Apache/BSD/ISC, and reporting those as unapproved buried the rows
carrying real risk. A package whose licence matches the repo's **own** list —
`PERMISSIVE_LICENSE_PATTERNS` in `.github/scripts/check_python_licenses.py`, already enforced
by the `check_python_licenses.sh` pre-commit hook — is `PERMISSIVE_AUTOCLEARED` and kept out of
the review comment. Reusing that list rather than copying it stops the two gates drifting apart
on the same package.

It only downgrades verdicts meaning "no approval row found" — it cannot touch `VERSION_DRIFT`
or `LICENSE_DRIFT`, which are disagreements with an approval that exists. Composites
(`MIT AND GPL-2.0-or-later`) and `UNKNOWN` are refused: exact match only. `license_denylist.txt`
wins. A failed import clears **nothing**, so the report gets noisier, never quieter.

| | before | now |
|---|---|---|
| rows needing human review | 2878 | **616** |
| permissive rows in that set | many | **0** |
| `MODULE_UNSUBMITTED` | 2359 | 168 |

546 of the remaining 616 are `UNKNOWN` licence — a resolution gap, not a risk finding, and the
largest remaining lever.

## What a reviewer should look at first

**12 shipped packages carry an OSRB refusal or condition.**

- `mkl` and `tbb` in `rt-embed` and `rt-vlm` — *"NOT cleared by OSRB, requires BU attorney
  confirmation"* (Intel Simplified Software License, not OSI-approved)
- **Two live refusals** (comments #93/#95): the RTVI VLM and RTVI Embed source batches were
  refused because *"the source still advertises proprietary licensing in the service
  metadata"*. The file headers are now Apache-2.0, but `gsttimestampfilter.cpp` in **both**
  services still declares `"Proprietary"` in `GST_PLUGIN_DEFINE`, and
  `libgsttimestampfilter.so` is installed into `/usr/lib/*/gstreamer-1.0/` by both Dockerfiles.
  It ships.
- `ffmpeg` in `services/agent` and `services/vios` — "NOT approved in this comment"

Those refusals needed a module-level reporting path: their `package` field is a prose batch
name, so package-level matching found nothing and the most serious finding in the comparison
would have been silently absent.

## Still report-only, deliberately

A ratchet was built here and removed after review found it unsound four ways — a cross-job
`steps.ctx` reference that resolved to `""` (and `git show ":path"` reads the *index* rather
than failing), a baseline computed from the PR's own `approved.csv`, count-netting that let a
new GPL-3.0 package through, and an ungated catch-all for unknown modules. A gate that cannot
fail is worse than no gate. The **delta** gate still blocks what a PR introduces; the state
comparison reports. Turning it on needs identity-based comparison, which is a follow-up.

Everything committed is scrubbed of reviewer names, ticket IDs and internal URLs, asserted by
the generator — which caught two real leaks the first pass missed.


zac-wang-nv:
> The source inventory still deduplicates rows without the module field before passing them to a module-aware diff, causing a dependency newly used by another module to disappear from the compliance output.

Confirmed and fixed in `281953e7`. This was a real bug, and it's the one my own review passes missed.

**Measured on this tree** rather than reasoned about: `inventory_at_ref` kept **363** rows against **450** module-aware rows — **87 rows (19%) silently dropped**, across **54 packages used by more than one module**:

| package | modules |
|---|---|
| `wheel` | 4 |
| `pipenv` | 3 |
| `dpkg-dev` | 3 |
| `curl`, `gnupg`, `eclipse-mosquitto`, `make`, `NVIDIA/DeepStream` | 2 each |

`_row_key` is now `(source_kind, language, package, new_version, module)`. `inventory.csv` goes 3904 → 3991, exactly the recovered rows.

Your point about the module-aware consumer is the sharper half: because `osrb_scan.diff_source_rows` keys on module, a dependency newly adopted by a *second* module produced no diff row at all — it read as unchanged. That's the failure mode that would actually have reached OSRB as a clean bill of health.

**One thing worth flagging for reviewers:** an existing test asserted the bug as correct behaviour. `test_missing_blob_is_normal_and_duplicates_collapse` used `a/Dockerfile` and `b/Dockerfile` — different modules — and required them to collapse to one row. So the defect was load-bearing in the suite and a regression test alone wouldn't have caught it. Rewritten to assert collapsing *within* a module, plus a counterpart asserting two modules pulling one image stay two rows.

For context, this is the third instance of the same mistake in this PR — `_inventory_at_ref` and `requirements_inventory` both had component-dropping keys and were fixed earlier (the latter was inflating the use-side false-positive rate from 11.5% to 52%). Worth a reviewer eye on any other dedup key here.

PR #1878, opened 2026-08-27
fix(skill-eval): schedule on catalog GPU counts and survive a refused box

zac-wang-nv:
## Summary

A leg whose spec needs two GPUs could be locked onto a one-GPU box and then killed by the very check that caught the mistake. Two independent causes, both fixed here.

Seen on #1863: `IN-3` died in 14s on `vss-eval-rtx-1g-1`, and again on the next head; `AT-1` and `IN-1` were lost the same way. **Three independent eval agents diagnosed the gate in their own traces** before exiting, e.g.

> *"the wrapper's bug is that it doesn't filter managed (non-registered) instances by name-hint gpu_count"*
> *"line 661 gates the gpu_count hint filter on `inst.get("_registered")` … They match on gpu_type alone and then fail at `brev_env.py`'s stricter validation."*

#1744 describes the same failure and reports it fixed, so this regressed or never landed.

## What was wrong

**1. Selection did not know how big a managed box is.** `pool_candidates()` applied its `gpu_count` filter only to registered nodes:

```python
if inst.get("_registered") and required_count > 0:   # managed boxes skip this
```

so a managed `vss-eval-rtx-1g-*` matched on `gpu_type` alone and stayed a candidate for a `gpu_count: 2` spec. `brev ls --json` carries no `gpu_count` — which is why fleet naming became the stand-in — but the catalog does, and it is the same source `envs/brev_env.py` consults before falling back to live `nvidia-smi`.

**2. A refusal ended the leg.** `brev_env` raises when a box cannot satisfy the task, but selection and reservation are one atomic step, so the refusal arrived *after* the lock was held. The leg exited with a `RuntimeError` having never run a trial — indistinguishable, in the PR comment, from a skill that failed its checks.

## What this does

- Resolve managed boxes through `brev search gpu --json`, keyed on `instance_type`. Registered nodes keep the name hint: the operator controls those names, and it is the only signal they carry.
- **An unlisted SKU stays unknown, and unknown never disqualifies.** Names are not authoritative for managed boxes — a resized box keeps its old name — and a catalog outage that filtered the fleet to nothing would be a worse failure than the one being fixed. A transient `brev search` failure is not cached, so it blinds one round rather than the whole leg.
- Treat a refusal as what it is: this box is unusable, the task untouched. Drop it, take the next candidate, and stop when the pool is exhausted rather than waiting out a lock timeout for a box already ruled out. Each retry spends only what is left of the leg deadline.
- Recognise the refusal from Harbor's structured `exception_info` on a trial that recorded no `agent_execution`/`agent_result` — **never** from log or transcript text. Eval agents quote these very sentences while diagnosing the failure, so matching prose would discard a healthy box and hide a real result behind a retry.

## Review

Reviewed by the OpenAI Codex CLI across three rounds, told up front the diff was AI-generated. It rejected the first two revisions and its findings are why this one looks the way it does:

| Round | Finding | Outcome |
|---|---|---|
| 1 | Promoting the regex name hint to a hard fleet-wide gate contradicts the documented design (`brev_env` validates the pick; #929 made scheduling permissive) | Rewritten to use the catalog |
| 1 | Detector scanning prose would false-positive on agent transcripts | Rewritten to structured `exception_info` |
| 2 | **`work_deadline` is monotonic but the retry used `time.time()`** — collapsing every attempt, including the first, to a 1s floor | Fixed; the helper now reads the clock itself so a caller cannot pass the wrong one |
| 2 | Unreachable terminal raise; pinned-instance refusal waited out the full timeout; `LockTimeoutError` handler judged the wrong budget | All fixed |
| 2 | `agent_execution` falsiness admits `{}`/`[]` | Now also checks `agent_result` |

## Testing

`.github/skill-eval/tests/test_run_leg.py` — **69 pass**. New coverage: catalog-filtered managed box; unknown SKU kept; registered node still name-filtered; the four refusal-detector cases including *an agent quoting the error is not a refusal*; stale-refusal window; and the per-attempt budget including the monotonic-clock contract.

**Pre-existing and unrelated:** three `test_4090_*` failures assert `_rtx4090_supports` for `vss-deploy-profile`. They fail identically on a clean `develop` checkout — verified — and are untouched here.

## Checklist
- [x] Pre-commit hooks run locally.
- [x] Every commit is DCO sign-off'd.

### DIFF
--- a/.github/skill-eval/run_leg.py
+++ b/.github/skill-eval/run_leg.py
@@ -1106,6 +1106,7 @@ def run_command(cmd: list[str], env: dict[str, str], timeout_sec: int) -> int:
     proc: subprocess.Popen | None = None
     pgid: int | None = None
     pending_signal: int | None = None
+    harbor_rc: int | None = None
     cleanup_started = False
     previous_handlers: dict[signal.Signals, object] = {}
 
@@ -1171,6 +1172,11 @@ def run_command(cmd: list[str], env: dict[str, str], timeout_sec: int) -> int:
                         cleanup_started = True
                         return rc
                     cleanup_started = True
+                    # Harbor finished on its own; only its transports linger.
+                    # Remember the real result: if the bounded cleanup below
+                    # reaps them, the trial stands and this is housekeeping,
+                    # not a timeout.
+                    harbor_rc = rc
                     outcome = 124
                     reason = (
                         "Harbor exited while detached transport groups remained: "
@@ -1201,6 +1207,18 @@ def run_command(cmd: list[str], env: dict[str, str], timeout_sec: int) -> int:
                 "preserving primary outcome",
                 flush=True,
             )
+        elif harbor_rc is not None:
+            # Harbor completed and the strays are now gone. Reporting 124
+            # here discards a finished trial -- and because rc == 124 also
+            # writes skip markers, it takes every later step of a multi-step
+            # spec with it. A step that scored 1.0 was recorded as a timeout
+            # and its successors never ran.
+            print(
+                f"[run-leg] detached transports reaped; keeping Harbor's "
+                f"result rc={harbor_rc}",
+                flush=True,
+            )
+            return harbor_rc
         return outcome
     finally:
         for sig, previous in previous_handlers.items():
--- a/.github/skill-eval/run_leg.py
+++ b/.github/skill-eval/run_leg.py
@@ -619,6 +619,77 @@ def _loose_gpu_match(want: str, have: str) -> bool:
     return want_tokens.issubset(have_tokens) or want in have
 
 
+# How many refusals a leg absorbs before reporting the last one. The pool is
+# small and `remaining_candidates()` already ends the loop when every eligible
+# box has refused; this only bounds a fleet that refuses indefinitely.
+_MAX_BOX_REJECTIONS = 4
+
+_CATALOG_GPU_COUNTS: dict[str, int] | None = None
+
+
+def _catalog_gpu_counts() -> dict[str, int]:
+    """`{instance_type: gpu_count}` from `brev search gpu --json`, cached.
+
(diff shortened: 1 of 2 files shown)


zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33029793086)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33033004962)

hugoverjus (approved):

zac-wang-nv:
Related harness observation from PR #1863's runs, **not** included in this PR because I cannot verify the SDK semantics involved.

Two legs in run 33059638119 each lost a single check to a judge that produced no verdict:

```
judge returned no compliant verdict (saw_result_message=True is_error=True
subtype='success' stop_reason='stop_sequence' text_chars=5028 text_blocks=24
retry_attempted=False)
```

That took `combined` from 19/19 to 18/19 and IN-1's step 2 to 12/13 — both on long, complex checks, and in both cases the judge clearly did the work (thousands of characters, 20+ text blocks) and simply never emitted the verdict JSON.

`generic_judge.py` has a retry for exactly this — its comment describes the model wandering *"into 'I'll investigate...' prose ... where the model gathered the right evidence but never emitted the JSON"*. But the retry is gated on `not result_is_error`, and these responses carry `is_error=True` alongside `subtype='success'` and a `stop_sequence` stop reason. The gate's stated rationale names rate limits, content policy, tool-use aborts and max-turns exhaustion — none of which look like what happened here.

If `is_error=True` with `subtype='success'` is a benign flag on a completed turn, then the gate is excluding the very case the retry was built for, and legs are being failed on it. Worth a look by someone who knows the SDK's contract for that combination; I did not want to loosen an error gate on a guess.

PR #1643, opened 2026-08-11
feat(mv3dt): build BEV Fusion image in GHCR

harshalnishar:
Committed-By: hnishar's Codex Agent

Claude Review: Approved

## Description
Enable GitHub-first GHCR builds for the MV3DT BEV Fusion container, vendor metadata generation, and route Compose coordinates through the shared GHCR tag.

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/HEAD/CONTRIBUTING.md).
- [x] I have installed and run [pre-commit hooks](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/HEAD/CONTRIBUTING.md#1-enable-pre-commit-hooks) locally (`uv run pre-commit install` once, then hooks run on every `git commit`).
- [x] Every commit on this PR is [DCO sign-off](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/HEAD/CONTRIBUTING.md#developer-certificate-of-origin-dco)'d (`git commit -s` adds a `Signed-off-by` trailer that certifies you have the right to submit the change under Apache-2.0).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #1894, opened 2026-08-27
fix(ci): clean up PR tags carrying variant and platform suffixes

zac-wang-nv:
## Summary

`cleanup_pr_tags.py` owned only `pr-<N>-<sha12>` and `pr-<N>-latest`. Every candidate tag
carrying a suffix fell outside that pattern and was never deleted.

Measured across six GHCR packages — **929 of 1886 `pr-*` tags leak**:

| shape | count | example |
|---|---|---|
| owned today | 927 | `pr-1885-fc5a4cbe98bf` |
| `-sbsa` | **879** | `pr-1885-latest-sbsa` |
| `-amd64` | 30 | `pr-1885-fc5a4cbe98bf-amd64` |
| `-arm64` | 30 | |
| `-sbsa-arm64` | 20 | |

Spot-checked against PR state: #1623, #1634, #1643, #1672, #1682, #1696 and #1710 all merged
two weeks ago and their staging tags are still in GHCR.

## The two suffixes, and where they come from

**`tag_suffix`** from `container-inventory.json`. `container_build_plan.metadata()` appends it
to candidate, content **and** alias tags, so it follows `latest` as well as a SHA:

```python
tag=f"{tag}{tag_suffix}"            # pr-1885-latest-sbsa
content_tag=f"tree-{tree_sha}{tag_suffix}"
```

**The per-architecture staging tag.** For the 12 images with `native_platform_build: true`,
each arch builds on its own runner and pushes `<tag>-<arch>`, which `imagetools create` later
merges into the multiarch manifest:

```yaml
- name: Compute platform staging tag
  run: echo "tag=${{ steps.meta.outputs.tag }}-${{ matrix.arch }}" >> "$GITHUB_OUTPUT"
```

Both together give `pr-1885-<sha12>-sbsa-arm64`.

## Why they were kept rather than merely skipped

A staging tag is usually the **only** tag on its version. Because the pattern did not own it,
`is_deletable` classified it as a *foreign* tag and refused to delete — so the rule that
protects shared digests was the thing keeping these alive forever.

## Approach

Both suffix sets are read from the same inventory that already supplies the package list, so a
new variant or platform is covered the day it is added instead of leaking until someone
notices.

They are **enumerated, never wildcarded**. This pattern decides ownership, and ownership
decides deletion. Over-claiming could delete a live `develop` image — unrecoverable without a
rebuild — while under-claiming only leaks storage. The costs are not symmetric, so the pattern
stays narrow:

```
^pr-1885-(?:[0-9a-f]{12}|latest)(?:-(?:sbsa))?(?:-(?:amd64|arm64))?$
```

Default arguments keep the previous behaviour, so a caller that passes no variants is never
made more aggressive by accident.

## Is deleting these safe?

Yes, for the images. The merged index references each platform manifest **by digest**:

```
sha256:3fe7ec14…   develop-7772b18705f0, tree-ef0350…      ← merged index
  ├── sha256:7b9a6960…   (no tags)                        ← amd64 image
  └── sha256:57f938ba…   (no tags)                        ← arm64 image

sha256:f4d063ee…   develop-7772b18705f0-amd64             ← staging wrapper
  └── sha256:7b9a6960…   (same amd64 image)
```

Removing a staging tag drops only the wrapper. The untagged child manifests stay alive because
the merged index depends on them.

One cosmetic caveat: the SLSA provenance attestation records
`subject.name = …@develop-<sha>-amd64`, frozen at staging time. Deleting the tag makes that
*name* a dangling reference; the digest beside it stays valid, which is what verification
actually uses.

## Verification

Run against the real 1894 `pr-*` tags currently in GHCR:

```
owned BEFORE (narrow)   :  932
owned AFTER  (variants) : 1894   (+962)
still unowned           :    0
```

Foreign shapes remain unowned — other PRs, `develop-*`, `tree-*`, unknown platforms, and prefix
overlaps such as `pr-12345` against `pr-1234`.

## Test plan

- [x] `python3 .github/scripts/test_cleanup_pr_tags.py` — 28 tests pass (21 existing + 7 new)
- [x] Widened pattern validated against the live GHCR tag set, both directions
- [x] `is_deletable` still keeps a version shared with a `develop-*` or `tree-*` tag
- [ ] After merge: close a PR touching a native-built image and confirm its `-amd64` / `-arm64`
      tags disappear alongside the plain candidate tag

## Checklist

### DIFF
--- a/.github/scripts/cleanup_pr_tags.py
+++ b/.github/scripts/cleanup_pr_tags.py
@@ -3,8 +3,11 @@
 # SPDX-License-Identifier: Apache-2.0
 """Delete a merged/closed PR's ``pr-<N>-*`` candidate images from GHCR.
 
-Candidate tags are build scaffolding: ``pr-<N>-<sha12>`` immutable candidates and
-the ``pr-<N>-latest`` alias. Once the PR closes they can never be rebuilt or
+Candidate tags are build scaffolding: ``pr-<N>-<sha12>`` immutable candidates,
+the ``pr-<N>-latest`` alias, and the two optional suffixes those carry — the
+inventory's ``tag_suffix`` variant marker (``pr-<N>-latest-sbsa``) and the
+per-architecture staging tag the native build publishes before merging the
+multiarch manifest (``pr-<N>-<sha12>-amd64``, ``…-sbsa-arm64``). Once the PR closes they can never be rebuilt or
 referenced, so retaining them only grows the package indefinitely. ``develop-*``
 tags are the opposite — they are the derivable coordinate set and are kept
 forever.
@@ -54,6 +57,7 @@ import sys
 import urllib.error
 import urllib.parse
 import urllib.request
+from collections.abc import Sequence
 from pathlib import Path
 from typing import Callable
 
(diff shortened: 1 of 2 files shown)


zac-wang-nv:
> the regex also recognizes platform suffixes on `latest` aliases that the build does not generate

Confirmed and fixed in `57fc4939`. You're right, and it contradicts the reasoning in my own commit message — I argued for enumerating these suffixes precisely so ownership cannot outrun what the build publishes, then wrote a pattern that outran it.

**Verified rather than assumed:** the moving alias is advanced onto the *merged* manifest, never onto a per-architecture staging image. Across the six native-built packages, GHCR holds **zero** `latest-<arch>` tags. `pr-<N>-latest-amd64` is not a shape this repository can produce.

Grammar is now split so the platform suffix follows a SHA candidate only:

```
^pr-<N>-(?:latest(?:-sbsa)?|[0-9a-f]{12}(?:-sbsa)?(?:-(?:amd64|arm64))?)$
```

The **variant** suffix still follows `latest` — `pr-1885-latest-sbsa` is real, and that class is 879 of the 929 leaked tags, so restricting it would have undone most of the fix.

**Re-validated against the live 1894 `pr-*` tags: still 1894 owned, 0 unowned.** The tightening costs no real tag; it only removes a shape that never existed.

Added `test_platform_suffix_only_follows_a_sha_candidate` so the grammar can't loosen again, with the reasoning and this PR referenced in the docstring. 29 tests pass.

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33107175396)

PR #1895, opened 2026-08-27
Feat/vss memory introspection

liamy-nv:
## Summary
- Added `vss memory introspect` to retrieve relevant memories, assess sufficiency, and run targeted VLM follow-ups when information is missing.
- Added `vss vlm run` as a reusable, persisted VLM job operating on VIOS sensor and time-window handles.
- Standardized `input.sensors[].id` on the human-readable VIOS sensor name while retaining internal identifiers in `info`.
- Added a bounded introspection workflow with one sufficiency check, up to three VLM calls, and one final answer synthesis.
- Introspection results are returned through stdout only, while internal VLM calls follow the configured persistence policy.
- Updated `vss-ask-video` to route requests between hot context, memory retrieval, memory introspection, and direct visual inspection.

Reference JIRA ticket: https://jirasw.nvidia.com/browse/VIA-2572

## Checklist

### DIFF
--- a/services/agent/packages/vss_cli/src/vss_cli/memory_cmd.py
+++ b/services/agent/packages/vss_cli/src/vss_cli/memory_cmd.py
@@ -323,7 +323,7 @@ def introspect_memory(
     group: str | None,
     pretty: bool,
 ) -> None:
-    """Answer one scoped question using memory and bounded VLM follow-ups."""
+    """Answer via the configured text judge and bounded RT-VLM follow-ups."""
     try:
         from vss_core.introspection import IntrospectionRequest
 
--- a/services/agent/packages/vss_cli/src/vss_cli/memory_cmd.py
+++ b/services/agent/packages/vss_cli/src/vss_cli/memory_cmd.py
@@ -11,6 +11,7 @@ from __future__ import annotations
 
 import asyncio
 import json
+import os
 import sys
 from typing import TYPE_CHECKING
 from typing import Any
@@ -133,17 +134,30 @@ async def _execute_introspection(request: IntrospectionRequest) -> tuple[Introsp
     from vss_core.introspection import introspect
 
     deployment = config_mod.load()
+    memory_config = deployment.memory
+    if memory_config is None or memory_config.introspection is None:
+        raise config_mod.ConfigError(
+            "memory introspection judge is not configured; run `vss configure memory introspection`"
+        )
+    judge_config = memory_config.introspection.judge
+    api_key: str | None = None
+    if judge_config.api_key_env is not None:
+        api_key = os.environ.get(judge_config.api_key_env, "")
+        if not api_key.strip():
+            raise config_mod.ConfigError(
+                f"introspection judge credential environment variable {judge_config.api_key_env!r} is missing or empty"
+            )
     memory = _memory(deployment)
     owns_memory = _TEST_MEMORY is None
     client: OpenAIIntrospectionClient | None = None
     try:
-        rt_vlm = deployment.services.get("rt_vlm")
-        if rt_vlm is None or not rt_vlm.url or not rt_vlm.models:
-            raise config_mod.ConfigError("the configured RT-VLM service reports no model")
         settings = IntrospectionSettings()
         client = OpenAIIntrospectionClient(
-            base_url=rt_vlm.url,
-            model=rt_vlm.models[0],
+            base_url=judge_config.endpoint,
+            model=judge_config.model,
+            backend_model=judge_config.backend_model,
+            api_key=api_key,
+            criteria_prompt=judge_config.criteria_prompt,
             settings=settings,
         )
         runner = IntrospectionVLMJobRunner(
--- a/services/agent/packages/vss_cli/src/vss_cli/memory_cmd.py
+++ b/services/agent/packages/vss_cli/src/vss_cli/memory_cmd.py
@@ -9,10 +9,12 @@ the underlying parent/child store across groups.
 
 from __future__ import annotations
 
+import asyncio
 import json
 import sys
 from typing import TYPE_CHECKING
 from typing import Any
+from typing import cast
 
 import click
 from pydantic import ValidationError
@@ -22,9 +24,17 @@ from . import memory as memory_mod
 from .exits import Exit
 
 if TYPE_CHECKING:
+    from collections.abc import Awaitable
+    from collections.abc import Callable
+
     from vss_cli.memory import Memory
+    from vss_core.introspection import IntrospectionRequest
+    from vss_core.introspection import IntrospectionResult
+    from vss_core.memory.models import MemoryGroup
+    from vss_core.memory.models import RecordType
 
 _TEST_MEMORY: Memory | None = None
+_TEST_INTROSPECT: Callable[[IntrospectionRequest], Awaitable[IntrospectionResult]] | None = None
 
 
 def set_test_memory(memory: Memory | None) -> None:
@@ -33,14 +43,29 @@ def set_test_memory(memory: Memory | None) -> None:
     _TEST_MEMORY = memory
 
 
-def _memory() -> Memory:
+def set_test_introspect(
+    function: Callable[[IntrospectionRequest], Awaitable[IntrospectionResult]] | None,
+) -> None:
+    """Inject the complete workflow boundary for hermetic CLI tests."""
+    global _TEST_INTROSPECT
+    _TEST_INTROSPECT = function
+
+
+def _memory(deployment: config_mod.Deployment | None = None) -> Memory:
     if _TEST_MEMORY is not None:
         return _TEST_MEMORY
-    return memory_mod.build(config_mod.load())
+    return memory_mod.build(deployment or config_mod.load())
 
 
 def _emit(value: Any, *, pretty: bool) -> None:
-    click.echo(json.dumps(value, indent=2 if pretty else None, default=str))
+    click.echo(
+        json.dumps(
+            value,
+            indent=2 if pretty else None,
+            separators=None if pretty else (",", ":"),
+            default=str,
+        )
+    )
 
 
 def _fail(prefix: str, error: BaseException, exit_code: Exit) -> None:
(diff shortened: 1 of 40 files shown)


liamy-nv:
**Known limitation:** `vss memory introspect` always uses `--query` as an Elasticsearch text filter, even when `--job-id` / `--record-id` already identify the row.

A paraphrase that is not in the stored wording can return `no_memory` (exit 5) even though `vss memory get` finds the record.

Example: event text says “yellow safety vest / hard hat”; query “What does event 1 say about PPE?” with that job/record id returned `no_memory`. `--sensor` recall is fine when the question overlaps stored text.

If we keep this: `--query` is both retrieval and the judge prompt; id-scoped asks should reuse stored wording. Follow-up: skip text match when identity uniquely selects the record.

liamy-nv:
/ok to test [854e851](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1895/commits/854e8519dff293ab6ee67d259126f6b79bfd9c32)

services/agent/packages/vss_core/src/vss_core/introspection/orchestrator.py
```diff
@@ -0,0 +1,354 @@
...
+    state.pending = list(valid_gaps)
+    for index, gap in enumerate(valid_gaps):
+        if index >= settings.max_vlm_queries:
+            state.unresolved.extend(
+                f"{_gap_label(remaining)}: not inspected because the VLM query cap is {settings.max_vlm_queries}"
+                for remaining in valid_gaps[index:]
+            )
+            break
+        try:
+            evidence = await vlm_runner.run(
```
maxervo: When you test, is it a bit more performant to run these introspective vlm queries concurrently? (`asyncio.gather` with max 3) 

RT-VLM could continuously batch the active requests. The speedup may be smaller during prefill because visual tokens can consume much of the per-iteration `VLLM_MAX_NUM_BATCHED_TOKENS` budget, but we may gain a bit more throughput during decoding. 

Hopefully, even if each concurrent query takes longer individually, all three may finish sooner overall. So just to check if there is total latency reduction and if concurrency is worth it with VLMs

skills/vss-ask-video/SKILL.md
```diff
@@ -1,749 +1,147 @@
...
-**One endpoint, one attempt.** Resolve the VLM endpoint once and send the request once. A `401`
-or `403` is an answer — the endpoint needs credentials you were not given — not a reason to try
-another host, port or auth header. Retry only to repair malformed structured output, and only
-once. Hunting for a combination that returns 200 turns one failed call into several and buries
-the actual problem, which is that the endpoint is gated.
+1. **Hot conversation context -> answer directly.** If current messages or
+   current-turn tool output already contain the answer, answer from that
+   evidence. Do not query memory or run a model.
+2. **Explicit stored summary/result -> `vss memory get` or `vss memory query`.**
+   For a known `job_id`, use `get`. To find or list stored results by text,
```
maxervo: Should we not rely on the harness native memory search (over markdown) first? And then mention it in the skill.

Meaning we would get this flow:
1. OK hot conversation -> answer directly
2. Let harness check with its native memory search tool over markdown (if harness does not have such capability, then can fall-back to `vss memory search`)
3. If markdown does not have answer, then `vss memory get` to get the full ES doc based on IDs
4. Harness decides if it can answer, or calls `vss memory introspect` for additional introspection
5. `vss memory introspect` would not perform its own memory retrieval internally => it takes the already retrieved memory above as input + the user query and creates gaps from judge etc.
6. follow up vlm queries etc.

So we make `vss memory introspect` not do memory searches internally and instead make it composable? Letting the harness chain `vss memory get` -> `vss memory introspect`

maxervo: note: in this composable approach, harness probably needs to write the memory record (retrieved from `vss memory get`) in its workspace before passing it to `vss memory introspect` via std input, because they are two separate turns (i.e. in between, model thinking if chaining to introspect is required or not). Also to avoid non-deterministically regenerating the input for `vss memory introspect`

maxervo: This also raises the question of what `vss memory search` is for vs the already present harness native memory search tool.
I remember it was discussed that it was good to have both so we can benchmark different retrieval methods and swap them in the future. We can confirm

services/agent/packages/vss_core/src/vss_core/introspection/orchestrator.py
```diff
@@ -0,0 +1,354 @@
...
+        group=request.group,
+        since=request.start_time,
+        until=request.end_time,
+        time_field="window",
+        include_children=True,
+        parents_only=False,
+        limit=settings.max_memory_records,
+    )
+    try:
+        state.records = await _retrieve_records(memory, query, settings.max_memory_records)
```
maxervo: So based on this [comment](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1895/changes#r3882902947), we would not make `vss memory introspect` perform its own memory searches internally and instead we would make it composable?

To rely on `vss memory get` that runs first then letting harness/model decide if it needs to chain it with `vss memory introspect`?

liamy-nv:
/ok to test [6d1c660](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1895/commits/6d1c660e352e39bb4ed5066409c62cc3dd84fd91)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33428772624)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33438663811)

services/agent/packages/vss_cli/src/vss_cli/vlm/runner.py
```diff
@@ -0,0 +1,434 @@
...
+    def close(status: Literal["failed", "partial", "timeout"], detail: str) -> Literal["absent", "closed", "stale"]:
+        if memory is None or job_id is None or created_at is None or input_data is None:
+            return "absent"
+        attempts = _TERMINAL_WRITE_ATTEMPTS
+        backoff_seconds = _TERMINAL_WRITE_BACKOFF_SECONDS
+        # Past the shared deadline, persistence is one best-effort write. The
+        # usual retry/sleep would return after the documented maximum.
+        if deadline - loop.time() <= 0:
+            attempts = 1
+            backoff_seconds = 0
```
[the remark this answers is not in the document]
zac-wang-nv: Valid — `close()` still calls synchronous `mark_terminal()` after the shared deadline. An Elasticsearch write can therefore block for the store's normal request timeout. This needs a bounded/off-loop terminal write before the documented end-to-end maximum can be trusted.

services/agent/packages/vss_cli/src/vss_cli/memory_cmd.py
```diff
@@ -181,6 +288,56 @@ def query_records(
...
+            sensor=sensor,
+            start_time=start_time,
+            end_time=end_time,
+            job_id=job_id,
+            record_id=record_id,
+            record_type=cast("RecordType | None", record_type),
+            group=cast("MemoryGroup | None", group),
+        )
+        has_time_range = request.start_time is not None and request.end_time is not None
+        if not (request.sensor or request.job_id or request.record_id or has_time_range):
```
zac-wang-nv: **P1** — Treating `--record-id` alone as useful identity scope is unsafe because a child is publicly identified by `(job_id, record_type, record_id)`; the store explicitly permits the same record ID across jobs and types. The orchestrator then treats any `record_id` as exact identity and removes text filtering, so this can send unrelated children to the judge. Require the full child identity, or keep text filtering until it is supplied.

services/agent/packages/vss_cli/src/vss_cli/vlm/runner.py
```diff
@@ -0,0 +1,434 @@
...
+                    answer=answer,
+                    sensor_name=sensor.name,
+                    start_time=start_time,
+                    end_time=end_time,
+                    model=model,
+                    persisted=persisted,
+                    record=success_record,
+                    persistence_error=persistence_error,
+                )
+        except TimeoutError as error:
```
zac-wang-nv, line 496: **P1** — This only handles this function's own deadline. `memory introspect` starts its outer 180-second timeout before calling this runner with a fresh 180-second budget, so outer expiry injects `asyncio.CancelledError` after the submitted write. Because that is a `BaseException`, it bypasses both handlers and leaves the job permanently `submitted` (reproduced against the in-memory store). Catch cancellation after job creation and perform a bounded terminal close, or pass the actual remaining workflow budget.

services/agent/packages/vss_cli/src/vss_cli/memory_cmd.py
```diff
@@ -69,6 +94,88 @@ def _output_options(function: Any) -> Any:
...
+        )
+    finally:
+        if client is not None:
+            await client.aclose()
+        if owns_memory:
+            close_memory = getattr(memory.service.store, "close", None)
+            if close_memory is not None:
+                close_memory()
+
+    if runner.persistence_errors:
```
zac-wang-nv: **P1** — Persistence warnings take precedence over the workflow's terminal failure here. If an early VLM follow-up records a persistence error and a later follow-up or synthesis times out or loses its backend, the command returns exit 6 instead of the required timeout/backend exit 7 or 3. Check timeout/backend signals first, and use `PARTIAL` only when the workflow otherwise succeeds.

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33576823050)

liamy-nv:
/ok to test [4b9fffc](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1895/commits/4b9fffc3eef26d46177771f520fff25fda38c4d0)

services/agent/packages/vss_cli/src/vss_cli/vlm/runner.py
liamy-nv: Memory introspect no longer hands the VLM runner a fresh 180s. The orchestrator passes the remaining shared deadline. If an outer asyncio.timeout_at still injects CancelledError after the submitted write, that is no longer a bypass. It is caught outside the inner work timeout (so it is not treated as a generic Exception) and the job gets a bounded terminal close before a timeout result is returned. test_cancellation_after_submission_closes_terminal_record reproduces the in-memory case: cancel after submit → status timeout, record closed.

services/agent/packages/vss_cli/src/vss_cli/memory_cmd.py
liamy-nv: --record-id alone is not identity. Child records are keyed by (job_id, record_type, record_id), and the store allows the same record_id across jobs and types. IntrospectionRequest now rejects a lone --record-id; the CLI requires --job-id, --record-type, and --record-id together. Until that full identity is present, the query keeps text filtering instead of treating any matching child id as exact identity.

services/agent/packages/vss_cli/src/vss_cli/memory_cmd.py
liamy-nv: CLI exit order is timeout (7) and backend unreachable (3) before persistence PARTIAL (6). An early VLM persist warning no longer masks a later timeout or backend failure; PARTIAL is used only when the workflow itself succeeded and only the write did not.

services/agent/packages/vss_cli/src/vss_cli/vlm/runner.py
```diff
@@ -0,0 +1,468 @@
...
+                    try:
+                        memory.service.upsert(
+                            adapter.terminal_record(
+                                job_id=job_id,
+                                created_at=created_at,
+                                status="completed",
+                                input_data=input_data,
+                                output=adapter.build_output(answer=answer, model=model),
+                            )
+                        )
```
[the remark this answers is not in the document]
liamy-nv: Introspection starts one 180s clock and each VLM follow-up gets only time left on that clock, not another 180s. After a submitted write, cancellation is converted into a bounded close("timeout", …) so the job does not stay submitted forever.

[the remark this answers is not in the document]
liamy-nv: Terminal and completion upserts run on a daemon thread so a slow Elasticsearch upsert does not block the event loop. The caller waits only for a reserved slice of the shared deadline (timeout_at / asyncio.timeout); if the write is still going, the job returns stale instead of overrunning the end-to-end budget.

services/agent/packages/vss_core/src/vss_core/introspection/orchestrator.py
```diff
@@ -0,0 +1,367 @@
...
+        parents = await asyncio.to_thread(
+            memory.query,
+            MemoryQuery(
+                job_id=job_id,
+                sensor_id=query.sensor_id,
+                group=query.group,
+                since=query.since,
+                until=query.until,
+                time_field=query.time_field,
+                include_children=False,
```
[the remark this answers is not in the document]
liamy-nv: Parent lookup no longer reapplies the child’s time_field="window" bounds. Expansion queries by job_id without a window filter, then _parent_matches_time_scope keeps parents that have no input.window and only drops parents whose known window does not overlap the request.

services/agent/packages/vss_core/src/vss_core/introspection/orchestrator.py
```diff
@@ -0,0 +1,374 @@
...
+    return (record.job.job_id, record.job.record_type, record.job.record_id)
+
+
+def _parent_matches_time_scope(parent: UnifiedMemoryRecord, query: MemoryQuery) -> bool:
+    """Keep scoped parents without windows, but reject known non-overlapping windows."""
+    window = parent.input.window if parent.input is not None else None
+    if window is None:
+        return True
+    if query.since is not None and window.end.timestamp < query.since:
+        return False
```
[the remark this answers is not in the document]
liamy-nv: _parent_matches_time_scope no longer assumes window.end exists. A start-only parent is kept unless a real end timestamp is before the query’s since.

services/agent/packages/vss_cli/src/vss_cli/vlm/runner.py
```diff
@@ -0,0 +1,501 @@
...
+
+            def report(result: Exception | None = outcome) -> None:
+                if not write_finished.done():
+                    write_finished.set_result(result)
+
+            with contextlib.suppress(RuntimeError):
+                loop.call_soon_threadsafe(report)
+
+        threading.Thread(target=write, name="vss-vlm-persistence-write", daemon=True).start()
+        outcome = await asyncio.shield(write_finished)
```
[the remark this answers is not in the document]
liamy-nv: If a persist write is still running when the deadline hits, the runner waits the reserved budget, then abandons further writes instead of starting a terminal-write thread. Late upserts are skipped so they cannot run after the caller closes Elasticsearch.

liamy-nv:
/ok to test [f3c96d7](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1895/commits/f3c96d7b4c3874fadaaa27c65e4341ed5b448acc)

liamy-nv:
/ok to test [f088748](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1895/commits/f088748f3b88ec95865d2d100d04c970c6161a85)

liamy-nv:
/ok to test [22ac255](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1895/commits/22ac255e0c329753d2377fd5dfb1b726b88a4fe8)

zac-wang-nv (approved):

PR #1903, opened 2026-08-28
feat(tools): BYO agent adapter — drive VSS from any agent harness

ashan-nv:
Lets a user bring their own agent or harness and drive VSS from the VSS UI,
instead of being tied to the built-in vss-agent:

  VSS UI --HTTP/SSE--> adapter --/v1/responses--> harness --skills--> VSS

The adapter exposes POST /chat/stream, taking an OpenAI-shaped {messages} body
and returning text/event-stream. That contract was not invented here: it is
already what the VSS UI speaks via pages/api/chat.ts, so pointing the UI at a
different agent backend is one env var and no UI code.

Harness-agnostic by construction. OpenClaw and Hermes both expose the OpenAI
Responses API at POST /v1/responses with identical request and SSE shapes, so
run_turn_responses is a single ~40-line driver parameterised by (url, token,
model); the named backends are thin wrappers. A third party arriving with their
own Responses-API agent needs no new driver at all. Verified against both with
an identical archive-search request, each returning real hits.

Also serves what a foreign agent needs to be useful:

  GET  /v1/skills                  manifest (name, path, requirements)
  GET  /v1/skills/<name>           one SKILL.md, fetched on demand
  GET  /v1/skills/bundle.tar.gz    all skills, matched to this deployment
  GET  /v1/skills/env              resolved VSS base URLs
  POST /v1/search                  archive search over HTTP

/v1/skills/env matters more than its size suggests: skills must know where VSS
is, and that is genuinely fiddly (HOST_IP in-sandbox, VSS_PUBLIC_URL on k8s,
Compose discovery fallback, per-profile path prefixes). Today that is prose in
deployment_resolution.md plus an export in ENV.md, so every BYO agent re-derives
it and gets it subtly wrong, with policy_denied as the failure mode.

Bootstrap is by prompt, not by install: the first turn of a session carries the
deployment context and skills index (~6.6 KB), while skill bodies (~345 KB
across 18 skills) stay remote until one is needed. This requires nothing of the
harness beyond accepting text, where `skill install` assumes an OpenClaw-shaped
workspace a hosted agent does not have.

/v1/search tolerates multi-document CLI output. The `vss` CLI prefixes log lines
and now also appends a job-completion event, so stdout holds more than one JSON
document and decoding from the first brace fails with "Extra data". It now walks
every document present and picks the one carrying `data`, rather than trusting
position -- the event line is last today, but ordering is not a contract. This
was caught by restarting against a freshly rebased tree.

The policy entry is required, not incidental: without host.openshell.internal
on the adapter's port, every call from the sandbox fails as policy_denied.

Known limits are recorded in the README's Status section rather than left to be
discovered: no auth beyond a caller allowlist and optional shared token; the
Responses stream carries text only so tool/step events are absent; and OpenClaw's
Responses endpoint answers from one shared session, so separate UI conversations
share memory -- history is carried in the prompt instead. That last one is the
multi-user concern arriving early and argues for agent-per-user before real use.

The OpenClaw gateway's undocumented WebSocket handshake quirks are recorded in
adapter.py's module docstring, next to the code that depends on them.

/v1/search/last is included but unused until the UI change that consumes it,
which is deliberately a separate PR: this one touches no UI files.

Signed-off-by: apanduwawala <apanduwawala@nvidia.com>

### DIFF
--- a/assets/vss_nemoclaw_policy.yaml
+++ b/assets/vss_nemoclaw_policy.yaml
@@ -238,6 +238,24 @@ network_policies:
           - 172.17.0.0/16
           - 172.18.0.0/16
           - 172.20.0.0/16
+      # BYO agent adapter: skills index, per-skill fetch, and archive search
+      # over HTTP (tools/vss-agent-adapter). Without this the sandbox cannot
+      # reach it and every call fails as policy_denied.
+      - host: host.openshell.internal
+        port: 9098
+        access: full
+        allowed_ips:
+          - 172.17.0.0/16
+          - 172.18.0.0/16
+          - 172.20.0.0/16
+      # Second adapter instance (e.g. AGENT_BACKEND=hermes) on its own port.
+      - host: host.openshell.internal
+        port: 9097
+        access: full
+        allowed_ips:
+          - 172.17.0.0/16
+          - 172.18.0.0/16
+          - 172.20.0.0/16
       - host: host.openshell.internal
         port: 18789
         access: full
--- a/deploy/docker/services/agent-adapter/compose.yml
+++ b/deploy/docker/services/agent-adapter/compose.yml
@@ -1,0 +1,97 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# BYO agent adapter: lets the VSS UI drive an external agent harness
+# (OpenClaw, Hermes, or anything speaking the OpenAI Responses API) instead of
+# the built-in vss-agent. See tools/vss-agent-adapter/README.md.
+#
+# Off unless the vss-agent-adapter profile is active, so existing deployments
+# are unaffected.
+
+services:
+
+  vss-agent-adapter:
+    build:
+      context: ../../../../tools/vss-agent-adapter
+      dockerfile: Dockerfile
+    image: ${VSS_AGENT_ADAPTER_IMAGE:-vss-agent-adapter}:${VSS_AGENT_ADAPTER_TAG:-local}
+    profiles: ["vss-agent-adapter"]
+    container_name: vss-agent-adapter
+    restart: always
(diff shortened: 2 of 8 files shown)


ashan-nv:
@coderabbitai review

ashan-nv:
Both review findings are addressed in b114a4c25.

**Query token leaks into logs (P1)** — the request-line logger now redacts `?token=`. Verified: `GET /health?token=[REDACTED]`.

**Disconnected turns keep running (P2)** — turns now take a cancel event that every driver checks once per stream iteration; the handler sets it when the client goes away. Verified by killing a client mid-turn: `client disconnected; cancelling turn`.

One deliberate non-change: the queue is still unbounded. Bounding it looked tidier but trades a short-lived queue for a wedged thread, since the driver's terminal `put()` would block forever once the consumer is gone. Cancelling the producer is the actual fix.

While testing the token path I also found that setting `ADAPTER_TOKEN` gates the agent's own callbacks (`/v1/skills*`, `/v1/search`) while the bootstrap hands the agent a URL with no token — so the agent 403s. Resolving it means either putting a credential in the model's context or scoping the token to `/chat/stream`, so it is documented in the README rather than guessed at.

Re-verified after the change: both harnesses returned real archive-search hits (98 and 50 frames).

ashan-nv:
Addressed in 13decaf8d.

**Cancellation waits behind upstream reads** — correct, the flag was only checked between iterations. Turns now carry a cancellation handle that drivers register their upstream connection with; cancelling closes it so the blocked read raises immediately.

While verifying that I found the disconnect detection itself was unreliable: a browser closing sends FIN, and writing to a half-closed socket still succeeds, so the write-error path missed departed clients entirely in testing. The handler now polls the client socket for EOF, bounding detection to one keepalive interval — measured at ~20s with the 15s default.

zac-wang-nv:
**P1** — Assuming OpenClaw installs and manages the VSS skills itself, the VSS UI should use OpenClaw's native `/v1/chat/completions` API directly instead of adding this adapter.

The UI already sends a Chat Completions-shaped `messages` payload and parses `choices[0].delta.content`, while OpenClaw exposes that same compatibility surface. A small change in the existing UI server-side proxy can attach gateway authentication, select `openclaw/default`, map `Conversation-Id` to OpenClaw's `user` or `x-openclaw-session-key`, and enable streaming.

Under this ownership model, OpenClaw should receive the VSS skills through its normal installation/configuration path and resolve VSS access from those skills. The adapter should not duplicate skill discovery, bundle serving, bootstrap injection, or deployment metadata. If a particular skill lacks a runtime dependency such as the host-side VSS CLI, that capability should be addressed at the skill/tool boundary rather than by retaining a general chat transport proxy.

This removes the custom Responses translation, legacy WebSocket driver, backend dispatch paths, skill-serving endpoints, and most of the session/cancellation plumbing introduced here. OpenClaw's native endpoint is documented at https://docs.openclaw.ai/gateway/openai-http-api.

ashan-nv:
Closing in favor of this PR https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1980/changes#diff-e92cf4f9b6ef03a124c1bb49e17b1fb564faeed107562497338e72ae3158d826R3075-R3077

PR #1904, opened 2026-08-28
feat(ui): toolkit-free VSS chat surface

ashan-nv:
Toolkit-free chat surface for the VSS UI, behind `NEXT_PUBLIC_USE_VSS_CHAT`.

**Stacked on #1903** — base is `feat/byo-agent-adapter`, not `develop`. GitHub will
retarget to `develop` automatically once #1903 merges. Reviewing against `develop`
would show the adapter twice.

### What is here

- New `@nv-metropolis-bp-vss-ui/chat` package: `ChatPanel`, `useChatStream`, SSE parsing,
  styles, and a test. No NeMo Agent Toolkit dependency.
- `Home.tsx` renders it for both the chat tab and the docked sidebar when the flag is on;
  unset, nothing changes, so the two can be compared in place.
- `pages/api/vss-chat.ts` proxies server-side (the configured backends are host-private
  ports a browser cannot reach) and serves the last search result so the Search tab can
  render hits without the model retyping them.
- Ingress: `/api/vss-chat` routed to the UI backend, mirroring the existing `/api/chat`
  rule. Without it the chat works on localhost and 404s through any ingress.

### Notes for review

- `package-lock.json` accounts for ~12k of the diff; the hand-written change is ~1,700 lines.
- `apps/vss-chat` (8 files) was the standalone proving ground before `ChatPanel` moved into
  the product UI. It may be redundant now — happy to drop it.
- Runtime dependency on the adapter is hard (the panel proxies to it); build dependency is none.

### Verification

Built against current `develop` and deployed: UI reachable on localhost, through haproxy,
and through the public ingress; end-to-end chat returned real archive-search hits, and the
search-artifact endpoint returned structured results for the Search tab.

### DIFF
--- a/deploy/docker/services/infra/haproxy/haproxy.cfg.template
+++ b/deploy/docker/services/infra/haproxy/haproxy.cfg.template
@@ -127,6 +127,12 @@ frontend fe_http
     acl known_host hdr(host) -i "${HOST_IP}"
     acl known_host hdr(host) -i "${HOST_IP}:${HAPROXY_PORT}"
     acl known_host hdr(host) -i localhost
+    # The UI's media proxy reaches the ingress container-to-container, so the
+    # Host it sends is the compose service name. Node's fetch will not let the
+    # proxy override Host, so the ingress has to accept it or every proxied
+    # thumbnail is denied 404 by the known_host guard.
+    acl known_host hdr(host) -i vss-haproxy-ingress
+    acl known_host hdr(host) -i "vss-haproxy-ingress:${HAPROXY_PORT}"
     acl known_host hdr(host) -i "localhost:${HAPROXY_PORT}"
     acl known_host hdr(host) -i 127.0.0.1
     acl known_host hdr(host) -i "127.0.0.1:${HAPROXY_PORT}"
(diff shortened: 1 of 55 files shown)


ashan-nv:
@coderabbitai review

ashan-nv:
Both findings addressed in the latest push.

**Proxy bypasses ingress authentication** — the catch-all now accepts only GET/HEAD to an allowlisted prefix (`/vst/` by default, `VSS_PROXY_ALLOWED_PREFIXES` to widen), and resolves the path before checking so `..` cannot escape it. Verified through the public ingress: allowed media 200, Elasticsearch 403, POST 405, traversal 403.

**Search results cross conversations** — the answer handler now carries the conversation that produced it and the fetch names it; the adapter attributes each search to the sole in-flight turn and serves only the owner. With two turns in flight attribution would be a guess, so it is withheld rather than risk handing over the wrong media. Verified: owner served; other conversation, missing id, concurrent turns and stale entries all withheld.

Routing the proxy also revealed it had never worked through an ingress at all — no haproxy rule, so it 404d everywhere except localhost, which is the opposite of what it exists for. Fixing that surfaced two more layers (the compose service name must satisfy both `known_host` and `h_main`, since Node fetch cannot override `Host`). All three are in the ingress template now, verified 200 end to end.

ashan-nv:
Reviewed the parity work in depth and corrected an overstatement in the commit message.

**What checks out:** all 15 toolkit `NEXT_PUBLIC_CHAT_*` variables the deployment sets are honoured; all five bridge props are declared, consumed in `ChatPanel` and passed from `Home.tsx`; SSE parsing holds partial lines across chunk boundaries, normalises CRLF, skips `:` keepalives and handles `[DONE]`; conversation storage is a faithful port of the toolkit's tab-scoped IndexedDB scheme, keyed per surface; streaming has a per-send `AbortController` with unmount cleanup and a busy guard. Package typecheck clean, 53 tests passing, Docker build compiles.

**Correction:** the message previously said this lets the NAT package go. It does not. Chat is no longer the blocker, but `@nemo-agent-toolkit/ui` is still imported by 14 files across five packages for non-chat reasons — `VideoModal`/`useVideoModal` in alerts and search, `chunkedUpload` in video-management, `copyToClipboard`, `ChatSidebarContent`, and the app's server entry points. Filed #1921 to track extracting those. This PR is a prerequisite, not the last step.

**One pre-existing risk, not introduced here:** message markdown renders through `rehypeRaw` with DOMPurify applied only to `callerInfo`, so raw HTML in model output reaches the tree. The toolkit does exactly the same, so this is faithful parity rather than a regression — but worth a follow-up (`rehype-sanitize`) independent of this PR.

PR #1905, opened 2026-08-28
fix(ci): scan SonarQube projects only when their trees change

akshayanv:
On PRs, still report every SonarQube Scan check so required develop contexts are not left pending, but only run the scanner for projects whose sources actually changed.

## Checklist

### DIFF
--- a/.github/scripts/detect_sonarqube_projects.py
+++ b/.github/scripts/detect_sonarqube_projects.py
@@ -1,0 +1,313 @@
+#!/usr/bin/env python3
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+"""Decide which SonarQube matrix legs a workflow run must execute.
+
+Emits a GitHub Actions matrix (JSON) with one entry per catalogued project
+that needs a scan.
+
+Diff-range rules:
+
+* ``pull_request`` → diff ``merge-base(PR base, HEAD)..HEAD`` so the matrix
+  reflects the whole PR, not just its last push. Every catalogued
+  ``SonarQube Scan (<name>)`` check is still *reported* (GitHub has no
+  path filter on required contexts, and the same no-op is used for the
+  rest so a later ruleset addition cannot stall merge). Untouched legs
+  run as no-ops (``scan=false``). A real scan runs only when that tree
+  changed — that is when the check can fail and actually gate the PR.
+* ``push`` to ``main`` / ``develop`` / ``release/**``, and
+  ``workflow_dispatch`` → scan every project. Branch analysis and a manual
+  run must not silently drop a leg.
+* Unresolvable PR base, or a failed ``git diff`` → **scan everything**.
+  Scanning too much is safe; skipping a project that changed is the
+  failure mode this replaces.
+* A change to this script or ``.github/workflows/sonarqube.yml`` also
+  scans every project (the scan contract changed).
+
+Do not put a top-level ``paths:`` filter on the workflow. GitHub evaluates
+that against the pushed commits, not the cumulative PR, and can skip the
+workflow entirely on an unrelated follow-up push.
+"""
+
+from __future__ import annotations
+
+import argparse
+import json
+import sys
+from pathlib import Path
+from typing import Any
+
+sys.path.insert(0, str(Path(__file__).resolve().parent))
+
+from detect_changed_images import (  # noqa: E402
+    changed_paths,
+    commit_exists,
+    paths_changed_under,
+    run_git,
+)
+
+CONTRACT_PATHS = (
+    ".github/workflows/sonarqube.yml",
+    ".github/scripts/detect_sonarqube_projects.py",
+)
+
+SONAR_RUNNER = "sonarqube-workflows-bp-sre"
+SKIP_RUNNER = "ubuntu-latest"
+
+# ``paths`` is the skip gate. ``sources`` / ``tests`` stay the scanner
+# configuration and may be a narrower list than the directory we watch.
+PROJECTS: list[dict[str, Any]] = [
+    {
+        "name": "agent",
+        "project_key": (
+            "TEGRASW_metropolis_video-search-and-summarization-agent"
+            "_video-search-and-summarization"
+        ),
+        "project_name": "video-search-and-summarization-agent",
+        "sources": "services/agent",
+        "tests": (
+            "services/agent/packages/vss_core/tests,"
+            "services/agent/packages/vss_agents/tests,"
+            "services/agent/packages/vss_cli/tests"
+        ),
+        "python_version": "3.13",
+        "paths": ["services/agent"],
+    },
+    {
+        "name": "alert",
+        "project_key": "TEGRASW_metropolis_vss-alert-verification_alert_agent",
+        "project_name": "metropolis-vss-alert-verification",
+        "sources": "services/alert/src",
+        "tests": "services/alert/test/unit",
+        "python_version": "3.13",
+        "paths": ["services/alert"],
+    },
+    {
+        "name": "ui",
+        "project_key": (
+            "TEGRASW_metropolis_video-search-and-summarization-ui"
+            "_video-search-and-summarization"
+        ),
+        "project_name": "video-search-and-summarization-ui",
+        "sources": "services/ui",
+        "tests": "",
+        "paths": ["services/ui"],
+    },
+    {
+        "name": "skills",
+        "project_key": (
+            "TEGRASW_metropolis_video-search-and-summarization-skills"
+            "_video-search-and-summarization"
+        ),
+        "project_name": "video-search-and-summarization-skills",
+        "sources": "skills",
+        "tests": "",
+        "paths": ["skills"],
+    },
+    {
+        "name": "behavior-analytics",
+        "project_key": "TEGRASW_metropolis_vss-behavior-analytics_py-analytics-stream",
+        "project_name": "metropolis-vss-behavior-analytics",
+        "sources": (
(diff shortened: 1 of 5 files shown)


zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33145002458)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33145860058)

hugoverjus:
Reviewed at `8e23820`. The design here is right: the planner, the fail-open defaults, the contract-change full scan, and scanning everything on push (which is what keeps Sonar's server-side branch data current) all check out. I ran the new tests locally — 18/18 in `test_detect_sonarqube_projects.py`, plus the migrated SDU catalog assertion in `test_prepare_downstream_release_set.py` (15/15). All steps in the rewritten matrix job are correctly gated on `matrix.scan == 'true'` — I checked all 18.

Two things to fix, one of which I'd treat as blocking.

## 1. Blocking: the required contexts stop being enforceable, and the gate that would catch it isn't required

`Protect develop` (ruleset `15323753`) requires exactly:

```
DCO Sign-off
Check folder structure
Copyright Headers
Security Scan (detect-secrets)
SonarQube Scan (agent)
SonarQube Scan (ui)
SonarQube Scan (skills)
```

After this change, `detect` is a single point of failure in front of all three of the Sonar contexts, and neither `SonarQube` nor `Detect SonarQube projects` is required. Both failure modes are bad:

- If the matrix expands but `detect` fails, the legs are skipped — and GitHub reports a conditionally-skipped job as success for required status checks, so Sonar enforcement disappears silently.
- If `fromJSON` receives a malformed or empty `matrix`, the per-leg check runs may never be created at all, leaving the three contexts pending forever. That is the exact trap the new file header sets out to avoid, reintroduced by a different route.

`sonarqube-gate` already handles both cases correctly (`if [ "$DETECT" != "success" ]; then exit 1`) — it just isn't wired to anything that can block a merge. Suggest adding `SonarQube` to the ruleset's required contexts, ideally *replacing* the three per-leg contexts: after this PR, `SonarQube Scan (ui)` passing no longer means a scan of `services/ui` ran, so it isn't the signal the ruleset thinks it is. Worth landing in the same change window as the merge rather than afterwards.

## 2. `CONTRACT_PATHS` is incomplete

The planner imports `changed_paths`, `commit_exists`, `paths_changed_under` and `run_git` from `detect_changed_images.py`, so a change to any of those helpers can alter every skip decision without tripping the full-scan safety valve:

```python
CONTRACT_PATHS = (
    ".github/workflows/sonarqube.yml",
    ".github/scripts/detect_sonarqube_projects.py",
    ".github/scripts/detect_changed_images.py",
)
```

## Worth knowing, not blocking

Because this PR touches both contract paths, its own run took the full-scan path. From the `detect` log:

```
PR merge-base with 9b59f10d8fde...: 9b59f10d8fde; scan contract changed; scanning all projects
"count": 10
```

All ten legs were real scans (agent 4m39s, behavior-analytics 8m45s), so the no-op leg behaviour — the only thing this PR actually changes about required-check semantics — has no end-to-end coverage yet, only unit coverage. The first live exercise will be the next unrelated PR after merge; worth checking that PR's three required contexts deliberately rather than assuming they report correctly.

## Minor

- `needs.detect.outputs.any` can never be anything but `'true'`: on `pull_request` the include list is padded to all ten projects, and on `push`/`workflow_dispatch` all ten are selected. The `if:` condition that consumes it is dead, and reads as if an empty-matrix case is being handled.
- `matrix_for()` takes a `_base_ref` parameter it never uses.

.github/scripts/detect_sonarqube_projects.py
```diff
@@ -0,0 +1,313 @@
...
+    changed_paths,
+    commit_exists,
+    paths_changed_under,
+    run_git,
+)
+
+CONTRACT_PATHS = (
+    ".github/workflows/sonarqube.yml",
+    ".github/scripts/detect_sonarqube_projects.py",
+)
```
hugoverjus, line 54: `CONTRACT_PATHS` is missing the module the skip decisions actually depend on. This planner imports `changed_paths`, `commit_exists`, `paths_changed_under` and `run_git` from `detect_changed_images.py`, so a change to any of those helpers can alter every skip decision without tripping the full-scan safety valve.

```suggestion
CONTRACT_PATHS = (
    ".github/workflows/sonarqube.yml",
    ".github/scripts/detect_sonarqube_projects.py",
    ".github/scripts/detect_changed_images.py",
)
```

.github/scripts/detect_sonarqube_projects.py
```diff
@@ -0,0 +1,313 @@
...
+    return entry
+
+
+def to_matrix(entries: list[dict[str, Any]]) -> dict[str, Any]:
+    return {"include": entries}
+
+
+def matrix_for(
+    selected: list[dict[str, Any]], event_name: str, _base_ref: str
+) -> list[dict[str, Any]]:
```
hugoverjus, line 219: `_base_ref` is never used - `matrix_for` branches only on `event_name`. Worth dropping so the signature does not imply a base-ref-dependent matrix that does not exist.

No suggestion block here on purpose: the parameter removal has to land together with the call site in `plan()` (`matrix_for(selected, event_name, base_ref)`) and the three `dsp.matrix_for(...)` calls in `test_detect_sonarqube_projects.py`, so a single-hunk commit would break the build.

.github/workflows/sonarqube.yml
```diff
@@ -21,83 +31,58 @@ permissions:
...
     name: SonarQube Scan (${{ matrix.name }})
-    runs-on: sonarqube-workflows-bp-sre
+    needs: detect
+    runs-on: ${{ matrix.runner }}
     timeout-minutes: 30
-    if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name == github.repository
+    if: >-
+      needs.detect.outputs.any == 'true'
+      && (github.event_name != 'pull_request'
+          || github.event.pull_request.head.repo.full_name == github.repository)
```
hugoverjus: `needs.detect.outputs.any` can never be anything other than `'true'`, so this first condition is dead. On `pull_request`, `matrix_for()` pads the include list to all 10 projects; on `push` / `workflow_dispatch`, all 10 are selected. `plan()` sets `any` from `len(include) > 0`, which therefore always holds. It is not a useful failure guard either: because the expression contains no `always()`, a failed `detect` already skips this job through `needs:`.

```suggestion
    if: >-
      github.event_name != 'pull_request'
      || github.event.pull_request.head.repo.full_name == github.repository
```

Applying just this hunk is safe - the `any` output simply becomes unused. To finish the removal in the same commit, also drop:

* `any: ${{ steps.plan.outputs.any }}` from the `detect` job's `outputs:` (line 39 of this file);
* `"any": len(include) > 0,` in `plan()` and the `any=` line in `write_github_output()` in `detect_sonarqube_projects.py`;
* the four `result["any"]` / `lines["any"]` assertions in `test_detect_sonarqube_projects.py`.

I deliberately did not put suggestion blocks on those, because order matters: removing the script side while this `if:` still reads the output would make the expression evaluate to false and skip every SonarQube leg.

hugoverjus (commented):
Two things here that cannot be code suggestions; the mechanical fixes are inline below.

**Blocking: the `Protect develop` ruleset needs updating alongside this PR.** Ruleset `15323753` requires exactly these contexts: `DCO Sign-off`, `Check folder structure`, `Copyright Headers`, `Security Scan (detect-secrets)`, `SonarQube Scan (agent)`, `SonarQube Scan (ui)`, `SonarQube Scan (skills)`. After this PR, `detect` becomes an unguarded single point of failure in front of the three Sonar contexts, and neither the new `SonarQube` gate job nor `Detect SonarQube projects` is required. Two bad failure modes follow:

1. If `detect` fails and the legs are skipped, GitHub reports conditionally-skipped jobs as *success* for required status checks, so Sonar enforcement silently disappears.
2. If `fromJSON` gets a malformed or empty `matrix`, the per-leg check runs may never be created at all, leaving the three contexts pending forever - the exact trap the new file header sets out to avoid.

`sonarqube-gate` already handles both correctly (`if [ "$DETECT" != "success" ]; then exit 1`), but it is not wired to anything that blocks a merge. Please add `SonarQube` to the ruleset's required contexts, ideally *replacing* the three per-leg contexts, since after this PR `SonarQube Scan (ui)` passing no longer means a scan of `services/ui` actually ran.

**The no-op leg path has no end-to-end coverage yet.** This PR touches both `CONTRACT_PATHS` entries, so its own run took the full-scan path: the `detect` job log reads `scan contract changed; scanning all projects` with `"count": 10`, and all ten legs were real scans (agent 4m39s, behavior-analytics 8m45s). So `scan=false` behaviour is covered by unit tests only, never by a real run.

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33197711621)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33228632908)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33451078086)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33451838125)

zac-wang-nv (approved):
LGTM

kaushikc-nvidia (approved):

PR #1906, opened 2026-08-28
fix(ci): stop leftover PR and downstream jobs after merge

akshayanv:
Closing a PR (merge or abandon) now cancels remaining GitHub PR checks, including optional ones like SonarQube, without touching develop/main. When the GitHub trigger job itself is cancelled, it also cancels the GitLab pipeline it already started.

## Checklist

### DIFF
--- a/.github/scripts/cancel_stale_pr_ci.py
+++ b/.github/scripts/cancel_stale_pr_ci.py
@@ -1,12 +1,11 @@
 #!/usr/bin/env python3
 # SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
-"""Cancel in-flight CI on ``pull-request/<N>`` after the source PR moves.
+"""Cancel leftover PR CI after the source PR moves or closes.
 
-PR CI is gated on copy-pr-bot updating the mirror (``/ok to test``). Until
-that push, GitHub concurrency cannot see a new run and will not cancel the
-old one. This helper is the source-PR side of that: on synchronize or close,
-cancel mirror runs that are not testing the current source head.
+On synchronize, cancel ``pull-request/<N>`` runs that are not testing the
+current source head. On close (merge or abandon), cancel every remaining
+PR run — required or not — except close-path workflows such as tag cleanup.
 
 copy-pr-bot often pushes a merge of the approved SHA into the PR base, so
 the run's tree is the merge result, not the source-head tree. A current run
@@ -29,6 +28,10 @@ GITHUB_API = "https://api.github.com"
 ACTIVE_STATUSES = frozenset(
     {"in_progress", "pending", "queued", "requested", "waiting"}
 )
+# Close handlers that must finish after merge. Everything else on the PR,
+# required or not, is leftover work.
+PROTECTED_WORKFLOWS = frozenset({"Cancel Stale PR CI", "Cleanup PR Tags"})
+PR_EVENTS = ("pull_request", "pull_request_target")
 
 
 class CancelError(RuntimeError):
@@ -114,19 +117,18 @@ def load_commit(
     return info
 
 
-def list_active_runs(
+def _collect_status_pages(
     repo: str,
-    branch: str,
-    get: Callable[..., Any] = github,
+    extra_query: str,
+    get: Callable[..., Any],
 ) -> list[dict[str, Any]]:
     seen: set[int] = set()
     runs: list[dict[str, Any]] = []
-    quoted = urllib.parse.quote(branch, safe="")
     for status in sorted(ACTIVE_STATUSES):
         page = 1
         while page <= 20:
             path = (
-                f"/actions/runs?branch={quoted}&status={status}"
+                f"/actions/runs?{extra_query}&status={status}"
                 f"&per_page=100&page={page}"
             )
             try:
@@ -146,6 +148,65 @@ def list_active_runs(
             if len(batch) < 100:
                 break
             page += 1
+    return runs
+
+
+def list_active_runs(
+    repo: str,
+    branch: str,
+    get: Callable[..., Any] = github,
+) -> list[dict[str, Any]]:
+    quoted = urllib.parse.quote(branch, safe="")
+    return _collect_status_pages(repo, f"branch={quoted}", get)
+
+
+def list_active_runs_for_event(
+    repo: str,
+    event: str,
+    get: Callable[..., Any] = github,
+) -> list[dict[str, Any]]:
+    quoted = urllib.parse.quote(event, safe="")
+    return _collect_status_pages(repo, f"event={quoted}", get)
+
+
+def run_belongs_to_pr(run: dict[str, Any], pr_number: int) -> bool:
+    for pr in run.get("pull_requests") or []:
+        number = pr.get("number") if isinstance(pr, dict) else None
+        if number == pr_number or str(number) == str(pr_number):
+            return True
+    return str(run.get("head_branch") or "") == f"pull-request/{pr_number}"
+
+
+def collect_runs_to_consider(
+    repo: str,
+    pr_number: int,
+    closed: bool,
+    get: Callable[..., Any] = github,
+) -> list[dict[str, Any]]:
+    """Mirror-branch runs always; on close, also native PR-event runs.
+
+    SonarQube and other optional checks use ``pull_request`` / 
+    ``pull_request_target`` on the contributor branch, not ``pull-request/N``.
+    """
+    branch = f"pull-request/{pr_number}"
+    seen: set[int] = set()
+    runs: list[dict[str, Any]] = []
+    for run in list_active_runs(repo, branch, get):
+        run_id = run.get("id")
+        if isinstance(run_id, int) and run_id not in seen:
+            seen.add(run_id)
+            runs.append(run)
+    if not closed:
+        return runs
+    for event in PR_EVENTS:
+        for run in list_active_runs_for_event(repo, event, get):
+            run_id = run.get("id")
+            if not isinstance(run_id, int) or run_id in seen:
+                continue
+            if not run_belongs_to_pr(run, pr_number):
+                continue
+            seen.add(run_id)
+            runs.append(run)
     return runs
 
 
@@ -181,6 +242,8 @@ def should_cancel_run(
 ) -> bool:
     run_id = run.get("id")
     if this_run_id is not None and run_id == this_run_id:
+        return False
+    if str(run.get("name") or "") in PROTECTED_WORKFLOWS:
         return False
     if run.get("status") not in ACTIVE_STATUSES:
         return False
@@ -234,7 +297,6 @@ def main() -> int:
     closed = os.environ.get("PR_CLOSED", "").strip().lower() in {"1", "true", "yes"}
     this_run_raw = os.environ.get("GITHUB_RUN_ID", "").strip()
     this_run_id = int(this_run_raw) if this_run_raw.isdigit() else None
-    branch = f"pull-request/{pr_number}"
     cache: dict[str, CommitInfo | None] = {}
 
     source_commit = load_commit(repo, source_sha, cache) if source_sha else None
@@ -247,7 +309,8 @@ def main() -> int:
         return 0
 
     cancelled = 0
-    for run in list_active_runs(repo, branch):
+    pr_id = int(pr_number)
+    for run in collect_runs_to_consider(repo, pr_id, closed):
         run_id = run.get("id")
         if not isinstance(run_id, int):
             continue
@@ -265,7 +328,7 @@ def main() -> int:
         print(f"Cancelling run {run_id} ({run.get('name')}) sha={head_sha[:12]}")
         cancel_run(repo, run_id)
         cancelled += 1
-    print(f"Cancelled {cancelled} stale run(s) on {branch}")
+    print(f"Cancelled {cancelled} leftover run(s) for PR {pr_number}")
     return 0
 
 
--- a/.github/scripts/cancel_stale_pr_ci.py
+++ b/.github/scripts/cancel_stale_pr_ci.py
@@ -169,12 +169,30 @@ def list_active_runs_for_event(
     return _collect_status_pages(repo, f"event={quoted}", get)
 
 
-def run_belongs_to_pr(run: dict[str, Any], pr_number: int) -> bool:
+def run_belongs_to_pr(
+    run: dict[str, Any],
+    pr_number: int,
+    source_branch: str = "",
+    source_sha: str = "",
+) -> bool:
+    """Attribute a run to a PR using keys that survive the PR closing.
+
+    GitHub empties ``pull_requests`` as soon as the PR closes, which is
+    exactly when the close path runs, so it cannot be the only key. The
+    ``pull-request/<N>`` fallback never matches a native ``pull_request``
+    run either — that run's ``head_branch`` is the contributor branch. The
+    source branch and head SHA carried by the close event are what remain.
+    """
     for pr in run.get("pull_requests") or []:
         number = pr.get("number") if isinstance(pr, dict) else None
         if number == pr_number or str(number) == str(pr_number):
             return True
-    return str(run.get("head_branch") or "") == f"pull-request/{pr_number}"
+    head_branch = str(run.get("head_branch") or "")
+    if head_branch == f"pull-request/{pr_number}":
+        return True
+    if source_branch and head_branch == source_branch:
+        return True
+    return bool(source_sha) and str(run.get("head_sha") or "") == source_sha
 
 
 def collect_runs_to_consider(
@@ -182,11 +200,16 @@ def collect_runs_to_consider(
     pr_number: int,
     closed: bool,
     get: Callable[..., Any] = github,
+    *,
+    source_branch: str = "",
+    source_sha: str = "",
 ) -> list[dict[str, Any]]:
     """Mirror-branch runs always; on close, also native PR-event runs.
 
-    SonarQube and other optional checks use ``pull_request`` / 
+    SonarQube and other optional checks use ``pull_request`` /
     ``pull_request_target`` on the contributor branch, not ``pull-request/N``.
+    ``source_branch`` / ``source_sha`` come from the close event; they are the
+    only attribution left once GitHub has emptied ``pull_requests``.
     """
     branch = f"pull-request/{pr_number}"
     seen: set[int] = set()
@@ -203,7 +226,9 @@ def collect_runs_to_consider(
             run_id = run.get("id")
             if not isinstance(run_id, int) or run_id in seen:
                 continue
-            if not run_belongs_to_pr(run, pr_number):
+            if not run_belongs_to_pr(
+                run, pr_number, source_branch=source_branch, source_sha=source_sha
+            ):
                 continue
             seen.add(run_id)
             runs.append(run)
@@ -310,7 +335,14 @@ def main() -> int:
 
     cancelled = 0
     pr_id = int(pr_number)
-    for run in collect_runs_to_consider(repo, pr_id, closed):
+    source_branch = os.environ.get("SOURCE_BRANCH", "").strip()
+    for run in collect_runs_to_consider(
+        repo,
+        pr_id,
+        closed,
+        source_branch=source_branch,
+        source_sha=source_sha,
+    ):
         run_id = run.get("id")
         if not isinstance(run_id, int):
             continue
--- a/.github/scripts/test_cancel_downstream_pipeline.py
+++ b/.github/scripts/test_cancel_downstream_pipeline.py
@@ -240,6 +240,25 @@ class SearchFallbackTest(unittest.TestCase):
                 open_func=open_func,
             )
 
+    def test_forbidden_variables_read_still_aborts_discovery(self):
+        def open_func(req):
+            raise HTTPError(
+                req.full_url,
+                403,
+                "Forbidden",
+                hdrs=None,  # type: ignore[arg-type]
+                fp=io.BytesIO(b"{}"),
+            )
+
+        with self.assertRaises(SystemExit):
+            module.fetch_pipeline_variables(
+                "https://gitlab.example/api/v4",
+                "token",
+                "1",
+                10,
+                open_func=open_func,
+            )
+
 
 class WorkflowWiringTest(unittest.TestCase):
     def test_ci_cancels_downstream_when_the_github_job_is_cancelled(self):
--- a/.github/scripts/test_cancel_downstream_pipeline.py
+++ b/.github/scripts/test_cancel_downstream_pipeline.py
@@ -353,6 +353,158 @@ class SearchFallbackTest(unittest.TestCase):
                 open_func=open_func,
             )
 
+    def test_variables_503_is_retried_then_finds_ours(self):
+        class JsonResponse:
+            def __init__(self, payload: object) -> None:
+                self._body = json.dumps(payload).encode("utf-8")
+
+            def read(self) -> bytes:
+                return self._body
+
+            def __enter__(self):
+                return self
+
+            def __exit__(self, *_args):
+                return False
+
+        reads = {"n": 0}
+
+        def open_func(req):
+            url = req.full_url
+            if "/pipelines?" in url:
+                if "status=running" in url:
+                    return JsonResponse(
+                        [{"id": 11, "created_at": "2026-08-28T08:12:31Z"}]
+                    )
+                return JsonResponse([])
+            if url.endswith("/pipelines/11/variables"):
+                reads["n"] += 1
+                if reads["n"] == 1:
+                    raise HTTPError(
+                        url,
+                        503,
+                        "Service Unavailable",
+                        hdrs=None,  # type: ignore[arg-type]
+                        fp=io.BytesIO(b"{}"),
+                    )
+                return JsonResponse(
+                    [{"key": module.CORRELATION_VARIABLE, "value": "gh-1-1-mine"}]
+                )
+            raise AssertionError(url)
+
+        self.assertEqual(
+            module.search_matching_pipeline_ids(
+                "https://gitlab.example/api/v4",
+                "token",
+                project="1",
+                ref="main",
+                correlation_id="gh-1-1-mine",
+                started_at="2026-08-28T08:12:00Z",
+                attempts=3,
+                delay=0,
+                open_func=open_func,
+            ),
+            [11],
+        )
+        self.assertEqual(reads["n"], 2)
+
+    def test_variables_5xx_exhausted_fails_cleanup(self):
+        class JsonResponse:
+            def __init__(self, payload: object) -> None:
+                self._body = json.dumps(payload).encode("utf-8")
+
+            def read(self) -> bytes:
+                return self._body
+
+            def __enter__(self):
+                return self
+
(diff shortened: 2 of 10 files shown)


zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33146152476)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33154463801)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33156349381)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33160801258)

hugoverjus:
Reviewed at `d03a8af`. Ran all the new tests locally — 22/22 in `test_cancel_downstream_pipeline.py`, 18/18 in `test_cancel_stale_pr_ci.py`, 11/11 in `test_trigger_downstream_pipeline.py`. Env plumbing resolves correctly in all three workflows: `DOWNSTREAM_CI_URL`, `DOWNSTREAM_CI_TOKEN`, `DOWNSTREAM_PROJECT_PATH` and `DOWNSTREAM_REF` are all set at job level, so the `cancelled()` cleanup step has what it needs.

The GitLab half is good work — the correlation-token design is the correct answer to cancel-racing-the-POST, and the handoff write is properly atomic (tmp + fsync + rename). The GitHub half has one design problem and one environmental one. Suggest splitting the two so the GitLab fix isn't held up behind them.

## 1. The close-path ownership check reads a field GitHub empties on close

`run_belongs_to_pr()` matches on `run["pull_requests"]`, and its fallback (`head_branch == f"pull-request/{pr_number}"`) cannot match a native `pull_request` run, whose `head_branch` is the contributor branch. So `pull_requests` is the only working path — and it is empty precisely when this code runs.

Sampling `GET /actions/runs?event=pull_request&per_page=100` on this repo and correlating each run's branch to its PR state:

- 22 of 22 runs with `pull_requests` populated → PR **OPEN**
- 9 of 10 runs with it empty → PR **MERGED**
- 9 of 9 `Cleanup PR Tags` runs (which only ever trigger on `closed`) → empty

GitHub drops the PR association when the PR closes, so on the `closed` event the SonarQube runs this is meant to cancel report no PR at all and are skipped.

The fixtures hide this: `test_closed_collects_native_pr_event_runs` hands closed-PR runs `"pull_requests": [{"number": 42}]`, which is a state that doesn't occur, so the suite passes while the production path can't work.

Suggest matching on something that survives close — `head_branch` against `github.event.pull_request.head.ref`, or `head_sha` against the `SOURCE_SHA` the workflow already passes in — and setting `"pull_requests": []` in the closed-PR fixtures.

## 2. `cancel-stale-pr-ci.yml` has never fired: `pull_request_target` needs the file on the default branch

- `GET /actions/runs?event=pull_request_target` → `total_count: 0`, against 6720 for `event=pull_request` and 819 for `event=workflow_run`, so the filter itself works.
- The workflow does not appear in `GET /actions/workflows` at all — GitHub has never seen it triggered.
- #1900 landed on `develop` on Aug 27, and #1911, #1908 and #1902 have merged since. Three `closed` events, zero runs.

The cause looks like default-branch presence rather than the event being disabled. Among workflows present on `develop` but absent from `main`, all six triggered by `push` or `pull_request` are registered and firing — including `cleanup-pr-tags.yml`, which is also absent from `main` — while both `pull_request_target` workflows, this one and `tracking-ref-check.yml`, are unregistered. `main` is currently 8 days behind `develop` (`9ba62a70`, Aug 20 vs `2eae638e`, Aug 28).

So this half activates only once `develop` reaches `main`, and cannot be validated before then. Two options, both your call: accept the delay, or follow the precedent already set in `cleanup-pr-tags.yml` and switch to `pull_request` — viable here because PRs are branched in-repo rather than from forks, but it gives up the fork-safety reasoning in the current file header, so it deserves a deliberate decision rather than a mechanical edit. Note that `tracking-ref-check.yml` is dormant for the same reason and probably wants its own issue. The header comment in `cleanup-pr-tags.yml` ("this repo has never fired a single pull_request_target run") already documents the symptom.

## GitLab half — nits only

- `persist_handoff` is called twice for the same data: once inside `trigger_pipeline()` right after the POST, then again in `main()`. Drop one. The in-`trigger_pipeline` call also moves file I/O into what was a pure request path.
- `persist_handoff` still records `commit_sha` and `variable_name`, which no consumer reads after `8f26390` switched matching to the correlation token.
- The comment at the end of `main()` — that `pipeline_id` / `project_id` "were already flushed above … so a cancel during this summary still has a handoff file" — conflates two mechanisms. Moving the `write_output` calls earlier buys nothing, because GitHub only publishes step outputs once the step completes; it's the handoff file that does the work. The comment reads as if the flush is what saves you.
- `DEFAULT_HANDOFF_PATH = ".ci/downstream-pipeline.json"` is workspace-relative, and these jobs run on `[self-hosted, downstream-pipeline]` where the workspace persists between runs. `ci.yml` wipes `.ci` in "Unpack source artifact", but a cancel during the preceding `download-artifact` step would have cleanup read a *previous* run's pipeline id. Mostly benign (a finished pipeline returns 404/409), but keying the filename by `GITHUB_RUN_ID`, or truncating it at job start, removes the ambiguity for free.
- Cosmetic: over-indented closing parens at lines 508 and 589 of `test_cancel_downstream_pipeline.py`, each followed by two blank lines inside the class body. Harmless as-is, since ruff is scoped to `services/agent` and doesn't lint `.github/scripts`.

## One process note

Nine of the ten commits map one-to-one onto a Greptile P1 comment. Each is individually defensible, but the result is 468 lines of production code and 685 of tests hardening a race window in a cancelled-job cleanup path, while the question of whether the workflow fires at all went unasked. Worth a conscious signoff on that complexity rather than inheriting it, since it's permanent maintenance surface.

.github/scripts/cancel_stale_pr_ci.py
```diff
@@ -149,6 +151,65 @@ def list_active_runs(
...
+    quoted = urllib.parse.quote(event, safe="")
+    return _collect_status_pages(repo, f"event={quoted}", get)
+
+
+def run_belongs_to_pr(run: dict[str, Any], pr_number: int) -> bool:
+    for pr in run.get("pull_requests") or []:
+        number = pr.get("number") if isinstance(pr, dict) else None
+        if number == pr_number or str(number) == str(pr_number):
+            return True
+    return str(run.get("head_branch") or "") == f"pull-request/{pr_number}"
```
hugoverjus: **Close-path attribution can never match: `pull_requests` is empty by the time this runs.**

`run_belongs_to_pr` has two keys and neither works on the close path:

- `run["pull_requests"]` — GitHub drops the PR association when the PR closes, and `PR_CLOSED=true` is precisely when native-event runs get collected.
- `head_branch == f"pull-request/{pr_number}"` — a native `pull_request` run's `head_branch` is the *contributor* branch, never the mirror ref, so this fallback cannot fire for the runs this PR is trying to reach.

Sampled on this repo via `GET /actions/runs?event=pull_request&per_page=100`, correlating each run's branch to its PR state: 22/22 runs with `pull_requests` populated had an **open** PR; 9/10 with it empty had a **merged** PR; and 9/9 `Cleanup PR Tags` runs — a workflow that only triggers on `closed` — had it empty.

So `collect_runs_to_consider` lists the native runs and then discards every one of them. The fix is to also match on keys the close event hands us: the source branch and the source head SHA. This is hunk 1 of 4; the rest are on `collect_runs_to_consider`, `main()`, and the workflow `env:`.

```suggestion
def run_belongs_to_pr(
    run: dict[str, Any],
    pr_number: int,
    source_branch: str = "",
    source_sha: str = "",
) -> bool:
    """Attribute a run to a PR using keys that survive the PR closing.

    GitHub empties ``pull_requests`` as soon as the PR closes, which is
    exactly when the close path runs, so it cannot be the only key. The
    ``pull-request/<N>`` fallback never matches a native ``pull_request``
    run either — that run's ``head_branch`` is the contributor branch. The
    source branch and head SHA carried by the close event are what remain.
    """
    for pr in run.get("pull_requests") or []:
        number = pr.get("number") if isinstance(pr, dict) else None
        if number == pr_number or str(number) == str(pr_number):
            return True
    head_branch = str(run.get("head_branch") or "")
    if head_branch == f"pull-request/{pr_number}":
        return True
    if source_branch and head_branch == source_branch:
        return True
    return bool(source_sha) and str(run.get("head_sha") or "") == source_sha
```

.github/scripts/cancel_stale_pr_ci.py
```diff
@@ -149,6 +151,65 @@ def list_active_runs(
...
+    for event in PR_EVENTS:
+        for run in list_active_runs_for_event(repo, event, get):
+            run_id = run.get("id")
+            if not isinstance(run_id, int) or run_id in seen:
+                continue
+            if not run_belongs_to_pr(run, pr_number):
+                continue
+            seen.add(run_id)
+            runs.append(run)
+    return runs
```
hugoverjus, line 235: Hunk 2 of 4 — thread the surviving keys through to `run_belongs_to_pr`.

`get` keeps its fourth positional slot so existing positional callers and tests (`collect_runs_to_consider("owner/repo", 42, True, get)`) still work; the new parameters are keyword-only.

```suggestion
def collect_runs_to_consider(
    repo: str,
    pr_number: int,
    closed: bool,
    get: Callable[..., Any] = github,
    *,
    source_branch: str = "",
    source_sha: str = "",
) -> list[dict[str, Any]]:
    """Mirror-branch runs always; on close, also native PR-event runs.

    SonarQube and other optional checks use ``pull_request`` /
    ``pull_request_target`` on the contributor branch, not ``pull-request/N``.
    ``source_branch`` / ``source_sha`` come from the close event; they are the
    only attribution left once GitHub has emptied ``pull_requests``.
    """
    branch = f"pull-request/{pr_number}"
    seen: set[int] = set()
    runs: list[dict[str, Any]] = []
    for run in list_active_runs(repo, branch, get):
        run_id = run.get("id")
        if isinstance(run_id, int) and run_id not in seen:
            seen.add(run_id)
            runs.append(run)
    if not closed:
        return runs
    for event in PR_EVENTS:
        for run in list_active_runs_for_event(repo, event, get):
            run_id = run.get("id")
            if not isinstance(run_id, int) or run_id in seen:
                continue
            if not run_belongs_to_pr(run, pr_number, source_branch, source_sha):
                continue
            seen.add(run_id)
            runs.append(run)
    return runs
```

.github/scripts/cancel_stale_pr_ci.py
```diff
@@ -247,7 +309,8 @@ def main() -> int:
         return 0
 
     cancelled = 0
-    for run in list_active_runs(repo, branch):
+    pr_id = int(pr_number)
+    for run in collect_runs_to_consider(repo, pr_id, closed):
```
hugoverjus: Hunk 3 of 4 — read `SOURCE_BRANCH` and pass both keys down. `source_sha` is already resolved above (line 296); `SOURCE_BRANCH` is new and needs adding to the workflow step `env:` (hunk 4).

Both stay optional here: with neither set, behaviour is identical to this branch today, so the script remains runnable against an older workflow file.

```suggestion
    pr_id = int(pr_number)
    source_branch = os.environ.get("SOURCE_BRANCH", "").strip()
    for run in collect_runs_to_consider(
        repo,
        pr_id,
        closed,
        source_branch=source_branch,
        source_sha=source_sha,
    ):
```

.github/workflows/cancel-stale-pr-ci.yml
```diff
@@ -3,8 +3,10 @@
 #
 # PR CI runs on copy-pr-bot `pull-request/<N>` pushes, not on the source PR
 # event. Concurrency in those workflows only fires when the *mirror* moves.
-# A rebase of the source PR otherwise leaves the previous mirror SHA running
-# until the next `/ok to test`.
+# A rebase or new commit on the source PR otherwise leaves the previous
+# mirror SHA running until the next `/ok to test`. Close (merge or
+# abandon) cancels leftover PR runs, including non-required jobs, but
+# not `develop`/`main` delivery CI.
```
hugoverjus, line 9: Hunk 4 of 4 — **the step `env:` needs `SOURCE_BRANCH`.** Anchored here because these header lines are the only part of this file inside the diff, so this one cannot be a clickable suggestion; the change belongs at line 46, beside the `SOURCE_SHA` already passed:

```yaml
          PR_NUMBER: ${{ github.event.pull_request.number }}
          SOURCE_SHA: ${{ github.event.pull_request.head.sha }}
          SOURCE_BRANCH: ${{ github.event.pull_request.head.ref }}
          PR_CLOSED: ${{ github.event.action == 'closed' }}
```

Without it, matching falls back to SHA only, which misses leftover runs from *earlier* pushes on the branch — the runs most likely to still be burning a runner at merge time. The job's `if:` guard already establishes that `head.ref` is not the mirror ref, so there is no ambiguity about what it holds.

.github/scripts/test_cancel_stale_pr_ci.py
```diff
@@ -178,6 +188,66 @@ def get(_method: str, _repo: str, path: str):
...
+                            "name": "CI",
+                            "status": "in_progress",
+                            "head_branch": "pull-request/42",
+                        }
+                    ]
+                }
+            return {"workflow_runs": []}
+
+        runs = module.collect_runs_to_consider("owner/repo", 42, True, get)
+        self.assertEqual(sorted(run["id"] for run in runs), [3, 8])
```
hugoverjus, line 262: **These fixtures encode a state that never occurs, which is why the suite is green while the close path cannot work.**

`test_closed_collects_native_pr_event_runs` hands closed-PR runs `"pull_requests": [{"number": 42}]`. On a closed PR that field is empty (evidence in the comment on `run_belongs_to_pr`), so the test exercises a branch production never reaches.

This rewrite drops the impossible fixture, adds a direct closed-state case for `run_belongs_to_pr`, and keys the collection test off `source_branch`. It goes with the three `cancel_stale_pr_ci.py` suggestions and only passes with them: on today's matcher, an empty `pull_requests` collects run 3 alone. Verified locally — full file green with all four applied.

```suggestion
    def test_run_belongs_to_pr_via_pull_requests_field(self):
        sonar = {
            "id": 8,
            "name": "SonarQube Analysis",
            "status": "in_progress",
            "head_branch": "feat/foo",
            "pull_requests": [{"number": 1900}],
        }
        self.assertTrue(module.run_belongs_to_pr(sonar, 1900))
        self.assertFalse(module.run_belongs_to_pr(sonar, 1))

    def test_run_belongs_to_pr_after_close_empties_pull_requests(self):
        # The shape the close path actually sees: GitHub drops the PR
        # association the moment the PR closes.
        sonar = {
            "id": 8,
            "name": "SonarQube Analysis",
            "status": "in_progress",
            "head_branch": "feat/foo",
            "head_sha": "a" * 40,
            "pull_requests": [],
        }
        self.assertFalse(module.run_belongs_to_pr(sonar, 1900))
        self.assertTrue(
            module.run_belongs_to_pr(sonar, 1900, source_branch="feat/foo")
        )
        self.assertTrue(
            module.run_belongs_to_pr(sonar, 1900, source_sha="a" * 40)
        )
        self.assertFalse(
            module.run_belongs_to_pr(sonar, 1900, source_branch="feat/other")
        )

    def test_closed_collects_native_pr_event_runs(self):
        def get(_method: str, _repo: str, path: str):
            if "event=pull_request&" in path and "status=in_progress" in path:
                return {
                    "workflow_runs": [
                        {
                            "id": 8,
                            "name": "SonarQube Analysis",
                            "status": "in_progress",
                            "head_branch": "feat/foo",
                            "pull_requests": [],
                        },
                        {
                            "id": 9,
                            "name": "SonarQube Analysis",
                            "status": "in_progress",
                            "head_branch": "feat/other",
                            "pull_requests": [],
                        },
                    ]
                }
            if "branch=pull-request%2F42" in path and "status=in_progress" in path:
                return {
                    "workflow_runs": [
                        {
                            "id": 3,
                            "name": "CI",
                            "status": "in_progress",
                            "head_branch": "pull-request/42",
                        }
                    ]
                }
            return {"workflow_runs": []}

        runs = module.collect_runs_to_consider(
            "owner/repo", 42, True, get, source_branch="feat/foo"
        )
        self.assertEqual(sorted(run["id"] for run in runs), [3, 8])
```

hugoverjus (commented):
Read the whole change. The GitLab half — per-attempt correlation token, fsynced handoff, the skip/retry semantics in discovery — is sound, and I only have nits on it; those and the one real bug are inline as clickable suggestions. Two findings that are not code changes:

**1. The close path has never actually fired.** `cancel-stale-pr-ci.yml` triggers on `pull_request_target`, and that workflow appears never to have been registered. `GET /actions/runs?event=pull_request_target` returns `total_count: 0` — against 6720 for `event=pull_request` and 819 for `event=workflow_run` — and the workflow does not appear in `GET /actions/workflows` at all. #1900 added it to `develop` on Aug 27, and #1911, #1908 and #1902 have merged since: three `closed` events, zero runs.

The cause looks like default-branch presence. Among the workflows that exist on `develop` but not on `main`, all six triggered by `push` or `pull_request` are registered and firing — including `cleanup-pr-tags.yml`, which is also absent from `main` — while both `pull_request_target` workflows, this one and `tracking-ref-check.yml`, are unregistered. `main` is eight days behind `develop` (`9ba62a70`, Aug 20 vs `2eae638e`, Aug 28).

So the GitHub half of this PR only activates once `develop` reaches `main`. Which way to go is your call: accept the delay, or follow `cleanup-pr-tags.yml`'s precedent and switch to `pull_request`. The second is viable here because PRs are branched in-repo rather than from forks, but it gives up the fork-safety reasoning in this file's own header, so it deserves a deliberate decision rather than a mechanical edit. `tracking-ref-check.yml` is dormant for the same reason and probably wants its own issue.

**2. Scope.** Nine of the ten commits map one-to-one onto a Greptile P1 comment. Each is individually defensible, but the result is 468 lines of production code and 685 lines of tests hardening a race window in a cancelled-job cleanup path. Worth a conscious signoff on that as permanent maintenance surface, rather than letting it accrete one comment at a time.

Inline below: four hunks for a close-path bug that means the new native-run cancellation cannot currently match anything, plus five nits on the GitLab side and two optional formatting ones.

.github/scripts/trigger_downstream_pipeline.py
```diff
@@ -167,8 +173,18 @@ def trigger_pipeline(
...
+    response = request_json(
+        "Pipeline trigger",
+        f"{base_url}/projects/{project_id}/pipeline",
+        token,
+        data=payload,
+    )
+    pipeline_id = str(response.get("id") or "")
+    if pipeline_id:
+        persist_handoff(pipeline_id=pipeline_id)
+    return response
```
hugoverjus, line 184: Nit: `persist_handoff` runs twice for the same data. `main()` calls `persist_handoff(project_id=..., pipeline_id=...)` on the statement right after this returns (line 579), so this write is redundant — and it puts file I/O inside what was a pure request helper.

Dropping it here keeps a single write, still immediately after the POST returns. No test calls `trigger_pipeline`, so nothing in the suite depends on the write happening here.

```suggestion
    return request_json(
        "Pipeline trigger",
        f"{base_url}/projects/{project_id}/pipeline",
        token,
        data=payload,
    )
```

.github/scripts/trigger_downstream_pipeline.py
```diff
@@ -378,6 +402,59 @@ def write_output(key: str, value: str) -> None:
...
+    """One-off token identifying a single trigger attempt."""
+    run_id = os.environ.get("GITHUB_RUN_ID", "local").strip() or "local"
+    attempt = os.environ.get("GITHUB_RUN_ATTEMPT", "1").strip() or "1"
+    return f"gh-{run_id}-{attempt}-{uuid.uuid4().hex}"
+
+
+def handoff_path() -> str:
+    return os.environ.get("DOWNSTREAM_HANDOFF_PATH", DEFAULT_HANDOFF_PATH).strip() or (
+        DEFAULT_HANDOFF_PATH
+    )
```
hugoverjus: Nit, stale-state risk: `DEFAULT_HANDOFF_PATH` is workspace-relative and these jobs run on `[self-hosted, downstream-pipeline]`, where the workspace survives between runs. `ci.yml` wipes `.ci` in "Unpack source artifact", but a cancel during the preceding `download-artifact` step leaves the *previous* run's file in place and cleanup would read that pipeline id. Mostly benign — a finished pipeline answers 404/409 — but it is a cross-run read.

Keying the default by run id closes the window. `cancel_downstream_pipeline.py` imports this function rather than duplicating the path, and both steps run in the same job, so the two sides stay in agreement; the `DOWNSTREAM_HANDOFF_PATH` override still wins first, so the tests are unaffected. Truncating `.ci/downstream-pipeline.json` at job start would do just as well.

```suggestion
def handoff_path() -> str:
    override = os.environ.get("DOWNSTREAM_HANDOFF_PATH", "").strip()
    if override:
        return override
    run_id = os.environ.get("GITHUB_RUN_ID", "").strip()
    attempt = os.environ.get("GITHUB_RUN_ATTEMPT", "").strip()
    if run_id:
        suffix = f"{run_id}-{attempt}" if attempt else run_id
        return f".ci/downstream-pipeline-{suffix}.json"
    return DEFAULT_HANDOFF_PATH
```

.github/scripts/trigger_downstream_pipeline.py
```diff
@@ -378,6 +402,59 @@ def write_output(key: str, value: str) -> None:
...
+
+
+def persist_handoff(**fields: object) -> None:
+    """Record GitLab trigger identity on disk for a cancelled cleanup step.
+
+    GitHub only publishes ``GITHUB_OUTPUT`` when the step finishes. The
+    pipeline id is written the instant GitLab returns it. ``ref`` /
+    ``commit_sha`` / ``trigger_started_at`` are written *before* POST so
+    cleanup can still find the pipeline if cancel lands in the HTTP window.
+    """
```
hugoverjus, line 436: Follow-on to the dead-fields nit below: the docstring names `commit_sha` as one of the pre-POST fields, but no consumer reads it, and `correlation_id` — which cleanup *does* read, and which is the whole point of the recovery path — goes unmentioned.

```suggestion
def persist_handoff(**fields: object) -> None:
    """Record GitLab trigger identity on disk for a cancelled cleanup step.

    GitHub only publishes ``GITHUB_OUTPUT`` when the step finishes, so the
    ids go to disk instead. ``ref`` / ``correlation_id`` /
    ``trigger_started_at`` are written *before* POST so cleanup can still
    find the pipeline if cancel lands in the HTTP window; ``pipeline_id``
    is written the instant GitLab returns it.
    """
```

.github/scripts/trigger_downstream_pipeline.py
```diff
@@ -469,6 +547,20 @@ def main() -> int:
...
+        persist_handoff(
+            project_id=project_id,
+            ref=ref,
+            commit_sha=commit_sha,
+            variable_name=variable_name,
+            correlation_id=correlation_id,
+            trigger_started_at=datetime.now(timezone.utc).strftime(
+                "%Y-%m-%dT%H:%M:%SZ"
+            ),
+        )
```
hugoverjus, line 565: Nit: `commit_sha` and `variable_name` are dead payload. Since `8f26390` moved matching to the correlation token, `cancel_downstream_pipeline.py` reads only `project_id`, `pipeline_id`, `correlation_id`, `ref` and `trigger_started_at` (`resolve_pipeline_ids`, plus `main`'s recovery path). Writing the submodule SHA and variable name only puts more into a file the cleanup step never consults.

```suggestion
        persist_handoff(
            project_id=project_id,
            ref=ref,
            correlation_id=correlation_id,
            trigger_started_at=datetime.now(timezone.utc).strftime(
                "%Y-%m-%dT%H:%M:%SZ"
            ),
        )
```

.github/scripts/trigger_downstream_pipeline.py
```diff
@@ -525,14 +622,11 @@ def main() -> int:
...
         write_summary("\n".join(summary_lines))
 
-        # Expose identifiers to the poll step in the same job. Do NOT
-        # write the pipeline URL here - it is a secret and would appear
-        # in any caller that echoes the output.
-        write_output("pipeline_iid", pipeline_iid)
-        write_output("pipeline_id", pipeline_id)
+        # Remaining identifiers for the poll step. pipeline_id / project_id
+        # were already flushed above, immediately after GitLab created the
+        # pipeline, so a cancel during this summary still has a handoff file.
```
hugoverjus: This comment conflates the two mechanisms. GitHub only publishes a step's outputs once the step completes, so moving these `write_output` calls earlier does nothing for a cleanup step that runs after a cancel — yet "already flushed above … so a cancel during this summary still has a handoff file" reads as if the flush is what provides the cover. It is the handoff file, and only the handoff file, that does.

```suggestion
        # Remaining identifiers for the poll step. Ordering these writes
        # buys nothing on cancel: GitHub only publishes a step's outputs
        # once the step completes. What covers a cancel during this summary
        # is the fsynced handoff file, written right after GitLab created
        # the pipeline.
```

.github/scripts/test_cancel_downstream_pipeline.py
```diff
@@ -0,0 +1,685 @@
...
+                ref="main",
+                correlation_id="gh-1-1-mine",
+                started_at="2026-08-28T08:12:00Z",
+                open_func=open_func,
+            ),
+            [11],
+            )
+
+
+    def test_empty_retry_after_transient_still_fails_cleanup(self):
```
hugoverjus, line 510: Optional nit, formatting only — ruff is scoped to `services/agent` and does not lint `.github/scripts`, so nothing flags this: the closing paren of `assertEqual` sits at 12 spaces instead of 8, and there are two blank lines inside the class body. Anchor includes the next `def` so the suggestion applies cleanly.

```suggestion
            [11],
        )

    def test_empty_retry_after_transient_still_fails_cleanup(self):
```

.github/scripts/test_cancel_downstream_pipeline.py
```diff
@@ -0,0 +1,685 @@
...
+                started_at="2026-08-28T08:12:00Z",
+                attempts=2,
+                delay=0,
+                open_func=open_func,
+            ),
+            [],
+            )
+
+
+    def test_paginates_past_the_first_twenty_on_a_busy_ref(self):
```
hugoverjus, line 590: Optional nit — same over-indented paren and double blank line as above.

```suggestion
            [],
        )

    def test_paginates_past_the_first_twenty_on_a_busy_ref(self):
```

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33457670505)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33458157892)

hugoverjus (approved):

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33559331262)

PR #1647, opened 2026-08-11
Onboard vss-rt-cv to the GitHub-first GHCR build

ashwind-nvidia:
Flips vss-rt-cv from external-pin to a GHCR-built image, following .github/github-first.md and the vss-behavior-analytics / vss-video-analytics-api reference implementations.

WORKAROUND -- the base image. The existing x86/aarch64/sbsa Dockerfiles layer this app onto nvcr.io/nvstaging/vss-core/vss-rt-cv, i.e. they build vss-rt-cv from vss-rt-cv. Every nvcr.io/nvidia/vss-core and nvstaging/vss-core repository reports canGuestPull=false, so that base needs NGC credentials, and by design no NGC credentials exist in GitHub Actions. docker/Dockerfile therefore builds on the public DeepStream SDK (canGuestPull=true, multi-arch). Verified locally: compiles clean on both linux/amd64 and linux/arm64 (the latter under QEMU, with -DPLATFORM_TEGRA actually exercised), producing a fully linked binary in each case.

CONSEQUENCE: the resulting image is a valid BUILD artifact but is NOT yet a runtime replacement for the NGC image -- it carries the compiled perception binary and none of the models, TRT engines, tracker configs or ds-start.sh. It is enough to put vss-rt-cv on the shared GHCR coordinate and exercise the build/tag/release-set flow. Do not treat develop-latest as deployable until that gap is closed. The sibling Dockerfiles are left untouched so the private-base local build path still works.

Changes:
- container-inventory.json: strategy build + ghcr_build true, with source_path/context/dockerfile and linux/amd64+linux/arm64.
- containers.env: VSS_RT_CV_IMAGE/_TAG, per-image first with the shared VSS_CONTAINER_* as fallback. Replaces PERCEPTION_IMAGE/PERCEPTION_TAG, which were unnamespaced and split across containers.env and rtvi.env.
- compose: three image lines onto the shared coordinate, develop-latest as the inline literal.
- PERCEPTION_* purged across 23 files (profiles, blueprint_config.yml, orchestrator MCP config, an agent unit test, skill reference docs). rtvi.env's shadowing declaration removed.
- .ignore_source_code_check added; vss-rt-cv registered in check_container_tag_source.py.
- compose-images.golden regenerated: three lines move nvstaging -> ghcr.io.

[...]

PR #1918, opened 2026-08-28
fix(skill-eval): redact secrets from everything the harness posts to GitHub

zac-wang-nv:
## Summary

A judge proved *"the agent printed NGC_CLI_API_KEY"* by quoting the key in its rationale, and the harness
shipped that rationale verbatim through every public surface: `verifier/judge.json` and
`verifier/test-stdout.txt` in the run artifact, and the PR comment `leg_report.py` renders from them.
**PR #1647 run 32535909071 published a full 70-char NGC key that way** — live in the public artifact, and in
the PR comment for six days until manually redacted. (Key rotation and artifact deletion are being handled
separately; this PR is prevention.) The tarball already excludes `agent/` trajectories for exactly this leak
class (PR #516) — the verifier files and rendered reports stayed in.

A sweep of all 292 live artifacts in the affected eval family found exactly one containing the full key and
one containing a 4-char prefix; 32 more contain only the harmless literal `nvapi-`.

## What this does

Gates every path from the harness to GitHub with TruffleHog (the same scanner the repo's pre-commit hook
already trusts) plus always-on builtin layers:

| Boundary | Gate |
|---|---|
| Results tree → PR comment | `leg_report.py` runs `redact_secrets.redact_tree()` **before reading anything** — the comment posts during the agent step, long before the workflow pack step, so gating only the tarball would leave the comment exposed |
| Results tree → run artifact | `skills-eval.yml` / `skills-eval-daily.yml` run the same CLI gate immediately before `tar czf`, covering legs that never reached a report |
| Rendered comment + leg summary | body shape-scrubbed; summary scrubbed **as a structure before serialization** (`json.dumps` escaping defeats substring replacement afterwards) |
| Judge's own outputs | self-contained scrub in `generic_judge.py` (the verifier container has neither the module nor a scanner): stdout header, per-check prints, and the whole `judge.json` object at serialization |

`redact_secrets.py` semantics:

- **`--no-verification`**: verification would send candidate secrets to their providers — itself exfiltration — and a candidate is worth masking whether or not it still works.
- **scan → replace literals → rescan → quarantine → verify**: findings that survive literal replacement (base64/UTF-16-encoded, composite `RawV2`, binary) get their file stubbed out, and a final scan must come back empty or the publish refuses.
- **Fail closed**: no scanner, timeout, scan error, or surviving findings → nonzero exit. The workflow step stops before `tar` (the upload finds nothing), and `skills_eval_agent.py` already fail-closes a leg whose report crashes, so a scanner outage reads as BLOCKED, not green.
- Builtin layers run unconditionally: the `nvapi-` token shape, and exact values of env vars whose underscore-delimited name contains KEY/TOKEN/SECRET/PASSWORD/CREDENTIAL (≥8 chars; component match, so `KEYCLOAK_URL` is not a secret).
- Docker fallback pins `trufflesecurity/trufflehog:3.94.2` — the pre-commit rev — not `latest`.
- Markers are `[REDACTED]` / `nvapi-[REDACTED]`; `<redacted>` would parse as an HTML tag in a PR comment.

## Verification

- The real leaked `judge.json` from run 32535909071, planted in a tree and run through the CLI with
  TruffleHog via docker: scanner finds the key, both occurrences replaced, zero `nvapi-` tokens remain, exit 0.
- 15 new tests (`test_secret_redaction.py`) cover each boundary end to end: judge `main()` with a
  secret-bearing spec asserting captured stdout + `judge.json` + reward; quarantine convergence with a
  binary survivor and the mandatory final verify scan; refuse-to-publish when findings survive quarantine;
  fail-closed CLI exit; the report gate running before `collect_leg`; pre-serialization structure scrubbing
  beating JSON escaping; docker mount path mapping; `KEYCLOAK_URL` not treated as a secret.
- Full harness suite: 242 pass; the 3 failures in `test_run_leg.py::PoolCandidates` pre-exist on pristine
  `origin/develop` (they are #1878's unmerged fixes — verified by stash/run/pop).

## Review

Reviewed twice by the OpenAI Codex CLI before posting, told the diff was Claude-generated; both rounds
reshaped it. Round 1 found the first version scrubbed the results list but still printed the raw query and
dumped it into `judge.json`, missed dict keys, used a substring env-name match (`KEYCLOAK_URL` would have
been masked), and used an HTML-swallowed marker. Round 2 found the structural bypass (comment posts before
the pack step), that `--results=verified,unknown` excludes unverified candidates, that scanner failures were
converted into success, that TruffleHog `Raw`/`RawV2` values are not necessarily literal substrings (hence
quarantine + verify), that binary findings were deliberately shipped, and that post-serialization scrubbing
misses JSON-escaped values. All addressed as described above.

- Enable GitHub secret scanning on the repo — it is currently disabled, so no partner auto-revocation fired
  on the leaked NGC key.
- The Harbor viewer on eval boxes serves `/tmp/skill-eval/results/` unauthenticated to the public internet;
  `leg_report`'s tree mutation covers what it serves after a report runs, but trajectories under `agent/`
  remain exposed there while a box lives.

## Checklist
- [x] New or existing tests cover these changes — 15 new boundary tests; full harness suite passes.

### DIFF
--- a/.github/skill-eval/leg_report.py
+++ b/.github/skill-eval/leg_report.py
@@ -34,6 +34,8 @@ import math
 import os
 import re
 import sys
+
+import redact_secrets
 from pathlib import Path
 
 TRACE_URLS_NAME = "trace-urls.tsv"
@@ -650,6 +652,14 @@ def main(argv: list[str] | None = None) -> int:
                          "use instead of parsing the rendered Markdown")
     args = ap.parse_args(argv)
 
+    # Redact the tree BEFORE reading it: this render is what the PR comment
+    # is posted from, during the agent step — the workflow's pack-step scan
+    # runs later and cannot protect the comment. Scanner failure propagates:
+    # skills_eval_agent fail-closes a leg whose report crashes, and an
+    # unscanned tree must read as BLOCKED, not as green.
+    for line in redact_secrets.redact_tree(args.results_root):
+        print(f"redact_secrets: {line}", file=sys.stderr)
+
     leg = collect_leg(args.results_root)
     if not leg["trials"]:
         print("no trials under results root", file=sys.stderr)
@@ -667,11 +677,19 @@ def main(argv: list[str] | None = None) -> int:
     body = render_comment(leg, spec_path=args.spec_path, platform=args.platform,
                           head_sha=args.head_sha, spec_sha=args.spec_sha,
                           declared=declared)
+    # Both writes leave the runner for GitHub (comment body, artifact
+    # summary). The results tree was already TruffleHog-redacted above; this
+    # covers what rendering itself could reintroduce. The summary is scrubbed
+    # as a STRUCTURE before serialization — json.dumps escapes quotes,
+    # backslashes and non-ASCII, after which an env value no longer occurs
+    # literally in the document and substring replacement would miss it.
+    body = redact_secrets.redact_text(body)
     if args.summary_json:
         args.summary_json.parent.mkdir(parents=True, exist_ok=True)
         args.summary_json.write_text(json.dumps(
-            leg_summary(leg, declared=declared, spec_path=args.spec_path,
-                        platform=args.platform), indent=2, default=str))
+            redact_secrets.redact_obj(
+                leg_summary(leg, declared=declared, spec_path=args.spec_path,
+                            platform=args.platform)), indent=2, default=str))
     if args.out:
         args.out.parent.mkdir(parents=True, exist_ok=True)
         args.out.write_text(body)
--- a/.github/skill-eval/redact_secrets.py
+++ b/.github/skill-eval/redact_secrets.py
@@ -180,6 +180,28 @@ def _finding_file(finding: dict, target: Path) -> Path:
     return resolved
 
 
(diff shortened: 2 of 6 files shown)


zac-wang-nv:
Follow-up pushed: a final Codex review of the shipped commit found five remaining fail-opens (multipart RawV2 blinding the rescan, a skippable verification scan, scan errors exiting 0 without `--fail-on-scan-errors`, reruns uploading a stale pre-redaction tarball, and binaries bypassing the builtin layer). All five closed in the second commit, with tests for each. Live end-to-end against the real leaked `judge.json` still passes.

zac-wang-nv:
Round-4 Codex review blocked again and was right: the multipart guard quarantined only zero-match findings, so the real AWS shape (Raw=id replaced, RawV2=id:secret unmatched) still blinded the rescan and shipped the secret half — and the prior test only covered zero-match. Third commit judges each finding on its complete representation, neutralizes symlinks before anything runs (a link would have let the gate write outside the tree), pins every finding to an in-tree file before any rewrite, and fixes a test that asserted its own mock. 30 boundary tests; live e2e with a planted symlink passes.

PR #1915, opened 2026-08-28
feat(skill-eval): let legs opt into prebaked VIOS packages

simings-nv:
## Description

[#1798](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1798) moves the VIOS runtime-media apt install from every container start to image build time, behind `VSS_VIOS_PREBAKE_PACKAGES`. Skill-eval legs cannot currently reach it.

Setting it in the workflow alone would not have worked. A leg's deploy does not run on the runner: it runs on the Brev instance, where the environment is `~/.eval_env` and only names on `brev_env.py`'s allowlist are written there. So this does both, and adds a test that they cannot drift apart.

```
skills-eval.yml     VSS_VIOS_PREBAKE_PACKAGES: ${{ vars.VSS_VIOS_PREBAKE_PACKAGES }}
brev_env.py         + "VSS_VIOS_PREBAKE_PACKAGES" on the forwarding allowlist
```

## Nothing changes until the variable is set

The repository variable is unset, which yields an empty value, which `brev_env.py` does not forward, which leaves `dev-profile.sh` on its default of `false`. Every leg behaves exactly as it does today. Driving it from a repository variable rather than a literal means turning it on or off later is a Settings change, not another PR.

## Why a test

The half-wired state is silent by construction: the workflow looks correct, the deploy quietly keeps its default, and nothing errors. `test_eval_env_forwarding.py` asserts that every `VSS_*` key in the workflow's `env:` block is on the forwarding allowlist, so this class of mistake fails in CI rather than showing up as a change that mysteriously does nothing.

Verified it fails when either side is removed. It reads both files as text, with no YAML dependency, so it runs in the harness-contracts step whose uvx environment carries only pytest.

## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/HEAD/CONTRIBUTING.md).
- [x] I have installed and run [pre-commit hooks](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/HEAD/CONTRIBUTING.md#1-enable-pre-commit-hooks) locally.
- [x] Every commit on this PR is [DCO sign-off](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/HEAD/CONTRIBUTING.md#developer-certificate-of-origin-dco)'d.
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.

PR #1920, opened 2026-08-28
feat(vios): complete the package-prebake loop — NVStreamer, nightly, and per-leg observability

simings-nv:
## Description

Rounds out [#1798](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1798) so `--prebake-vios-packages` covers everything that installs the runtime-media packages at container start, and so the effect is observable where it runs. Measured on an L40, per container start:

| container | install at start | prebaked |
|---|---|---|
| vss-vios-streamprocessing | 116.3 s | 0.17 s |
| vss-vios-nvstreamer | 144.2 s | 0.15 s |

Confirmed on live CI (GitLab pipeline 65243977, the first run where the flag reached `dev-profile.sh`): `test-search`'s container went Started→healthy in **2 s against a 30 s baseline** built from 74 prior jobs.

### NVStreamer prebake
A counterpart overlay to the streamprocessing one resolves `nvstreamer-base` **and every variant** to a local prebaked image — every variant, because compose resolves `extends` at file load, before overlays merge, so overriding only the base leaves the variants on the registry image. The entrypoint needs no change: the baked marker makes its install branch skip. The Dockerfile's base ARG becomes `VSS_VIOS_PREBAKE_BASE` now that it bakes more than one image. Golden image refs regenerated for the new overlay.

Verified by resolving the full compose model with and without the overlays: with them, `nvstreamer` and `streamprocessing-ms` land on the `-prebaked` images while `vios-apt-cache-init` stays on the registry image; without them, byte-identical to today.

### Warehouse path
`blueprint-deploy.sh` assembles its own compose file list, so the warehouse eval never saw the flag regardless of settings. It now applies the same two overlays under the same environment opt-in (it has no flag surface, and CI reaches it through two containers where only environment survives). This is the heaviest deploy in CI — its container-readiness baseline is 247 s.

### Nightly sweep
`skills-eval-daily.yml` is a separate workflow with its own `env:` block, so #1915 left the sweep where the VIOS specs actually run on the old path. Same repository variable; a parity test asserts both workflows declare the same `VSS_*` knobs and fails when either side is removed.

### Per-leg observability
A leg's deploy runs on the Brev instance and none of its output reaches the job log or artifact, so the flag was enabled on GitHub and unverifiable. `vios_probe.py` records the container's own verdict (`prebaked` / `installed-at-start` / `unknown`, read from its logs) and timing to `/logs/artifacts/vios-ready.log`, written at collection time so it cannot land after its own pickup, and never exits non-zero so a diagnostic cannot fail the trial it observes.

Nothing changes until the repository variable is set; unset is not forwarded and every default stands.

## Checklist

### DIFF
--- a/.github/skill-eval/envs/brev_env.py
+++ b/.github/skill-eval/envs/brev_env.py
@@ -765,6 +765,39 @@ echo "synced $REPO to $(git rev-parse --short HEAD)"
             self._instance_name, (result.stdout or "").strip().splitlines()[-1] if result.stdout else "<no output>",
         )
 
+    async def _probe_vios_ready(self, label: str) -> None:
+        """Record what the VIOS container cost to start, into the trial artifact.
+
+        Prebake (#1798) moves a 224-package apt install out of container start.
+        Enabling it was one variable; seeing whether it did anything was not,
+        because the deploy runs here and none of its output reaches the job log
+        or the artifact. This writes the container's own verdict and timing to
+        /logs/artifacts/vios-ready.log, which harbor already collects.
+
+        NEVER raises: a probe fault must not perturb the trial it observes.
+        """
+        try:
+            try:
+                from envs import vios_probe  # normal harbor import path
+            except ImportError:
+                import vios_probe  # PYTHONPATH=.github/skill-eval fallback
+            result = await _run_brev_exec(
+                self._instance_name, vios_probe.build_probe_command(label),
+                timeout=90,
+            )
+            lines = vios_probe.parse_probe_lines(result.stdout or "")
+            if not lines:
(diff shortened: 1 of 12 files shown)


zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33277128351)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33277470060)

PR #1738, opened 2026-08-17
skill-eval: align NemoClaw sandbox lifecycle with remote inference

gbau1:
## Scope

Fix the NemoClaw-specific sandbox onboarding failure in the Harbor skill-eval harness.

The final diff only:
- pins the harness to NemoClaw v0.0.109,
- makes the NemoClaw agent sandbox CPU-only (`NEMOCLAW_SANDBOX_GPU=0`),
- retires the harness-owned sandbox through NemoClaw before the shared Brev/Docker reset, and
- covers these contracts with harness tests.

No worker quarantine, retry loop, worker-selection override, Docker cleanup/repair, skill, adapter, compose, notebook, workflow, or verifier changes are included.

## Root cause

The harness uses a remote inference endpoint, but the coordinator supplied `NEMOCLAW_SANDBOX_GPU=1`. VSS workloads use the Brev host GPUs through the orchestrator MCP; the NemoClaw agent sandbox itself does not need GPU passthrough.

A cancelled/reset run could also leave NemoClaw lifecycle state after Docker state disappeared. The preflight teardown searched the host user's PATH, while the notebook installs `nemoclaw` in the isolated harness home (`~/.skill-eval/nemoclaw-home/.local/bin`). It therefore treated a missing CLI as a clean worker and the next onboarding resumed stale GPU-enabled state.

The harness now owns a CPU-only sandbox topology, resolves the CLI from its isolated home, and invokes NemoClaw's supported idempotent `destroy --force --cleanup-gateway` lifecycle before the shared reset.

## Validation

- Harness tests: 20 passed
- Python compile check passed
- `git diff --check` passed
- Merged current `develop` without conflicts; net diff remains three NemoClaw harness files.
- [Targeted validation 32180482692](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32180482692), [base-profile job](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32180482692/job/95855326071): reran on the exact previously failing worker `vss-eval-l40s-1g`. NemoClaw completed with no exception, produced token metrics (3.21M input / 3.13M cached / 16.4K output), and the standard verifier returned reward 0.6. The job is red only because two skill checks failed; the harness and reporting path completed normally.
- [Merged-head all-skills sweep 32183708669](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/32183708669) is in progress.

[...]

PR #1942, opened 2026-09-01
feat(vss-build-vision-ai): add the host-side NemoClaw agent harness

nvleeryan:
## Summary

Fixes VIA-2737

Adds a **host-side NemoClaw sandbox** as a second way to drive a `vss-build-vision-ai`
build, alongside the in-stack `vss-agent`, and fixes the environment resolution
problems found by actually running the skill end to end on Brev and Compose hosts.

The harness is **additive**: it changes no service key, never prunes the agent tier,
and `nemoclaw` stays a harness label that must never appear in `COMPOSE_PROFILES`,
`compose.yml`, or `patches/`. Provisioning stays agent-owned, so such a build is
never headless.

Bring-up reuses the checked-in `deploy/docker/scripts/deploy_nemoclaw.ipynb` rather
than reimplementing onboarding, executed non-interactively through a new shared
runner, and defaults the harness LLM to a remote OpenAI-compatible endpoint.

### The harness contract (skill)

- `skills/vss-build-vision-ai/references/agent-harness.md` (new) owns the harness
  contract: the two harnesses, why NemoClaw is additive rather than a substitute,
  its default model provider, and its post-readiness bring-up.
- `SKILL.md` gains a **Q3 harness selection** gate (asked only when the build has an
  interactive surface to choose), two routing rows, a new Step 10 for bring-up, and
  a diagram requirement that shows the sandbox as a host-side box outside the Compose
  project so a user can catch an unintended harness or an unbudgeted redundant agent.
- The skill `description` now names its exclusions — operate requests (search, ask,
  summarize, alerts, reports) and single-microservice deploys belong to sibling
  `vss-*` skills — so a host matching on the description alone stops routing them here.

### Shared notebook runner

- `deploy/docker/scripts/run_setup_notebook.py` (new) takes notebook parameterization
  and `nbclient` execution out of the skill-eval NemoClaw CI adapter, so the
  per-notebook parameter contract lives beside the notebooks it describes instead of
  drifting in a CI-only copy, and any caller can execute them non-interactively.
- `.github/skill-eval/nemoclaw/notebook_setup_adapter.py` keeps only what is specific
  to CI: provider mapping, scoped MCP cleanup, venv repair, and the runtime env file.
- `deploy/docker/scripts/tests/test_run_setup_notebook.py` (new) covers the runner
  beside the module rather than from `.github`, and runs in the existing CI step.
- Because notebook section 1.2's provider cells (a)/(b)/(c) are mutually exclusive
  human choices assigning plain literals, running top to bottom would let the last one
  win. The runner injects `NAME = os.environ.get("NAME", NAME)` just before the
  derived-settings marker to give the environment the last word. Injection touches
  only the in-memory copy and executed notebooks are never written back, so a run
  cannot persist credentials into the checkout.

### Sandbox egress derived from Docker, not guessed

- `orchestrator_mcp_helper.sandbox_host_cidrs()` (new) returns the IPv4 subnets of the
  networks the sandbox is actually attached to. The checked-in guess predates the
  sandbox: the harness comes up after the build it drives, so Docker has already taken
  172.17 and 172.18 and `openshell-docker` lands on 172.19 — `host.openshell.internal`
  then falls outside `allowed_ips` and every VSS backend port answers 403.
- Notebook section 3.2 rewrites each `allowed_ips` block from those subnets and keeps
  the checked-in list when Docker cannot be inspected, so a hand-applied policy still
  works. `assets/vss_nemoclaw_policy.yaml` adds 172.19.0.0/16 to that fallback and the
  HAProxy ingress origin (7777) — the only origin a build publishes.
- `VSS_PUBLIC_URL` is left unset on a Compose build. It is a Kubernetes setting;
  pointing it at the build's own origin makes `vss-k8s-ingress` collide with
  `vss-backend-readwrite` on `host.openshell.internal:7777`, and because the add is
  atomic the sandbox ends up with no VSS egress at all.

### Ingress ACL admits the sandbox

`agent-harness.md` prescribes `EXTERNAL_IP=host.openshell.internal` in the build
override before resolving. HAProxy's frontend serves only its `known_host` allowlist,
and the sandbox reaches the host as `host.openshell.internal`, which matches none of
them — so a Compose NemoClaw build 404s on its own origin while egress succeeds and the
proxy answers. **Not on alerts**, where `alert-bridge`'s `url_transformer` rewrites
`INTERNAL_IP` to `EXTERNAL_IP` in clip URLs bound for Elasticsearch and the UI: alerts
would keep firing and their evidence would stop opening. That profile gets a curated
`haproxy.cfg` or accepts the origin 404.

### Environment resolution fixes, all found by running the docs

- **Brev addressing** — `prerequisites.md` detects Brev by file check before offering
  any address, and `brev.md` takes the ingress FQDN and public port from the
  environment context file rather than a hostname pattern. A public/elastic IP is wrong
  here: the browser arrives over the secure link, so a host-IP origin serves
  `http://…:7777` URLs into an `https://` page and gets blocked as mixed content.
- **RT-VLM co-location** — `sizing.md` makes the one-GPU vs two-GPU question mandatory
  when a GPU is free, instead of silently inheriting the Foundation's placement, and
  has each answer set `RTVI_VLLM_GPU_MEMORY_UTILIZATION` explicitly. Left blank it
  means the dedicated 0.7, which overcommits a card shared with the NIM's 0.3 and kills
  the second model at init.
- **Agent UI link** — the final summary reports the notebook's `Agent UI:` line as a
  markdown link rather than a bare URL, with the target copied verbatim: the OpenClaw
  target carries the gateway token in its `#token=` fragment, and a link that drops it
  opens an unauthenticated page. The SSH tunnel goes alongside a `127.0.0.1` link, not
  in place of it.
- **Probe corrections** — `brev.md` step 1 probed `/health`, which HAProxy's route
  table does not have, so a healthy proxy returned 404 and the check reported failure;
  it probes `/` instead. `credentials.md`'s NGC model probe omitted `--org`/`--team`
  and failed "Missing org" on a fully entitled key, which read as a failed entitlement.
- **Readiness pattern** — the bring-up snippet's `--require-output` repeated the
  literal `<build-name>` placeholder after the export had already set
  `NEMOCLAW_SANDBOX_NAME`; copied verbatim, the gate waited for a line the notebook
  never prints.

## Test plan

- [ ] `python -m pytest deploy/docker/scripts/tests/test_orchestrator_mcp_helper.py` —
      six new `SandboxHostCidrsTests` cover the happy path, sort order, IPv6 and
      unparseable-subnet rejection, and the three empty-list fallbacks (absent sandbox,
      `docker inspect` failure, no attached networks).
- [ ] `python -m pytest deploy/docker/scripts/tests/test_run_setup_notebook.py` —
      25 new tests over the runner: the parameter contract and repo-root resolution,
      the injection semantics (environment wins, literal stands when unset, all three
      mutually exclusive provider cells lose to the caller, marker cell only,
      list-valued sources, empty contract, missing marker), output collection and
      `require_output`, `run_notebooks` ordering plus its fail-before-kernel check on a
      missing file and its aggregate marker report, the CLI's forwarding and its three
      rejections (unknown notebook, absent file, sub-minimum timeout), and the
      no-write-back guarantee. One test parameterizes the real checked-in notebooks, so
      a renamed derived-settings marker fails the suite instead of silently disabling
      every override.
- [ ] Skill-eval `nemoclaw` leg green, exercising `notebook_setup_adapter.py` through
      the extracted runner against `deploy_nemoclaw.ipynb` and
      `deploy_vss_orchestrator.ipynb`.
- [ ] End-to-end on a Brev instance: build with the NemoClaw harness, confirm the
      ingress origin answers from inside the sandbox (no 403, no origin 404), and open
      the Agent UI from the reported markdown link with its `#token=` fragment intact.
- [ ] End-to-end on a plain Compose host with a free GPU: confirm Q3 and the
      co-location question are both asked and that `RTVI_VLLM_GPU_MEMORY_UTILIZATION`
      lands in `override.env`.
- [ ] Re-run with `NEMOCLAW_RECREATE_SANDBOX=0` against an already-deployed build
      (bring-up alone, Steps 5–8 skipped).

## Note for reviewers

The merge from `develop` accepted #1738's deletion of
`.github/skill-eval/nemoclaw/tests/test_nemoclaw_harness.py`, where this branch's
`NotebookRunnerTests` lived. Rather than restore that file, the runner's coverage is
re-established as `deploy/docker/scripts/tests/test_run_setup_notebook.py`, next to the
module and the notebooks it executes — the same reason the runner moved out of the CI
adapter in the first place — and added to the CI step that already runs
`test_orchestrator_mcp_helper.py`. The dropped tests that were specific to the CI
adapter (provider mapping, runtime env file) are not reinstated here.


nvleeryan:
/ok to test https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1942/commits/0e2079c62ea97c322b5effa1275f511c1133f2b0

zac-wang-nv (changes requested):

## Automated OSRB Review — GAP HARD

License Diff CSV empty and no dependency manifest touched, but source files were added/modified, firing the thin-overlay import safety net. run_setup_notebook.py introduces genuine third-party imports (nbformat, nbclient, ipykernel) executed as notebook-run tooling; none is strict-matched in the approval index across any NVBug comment, so they are IMPORT_UNAPPROVED. All other imports are Python stdlib or internal/test modules.

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33540558698)

nvleeryan:
/ok to test https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1942/commits/007bf42fc73a976af4705d5da8d057b3f81be3b7

zac-wang-nv (changes requested):

## Automated OSRB Review — GAP HARD

License Diff CSV empty and no dependency manifest touched, but source files add third-party imports nbformat and nbclient (notebook execution runtime) with no matching entry in any approved OSRB sheet or comment; strict import-match yields IMPORT_UNAPPROVED.

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33547910070)

nvleeryan:
/ok to test https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1942/commits/69f6241c1987c0d7358ee2ae38ebd301f703e8da

skills/vss-build-vision-ai/references/agent-harness.md
```diff
@@ -0,0 +1,381 @@
...
+
+**Default to notebook option (a) — the remote OpenAI-compatible endpoint —
+serving Claude Opus.** Do not ask which provider to use, and do not fall through
+to the notebook's own (c) build.nvidia.com path:
+
+| Variable | Default | Note |
+|---|---|---|
+| `NEMOCLAW_PROVIDER` | `custom` | option (a); the only provider that consumes `NEMOCLAW_ENDPOINT_URL` |
+| `NEMOCLAW_MODEL` | `claude-opus-4-6` | matches the notebook's own (a) example; override for a different Opus revision, or for a router **route id** when the endpoint is a model router |
+| `NEMOCLAW_ENDPOINT_URL` | `https://api.anthropic.com/v1/` | any OpenAI-compatible base URL; point it at an internal gateway or router instead when one is in use |
```
zac-wang-nv: **P1** — `custom` is NemoClaw’s OpenAI-compatible route, but `api.anthropic.com/v1` is Anthropic’s native Messages API. NemoClaw validates custom providers through `/v1/responses` or `/v1/chat/completions`, so this documented default fails before sandbox creation. Please either use the native Anthropic provider and `ANTHROPIC_API_KEY`, or default to an actual OpenAI-compatible gateway; the notebook’s option (a) example needs the same correction. See NVIDIA’s [inference-provider contract](https://docs.nvidia.com/nemoclaw/latest/reference/inference-profiles.html).

skills/vss-build-vision-ai/references/agent-harness.md
```diff
@@ -0,0 +1,381 @@
...
+# From the environment or the secret store; never a literal here.
+: "${COMPATIBLE_API_KEY:?bearer token for NEMOCLAW_ENDPOINT_URL is required}"
+export COMPATIBLE_API_KEY
+
+uv run --isolated --no-project --python 3.12 \
+  --with nbformat --with nbclient --with ipykernel -- \
+  python "$REPO/deploy/docker/scripts/run_setup_notebook.py" \
+    --notebook "$REPO/deploy/docker/scripts/deploy_nemoclaw.ipynb" \
+    --require-output "Sandbox '${NEMOCLAW_SANDBOX_NAME}' ready." \
+  2>&1 | tee "$REPO/_builds/${NEMOCLAW_SANDBOX_NAME}/nemoclaw-setup.log"
```
zac-wang-nv, line 316: **P1** — This pipeline reports `tee`’s exit status. A notebook exception or a missing readiness marker can therefore look successful and let the skill continue with an incomplete harness. Enable `set -o pipefail` before this pipeline or explicitly capture and check `${PIPESTATUS[0]}` before treating setup as complete.

deploy/docker/scripts/deploy_nemoclaw.ipynb
```diff
@@ -1141,18 +1141,51 @@
...
+        "\n",
+        "    rendered, filled = _POLICY_ALLOWED_IPS.subn(_rewrite, text)\n",
+        "    print(f\"allowed_ips: {', '.join(cidrs)} ({filled} entries)\", flush=True)\n",
+        "    return rendered\n",
+        "\n",
+        "\n",
+        "def _render_policy(policy: Path, staging: Path) -> Path:\n",
+        "    \"\"\"Fill the Ingress host and the bridge CIDRs, returning what to apply.\"\"\"\n",
+        "    source = policy.read_text(encoding=\"utf-8\")\n",
+        "    rendered = _narrow_allowed_ips(_fill_k8s_ingress(source, policy))\n",
```
zac-wang-nv: **P1** — The detected sandbox CIDRs are applied only to `POLICY_PATH`. `MODEL_ROUTER_POLICY` is later passed directly to `policy add`, while the switchyard-generated preset still hard-codes 172.17/18/20. On the 172.19 OpenShell network this change is intended to support, the model-router endpoint remains denied. Please render the router preset through the same CIDR rewrite, or generate it from the detected networks, before applying it.

skills/vss-build-vision-ai/references/brev.md
```diff
@@ -54,39 +74,50 @@ Ports that should NOT get their own secure link (they're behind HAProxy):
...
+curl -sf -o /dev/null http://localhost:7777/ && echo "proxy OK"
 
 # 2. UI reachable through the proxy (internally)
 curl -sfI http://localhost:7777/ | head -1
 
-# 3. Print the browser URL the user should open
-brev_env_id=$(awk -F= '/^BREV_ENV_ID=/ {gsub(/"/, "", $2); print $2; exit}' /etc/environment)
-echo "https://7777-${brev_env_id}.brevlab.com"
+# 3. Print the browser URL the user should open — resolved, never constructed
+echo "https://$(brev_fqdn 7777)"
```
zac-wang-nv: **P1** — The setup now correctly reads `public_port`, but this printed URL drops it. For any Brev context whose public port is not 443, verification hands the user the wrong origin; the troubleshooting `curl` has the same omission. Please include `brev_public_port 7777` in the URL, either always or conditionally when it is not 443.

skills/vss-build-vision-ai/SKILL.md
```diff
@@ -152,14 +155,39 @@ Rules for the multi-select:
...
+**Ask Q3 only when the build has an interactive surface to choose.** Skip it entirely, with no question and no harness, when no requested capability reaches the Agent owner and the user asked for none: an ingest-, index-, or API-only build is legitimately headless, and Q3 would invent a requirement. When the user has already said "headless", that *is* the answer — do not re-ask.
+
+A request that asks for **both** "headless" and NemoClaw is a contradiction, since a NemoClaw build retains the agent tier: take it to the clarification gate rather than silently dropping one side.
+
+**Q3 — Harness (single-select).** *"What should drive this deployment?"*
+
+| Option | Harness | Effect on the build |
+|---|---|---|
+| **In-stack VSS agent** *(default)* | `vss-agent` | The agent tier stays: agent REST API (`/generate`) and Web UI, served from the build's own origin. |
+| **NemoClaw sandbox (chat UI + skills)** | `nemoclaw` | **Purely additive** — no service key changes, and the agent tier always stays. A host-side sandbox with the VSS skills installed drives the build over its public origin, alongside the in-stack agent. |
```
zac-wang-nv: **P1** — Please deploy exactly one harness and make NemoClaw the default. Selecting NemoClaw should replace the in-stack `vss-agent` harness rather than adding a second harness and conversational surface alongside it. Update Q3 and the downstream additive/retained-agent instructions so a generated deployment cannot include both `vss-agent` and NemoClaw.

zac-wang-nv (changes requested):
Requesting changes for five P1 findings covering the default provider contract, notebook failure propagation, model-router egress, Brev URL construction, and mutually exclusive harness deployment. The focused contract suite passes (40 tests), but the main repository CI workflow has not run on this head.

nvleeryan:
/ok to test https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1942/commits/361cc74a13aba535f5d76ec6f6e7528104de1676

skills/vss-build-vision-ai/references/agent-harness.md
nvleeryan: https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1942/changes#diff-55db1dd5e9326f8c675e975d3afcfa8a65d8e24eea47bb8d0177b5d022c24e8eR278

nvleeryan:
/ok to test https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1942/commits/610be02faccca985fd17301b0a81940c336488de

skills/vss-build-vision-ai/SKILL.md
```diff
@@ -152,14 +165,50 @@ Rules for the multi-select:
...
+- Removing a service key makes it a **Delta build**, never a Stock deploy — on a no as much as a yes, and on a quickstart as much as a custom build.
+- **A capability only the Agent owner serves contradicts a no.** Agentic natural-language decomposition (`/api/v1/search`) and `/generate` have no non-agent provider. When the request needs one, a no drops a requested capability: take it to the clarification gate rather than resolving it either way.
+
+Apply these on a **yes** only:
+
+- **Preflight the host before accepting the yes**, per the Prerequisites section of [`references/agent-harness.md`](references/agent-harness.md): `COMPATIBLE_API_KEY`, the NemoClaw CLI, and Python 3.11+. A missing piece is a **blocker at this step, not at Step 10** — name what is missing and ask whether to supply it, proceed with no harness, or name the in-stack agent instead. Deploy nothing until that is answered, and never substitute a harness silently.
```
zac-wang-nv: **P1** — These unconditional checks reject supported paths: the documented NVIDIA build flow uses `NVIDIA_API_KEY`, a local or self-hosted compatible endpoint may not require that credential, and a fresh host is expected to install the pinned NemoClaw CLI in notebook cell 3.1. Please derive credential checks from the selected provider and preflight the installer's actual prerequisites (for example `curl` and Python) instead of requiring the post-install CLI.

nvleeryan:
/ok to test https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1942/commits/44638890c90b78a59824c771e2443b1a8eed13ed

zac-wang-nv:
**TODO (non-blocking follow-up): remove the GHCR post-create read race.**

`advance_ghcr_alias.py` currently creates an alias for a digestless/reuse-pinned entry and immediately inspects the new target tag. In the failed run, Buildx completed the create but GHCR briefly returned `not found`; the tag later resolved to the expected digest, and the rerun passed.

Please have `docker buildx imagetools create` write `--metadata-file` and consume its `containerimage.descriptor.digest` instead of reading the just-created tag back from GHCR. Keep the existing pre-create source verification when the release set has an expected digest, carry prior digests forward for reused entries where possible, and add a regression test asserting that the digestless path never inspects the target tag.

zac-wang-nv (approved):

PR #1917, opened 2026-08-28
docs: rewrite internal links to Fern slug URLs

shruthiraghu23:
## Summary
- Rewrite Sphinx filename and leftover HTML paths to Fern's slugified `/vss/...` URLs after the MDX conversion.
- Pin navigation `slug:` values in `fern/docs.yml` (and `skip-slug` on the nested Industry-Specific Examples wrapper) so autogenerated segments cannot drift from those URLs.
- Restore mashed headings in Agent Customization that had swallowed a report-generation link, and convert remaining unpublished-page leftovers to same-page anchors or relative MDX paths.

## Test plan
- [x] `fern docs broken-links --strict` reports all checks passed
- [ ] Confirm a sample of previously broken paths (`/faq`, `/prerequisites`, warehouse Prerequisites, NvSchema) resolve in a Fern preview
- [ ] Spot-check warehouse, smart city, and agent workflow pages for working in-docs navigation

PR #1943, opened 2026-09-01
docs: retire the removed LLMs from the product docs

zac-wang-nv:
## What

Retires the removed local LLMs from the product documentation.

**Split out of #1854** so that PR carries deploy config, scripts, tests and skills only.

#1854 reduces the bundled local LLM catalogue to `nvidia/nemotron-3.5-lightning-30b-a3b` — removing `nvidia-nemotron-nano-9b-v2`, `nemotron-3-nano`, `llama-3.3-nemotron-super-49b-v1.5` and `openai/gpt-oss-20b` — and makes Lightning the default on every hardware profile, including DGX Spark and AGX/IGX Thor. The Fern pages, the model tables and the profile READMEs still offered the removed models as selectable options.

## Scope

Documentation only — 38 files, +120/−105:

- `docs/**` — the Fern site (`.mdx`), including `edge-deployment`, `quickstart`, `configure-llm`, and the per-vertical doc sets
- `docs/assets/static/default_models.csv` and `llm_params.csv` — the generated model tables
- `README.md` files under `deploy/`, `services/` and `.github/skill-eval/`

No deploy config, script, test or skill changes ride along.

**License pages are excluded on purpose.** `docs/licenses.mdx` and all four `License-Information.mdx` pages still list the removed models; which models appear on a license page is a legal-review question, not a rename. They need a follow-up with OSRB/legal sign-off (also flagged as a known gap on #1854).

## Ordering

This can land before or after #1854. Until #1854 merges, these pages describe the intended end state rather than what is in `develop`, so the two are best landed together.

## Why the skills tree stayed in #1854

`skills/**` is not documentation in the CI sense. Its evals are **agentic** — they run an agent against the skill references and grade the answer — so an eval asserting Lightning-on-DGX-Spark only passes once the routing change is in the same tree. `skills-eval`, `skills-review` and `skills-signatures` all run per-PR, so splitting the two halves would fail whichever PR lacked the other.

## Note on the diff

The first push of this branch was built by checking the files out of the feature branch wholesale, which reverted four newer `develop` commits that touch the same pages — notably `b33c7634b docs: rewrite internal links to Fern slug URLs (#1917)`. It was rebuilt as a patch applied onto `develop`, and the Fern slug URLs are verified present. The corrected diff is +120/−105; the bad one was +369/−378.

### DIFF
--- a/.github/skill-eval/README.md
+++ b/.github/skill-eval/README.md
@@ -46,7 +46,7 @@ Per-CI-run hygiene is the trial's own responsibility: each spec's first agent tu
 | `NGC_CLI_API_KEY` | Pull VSS NIM containers from `nvcr.io` |
 | `LLM_REMOTE_URL` / `LLM_REMOTE_MODEL` | Remote-LLM endpoint used by `remote-*` deploy modes |
 | `VLM_REMOTE_URL` / `VLM_REMOTE_MODEL` | Remote-VLM endpoint used by `remote-*` deploy modes |
-| `HF_TOKEN` | Required by the Edge 4B vLLM on SPARK / Thor `shared` mode |
+| `HF_TOKEN` | Required by RT-VLM / RT-Embed when loading Hugging Face checkpoints |
 | `GITHUB_TOKEN` | Issued to `gh pr comment` when the agent posts results |
 | `BREV_REGISTERED_POOL` | Comma/space-separated registered-node names approved for automatic pool selection |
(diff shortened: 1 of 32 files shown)


kaushikc-nvidia (approved):

kaushikc-nvidia:
/ok to test [199d559](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1943/commits/199d5599d24732ddf9066816d8e4787e4e338681)

PR #1958, opened 2026-09-01
fix(deploy): make the FP8 fallback honor its hw-*.env sizing keys

zac-wang-nv:
## What

The `nvidia-nemotron-nano-9b-v2-fp8` fallback LLM loads `hw-${HARDWARE_PROFILE}[-shared].env` via `env_file` — and `hw-OTHER-shared.env` has carried `NIM_GPU_MEM_FRACTION=0.4`, `NIM_MAX_MODEL_LEN=128000`, `NIM_MAX_NUM_SEQS=4` since it was written — but **nothing ever read those keys**. `env_file` only sets container environment, and the compose `command:` was an exec-form array with every vLLM flag as a literal (`"0.85"` dedicated / `"0.40"` shared). The troubleshooting docs compensated by telling users to edit the checked-in `compose.yml`.

The command now invokes `sh -c` and builds the argv inside the container, where `env_file` values exist:

- `NIM_GPU_MEM_FRACTION` → `--gpu-memory-utilization` (defaults preserved: 0.85 / 0.40)
- `NIM_MAX_MODEL_LEN` → `--max-model-len`, `NIM_MAX_NUM_SEQS` → `--max-num-seqs` — emitted only when set
- `$$` defers expansion to the container shell (same escaping the file's own healthcheck uses); the argv is assembled with `set --` and **quoted** expansions, so a malformed value like `4 --disable-log-requests` stays one argument for vLLM to reject rather than becoming injected flags — containment tested against the committed YAML
- both variants gain the `${LLM_ENV_FILE:-fallback-override.env}` second `env_file` the Lightning service already has, so **`--llm-env-file` now works for this model** instead of being silently ignored

## Runtime behavior: unchanged by design

The shared fraction the env sets (0.4) equals the old literal. `hw-OTHER-shared.env`'s max-model-len / max-num-seqs values were written when nothing read them and never exercised at runtime, so they are **commented out rather than silently activated** — one uncomment away, after validation, instead of a compose.yml edit. Each env file now states which keys are live (the NIM-only keys — `NIM_KVCACHE_PERCENT`, `NIM_LOW_MEMORY_MODE`, `MAX_JOBS` — have no vLLM equivalent here).

Second commit repositions the model's framing from "the edge build" (a role #1854 removed — every platform now defaults to Lightning) to the smaller-GPU alternative: Lightning's INT4 profile refuses to load below its 32 GB metadata floor and observes ~45 GB, this build is 11.7 GB.

## Review

Codex-reviewed pre-PR; acted on: quoted `set --` argv (argument-injection containment), `LLM_ENV_FILE` wiring gap, opt-in rather than silent activation of the never-validated values, and dropping an over-precise "32 GB floor" claim (the floor is profile metadata; observed footprint is ~45 GB). Declined: none material.

## Verification

- The exact committed command string exercised outside compose: env loaded → flags appear; empty env → byte-identical flags to the old literals; injection payload → single argv element
- `compose.yml` parses; compose image golden 13/13
- dev-profile suite: **229 passed / 10 failed, identical failure names** to a develop baseline run in the same worktree (the 10 are pre-existing on develop)

## Checklist


zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33550706582)

zac-wang-nv (commented):

- Public evidence: [License Diff Actions run](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33551548687)

kaushikc-nvidia (approved):

PR #1813, opened 2026-08-24
chore(skills): refresh NVSkills signature for vss-manage-alerts

zac-wang-nv:
## Why

PR #1749 raised the `python-dotenv` floor to `>=1.2.2` for CVE-2026-28684 and merged on 18 Aug. The skill's detached OMS signature, however, is still the one attached on **2026-06-14** and covers the pre-fix content:

| | sha256 of `scripts/alert-notify/requirements.txt` |
|---|---|
| recorded in `skill.oms.sig` | `a6dba0f99113cac1fe96c14deaa29c896f8f57c0c7bcab54d8ea590c100c14e7` |
| actual on `main` | `99ff9ac5f6a7498fa7e5e7bd9faf0a24624ca9bdde960790a7a11f5ee5796fc2` |

Because the content no longer matches the signature, catalog sync falls back to the last signed copy rather than publish unverifiable content. `NVIDIA/skills` has therefore been serving the version with the vulnerable floor since the merge.

## What this changes

One line: a comment in `requirements.txt` recording *why* the floor exists, so it is not lowered again. No behavioural change — the pins are identical.

The change is deliberately inside `skills/`: `team-request.yml` filters the PR's changed paths and skips when nothing under `skills/` or `team-skills/` changed, so a no-op PR would not trigger signing.

## Next step

Commenting `/nvskills-ci` so the signing bot attaches a fresh `skill.oms.sig` covering current `main`. The next hourly sync should then publish the fixed floor.

cc @moshe — flagged this as seven days stale.

PR #1962, opened 2026-09-01
backport(main → develop): merge main into develop to reconcile content + ancestry

zac-wang-nv:
## Why

`main` accumulated **20 commits** that went main-first (release hot-fixes, CI config, docs) and were never propagated back, so `main` and `develop` diverged (main: 20 ahead / 1156+ behind). This back-merge does two things at once:

1. **Content propagation** — bring the 20 main-only commits into `develop`. Most were already cherry-picked to `develop` under different SHAs and auto-resolved to identical content; the net delta is 13 files.
2. **Ancestry reconciliation** — after the merge, `develop` *contains* `main` (`git merge-base --is-ancestor main develop` becomes true), eliminating the merge-base drift. Cherry-picks alone copy content but never fix ancestry.

## Landing (merge-commit topology, done out-of-band)

The squash/rebase buttons must never be used on this PR — either would copy content while destroying the shared ancestry this PR exists to create. Because the PR is conflicting and the repo disables the merge-commit button (`allow_merge_commit: false`, `required_linear_history` on `develop`), the merge commit was built locally and lands via an admin push with ruleset bypass:

- **Merge commit: `6cd4de08d`** — parent 1 `d43fd1572` (develop), parent 2 `b5cf39f32` (main), DCO-signed.
- After the push, GitHub may or may not auto-mark this PR merged (indirect-merge detection is documented for the default branch, which is `main`, not `develop`). If it stays open, it will be closed manually with a pointer to the merge commit — the topology on `develop` is identical either way.

## Conflict resolutions (15 files)

| Group | Resolution |
|---|---|
| `skills/operations/vss-manage-alerts/*` (5 files + sig) | main's newer #1813 content (BENCHMARK, skill-card, requirements without `python-dotenv`, `config/skillspector-baseline.yml`, `.env.example` deletion) taken at develop's renamed path. `references/alert-notify.md` is a hand-merged **union**: develop's multi-backend docs (which match the merged `server.py`) + main's `.env`→shell-export flow and loopback-stop notes (both verified against the merged code). |
| `skill.oms.sig` | **Dropped.** The merged skill tree is a superset of the signed manifest (develop added 12 files and evolved 8 others), so main's signature can never validate; develop had no sig at this path, and the NVSkills signing service re-signs post-merge. |
| `.github` CI/OSRB files (8 files) | develop's evolved versions win — every main-side change was already ported (#1454→#1672, #1456/#1523/#1584→develop twins). One genuinely unported fix from **#1735** was applied: `osrb-review.yml` now passes `DOWNSTREAM_COMMIT_SHA` instead of the runner-owned `GITHUB_SHA` env override (which GitHub silently ignores). `test_osrb_dispatch.py`: 14/14 pass. |
| `services/video-summarization/tests/unit/test_vss_api_models.py` | develop's full 1848-line suite already contains every test in main's 133-line file by name. |

Not ported: #1735's two `downstream_commit_sha()` unit tests — develop implements the override inline without that helper function.

## Commits being backported (main-only)

`dc3746db5`, `7732edf8f`, `62d587c30`, `50c08da79`, `fb919a14c`, `18cf989ed`, `c5aa30a36`, `1917f32d8`, `8cdf75898`, `f7128c711`, `87cf66a63`, `90991b416`, `3d8590936`, `3b47570cd`, `f57db3245`, `27bc8aa04`, `cbb787804`, `e024bef82`, `9ba62a70e`, `b5cf39f32`


zac-wang-nv:
Superseded by #1965 — the head of a PR based on `main` itself can never carry the conflict resolution. The resolved merge commit `1917f7d2f` (parents: develop `f68ec2cdf`, main `b5cf39f32`) now lives on `backport/main-into-develop`; see #1965 for the two landing options (squash = content-only, or FF-push of the merge commit = content + ancestry).

PR #1990, opened 2026-09-03
fix(ci): read created GHCR alias digest from buildx metadata, never the fresh tag

zac-wang-nv:
## Why

The digestless (reuse-pinned) path in `advance_ghcr_alias.py` inspected the **target tag immediately after** `imagetools create`. GHCR reads are eventually consistent right after a write, so that inspect can return the old manifest (or 404) and fail a run whose retag actually succeeded — the flake this PR removes.

## What

`docker buildx imagetools create` reports the created manifest digest itself via [`--metadata-file`](https://docs.docker.com/reference/cli/docker/buildx/imagetools/create/#write-create-result-metadata-to-a-file---metadata-file) (`containerimage.descriptor.digest`) — a local file immune to registry read-after-write lag. The script now:

- **keeps** the pre-create SOURCE (content-tag) verification when the release set records an expected digest — unchanged;
- creates every alias with `--metadata-file` and reads the resulting digest from the metadata;
- **never inspects the newly created target tag**;
- fails the run if create leaves no usable metadata, rather than reporting a digest it did not observe.

The created digest is deliberately **not** compared against the release-set digest: when the source is a single-platform manifest, `imagetools create` wraps it in a fresh index, so the created digest legitimately differs from the source manifest digest (an earlier draft had this comparison; Codex review flagged it as a false-failure source and it was removed). Correctness is asserted **before** the write, on the stable source.

## Tests

- New regression test `test_digestless_entry_never_inspects_the_target`: digestless/reused entries issue exactly one `create` call and **zero** `inspect` calls.
- New fail-closed test for the missing-metadata path.
- Fake runner now honours `--metadata-file` the way buildx does (digest written to the file, not stdout).
- `python3 .github/scripts/test_advance_ghcr_alias.py` → **25/25 pass**; ruff clean.

## Not in scope (follow-up candidate)

Carrying a prior digest forward on reused release-set entries: `release_set.py` builds reuse-pinned entries from the committed compose pins, which are tag-only — a digest would have to come from the previous release set or a registry read, a design question rather than a patch.

### DIFF
--- a/.github/scripts/advance_ghcr_alias.py
+++ b/.github/scripts/advance_ghcr_alias.py
@@ -43,6 +43,7 @@ import json
 import re
 import subprocess
 import sys
+import tempfile
 from dataclasses import dataclass
 from pathlib import Path
 from typing import Callable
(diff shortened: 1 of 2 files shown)


kaushikc-nvidia (approved):

kaushikc-nvidia:
/ok to test [03f3abe](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/1990/commits/03f3abeebed5e63ce69f0361005eec25fdb4d539)

PR #1993, opened 2026-09-03
[NVBug 6706112] fix(rtvi-vlm): keep live stream assets picklable

nv-amkale:
## Summary

- omit the process-local `Asset._lifecycle_lock` when serializing an asset for decoder commands
- recreate the lock when the asset is deserialized in the decoder process
- cover the exact `start-live-stream` payload with `multiprocessing.reduction.ForkingPickler`

## Root cause

An `RLock` was added to `Asset`, while live-stream startup continued to pass the complete asset through a `multiprocessing.Queue`. The queue feeder cannot pickle the lock, so `start-live-stream` never reaches the decoder and the caller waits indefinitely for a response.

## Verification

Verified on NVIDIA B200 with Cosmos3 Nano Reasoner NVFP4.

- single stream with four concurrent prompts: 4/4 requests returned HTTP 200 and produced SSE captions
- four streams with four concurrent prompts each: 16/16 requests passed per round across four rounds
- total: 68/68 requests produced captions
- zero `RLock` serialization errors and zero tracebacks
- RTVI remained ready after every round

### DIFF
--- a/services/rtvi/rt-vlm/src/utils/asset_manager.py
+++ b/services/rtvi/rt-vlm/src/utils/asset_manager.py
@@ -341,6 +341,15 @@ class Asset:
         self._sensor_name = sensor_name
         self._camera_id = camera_id
 
+    def __getstate__(self):
+        state = self.__dict__.copy()
+        state.pop("_lifecycle_lock", None)
+        return state
+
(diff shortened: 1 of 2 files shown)


nv-pnikam (approved):
LGTM

PR #2005, opened 2026-09-03
fix(helm): Fix Helm storage ingress rewrite matching

munjalp6:
### Summary

Anchor the shared HAProxy `/storage` rewrite rules so they only match paths beginning with `/storage`.

Previously, the unanchored rule could also match `/storage` within unrelated API paths. For example, an NVStreamer request:

```text
/api/v1/storage/file
```

could be rewritten incorrectly as:

```text
/api/v1/vst/storage/file
```

This caused video uploads through the NVStreamer ingress to return 404.

### Changes

- Mark the shared `/storage` route as anchored.
- Generate these HAProxy rewrite rules:

```text
^/storage/(.*) /vst/storage/\1
^/storage /vst/storage
```

- Update ingress route verification to support anchored rules.
- Update the alerts, base, and LVS example ingress manifests.

Only `/storage` is anchored. All other ingress rewrite rules remain unchanged.

### Impact

Valid VST media paths continue to work:

```text
/storage/<path> → /vst/storage/<path>
```

Unrelated paths containing `storage` are no longer modified:

```text
/api/v1/storage/file
/vst/api/v1/storage/file
```

The shared rule is rendered by the alerts, base, LVS, and search developer profiles. No service code, internal Kubernetes service URL, storage data, or Docker deployment is changed.

### Validation

- Ran `helm lint` for the alerts real-time profile.
- Ran the ingress route verifier across base, alerts, LVS, and search.
- Verified all 22 ingress mounts and six disabled-component cases.
- Confirmed generated example manifests are current.
- Validated a Playwright video upload through the NVStreamer ingress.
- Confirmed NVStreamer received `POST /api/v1/storage/file` without unintended rewriting.
- Verified RTSP stream creation and cleanup succeeded.

### DIFF
--- a/deploy/helm/developer-profiles/dev-profile-alerts/vss-ingress-example-rewrites.yaml
+++ b/deploy/helm/developer-profiles/dev-profile-alerts/vss-ingress-example-rewrites.yaml
@@ -32,8 +32,8 @@ metadata:
   namespace: <NAMESPACE>
   annotations:
     haproxy.org/path-rewrite: |
-      /storage/(.*) /vst/storage/\1
-      /storage /vst/storage
+      ^/storage/(.*) /vst/storage/\1
+      ^/storage /vst/storage
       /va-mcp/(.*) /\1
       /va-mcp /
       /alert-bridge/(.*) /\1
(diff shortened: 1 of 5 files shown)


trannguyenanhhien (approved):
LGTM

PR #2013, opened 2026-09-03
refactor(model-routing): format Switchyard notebook for line diffs

simings-nv:
## Summary

- convert every `deploy_vss_switchyard.ipynb` cell source from a multiline JSON string to a per-line source array
- preserve the joined source text, cell order, cell IDs, metadata, outputs, and execution counts exactly
- keep notebook behavior changes out of this PR

The only non-format hunk is `persist-credentials: false` on the workflow checkout. The existing end-to-end test clones public Switchyard from inside the VSS checkout. Without this setting, Git carries the VSS-scoped authentication header into that clone and GitHub rejects it. This changes CI authentication only, not notebook behavior.

Tracking: [VIA-2644](https://jirasw.nvidia.com/browse/VIA-2644)

## Validation

- compared the parsed notebook to `develop` after joining each source array; every cell is semantically identical
- compiled every code cell
- confirmed execution counts remain null and outputs remain empty
- `git diff --check` and commit hooks passed

## Separate follow-up

The NemoClaw `0.0.0.0` bind and execution-sequence guidance are intentionally excluded. They are isolated in stacked PR #2021.


simings-nv:
@greptileai review

zac-wang-nv (approved):

PR #2015, opened 2026-09-03
fix(vios): mint media URLs on an origin the caller can reach

hugoverjus:
## What this does now

VIOS stops stamping the internal container name into the media URLs it hands
out. `imageUrl` and `videoUrl` come back on the deployment's public origin, so
the browser, webhook subscriber or alert consumer that receives one can
actually fetch it. In-deployment consumers that fetch the same media translate
back to the internal origin, so nothing inside the network starts depending on
the public one resolving.

Pre-existing on `develop`; not introduced by the gateway work. Every file
touched here except `ci.yml` is byte-identical between `develop` and the
gateway branch.

## The defect

`getIngressBaseUrl()` (`services/vios/src/framework/utilities/utils.cpp`) built
every media URL from `VST_INGRESS_ENDPOINT`, which was the internal container
name. Four call sites feed it, covering the snapshot and recording-URL
responses:

```
GET /vst/api/v1/replay/stream/<id>/picture/url
  "imageUrl" : "http://vst-ingress:30888/vst/storage/temp_files/....jpg"
GET /vst/api/v1/storage/file/<id>/url
  "videoUrl" : "http://vst-ingress:30888/..."
```

Fetching either verbatim off-host gives `curl: (6) Could not resolve host`,
while the same path on the public origin returns 200 with real bytes.

Nothing in a single-host run notices. The `vss` CLI is immune by construction —
`normalise_media_url` discards whatever origin VIOS returned and re-attaches the
configured one — and the agent likewise keeps only the path, so a clean CLI run
proves nothing about the API beneath it.

## Why the one-line repoint does not work

**The value that looked authoritative was not the one that ran.** Compose gives
`environment:` precedence over `env_file:`, and no profile in this tree sets
`VST_INGRESS_ENDPOINT` in its `--env-file` chain, so the inline default in
`services/vios/streamprocessing/docker-compose.yaml` — not the definition in
`services/vios/vst.env` — is what the container ran with. Both are updated.

**The scheme has to travel with the authority.** The scheme came from
`security.use_https`, which cannot express "TLS terminates upstream": it also
decides whether VIOS serves TLS itself (`webServer.cpp:86`) and is echoed to
WebRTC and RTSP clients as `useHttps` in seven other responses. It is also not
settable by environment — only from `vst_config.json`. Pointing the endpoint at
an https origin without it mints `http://host:443/...`, which the TLS listener
answers 400. `getIngressBaseUrl()` now takes the scheme from the endpoint when
the endpoint carries one; values without a scheme behave exactly as before.

**An in-network consumer still has to reach the bytes.** The alert service
fetches the URL VIOS minted and hands it to the VLM, so minting on the public
origin fixes external callers by breaking alert verification. Demonstrated
live: with VIOS on an https public origin and the unmodified alert service,
`URL validation failed after 8 attempts — [Errno -2] Name or service not
known`, and the alert is recorded as a verification failure with no VLM call at
all. It now translates back to `VST_INTERNAL_URL` for the copies fetched from
inside, keeping the minted origin for the copies read from outside — the
direction the agent already translates in `rewrite_to_internal_vst_url`.

## Immediate behaviour change

- VIOS media URLs carry the configured public origin instead of `vst-ingress`.
- The alert service fetches VST media on the internal origin.
- `compose-golden` gains the `check_vios_media_origin.py` guard and its
  self-test.

**Nothing is dormant.** The C++ change and the configuration must land
together: the endpoint is only safe to point at a scheme-carrying origin once
the binary reads the scheme from it, and separating them leaves an intermediate
that mints `http://http://host:7777/...`. VIOS images are built from this repo,
so both arrive in the same image build. A deployment pinning an older VIOS
image tag alongside this compose file would hit that doubled scheme.

**Fallback path unchanged.** With no public origin configured,
`${VST_EXTERNAL_URL:-${VST_INTERNAL_URL}}` resolves to the internal origin and
behaviour is exactly as before.

**Rollback** is a revert of this commit; it restores the leak and nothing else.

## Validation

Live on a healthy 21-container `dev-profile-alerts` deployment, VIOS rebuilt
from this source (`services/vios/docker/Dockerfile`, 139 translation units).

Reproduced before the change on the internal, loopback and host-IP origins, and
on a simulated public HTTPS origin.

After, minted URLs fetched off-host from a separate workstation:

| field | result |
|---|---|
| `imageUrl` | 200, 72580 bytes, JPEG 1920x1080 |
| `videoUrl` | 200, 345145 bytes, ISO Media MP4 |

Both origins in one running configuration: the same clip fetched 200/345145
from inside the bridge on the public origin and on the internal one.

Simulated HTTPS public origin — previously inexpressible — mints
`https://vss-sim.brev.test:443/vst/storage/...` with the correct scheme.

**Alert payload, live.** A real alert through the Kafka-side pipeline (VIOS clip
URL, URL validation, VLM verdict, Elasticsearch sink):

- before: `info.videoSource = http://vst-ingress:30888/vst/storage/...`
- after, on the https origin: `info.videoSource =
  https://vss-sim.brev.test:443/vst/storage/...`, validation successful on
  attempt 1

The alert service's existing outward substitution was a no-op here:
`INTERNAL_IP=10.63.177.97` never appears in `http://vst-ingress:30888/...`, so
the internal origin reached the sink unrepaired.

**Guard canaried** by reintroducing the defect four ways — the internal-only env
definition, the exact shipped Compose default, a schemeless public origin, and
the internal origin ahead of the public one. All four fail the lint.

## Known limitation

The minted origin is the *configured* public origin, not the origin the caller
used. A deployment serving two valid external origins hands both callers the
configured one — resolvable, but cross-origin for the second. Only a relative
URL, or deriving the origin from the request's `Host`/`X-Forwarded-Proto`,
makes it correct per caller; both change the response contract and are left as
follow-ups.


hugoverjus:
## Validation evidence

Validated on a live `dev-profile-alerts` deployment (21 containers, 0 unhealthy), serving internal and external callers in **one** running configuration. Full artefacts on that box under `/home/hverjus/vss-gw-evidence/vios-origin/` (`SUMMARY.md` indexes them).

### Before

`imageUrl`/`videoUrl` minted `http://vst-ingress:30888/vst/...` on all three origins — bridge, loopback and host-IP. Fetched off-host from a workstation: `curl: (6) Could not resolve host: vst-ingress`, while the identical path served `200` through the gateway. So only the minted origin was wrong.

### The C++ change is required, not optional

The configuration change alone, against the **stock** binary, mints:

```
http://http://10.63.177.97:7777/vst/storage/...
```

`getIngressBaseUrl()` prepended its own scheme to a value that already carried one. Captured in `interim-doubled-scheme-stock-binary.txt`. This is why the fix cannot be configuration-only.

`use_https` is not an alternative. It also decides whether VIOS serves TLS itself (`webServer.cpp`) and is echoed to WebRTC/RTSP clients as `useHttps`, so raising it to correct a URL scheme breaks the listener and misinforms other clients. It cannot express "TLS terminates upstream". Carrying scheme and authority in one value is what stops them disagreeing.

### After

Every minted URL fetched from the workstation off-host returns `200` with real bytes, and the in-bridge consumer still reaches the media it needs — internal consumers translate inward (`url_translation.py` in the agent, `url_transformer.py` in the alert service) rather than following the public origin outward.

### Alert payload — live-confirmed

Previously code-confirmed only. A real alert was fired and its Elasticsearch document captured both before and after:

- `before-alert-payload-live.txt` — `info.videoSource` on `vst-ingress`, unusable by the webhook subscriber that receives it.
- `after-alert-payload-https-live.txt` — the same path on an `https` public origin, usable.

This is the path that matters most, since alert payloads go to external webhooks where nothing re-anchors them.

### Harnesses

| harness | result | delta |
|---|---|---|
| `gateway-proof.sh` | 70 / 0 / 2 | unchanged |
| `gateway-adversarial.sh` | 11 / 0 | unchanged |
| `gateway-brev.sh` canonical | 52 / 0 / 4 | unchanged |
| `gateway-brev.sh` simulated-tls | 48 / 4 / 4 | unchanged before vs after |
| `gateway-dns-host.sh` | **27 / 0 / 0** | **+1 pass, −1 known failure** |

`gateway-dns-host.sh` is the acceptance test: its one known failure *was* this leak, and it now passes.

The 4 `simulated-tls` failures are **not** a regression. An isolation run with the stock image *and* stock configuration produces an identical 48/4/4 (`before-brev-simtls-isolation.txt`). The documented 53/0/3 baseline was not reproducible in that mode even pre-change; the failures are the harness comparing declared `https` env vars against its own plain-HTTP transport base.

### Regression guard, and what canarying it found

`check_vios_media_origin.py` / `test_check_vios_media_origin.py`, wired into `compose-golden`.

It checks the default *inside* a passthrough, because Compose gives `environment:` precedence over `env_file:` — so the inline `${VST_INGRESS_ENDPOINT:-<default>}` is the value the container actually runs with, and a lint that skips passthroughs checks the file that loses.

Three canaries:

1. `vst.env` back to `${VST_INTERNAL_IP}/vst` → fails, both diagnostics.
2. Compose inline default back to internal-IP → fails, all four occurrences.
3. Revert **only** the `getIngressBaseUrl()` passthrough → **passed the first time.**

Canary 3 was a real hole, and worth stating plainly: configuration and C++ are one contract. With the passthrough reverted, the scheme-carrying value the lint *demands* becomes the broken input and mints `http://http://...` — the failure actually observed. Guarding configuration alone called that combination clean. Commit 8a513c41d adds `scan_minter()`; canary 3 now fails. A missing file or renamed function is a finding rather than a skip.

The guard reports `5 resolvable definition(s) ... in 2 file(s), scheme passthrough intact in utils.cpp` and fails when it has nothing to check, so it cannot pass vacuously.

CI: `Container Coordinates Golden` → `Unit-test static CI and deployment guards` = success at head `8a513c41d`.

### Placement

Separate PR off `develop` rather than folding into #1983. Verified rather than assumed: every file touched here except `ci.yml` is byte-identical between `develop` and the gateway branch head `d056377c2`, so the defect is pre-existing on `develop` and not introduced by the gateway work. Per `reviewable-migration-prs`, it is independently safe and independently reviewable, and #1983 is already large.

### Residual origin leak — plainly

One remains, by design rather than oversight: the minted origin is the *configured* public origin, not the caller's. A deployment serving two valid external origins hands both callers the configured one — resolvable for both, but cross-origin for the second. Closing that needs either a relative URL or an origin derived from the request `Host`, both of which change the response contract and belong in their own change.

No gateway-side body rewriter was added.

hugoverjus:
## Addendum: a second scheme leak, in the Helm charts

A follow-up consumer/config trace turned up something the Compose validation could not have shown, because it only manifests under Helm. Worth reading before review, since it changes what the C++ hunk is doing.

**The two VIOS charts disagreed about the same variable.** Rendered with `global.externalHost=vss.example.com`, `externalPort=443`, `externalScheme=https`:

| chart | rendered `VST_INGRESS_ENDPOINT` | what the stock binary minted |
|---|---|---|
| `vios-sensor` | `https://vss.example.com:443/vst` | `https://https://vss.example.com:443/...` |
| `vios-streamprocessing` | `vss.example.com:443/vst` | `http://vss.example.com:443/...` |

Both broken, in opposite directions. The sensor chart already assumed the scheme travels with the value; streamprocessing assumed the opposite, under a comment saying "no scheme; app prepends `http://`".

Two things follow:

1. **The `getIngressBaseUrl()` passthrough fixes the sensor case with no chart change.** That is a pre-existing, live defect in every Helm sensor deployment with an external host, and it is independent of the Compose work in this PR. It is also independent corroboration that the passthrough is the right shape — one of the two charts was already written expecting it.
2. **Streamprocessing needed the chart fix** in `c67ac8097`: build the value from `global.externalScheme`, the way the sensor helper and every other browser-facing URL in this chart already do. An explicit `vstIngressEndpoint` is still passed through untouched, with or without a scheme, since the minter now accepts both.

Rendered after the fix, all five cases:

```
internal (no externalHost)  http://vss-vios-ingress:30888/vst
externalHost only           http://vss.example.com/vst
externalHost + port         http://vss.example.com:8443/vst
https + host + port         https://vss.example.com:443/vst
explicit override           https://custom.example/vst
```

`helm lint` clean on both charts; the `statefulset` path renders identically to `deployment`.

### Guard extended, and canaried again

`scan_helm()` now checks both helpers: any branch that emits the endpoint, or the in-cluster default it falls back to, has to carry a scheme. A missing chart or renamed helper is a finding rather than a skip.

Canaries 4 and 5 reintroduce the schemeless form in each chart; both fail the guard, including the sensor chart this PR does not otherwise modify. The guard now reports what it actually examined:

```
VIOS media origin lint passed (5 resolvable definition(s) of VST_INGRESS_ENDPOINT
in 2 file(s), 2 Helm helper(s) emitting a scheme, scheme passthrough intact in
utils.cpp).
```

### Consumer trace: no internal consumer breaks

The audit cleared the one case I was most worried about. LVS and RT-VLM never parse `imageUrl`/`videoUrl` — they fetch whatever URL a caller hands them, and the only route from a minted URL into them is the agent's `lvs_video_understanding.py:662`, which calls `rewrite_to_internal_vst_url` first. The CLI `summarize` path takes a user-supplied `inputs.url`, not a minted field. So the inward-translation set is complete: the agent, and the alert service this PR adds it to.

CI at head `418003e84`: `Container Coordinates Golden`, `Lint (Python)`, `Copyright Headers` and `DCO Sign-off` all pass, 0 failures.

### Revised residual statement

Two things remain, both response-contract questions rather than misconfiguration:

- The minted origin is the **configured** public origin, not the caller's. A deployment serving two valid external origins hands both callers the configured one — resolvable, but cross-origin for the second.
- Closing that needs a relative URL or a `Host`-derived origin.

No scheme/host disagreement remains in either Compose or Helm.

hugoverjus:

all four commits were cherry-picked with their original messages and authorship,
and landed on `feat/via-2713-gateway-consolidated` as:

|---|---|
| `2bba81291` fix(vios): mint media URLs on an origin the caller can reach | `e8b474e1f2071bbf7e1c3834bcabe5e28de334cb` |
| `8a513c41d` fix(ci): guard the scheme passthrough alongside the media origin | `f42071b8c7badaf2300d0dc93e81d745361605f6` |
| `c67ac8097` fix(helm): carry the scheme in the streamprocessing media origin | `380ea33b737d1b30a68bf9b399b9b11180995642` |
| `418003e84` style: apply ruff format to the media origin guard | `9a06e19f57adb1be1f75e5ea85daf23122346d49` |

Consolidated head: `de6c01cfcdb493607f6562793496b769f5e1ec74`.

Every file landed **byte-identical** to this branch's head, and #1983 records
this as a pre-existing `develop` defect rather than part of the gateway feature,
so it stays reviewable and revertable on its own.

`check_vios_media_origin.py` was re-canaried after consolidation, both arms
separately, since guarding configuration alone previously certified the broken
combination:

- dropping the `getIngressBaseUrl()` early return → exit 1 on the minter arm;
- reverting `VST_INGRESS_ENDPOINT` to the internal name while **keeping** the C++
  fix → exit 1 on both the no-public-origin and no-scheme findings.

**This branch is deliberately not deleted**, so the change can be restored here
if the consolidation is reverted.

PR #2021, opened 2026-09-03
fix(model-routing): make NemoClaw handoff explicit

simings-nv:
## Summary

- add `ROUTE_NEMOCLAW`, defaulting to `false`
- select `ROUTER_BIND="0.0.0.0"` before router startup when NemoClaw routing is enabled
- state the required order at the beginning of the notebook: run Switchyard through section 6 and leave the router running, then run `deploy_nemoclaw.ipynb`
- state that section 6 writes the router policy and NemoClaw section 3.2 applies it
- make the existing end-to-end adapter enable and verify the NemoClaw route

#2013 supplied the line-array notebook formatting and CI checkout fix separately. This PR is now rebased directly onto `develop` and contains only the behavior change.

Tracking: [VIA-2644](https://jirasw.nvidia.com/browse/VIA-2644)

## Security boundary

The existing default is unchanged: `ROUTE_NEMOCLAW=false` keeps `ROUTER_BIND=local`, publishing only on loopback and the Docker bridge gateway. The broader `0.0.0.0` bind requires an explicit opt-in and the notebook tells the operator to use it only on a trusted or firewalled host.

The bridge publishing decision has executable assertions for both modes:

- default `local`: loopback plus the Docker bridge gateway
- NemoClaw opt-in: `0.0.0.0:<router-port>:4000`

## Validation

The existing notebook adapter ran from beginning to end against an isolated mock upstream, container, port, and temporary work directory:

- `NEMOCLAW_ROUTING: True; bind=0.0.0.0`
- 12/12 routing probes passed across both targets
- the routed VSS Compose build validated and added no service
- grant and revoke policy lifecycle checks passed
- the owned router container was removed

All notebook cells retain per-line source arrays, null execution counts, and empty outputs. Every code cell compiles, `git diff --check` passes, and path-scoped commit hooks pass.

No dependency or third-party package is added.


simings-nv:
@greptileai review

deploy/docker/scripts/deploy_vss_switchyard.ipynb
```diff
@@ -114,11 +123,13 @@
...
-    "# (or \"0.0.0.0\") only deliberately, on a network you trust to reach it.\n",
+    "# Router runtime. The router is unauthenticated and forwards the caller's\n",
+    "# upstream credential. The default \"local\" publishes it on loopback plus\n",
+    "# the Docker bridge gateway: reachable from this host and its containers,\n",
+    "# unreachable from the outside network. Set this before the first run when\n",
+    "# the router must also serve NemoClaw. It selects the required broader bind;\n",
+    "# section 6 writes the handoff policy, and deploy_nemoclaw.ipynb applies it.\n",
+    "ROUTE_NEMOCLAW = \"false\"\n",
     "ROUTER_PORT = \"4000\"\n",
     "ROUTER_BIND = \"local\"\n",
```
nvleeryan, line 135: might be good to say a few words to let user know that ROUTER_BINE will be rewritten to 0.0.0.0 when ROUTE_NEMOCLAW = true so they won't be suprised

deploy/docker/scripts/deploy_vss_switchyard.ipynb
nvleeryan: multiline instead of multiple prints?

nvleeryan (changes requested):
Only NitPick.  Other then that looks good

deploy/docker/scripts/deploy_vss_switchyard.ipynb
simings-nv: Addressed in c40c55465a86c9200ed8bcdc06ebf80884f1cd95. The setting now says explicitly that a true ROUTE_NEMOCLAW value rewrites ROUTER_BIND to 0.0.0.0.

deploy/docker/scripts/deploy_vss_switchyard.ipynb
simings-nv: Addressed in c40c55465a86c9200ed8bcdc06ebf80884f1cd95. The teardown guidance now uses one multiline print call, with its stdout preserved exactly.

nvleeryan (approved):

zac-wang-nv (approved):
lgtm

PR #2022, opened 2026-09-04
Fix/mv3dt standalone bugs

ByronHernandez:
## Summary

Six P0s from SQA against the RT-CV-3D (MV3DT) standalone deployment, one commit per bug.
Scripts and documentation only — no image changes.

Four of them share a shape: the deployment accepted a configuration it could not run, and only
failed later where the cause was no longer visible. Staging now refuses instead, at the point
the problem can still be fixed.

## Related Issue

- Bug 6553034 — SEI check silently bypassed on a non-default VST port
- Bug 6636526 — the wrong stream is blamed for a stalled batch
- Bug 6551512 — OSD fails unless the container is run `privileged: true`
- Bug 6564660 — staging accepts an inconsistent camera set
- Bug 6565438 — Option B omits the restage the Kafka endpoint needs
- Bug 6646812 — software encoder install not automated for GPUs without NVENC

Tracker: Bug 6124541 (build drops). Related but **not** fixed here: Bug 6636594 (perception
segfaults during dynamic stream removal — reproduced during this work, but the fault is in
DeepStream's `nvstreammux` and not addressable from the deployment scripts).

## Changes

**`scripts/add-streams.sh`**
- **6553034** — the SEI prerequisite check assumed the VST proxy on port 30000 and ended in
  `|| return 0`, so a 404 was indistinguishable from "no VST here". The port is now discovered
  (30000-30005, 31000-31005), the check applies only when the staged config requires the SEI,
  and it stops rather than passing silently. A config-level fallback was tested and ruled out:
  a SEI-less live source does not activate under host-clock timestamps either.
- **6636526** — only the cameras passed to that invocation were judged, so a stalled source
  caused the last camera *added* to be reported. Every registered stream is judged once the
  batch completes.

**`scripts/stage-configs.sh`**
- **6551512** — `NVIDIA_VISIBLE_DEVICES` is gated by `GPU_DEVICE`, so on a multi-GPU host the
  GPU driving `DISPLAY` was usually not exposed and the OSD died at PAUSED; `privileged: true`
  "worked" only by bypassing the device cgroup. Staging resolves the display GPU, adds it to
  `GPU_DEVICE`, and derives `gpu-id` from the resulting visible set (a CUDA ordinal, not the
  `nvidia-smi` index). Inference stays on the requested GPU.
- **6564660** — `NUM_CAMS`, `calibration.json`, `camInfo` and `pub_sub_info_config.yml` were
  never compared. Staging now refuses a mismatch, writing nothing.
- **6565438** — the Kafka endpoint is baked into the staged config and configs are mounted
  read-only, so editing `docker/.env` without restaging left the old broker in place.
- **6646812** — `SAVE_VIDEO=1` on a GPU with no hardware encoder staged `enc-type=1`, which no
  shipping image can satisfy. Staging prepares the encoder into the image and points
  `PERCEPTION_TAG` at the result, or refuses.

**`scripts/prepare-sw-encoder.sh`** (new) — prepares and commits the software encoder.

**`docker/init-scripts/ds-start-mv3dt.sh`** — startup refuses a stale staged Kafka endpoint
(6565438) or camera set (6564660), so a refused staging cannot be launched anyway.

**`README.md`** — restage requirements, OSD/GPU selection, and the NVENC-less path. The previous
NVENC-less instructions could not work: the image runs as uid 1000, so the documented
`docker exec` needed `-u 0` for `apt` to write.

### Two notes for review

**Encoder detection asks the hardware, not a GPU name.** `nvidia-smi` reports encoder *usage*
but never *capability* — an NVENC-less H100 is indistinguishable from an NVENC-capable
RTX PRO 6000 in every field it exposes, and NVML's `nvmlDeviceGetEncoderCapacity` returns 100%
on the H100. Detection runs DeepStream's own encoder element in the image on the selected GPU,
falls back to `/dev/v4l2-nvenc` (Tegra, positive-only), then to an `ffmpeg` probe, and refuses
if none can answer. A name list would already miss this generation: a GB200 node reports its
GPUs as B200.

**Committing a prepared image needs three guards**, each of which otherwise produced an image
that passed every static check and still broke the deployment: restore the base entrypoint
(`docker commit` captures the prep container's `sleep` override); drop the GStreamer registry
cache before committing (the prep container has no GPU, so probing there blacklists every
DeepStream plugin); and re-validate an image left by an earlier run. Both properties are
verified after commit, with the GPU attached, and the image is discarded if either fails.

### Verification

Syntax and behaviour checks pass at **every commit checked out standalone**, on six machines:
H100 NVL (no hardware encoder), RTX PRO 6000 Blackwell, Orin (aarch64/Tegra), a GPU-less host,
a hybrid-graphics laptop, and RTX 4090.

Full pipeline on four cameras for both encoder paths — software (`enc-type=1`) and hardware
(`enc-type=0`), each producing `h264 1920x1080 30/1` output.

Not covered: end-to-end runs for 6553034 and 6636526, which need live RTSP through the SQA VST;
and the 6646812 prepare-and-commit path on aarch64 (GB200/GB300 are aarch64; the mechanism is
architecture-neutral and the package names match, but it has only been exercised on x86_64).

## Checklist

> On tests: `rt-cv-mv3dt` carries no test suite in this repo, and these changes are shell
> scripts. They are covered by an external harness (11 script-level tests plus a 20-case
> regression suite for the encoder path) run at every commit on the six machines above, but
> nothing lands in-tree here, so the box stays unticked. Happy to add an in-repo suite if that
> is wanted for merge.

### DIFF
--- a/services/rtvi/rt-cv-3d/rt-cv-mv3dt/README.md
+++ b/services/rtvi/rt-cv-3d/rt-cv-mv3dt/README.md
@@ -46,7 +46,8 @@ includes a 4-camera warehouse **sample dataset** you can run end-to-end.
 - [3. Launch](#3-launch)
   - [3.1 Option A — bundled brokers](#31-option-a--bundled-brokers)
   - [3.2 Option B — your own brokers](#32-option-b--your-own-brokers)
-  - [3.3 Verify startup](#33-verify-startup)
+  - [3.3 On screen, display, and GPU selection](#33-on-screen-display-and-gpu-selection)
+  - [3.4 Verify startup](#34-verify-startup)
 - [4. Add streams dynamically (RTSP)](#4-add-streams-dynamically-rtsp)
 - [5. Check logs and receive metadata from Kafka](#5-check-logs-and-receive-metadata-from-kafka)
 - [6. Visualization](#6-visualization)
@@ -221,7 +222,21 @@ docker compose up -d
 #   docker compose --profile "*" down
 ```
 
-### 3.3 Verify startup
+### 3.3 On screen, display, and GPU selection
+
+`GPU_DEVICE` in [docker/.env](docker/.env) picks the GPU the pipeline computes on. With `OSD=1` the container also has to reach the GPU that drives the X display, which is often a different one on a multi-GPU host.
+
+`stage-configs.sh` resolves this: it adds the display GPU to `GPU_DEVICE` when it is missing, and writes `gpu-id` into the staged config as the compute GPU's ordinal within that set. The two are not the same number, because the container runtime orders the visible devices by PCI address rather than by the order you list them.
+
+```
+GPU_DEVICE=7   OSD=1   ->   GPU_DEVICE=3,7   gpu-id=1     # compute on 7, display via 3
+```
+
+When the display GPU cannot be determined, staging says so and changes nothing. The on-screen display may then fail with `Failed to set pipeline to PAUSED`; add the GPU driving the display to `GPU_DEVICE` and restage.
+
+On a laptop or any host whose display is rendered by Mesa rather than by an NVIDIA GPU, this is a different problem with a different fix: uncomment the `devices` block in [docker/compose.yml](docker/compose.yml). The preflight names whichever of the two it sees.
+
+### 3.4 Verify startup
 
 Either option — follow the perception logs until the pipeline reports ready:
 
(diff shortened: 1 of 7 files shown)


zac-wang-nv (approved):

ByronHernandez:
/ok to test [6d3c2ae](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/2022/commits/6d3c2ae1a49f6462d089eda1369f914f70519364)

PR #2023, opened 2026-09-04
fix(deploy): clamp GPU device placement to the GPUs a host has

hugoverjus:
## The defect

Profiles commit the GPU layout they were validated on. `dev-profile-alerts`
places the LLM, the VLM and RT-VLM on **device 1** and reserves device 0 for
RT-CV; `dev-profile-search` splits RT-CV + RT-VLM on 0 from RT-Embed + the LLM
on 1; `warehouse-operations` goes as far as **device 2**.

Compose passes those indices straight into `device_ids`, so on a host with fewer
GPUs than the profile assumes the request names a device that does not exist and
containers fail to start. There is no fallback — the stack simply never comes up.

This is not hypothetical: much of the eval fleet is single-GPU
(`g7e.2xlarge` / `g7e.4xlarge`), so any one-GPU box hits it. It was found on a
Brev `g7e.4xlarge` where `dev-profile-alerts` could only be deployed after
hand-editing the device IDs on disk.

`get_nvidia_smi_gpu_count()` already existed, but was called from exactly one
place — a search-and-Brev-specific pre-flight — so nothing caught the general
case.

### Scope across the tree

The lint added here reports it directly:

| Profile | GPUs implied by committed placement |
|---|---|
| `dev-profile-alerts` | 2 |
| `dev-profile-search` | 2 |
| `warehouse-operations` | 3 |
| `dev-profile-base` | 1 |
| `dev-profile-lvs` | 1 |
| `smartcities` | no device placement committed |

**Not a capacity problem.** The LLM and the VLM co-resident on a single RTX PRO
6000 measured **71 GiB of 96 GiB** in use, so collapsing placement onto a device
that exists genuinely runs. What breaks is the *selection* of an index that is
not there.

## The fix

Consult the GPU count where device IDs are resolved and clamp any index at or
above it onto an existing device, rather than changing the committed defaults.
Two properties are deliberate:

- **A host with at least as many GPUs as the profile's highest index clamps
  nothing**, so every already-validated multi-GPU layout renders
  byte-identically. The clamp is dormant there by construction, not by a
  hardware-profile special case.
- **A GPU count of 0** — no `nvidia-smi`, a CI or dry-run host — leaves
  placement alone. An unknown count must never silently rewrite it.

`RESERVED_DEVICE_IDS` is **filtered rather than clamped**: it names GPUs a model
must stay *off*, so mapping index 1 down to 0 would reserve the only device the
profile then has to run on. That is exactly why the manual workaround had to
blank it by hand.

Every remap is logged. A quiet remap is barely better than the crash, because
the next person reading a bug report cannot tell which GPU the model actually
ran on.

`blueprint-deploy.sh` gets the same helpers — it already keeps private copies of
`get_env_value` and friends, and `warehouse-operations` is the most exposed
profile in the tree.

## Validation

- **Deployed on a real one-GPU box** (Brev `g7e.4xlarge`, one RTX PRO 6000
  Blackwell, 96 GB) with `dev-profile-alerts`: 23 containers healthy, the LLM
  NIM placed on `DeviceIDs: ["0"]`, 71358 / 97887 MiB in use with both models
  co-resident.
- **Two-GPU placement unchanged**: with a GPU count at or above each profile's
  highest committed index, the clamp is inert and the rendered configuration is
  identical.
- `deploy/docker/test-scripts/test-dev-profile.sh` extended to cover both
  branches; `VSS_GPU_COUNT` is the test hook that makes the single-GPU and
  multi-GPU paths reachable on any host, including a CI runner with no GPU
  (the same role `SKIP_HARDWARE_CHECK` plays for the hardware-profile match).

### Regression guard, and its canary

`.github/scripts/check_gpu_device_clamp.py` reports each profile's implied GPU
count and fails when a profile commits an index above 0 under a placement key
the clamp does not walk — a new placement key that nothing clamps is how this
reaches a single-GPU host in the first place.

Canaried rather than assumed: removing `RT_EMBED_DEVICE_ID` from
`DEVICE_ID_KEYS` makes it **exit 1**, naming
`dev-profile-search/.env:59`; restoring it returns **exit 0**. 22 unit tests
accompany it.

## Why this is a separate PR

The defect is pre-existing on `develop` (verified, not taken on trust) and has
nothing to do with gateway routing, so it does not belong in the large gateway
consolidation PR it was found alongside.

### DIFF
--- a/.github/scripts/check_gpu_device_clamp.py
+++ b/.github/scripts/check_gpu_device_clamp.py
@@ -1,0 +1,357 @@
+#!/usr/bin/env python3
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
+"""Keep every committed GPU device index reachable through the single-GPU clamp.
+
+The profile env files under ``deploy/docker`` place each service on a specific
+GPU, and those indices describe the host the profile was validated on rather
+than the host it will run on:
+
+  dev-profile-alerts    LLM, VLM and RT-VLM on device 1, RT-CV on 0   -> 2 GPUs
+  dev-profile-search    RT-CV + RT-VLM on 0, RT-Embed + LLM on 1      -> 2 GPUs
+  warehouse-operations  RT-CV on 0, RT-VLM on 1, LLM on 2             -> 3 GPUs
+
+Compose passes those values straight into ``device_ids``, and a request for an
+index the host does not have is a hard container-start failure, not a fallback.
+So on a single-GPU machine the stack never comes up, and the reason is invisible
+in the configuration: every value in the files is individually correct.
+
+What this lint asserts is therefore not a value but a route. The deploy scripts
+resolve committed placement against the GPU count the host reports
+(``clamp_device_ids_to_gpu_count``), folding any index at or above that count
+onto the last real device and logging every remap. This checks that:
+
+  1. every device index committed in a profile parses as a device index at all;
+  2. every key a profile uses to place a service is one the clamp knows about;
+  3. the clamp is actually invoked, and is not gated behind one profile or one
+     environment marker.
+
+(3) is the shape the defect originally had. ``get_nvidia_smi_gpu_count`` already
+existed in ``dev-profile.sh`` and was called from exactly one place: a
+search-only, ``BREV_ENV_ID``-gated pre-flight. The helper was there, the general
+case was not covered, and nothing in the tree said so.
+
+Deliberately NOT asserted: that no profile commits an index above 0. A two-GPU
+default is the validated configuration for alerts and search, and forbidding it
+would push those profiles onto a layout nothing has been tested on. Nor is a
+declared "this profile needs N GPUs" value required: the highest committed index
+already is that number, so a second copy of it would only be something to keep
+in sync. The requirement is reported below instead, derived from the files.
+"""
+
+from __future__ import annotations
+
+import argparse
+import re
+import sys
+from dataclasses import dataclass
+from pathlib import Path
+
+ROOT = Path(__file__).resolve().parents[2]
+ENV_SUFFIX = ".env"
+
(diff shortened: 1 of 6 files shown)


hugoverjus:

the commit was cherry-picked with its original message and authorship:

|---|---|
| `c3a57adac` fix(deploy): clamp GPU device placement to the GPUs a host has | `dcdabfb6b9396d0643f6d4fb4115730d9d085b8a` |

Consolidated head: `de6c01cfcdb493607f6562793496b769f5e1ec74`.

It landed with the same 1259 insertions / 17 deletions as here. Two conflicts,
both additive and both resolved as unions: the guard list in `ci.yml`, and the
new helper block in `dev-profile.sh` where the gateway branch had added its Brev
secure-link helpers at the same point. `clamp_device_ids_to_gpu_count` is
confirmed still called from both `dev-profile.sh` and `blueprint-deploy.sh`.

#1983 records this as a pre-existing `develop` defect rather than part of the
gateway feature, so it stays reviewable and revertable on its own.

`check_gpu_device_clamp.py` was re-canaried after consolidation using the canary
documented in the commit message: dropping `RT_EMBED_DEVICE_ID` from
`DEVICE_ID_KEYS` makes it exit 1 naming `dev-profile-search/.env:59`, and
restoring the key returns exit 0.

One thing carried in unchanged, for whoever picks this up: three of this PR's
`test-dev-profile.sh` cases fail on a GPU host, identically here and on the
consolidated branch, so they are not a consolidation regression. Two of them
(`names the profile's GPU assumption`, `logs which key moved`) are a harness
defect rather than a clamp defect — `assert_stdout_contains` reads `TEST_STDOUT`,
which `run_dry_run_test` never sets and has already deleted, so those assertions
can never pass after it. The clamp's actual output was checked by hand and is
correct, emitting both expected strings. The third
(`search on one GPU folds RT-Embed and the LLM onto device 0`) wants an
`RT_CV_DEVICE_ID` that the `search` profile does not render on this host.

**This branch is deliberately not deleted**, so the change can be restored here
if the consolidation is reverted.

PR #2026, opened 2026-09-04
Fix DASH overlay, replay timing, video wall tiles and encoder bitrate

gous-75:
## What this fixes

DASH sessions that draw an overlay were unusable in several ways. Each item below was found by measurement, and every claim here was verified on hardware.

**The overlay discarded whole streams.** Suppressing a frame the overlay drew nothing on was keyed on which branch found the metadata rather than on whether metadata matched. Recorded playback reaches the advancing branch for every frame, so the overlay threw away everything: the encoder was handed nothing and the session looked like it had never started. Suppression now also requires the metadata to be near the frame — replaying a window whose analytics have aged out returns the oldest surviving record, hours away, and video without boxes beats no video.

**Replay played fast.** The replay timeline was counted at a fixed rate, which is only honest while every frame arrives; with half of them suppressed a second of recording became half a second. It now uses the frames' own timestamps, which is what the WebRTC path always did — and why WebRTC replay was never fast. The gap-closer also treated every deliberate drop as a stall, carrying 12.8 s out of a 20 s replay.

**Nothing published on hosts without a hardware encoder.** There, the overlay and composite pipelines end at the transform, which hands its consumer a decoded picture. That is what libwebrtc wants; the DASH packager cannot publish a picture, so it dropped every one. The packager now encodes with x264enc, and a consumer states whether it wants decoded pictures rather than having it inferred from its type.

**Video wall tiles drew empty overlays.** The tile label carried the stream id, and the compositor passes it to the tile's overlay, which matches it against the sensor *name* on each detection. Nothing matched, so tiles showed their name and timestamp but never a box.

**Long-GOP recordings were never transcoded.** The scan path read the keyframe interval under a name the parser does not publish, silently took the fallback, and every scanned file looked like its keyframes were one frame-rate apart — so the "longer than five seconds" test could never fire. The upload path reads the correct name, which is why uploading a file transcoded it and scanning the same file did not.

**Startup scaled with segment length.** Preroll asked for a multiple of the segment length, reintroducing exactly what the segment count is documented as preventing: 256-frame keyframes held the first frame for 36 s, past the point the client stops waiting.

**Bitrate was wrong on both encoders.** A DASH session has no congestion controller, so the hardware encoder sat at the floor of the WebRTC range (2 Mbit/s at 1080p) and the software encoder used x264's own 2048 kbit/s default at every resolution — 4K included. Both now take one setting, `dash.bitrate_kbps`, derived from picture height when left at zero.

## Verification

Measured with a browser driving the real UI, reporting freezes, buffer and playback rate.

| | Thor (hardware encoder) | x86 (software encoder) |
|---|---|---|
| live / live+overlay | 1.000x / 1.000x | 1.000x / 1.000x |
| replay / replay+overlay | 1.000x / 1.000x | 1.000x / 1.000x |
| wall / wall+overlay | 1.000x / 1.000x | 1.000x / 1.000x |
| WebRTC live / replay / wall +overlay | 0.998–1.002x | 1.000–1.006x |

Zero freezes and zero GStreamer errors in every case. Video-wall bounding boxes were confirmed visually on all four tiles. A three-minute soak on the software path ran clean, and repeated replay runs did not reproduce a one-off startup hitch seen earlier on a loaded host.

Bitrate was audited per case on both encoders: live+overlay, replay+overlay, wall and wall+overlay all receive it; the passthrough paths have no encoder and correctly receive none.

## Notes for reviewers

- Per-session memory growth of roughly 30–70 MB exists on this branch, and identical measurements from fresh containers show stock `develop` growing the same amount within ~2 MB. It is pre-existing and worth its own investigation, not a regression from this work.
- The keyframe interval is set to 30 in every configuration. DASH can only cut a segment on a keyframe, so the previous 256 produced eight-second segments and a twenty-second presentation delay.
- Startup latency of roughly 10–15 s is inherent to segmented delivery: at least one full segment plus a preroll cushion must exist before a manifest can be published. Sources with very long keyframe intervals are slower until the transcode fix above applies to them.


dsitlani-nv (approved):

gous-75:
### The overlay fix is not DASH-only: WebRTC replay with an overlay is broken on develop

Asked whether the `overlay_internal.cpp` change is needed beyond DASH, I tested it rather than reasoned about it. Stock `develop` and this branch, same host (Jetson/Thor, `enable_overlay_skip_frame: true`), same UI, same sensor:

| case | develop | this PR |
|---|---|---|
| WebRTC replay + overlay | **no playback (3/3 runs)** | OK, 1.000x, 0 freezes (3/3) |
| WebRTC replay, no overlay (control) | OK 1.002x | OK |
| WebRTC live + overlay | OK 1.000x | OK 0.999x |
| WebRTC video wall + overlay | OK 0.999x | OK 0.998x |

Enabling an overlay on a WebRTC replay stream stops playback outright on develop. The control shows it is the overlay, not the replay path.

**Why**: the suppression guard fired even when the re-check had just matched and drawn the boxes. Live and the wall usually match on the first branch and survive; recorded playback reaches the advancing branch for *every* frame, so every frame is discarded — independent of transport. DASH made it look DASH-specific because with no frames there is no segment to write and nothing ever appears, but WebRTC fails the same way.

**Scope**: only where `enable_overlay_skip_frame` is true, which is Jetson/Thor. On x86 it is false and the condition never evaluates, which is why this looked fine when tested there.

PR #2027, opened 2026-09-04
(fix)nemoclaw + vss with nemotron 3. 5 lightning deployed locally

ayyappa-dev:
## Summary

   - Make NemoClaw + VSS search validation work end-to-end with a local Nemotron 3.5 Lightning vLLM.
   - Nemotron 3.5 Lightning is NOT a NIM — only the managed `install-vllm` recipe exists, capped at 65536.
   - Deploying vLLM separately, is the only path to native 262144.
   - Four sandbox→host fixes so the agent can actually reach VSS and call tools.
  
   ## Plan

   - BYO vLLM: user runs `vllm serve` with `--max-model-len 262144`; NemoClaw attaches at `host.openshell.internal:18000`.
   - Pass tool-call + reasoning parsers explicitly (the image does NOT bake them).
   - Dedup the 8081 policy conflict; fix the shallow skills glob; clear the device-id hardcodes.
   - Allowlist `host.openshell.internal` in HAProxy; honor `http_proxy` in aiohttp; probe the sandbox origin in the search skill.

   ## Related Issue

   - VIA-2729

   ## Changes

   **`deploy/docker/scripts/deploy_nemoclaw.ipynb`**
   - §1.2 (a): BYO vLLM — `custom`, `host.openshell.internal:18000`, `nvidia-nemotron-3.5-lightning-30b-a3b-nvfp4`.
   - §2.2 (new): start the vLLM container — pinned image, flags-only, `--max-model-len 262144`, parsers passed explicitly, conservative `1/4096/0.65`, 1800s
 readiness wait.
   - §3.2: `policy remove local-inference` before `policy-add` — 8081 dup.
   - §3.3: `glob("**/SKILL.md")` — 19 skills, not 1.

   **`deploy/docker/scripts/deploy_vss_orchestrator.ipynb`**
   - §1: clear `LLM_DEVICE_ID`/`VLM_DEVICE_ID` hardcodes → profile defaults win.

   **`deploy/docker/scripts/nemoclaw/README.md`**
   - `find` for SKILL.md discovery (matches the notebook glob).

   **`deploy/docker/services/infra/haproxy/haproxy.cfg.template`**
   - Add `host.openshell.internal` to `known_host` + `h_main` — sandbox `Host:` header was 404'd.

   **`services/agent/packages/vss_core/src/vss_core/vios/client.py`**
   - `trust_env=True` on all 11 `ClientSession` — aiohttp ignored `http_proxy` → `vss vios add` silently failed.

   **`skills/operations/vss-search-archive/SKILL.md`**
   - Probe `host.openshell.internal:7777` before asking for origin — no `~/.vss/config.json` on a fresh sandbox.
 
## Checklist


deploy/docker/scripts/deploy_nemoclaw.ipynb
```diff
@@ -120,12 +120,15 @@
...
+        "# Local Nemotron 3.5 Lightning (BYO vLLM): start the vLLM server with §2.2 below, then point NemoClaw at it here.\n",
+        "# Nemotron 3.5 Lightning is NOT a NIM (nim-local has no NIM for it) and the managed install-vllm recipe is a FIXED recipe\n",
+        "# capped at 65536, so this BYO vLLM path is the only way to get the model's native 262144 context without overflow.\n",
         "NEMOCLAW_PROVIDER     = \"custom\"\n",
-        "# Cloud: \"https://api.anthropic.com/v1/\" | self-hosted: \"http://host.openshell.internal:8000/v1\"\n",
-        "NEMOCLAW_ENDPOINT_URL = \"\"  # OpenAI-compatible base URL\n",
-        "NEMOCLAW_MODEL        = \"\"  # Model id at that endpoint, e.g. \"claude-opus-4-6\";\n",
-        "                            # for a router, the route id it serves, e.g. \"switchyard/random\"\n",
-        "COMPATIBLE_API_KEY    = \"\"  # Bearer token (sk-ant-..., sk-proj-...); blank is fine for a self-hosted server"
+        "# The sandbox reaches the host vLLM via the canonical host alias (whitelisted in the egress policy);\n",
```
nvleeryan: option (a) is used for both cloud and self-hosted llm.  Cloud llm capability (e.g. Opus) should not be removed from the comment.  Instead, a comment to let user know that they can point to the local LLM deployed by VSS from here is more appropriate.

deploy/docker/scripts/deploy_nemoclaw.ipynb
```diff
@@ -120,12 +120,15 @@
...
+        "# host.openshell.internal resolves inside the sandbox to the docker bridge gateway -> host:18000.\n",
+        "NEMOCLAW_ENDPOINT_URL = \"http://host.openshell.internal:18000/v1\"\n",
```
nvleeryan: likewise, the actual values should not be defaulted to vss local llm settings but have a comment to let user know how they can do it here instead

deploy/docker/scripts/deploy_nemoclaw.ipynb
```diff
@@ -940,6 +943,49 @@
...
       ]
     },
+    {
+      "cell_type": "code",
+      "execution_count": null,
+      "metadata": {},
+      "outputs": [],
+      "source": [
+        "%%bash\n",
+        "# §2.2 Start the agent vLLM (BYO, local Nemotron 3.5 Lightning) — run before §3.1 onboard.\n",
```
nvleeryan: where is 2.2?

deploy/docker/scripts/deploy_nemoclaw.ipynb
```diff
@@ -940,6 +943,49 @@
...
+        "# Nemotron 3.5 Lightning is NOT a NIM; the managed install-vllm recipe is a FIXED recipe\n",
+        "# capped at 65536, so the only path to its native 262144 context is to run our own vLLM\n",
```
nvleeryan: i don't think we need to mention the limitation of the install-vllm here as it will not help as an instruction and might even confuse users

deploy/docker/scripts/deploy_nemoclaw.ipynb
```diff
@@ -940,6 +943,49 @@
...
+        "# and point NemoClaw at it (option (a) above). NemoClaw §3.1 attaches to a server\n",
+        "# already answering on NEMOCLAW_VLLM_PORT (default 18000), so this must be up first.\n",
+        "# The pinned vllm/vllm-openai image entrypoint bakes \"vllm serve --model <model>\"\n",
```
nvleeryan: <model> was not baked by the pin, it was specified by the user

deploy/docker/scripts/deploy_nemoclaw.ipynb
```diff
@@ -940,6 +943,49 @@
...
+        "# and loads the Mamba-2 backend + NVFP4 by default, but it does NOT bake the\n",
```
nvleeryan: where did you see / confirm Mamba-2 + NVFP4 is loaded by default?

deploy/docker/scripts/deploy_nemoclaw.ipynb
```diff
@@ -940,6 +943,49 @@
         "<span style=\"color:red\"><strong>This cell replaces an existing sandbox</strong> whenever <code>NEMOCLAW_RECREATE_SANDBOX</code> is <code>True</code>, the default in section 1.3. The sandbox and its agent sessions are discarded, and 3.2–3.6 reinstall the policy, skills, and workspace docs. Set it to <code>False</code> in 1.3 to keep an existing sandbox.</span>\n"
       ]
     },
+    {
```
nvleeryan: This cell will be confusing as it seems to be a required step to run.  What if user is not using the local LLM?

Besides, it might not be a prefer approach to create a fork of the NemoClaw managed recipe (install-vllm) manually like that.

It looks more like an `vllm` instead of `install-vllm` to me.  And if user chose to use `vllm` (self-hosted vlllm) they can do so but i dont thin we should bake that into our notebook as that's not the recommended default

deploy/docker/scripts/deploy_vss_orchestrator.ipynb
```diff
@@ -98,9 +98,9 @@
         "\n",
         "\n",
         "# ================== VSS settings ==================\n",
-        "LLM_NAME = \"\"\n",
-        "LLM_ENDPOINT_URL = \"\"\n",
-        "LLM_MODEL_TYPE = \"\"\n",
+        "LLM_NAME = \"nvidia-nemotron-3.5-lightning-30b-a3b-nvfp4\"  # MUST match the agent vLLM /v1/models id (NEMOCLAW_MODEL in deploy_nemoclaw.ipynb)\n",
```
nvleeryan: Similar to above, we shouldn't force the local lightning here.  it should be left black to use whatever the docker compose / helm uses

deploy/docker/scripts/deploy_nemoclaw.ipynb
```diff
@@ -1205,6 +1251,11 @@
         "with tempfile.TemporaryDirectory(prefix=\"vss-policy-\") as _staging_dir:\n",
         "    _policy = _render_policy(POLICY_PATH, Path(_staging_dir))\n",
         "    _policy_add_cmd = AGENT_POLICY_ADD_CMD.format(sandbox=NEMOCLAW_SANDBOX_NAME, path=_policy)\n",
+        "    # install-vllm onboard applies the `local-inference` preset, which opens port 8081.\n",
```
nvleeryan: the install-vllm port can be configured by NEMOCLAW_VLLM_PORT (see 1.3 in this notebook).  I don't see the need to remove the local-inference policy manually

deploy/docker/scripts/deploy_vss_orchestrator.ipynb
```diff
@@ -109,8 +109,8 @@
         "VLM_ENDPOINT_URL = \"\"\n",
         "VLM_MODEL_TYPE = \"\"\n",
         "\n",
-        "LLM_DEVICE_ID = \"0\"\n",
-        "VLM_DEVICE_ID = \"1\"\n",
+        "LLM_DEVICE_ID = \"\"  # was \"0\"; cleared so the profile default places the VSS LLM (GPU placement is owned by the skill-based deploy, not this notebook)\n",
```
nvleeryan: we want to keep the default because user could be running the notebook manually instead of through the skill.  we can override that from the skill itself so user doesn't need to do or know anything extra

deploy/docker/scripts/deploy_vss_orchestrator.ipynb
```diff
@@ -153,7 +153,11 @@
         "LLM_DEVICE_ID = str(LLM_DEVICE_ID or os.environ.get(\"LLM_DEVICE_ID\", \"\")).strip()\n",
         "VLM_DEVICE_ID = str(VLM_DEVICE_ID or os.environ.get(\"VLM_DEVICE_ID\", \"\")).strip()\n",
         "LLM_NAME = (LLM_NAME or os.environ.get(\"LLM_NAME\", \"\")).strip()\n",
+        "# VSS reuses the NemoClaw agent vLLM (no dedicated VSS LLM NIM): a non-empty LLM_ENDPOINT_URL\n",
```
nvleeryan: again, this is not the default for the notebook (recommend to remove the comment to avoid ambiguity

deploy/docker/scripts/deploy_vss_orchestrator.ipynb
```diff
@@ -153,7 +153,11 @@
...
+        "# auto-derives LLM_MODE=remote + LLM_NAME_SLUG=none -> no VSS LLM NIM container.\n",
         "LLM_ENDPOINT_URL = (LLM_ENDPOINT_URL or os.environ.get(\"LLM_ENDPOINT_URL\", \"\")).strip()\n",
+        "if not LLM_ENDPOINT_URL and EXTERNAL_IP:\n",
+        "    LLM_ENDPOINT_URL = f\"http://{EXTERNAL_IP}:18000\"\n",
```
nvleeryan: first of all, we should not hard code 18000, user can use any arbitrary port.
second, while it's ok for (a) custom + BYO vLLM on 18000, it will break (a) custom + a cloud endpoint | (c) build.  Recommend to remove unless there is a reason this is absolutely necessary.

PR #1828, opened 2026-08-25
feat(nemoclaw): add standalone NeMo Relay notebook

simings-nv:
## Summary

- add `deploy_nemo_relay.ipynb` as a standalone, optional post-NemoClaw notebook
- remove Relay settings and setup logic from the crowded `deploy_nemoclaw.ipynb`; that notebook now contains only a short handoff
- keep the operator order explicit: NemoClaw, optional NeMo Relay, then the VSS Orchestrator
- run the Relay notebook in a fresh kernel in the existing skill-eval setup path
- add focused notebook contracts and a scoped third-party license inventory

Tracking: [VIA-2644](https://jirasw.nvidia.com/browse/VIA-2644)

## Operator flow

1. Complete `deploy_nemoclaw.ipynb`.
2. For an OpenClaw sandbox that needs trajectory capture, run `deploy_nemo_relay.ipynb`.
3. Continue with `deploy_vss_orchestrator.ipynb`.

The Relay notebook must be rerun after the sandbox is recreated. Hermes users do not run it because Hermes already includes NeMo Relay.

The notebook is the opt-in boundary. `RELAY_OBSERVABILITY=True` installs or refreshes the pinned plugin. Setting it to false and rerunning disables an earlier installation.

## What the Relay notebook does

1. Verifies the named sandbox and required OpenClaw/npm commands.
2. Applies NemoClaw's built-in `npm` egress preset.
3. Installs `nemo-relay-openclaw@0.7.3` with `--force`.
4. Resolves the active plugin generation with `openclaw plugins inspect` and installs the matching native binding there. This avoids writing into a stale generation left by a prior refresh.
5. Writes `plugins.entries.nemo-relay`, restarts or recovers the gateway, and fails unless the runtime reports the plugin loaded.

It does not modify `plugins.allow`, which would become an exclusive allowlist and could disable other non-bundled plugins.

## Data handling

Prompt text, per-call response text, tool arguments, and tool results are excluded. OpenTelemetry export is disabled.

NeMo Relay 0.7.3 still records the agent's final message in its session-close summary. The notebook states this before installation and tells operators not to enable Relay when that output must not be stored. ATIF files are written when a session closes or the gateway restarts.

## Validation

The new notebook was executed from beginning to end in a fresh kernel against the live `demo` NemoClaw sandbox on September 3, 2026:

```text
Executed deploy_nemo_relay.ipynb from beginning to end; outputs were not persisted.
```

Post-run checks showed:

[...]

PR #2028, opened 2026-09-04
fix(agent): treat an already-registered rt-embed asset as a successful upload

hugoverjus:
> **Caller-visible behaviour change — please read before reviewing the diff.**
> On a deployment where VIOS streamprocessing owns embedding generation, this upload now returns **200 in ~0.13s with the video not yet searchable**, reporting `chunks_processed: null`. Previously the agent's own synchronous call meant a 200 implied "indexed and searchable". **Clients that upload and then search must now poll for the embeddings** rather than assume they exist. Why the barrier cannot be restored here is in [Contract change](#the-contract-change-and-why-the-barrier-cannot-be-restored-here) — that section is the substance of this PR, not an afterthought.

## The defect

`PUT /api/v1/videos-for-search/<file>` (and `POST /api/v1/videos/{sensor_id}/complete`) return **502 while the embeddings are generated and indexed correctly**.

Where VIOS streamprocessing fans `camera_streaming` out to rt-embed's `/v1/stream/add`, that webhook registers the asset **and starts generation itself**. The agent's own `generate_video_embeddings` for the same asset id then loses and rt-embed rejects it:

- `400 AssetAlreadyExists` once the asset is published — `check_asset_exists` in `generate_video_embeddings`
- `409 DuplicateAssetId` while it is still reserved mid-download — `_reserve_asset_slot`

`_run_rtvi_embedding` raised on any non-200, so the API reported failure for work that had completed. An API that reports failure while completing the work is the worst shape for anything automated: every caller either retries work that already happened or reports a false failure, and both were happening.

Pre-existing on `develop` and unrelated to the gateway work: `video_ingest.py` is byte-identical (md5 `123ffb2948ad80b837b08c437e6b25af`) on `develop`, on #1983's merge-base, on its head, and in the running container. Sibling pre-existing splits: #2015, #2023.

## Ownership: streamprocessing does *not* always register

The obvious structural fix — drop the agent's call and let streamprocessing own registration — is **wrong**, because that registration is deployment configuration, not an invariant of the code path:

| Deployment | agent `rtvi_embed_base_url` | streamprocessing → rt-embed `stream/add` |
|---|---|---|
| Docker `dev-profile-search` | set (`COSMOS_EMBED_ENDPOINT`) | **yes** — `deploy/docker/developer-profiles/dev-profile-search/vios/configs/notification_config.json`, on `camera_streaming`, with `user_defined_metadata: {model, chunk_duration: 5}` |
| Helm `dev-profile-search` | set (`RTVI_EMBED_BASE_URL`) | **no** — the chart renders `deploy/helm/services/vios/charts/vios-streamprocessing/configs/notification_config.json` verbatim (only redis/kafka/mqtt address substitution), and that file is `webhooks.enabled: false` with a single disabled dummy item |

Those are the only two deployments in the tree that configure the agent's embed call. On Helm the agent's call is the **only** thing that generates embeddings for an uploaded file, so deleting it would turn this 502 into a 200 over an empty index — the same defect, now silent.

Two further asymmetries found while checking whether the two paths are equivalent:

- rt-embed only auto-starts generation when the webhook payload carries `metadata.model` (`StreamMetadata.has_embed_inference_params`). Registration and generation are separate steps in `cv_stream_add`, so a webhook without `model` registers the asset and generates nothing — and the agent could never retry, because the id is taken.
- `cv_stream_add` swallows a generation-start failure and still answers `200 {"status": "added", "inference": false}`. Full delegation therefore discards the only failure signal on this path.

So this PR takes the narrow option.

## The change

`_run_rtvi_embedding` treats exactly two `(status, code)` pairs as success and leaves every other non-200 fatal. The error **code** decides, not the status:

- `400 BadParameters` / `400 InvalidParameters` — malformed request, still fatal
- `409 DuplicateCameraId` / `409 ResourceInUse` — a *different* asset holds that camera id, still fatal
- a body we cannot parse — not attributable to the duplicate case, still fatal

That is deliberately the same discipline #1983 used for a marked vs. an unmarked 503; tolerating the status alone would re-create the class of bug that branch had just fixed.

## The contract change, and why the barrier cannot be restored here

The first version of this fix returned `200` with `chunks_processed: 0` and *"embeddings generated"*. That was still dishonest, and the downstream run proved it: `test-search` got a clean upload and then failed its `search` and `embed_search` groups against a corpus that was **3 documents** where healthy runs have 12 and 42.

The reason is that the agent's synchronous call was doing more than generating embeddings — **it was the only thing making a 200 mean "indexed and searchable"**. The owning path is asynchronous by design: `cv_stream_add` starts generation and answers `status: "processing"` without waiting for it. Measured on hardware:

```
http_code=200 time=0.133530
{"message":"…not searchable yet — poll…","chunks_processed":null}
index at the moment of the 200:  117 (delta 0)      <- zero embeddings exist
index +15s:                      126 (delta 9)      <- generation lands later
```

So a caller that uploads and then searches was correct against the old behaviour and now races — intermittently, which is a worse debugging experience than the 502 was.

**Restoring the barrier is not feasible without a change I don't think this PR should carry.** Both candidate completion conditions fail:

- **Ask rt-embed.** There is no signal. `GET /v1/stream/get-stream-info` skips non-live assets (`if not asset.is_live: continue`) and a VIOS file sensor is not live — the live deployment answers `{"status":"ok","stream_count":0,"stream_list":[]}` while assets exist.
- **Predict the document count and wait for it.** Only a heuristic. `FileSplitter` emits `ceil(duration / (chunk_duration − overlap))` chunks, *minus* a trailing sliver shorter than two video frames (`min_video_frame_duration_ns = 2e9/fps`) — so the count depends on fps and on sub-frame duration precision. A 40.1s clip at `chunk_duration: 5` yields 9 documents, and its 0.1s tail survives only because 0.1 > 2/30. Worse, the `chunk_duration` actually in force is the one in VIOS's `notification_config.json`, which the agent never reads; the agent's own `rtvi_embed_chunk_duration` matches it only by coincidence of defaults. Waiting for a predicted count would hang forever whenever the prediction was off by one, turning a race into a guaranteed timeout.
- **"No new documents for N seconds"** is a guess, and not a conservative one: generation is chunk-by-chunk through a VLM, so any GPU stall longer than N reports false completion.

The honest structural answer is a generation-status signal on rt-embed that the agent (or the caller) can wait on. That is an API addition with its own review and is **not** in this PR.

So what this PR does instead is stop overstating what happened. `chunks_processed` distinguishes three outcomes that were previously collapsed into `0`:

| value | meaning |
|---|---|
| positive `int` | generation completed inside this request — the video **is** searchable |
| `null` | the asset was already registered elsewhere; generation is running there and is **not** finished — poll before querying |
| `0` | no embedding work was requested at all (rt-embed not configured) |

and the message says so in words: *"embedding generation is owned by another registrant and is still in progress. The video is not searchable yet — poll for its embeddings before querying."*

**The status stays 200 deliberately.** A 202 would be more expressive, but `upload_video_via_agent` in the `metromind/ci-vss-oss` fixture treats anything other than 200 as a failure and does not retry a 202 — so it would re-break `test-search` in the exact way this branch set out to fix.

I have deliberately **not** touched the fixture's `min_docs=1` readiness gate. Strengthening a test so it stops noticing a weakened contract is backwards, and that gate should only be revisited once the product behaviour is right.

## Guard

`TestRtEmbedAssetAlreadyRegistered` drives the real ingest path (`_run_post_upload_processing`) over `httpx.MockTransport`, so real `httpx.Response` objects and real `response.json()` parsing are exercised rather than mocked away — following the `TestOptionalServiceBehindTheGateway` precedent. That matters here because the fix reads a JSON body and a `MagicMock` would answer whatever we told it to.

`test_the_three_outcomes_are_distinguishable` pins the three `chunks_processed` states together rather than apart, because the bug is a *collision* between them and only a shared test catches that.

**Canaried five ways**, each reintroducing a specific wrong answer:

| Canary | Failures |
|---|---|
| the original `raise` on any non-200 | 2 — both tolerated cases |
| tolerate all 4xx | 9 — every real rejection and every unparseable body |
| drop the agent's embed call entirely (the structural option) | 19 |
| report `0` for the deferred case (the state collision) | 3 |
| keep `null` but let the message still claim "embeddings generated" | 2 |

## Proven live

`dev-profile-search` on hardware, `docker cp` into the running agent (the box's NGC credential is expired, so no image can be rebuilt). The agent carries #1983's gateway patch throughout.

| Arm | HTTP | body | index `mdx-embed-filtered-2025-01-01` |
|---|---|---|---|
| before | **502** | `Embedding generation failed with status 400: {"code":"AssetAlreadyExists"…}` | 81 → **90** (9 docs) |
| first fix | **200** | `{"message":"…embeddings generated","chunks_processed":0}` | 90 → **99** (9 docs) |
| this fix | **200** | `{"message":"…not searchable yet — poll…","chunks_processed":null}` | 117 → **126**, and **delta 0 at the moment of the 200** |

Identical work in all three; the last one is the only one whose response describes it correctly.

## Downstream evidence, including where the original claim was wrong

Pipeline [66190921](https://gitlab-master.nvidia.com/metromind/ci-vss-oss/-/pipelines/66190921) at this branch's head, on a stock image rather than a patch:

- **Upload fixed.** `✓ Video uploaded: warehouse_sample_video` first attempt. A `test-search` on an unrelated branch shows the defect and its accidental workaround — `✗ Upload attempt 1/3 failed with HTTP 502` then `✓ Video uploaded` — because the fixture retries 502 and the retry re-uploads under a fresh sensor id. That is why this looked intermittent.
- **`attribute_search` cascade cleared.** All three pass, and `RTVI CV response missing or empty 'data' field` appears nowhere in the trace.
- **The `search` timeouts were never downstream of this 502 — I claimed otherwise and was wrong.** A `test-search` on branch `ed9bb689…`, without this fix, uploaded cleanly with 12 documents indexed and `embed_search` returning 42 results, and still shows the same four `search` timeouts. That is a separate pre-existing defect. **#2021 and #1828 will not be fixed by this PR**, and the !498 / !499 workarounds should not be dropped on the strength of it.

## Checks

`196 passed` across `tests/unit_test/api/`, `ruff check` and `mypy` clean.

### DIFF
--- a/services/agent/packages/vss_agents/src/vss_agents/api/video_ingest.py
+++ b/services/agent/packages/vss_agents/src/vss_agents/api/video_ingest.py
@@ -188,7 +188,15 @@ class VideoIngestResponse(BaseModel):
     message: str = Field(..., description="Status message indicating completion")
     sensor_id: str = Field(..., description="VST sensor id for the uploaded video (matches the {sensor_id} path param)")
     filename: str = Field(..., description="The filename returned by VST after upload")
-    chunks_processed: int = Field(default=0, description="Number of chunks processed")
+    chunks_processed: int | None = Field(
+        default=0,
+        description=(
+            "Number of chunks embedded by this request. ``null`` means the asset was already"
+            " registered by another party — generation is running there and is still in"
+            " progress, so no count is available and the video is not yet searchable."
+            " ``0`` means no embedding work was requested at all."
+        ),
+    )
 
 
 class VideoUploadUrlInput(BaseModel):
@@ -442,17 +450,23 @@ async def _run_rtvi_embedding(
     rtvi_embed_chunk_duration: int,
     start_timestamp: str,
     timeout_seconds: float = DEFAULT_RTVI_EMBED_TIMEOUT_SECONDS,
-) -> int:
+) -> int | None:
     """POST ``/v1/generate_video_embeddings``. Returns ``total_chunks_processed``.
 
     The call is synchronous on the RTVI-Embed side — it blocks until the
-    generation completes (up to the 600s client timeout). Any non-200 raises
-    ``HTTPException(502)`` so the caller can surface the failure, except the
-    two responses that mean the asset id is already registered — see
-    :func:`_already_registered_code`. Those return 0 chunks: generation is
-    someone else's, so we cannot count it, but the upload did succeed and
-    reporting a failure would make every caller either retry work that already
-    happened or report an upload that actually worked as broken.
+    generation completes (up to the 600s client timeout), so an ``int`` return
+    means the embeddings exist and the video is searchable.
+
+    Any non-200 raises ``HTTPException(502)`` so the caller can surface the
+    failure, except the two responses that mean the asset id is already
+    registered — see :func:`_already_registered_code`. Those return ``None``,
+    which is *not* the same as zero: the upload succeeded and generation is
+    running, but it belongs to the other registrant, so there is no count to
+    report and — crucially — no completion to report either. The owning path is
+    asynchronous: rt-embed's ``stream/add`` starts generation and answers
+    ``status: "processing"`` without waiting for it, so nothing in that flow
+    blocks until the embeddings land. ``None`` is what carries that "started
+    elsewhere, not finished" state to the caller; see the response model.
     """
     rtvi_embed_url = rtvi_embed_base_url.rstrip("/")
     embedding_url = f"{rtvi_embed_url}/v1/generate_video_embeddings"
@@ -486,12 +500,14 @@ async def _run_rtvi_embedding(
                 logger.warning(
                     "RTVI-Embed already holds asset %s (%d %s) — another registrant, normally"
                     " VIOS streamprocessing's stream/add webhook, owns it and drives generation."
-                    " Treating the upload as successful; the chunk count is not ours to report.",
+                    " The upload succeeded, but that generation is asynchronous and still in"
+                    " flight: the embeddings are not queryable yet and this request cannot tell"
+                    " when they will be. Callers that search straight after uploading must poll.",
                     sensor_id,
                     response.status_code,
                     already_registered,
                 )
-                return 0
+                return None
             error_msg = f"Embedding generation failed with status {response.status_code}: {response.text}"
             logger.error(error_msg)
             raise HTTPException(status_code=502, detail=f"Embedding generation failed: {error_msg}")
@@ -622,7 +638,7 @@ async def _run_post_upload_processing(
     else:
         logger.info("RTVI-CV not configured, skipping")
 
-    chunks_processed = 0
+    chunks_processed: int | None = 0
     if parsed_embed is not None:
         rtvi_tasks.append(
             (
@@ -656,13 +672,26 @@ async def _run_post_upload_processing(
                 logger.error("%s task failed: %s", label, result)
                 raise result
             if label == "rtvi-embed":
-                chunks_processed = result or 0
+                # ``None`` is meaningful here (generation deferred to whoever
+                # already owns the asset) so it must survive to the response —
+                # ``or 0`` would collapse it into "no work requested".
+                chunks_processed = result
 
-    message = (
-        f"Video {filename} successfully uploaded to VST and embeddings generated"
-        if parsed_embed is not None
-        else f"Video {filename} successfully uploaded to VST"
-    )
+    # Three distinct outcomes, and the message says which. The deferred one is
+    # the only case where the upload succeeded but the video is *not* yet
+    # searchable, so it must not read like the completed one — a caller that
+    # uploads and immediately searches was correct against the old synchronous
+    # behaviour and now races, and this is what tells it so.
+    if parsed_embed is None:
+        message = f"Video {filename} successfully uploaded to VST"
+    elif chunks_processed is None:
+        message = (
+            f"Video {filename} successfully uploaded to VST; embedding generation is owned by"
+            " another registrant and is still in progress. The video is not searchable yet —"
+            " poll for its embeddings before querying."
+        )
+    else:
+        message = f"Video {filename} successfully uploaded to VST and embeddings generated"
     return VideoIngestResponse(
         message=message,
         sensor_id=sensor_id,
--- a/services/agent/packages/vss_agents/src/vss_agents/api/video_ingest.py
+++ b/services/agent/packages/vss_agents/src/vss_agents/api/video_ingest.py
@@ -384,6 +384,55 @@ async def _register_with_rtvi_cv(
         logger.warning("RTVI-CV timed out at %s, skipping", rtvi_cv_add_url)
 
 
+# rt-embed refuses a ``generate_video_embeddings`` call whose asset id is
+# already taken, and the status it uses depends on how far the other
+# registration has got: ``check_asset_exists`` rejects a published asset with
+# ``400 AssetAlreadyExists``, while ``_reserve_asset_slot`` rejects one that is
+# still reserved (mid-download) with ``409 DuplicateAssetId``. Both mean
+# something else owns this asset id, which on a deployment where VIOS
+# streamprocessing fans ``camera_streaming`` out to rt-embed's ``/v1/stream/add``
+# is the normal outcome: that webhook registers the asset *and* starts
+# generation, so the agent's call for the same id is redundant rather than
+# unlucky.
+#
+# It is only redundant where that webhook exists, which is why the call is not
+# simply dropped. The Docker ``dev-profile-search`` ships it; the Helm
+# ``dev-profile-search`` sets ``RTVI_EMBED_BASE_URL`` for the agent but renders
+# streamprocessing's ``notification_config.json`` with ``webhooks.enabled:
+# false``, so there the agent's call is the *only* thing that generates
+# embeddings for an uploaded file. Deleting it would turn this 502 into a 200
+# with an empty index — the same bug, now silent.
+_ALREADY_REGISTERED_CODE_BY_STATUS = {
+    400: "AssetAlreadyExists",
+    409: "DuplicateAssetId",
+}
+
+
+def _already_registered_code(response: httpx.Response) -> str | None:
+    """Return rt-embed's error code when it says the asset id is already owned.
+
+    ``None`` for anything else, so every other non-200 stays fatal. The pairing
+    is checked both ways round: a 400 must carry ``AssetAlreadyExists`` and a
+    409 must carry ``DuplicateAssetId``, because the whole point is to
+    distinguish "already done" from a request rt-embed rejected on its merits.
+    ``400 BadParameters`` (bad model, bad url) and ``409 DuplicateCameraId``
+    (a *different* asset already holds this camera id) are real failures that
+    share a status code with the tolerated pair, so the code is what decides.
+    An unparseable body is not tolerated either — same discipline as the
+    gateway's marked-vs-unmarked 503.
+    """
+    expected_code = _ALREADY_REGISTERED_CODE_BY_STATUS.get(response.status_code)
+    if expected_code is None:
+        return None
+    try:
+        body = response.json()
+    except ValueError:
+        return None
+    if not isinstance(body, dict) or body.get("code") != expected_code:
+        return None
+    return expected_code
+
+
 async def _run_rtvi_embedding(
     *,
     rtvi_embed_base_url: str,
@@ -398,7 +447,12 @@ async def _run_rtvi_embedding(
 
     The call is synchronous on the RTVI-Embed side — it blocks until the
     generation completes (up to the 600s client timeout). Any non-200 raises
-    ``HTTPException(502)`` so the caller can surface the failure.
+    ``HTTPException(502)`` so the caller can surface the failure, except the
+    two responses that mean the asset id is already registered — see
+    :func:`_already_registered_code`. Those return 0 chunks: generation is
+    someone else's, so we cannot count it, but the upload did succeed and
+    reporting a failure would make every caller either retry work that already
+    happened or report an upload that actually worked as broken.
     """
     rtvi_embed_url = rtvi_embed_base_url.rstrip("/")
     embedding_url = f"{rtvi_embed_url}/v1/generate_video_embeddings"
@@ -427,6 +481,17 @@ async def _run_rtvi_embedding(
             )
 
         if response.status_code != 200:
+            already_registered = _already_registered_code(response)
+            if already_registered is not None:
+                logger.warning(
+                    "RTVI-Embed already holds asset %s (%d %s) — another registrant, normally"
+                    " VIOS streamprocessing's stream/add webhook, owns it and drives generation."
+                    " Treating the upload as successful; the chunk count is not ours to report.",
+                    sensor_id,
+                    response.status_code,
+                    already_registered,
+                )
+                return 0
             error_msg = f"Embedding generation failed with status {response.status_code}: {response.text}"
             logger.error(error_msg)
             raise HTTPException(status_code=502, detail=f"Embedding generation failed: {error_msg}")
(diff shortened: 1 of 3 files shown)


hugoverjus:
### CI authorization and status

Authorized with `/ok to test 61eedb0ebd1d648b64a54fc3d67e603ef298c046`; copy-pr-bot mirrored it to `pull-request/2028` at exactly that SHA and 12 workflows started.

Green: **Build Dev Images (GHCR)**, **OSRB Scan**, **Fern Docs CI**, **Helm Sync**, **Spatial AI Data Utils**, **Model Routing Notebook**, **vss-playbook-compliance**, all four Skills workflows. Within **CI**: `Test (pytest)` (3613 passed, including all 15 cases of the new guard), `Type Check (mypy)`, `Lint (Python)`, `Lint & Typecheck (UI)`, `DCO Sign-off`, `Copyright Headers`, `Security Scan (detect-secrets)`, `Test (vss-cli)`, `No Patented Codecs (agent image)`.

`Build (UI)` was **cancelled by its 10-minute job timeout one second after the build itself succeeded** (`Tasks: 10 successful, 10 total` at 07:18:12, timeout at 07:18:13, with `Cached: 0 cached, 10 total` — an uncached `npm ci`). It succeeds on every other recent CI run, and this PR touches no UI file. Needs a re-run, not a fix here.

`Trigger Downstream Pipeline` is still running — that is the GitLab pipeline carrying `test-search`, the job this PR is meant to unblock.

### Live proof, in full

`dev-profile-search` on hardware, both arms `docker cp`'d into the running agent (the box's NGC credential is expired, so no image can be rebuilt). The agent carries #1983's gateway live patch throughout, visible in both logs as `RTVI-CV not deployed behind … (gateway reports rtvi-cv absent)`; `utils/gateway.py` was untouched.

Before — `video_ingest.py` md5 `56edf56484a6054135f72e3a2caa8718` (gateway patch only):

```
$ curl -X PUT http://localhost:7777/api/v1/videos-for-search/dupfix1.mp4 -H 'Content-Type: video/mp4' --data-binary @dupfix1.mp4
http_code=502
{"detail":"Embedding generation failed: Embedding generation failed with status 400:
 {\"code\":\"AssetAlreadyExists\",\"message\":\"Asset with id f0ca4ecd-… already exists.\"}"}

ERROR - Embedding generation failed with status 400: {"code":"AssetAlreadyExists"…}
ERROR - rtvi-embed task failed: 502: Embedding generation failed…
INFO:  "PUT /api/v1/videos-for-search/dupfix1.mp4 HTTP/1.1" 502 Bad Gateway
```

`mdx-embed-filtered-2025-01-01`: **81 → 90**. Nine documents indexed while the API said 502.

After — md5 `8b822553b10dc9b732122e5682186716` (gateway patch **plus** this fix):

```
$ curl -X PUT http://localhost:7777/api/v1/videos-for-search/dupfix2.mp4 -H 'Content-Type: video/mp4' --data-binary @dupfix2.mp4
http_code=200
{"message":"Video dupfix2 successfully uploaded to VST and embeddings generated",
 "sensor_id":"8343df9b-f3ac-455c-b332-5ef8c3ad9194","filename":"dupfix2","chunks_processed":0}

WARNING - RTVI-Embed already holds asset 8343df9b-… (400 AssetAlreadyExists) — another
 registrant, normally VIOS streamprocessing's stream/add webhook, owns it and drives
 generation. Treating the upload as successful; the chunk count is not ours to report.
INFO:  "PUT /api/v1/videos-for-search/dupfix2.mp4 HTTP/1.1" 200 OK
```

`mdx-embed-filtered-2025-01-01`: **90 → 99**. Nine documents again — identical work, only the reported status changed.

Full artefacts in `~/vss-dupfix-evidence/` on the box (`SUMMARY.md`, both responses, both agent logs, the four index counts, and `video_ingest.py.gateway-only-backup` for reverting the agent to gateway-patch-only).

hugoverjus:
### Downstream verdict: the upload and `attribute_search` cascade are fixed; two things remain, one of them mine

Downstream pipeline [66190921](https://gitlab-master.nvidia.com/metromind/ci-vss-oss/-/pipelines/66190921) (`VSS_SUBMODULE_HASH = 61eedb0ebd1d648b64a54fc3d67e603ef298c046`), job `test-search` — running on a **stock image built from this branch**, not a `docker cp` patch.

**Fixed.** The upload succeeds on the first attempt and the corpus is present:

```
[2] Upload test video
  ✓ Video uploaded: warehouse_sample_video
  ✓ ES index 'mdx-embed-filtered-2025-01-01' ready with 3 doc(s)
```

All three `attribute_search` tests pass, and `RTVI CV response missing or empty 'data' field` — the symptom of the cascade — appears nowhere in the trace.

For contrast, a `test-search` on an unrelated branch 20 minutes earlier shows the defect and its accidental workaround:

```
✗ Upload attempt 1/3 failed with HTTP 502: Bad Gateway
✓ Video uploaded: warehouse_sample_video
```

`upload_video_via_agent` retries 502, and the retry re-uploads under a fresh sensor id which sometimes wins the race. That is why this defect looked intermittent.

**Not caused by this PR: `POST /api/v1/search` timing out.** A `test-search` on branch `ed9bb689…`, without this fix, uploaded cleanly with 12 documents indexed and `embed_search` returning 42 results, and still shows four `search` timeouts (`a person carrying boxes (agent_mode=True)`, `required fields`, `similarity range and ordering`, `video_sources filter`). So the `search` timeouts are a **separate pre-existing defect**; they were never downstream of the 502.

**Caused by this PR, and it needs addressing: the fix removes a synchronous barrier.** `_run_rtvi_embedding` used to block until generation completed, so `PUT /api/v1/videos-for-search` only returned 200 once the corpus was fully indexed. Deferring to streamprocessing returns immediately. Measured on hardware:

```
--- PUT start 08:30:12.865 ---
http_code=200 time=0.126578
--- PUT returned 08:30:13.002 ---
08:30:13 count=99  (delta 0)      <- 200 already returned, zero embeddings
08:30:16 count=108 (delta 9)      <- generation lands ~3s later
```

So the response says `embeddings generated` when none exist yet. The fixture's readiness gate is `_wait_for_es_index(..., min_docs=1)`, which is satisfied by the first chunk, so the searches run against a corpus that is still building — 3 documents here versus 12 and 42 on the two comparison runs. `embed_search` returns `ExecutionFailed` at 3 documents and passes at 12 and 42, which makes the incomplete corpus the leading explanation for that group.

That is a real gap in this PR, not a fixture problem: an API that returns 200 saying "embeddings generated" before any exist is a weaker version of the same dishonesty this PR set out to fix.

**rt-embed offers no completion signal for this case.** `GET /v1/stream/get-stream-info` skips non-live assets (`if not asset.is_live: continue`), and a VIOS file sensor is not live — the live deployment reports `{"status":"ok","stream_count":0,"stream_list":[]}` while assets exist. So the agent cannot currently wait on rt-embed for a file-sensor generation to finish.

Options, smallest first:

1. Stop claiming completion in the deferred case — distinct message (and arguably 202) so callers know the corpus is still building — and strengthen the fixture's gate to wait for a stable count rather than `min_docs=1`.
2. Add a file-sensor generation-status signal to rt-embed and have the agent wait on it, restoring the barrier honestly.

I would rather land (1) here and raise (2) separately than leave the response overstating what happened. Holding this PR pending a call on which.

hugoverjus:
### Second commit: the deferred response now says what actually happened

The downstream run showed the first fix was honest about the *status* and still dishonest about the *state* — a clean upload followed by `search`/`embed_search` failures against a 3-document corpus where healthy runs have 12 and 42. Root cause: `_run_rtvi_embedding`'s synchronous call was the only thing making a 200 mean "indexed and searchable", and deferring to streamprocessing removes that. `cv_stream_add` starts generation and answers `status: "processing"` without waiting, so nothing in the owning path blocks.

I looked at restoring the barrier and concluded it is not possible here without a change this PR should not carry — details in the PR body. Briefly: rt-embed exposes no completion signal (`stream/get-stream-info` skips non-live assets, and a VIOS file sensor is not live — the live box answers `stream_count: 0` while assets exist), and the document count cannot be predicted, because `FileSplitter` drops a trailing sliver shorter than two video frames and the `chunk_duration` in force belongs to VIOS's `notification_config.json`, which the agent never reads. Waiting for a predicted count would convert an intermittent race into a guaranteed timeout.

So `chunks_processed` now separates the three states that were collapsed into `0` — a positive count means searchable, `null` means generation is running elsewhere and is not finished, `0` means no embedding work was requested — and the message says the video is not searchable yet and the caller must poll. Live:

```
http_code=200 time=0.133530
{"message":"…not searchable yet — poll for its embeddings before querying.","chunks_processed":null}
index at the moment of the 200:  117 (delta 0)
index +15s:                      126 (delta 9)
```

Status stays 200 on purpose: `upload_video_via_agent` treats anything else as a failure and does not retry a 202, so 202 would re-break `test-search` in the exact way this branch set out to fix. I have not touched the fixture's `min_docs=1` gate — strengthening a test so it stops noticing a weakened contract is backwards.

Two guards added and canaried: reporting `0` for the deferred case fails 3 tests, and keeping `null` while letting the message still claim "embeddings generated" fails 2.

**`Build (UI)` is green** on this SHA — `npm ci` 4m38s, `Build` 3m31s, 8m23s total against the 10-minute cap. Third data point for that job living at its limit (6m21s / 10m20s / 8m23s); worth its own PR to raise the timeout, with a turbo task cache as the better follow-up.

**Correction carried into the PR body:** the `search` timeouts are a separate pre-existing defect, not a consequence of this 502 — a run without this fix uploaded cleanly with 12 documents and still timed out the same four ways. #2021 and #1828 are not fixed by this PR and the !498 / !499 workarounds should not be dropped on the strength of it.

hugoverjus:

both commits were cherry-picked with their original messages and authorship:

|---|---|
| `61eedb0eb` fix(agent): treat an already-registered rt-embed asset as a successful upload | `416f7e01891a1345349c9e21cb14b62d998cfc46` |
| `7e6437940` fix(agent): say plainly when embedding generation was deferred | `de6c01cfcdb493607f6562793496b769f5e1ec74` |

Consolidated head: `de6c01cfcdb493607f6562793496b769f5e1ec74` (also the second
commit).

### The `video_ingest.py` overlap

This was the one real conflict, and it resolved the opposite way round from what
was expected: `video_ingest.py` auto-merged and **`test_video_ingest.py`
conflicted**. A clean merge is not evidence in this file — during #1983's own
`develop` merge it merged with no textual conflict and then three tests failed —
so the source auto-merge was checked algebraically rather than by eye: the delta
applied on top of the gateway branch is line-for-line identical to this PR's own
delta against its parent, for both commits and for `video_search_ingest.py`. No
gateway content was displaced.

All three behaviours are preserved and individually tested:

- the gateway branch's `rewrite_to_internal_vst_url` translation and its
  marked-`503` tolerance in `_register_with_rtvi_cv`;
- `develop`'s hoisting of the VST rewrite into `_run_post_upload_processing`
  (which removed the `vst_url` parameter);
- this PR's error-code tolerance in `_run_rtvi_embedding`.

The test-file conflict was two independent test classes interleaved by git —
`TestRtEmbedAssetAlreadyRegistered` from here and
`TestOptionalServiceBehindTheGateway` from #1983. Both are kept whole; this PR's
class is byte-identical to its version here. All 22 tests across the two classes
pass, 108 in the file, 3641 in the agent suite with mypy clean.

### Canaries

Re-run after consolidation, both arms:

- disabling the tolerance (the original defect) → 3 tests fail and the `502`
  returns;
- keying the tolerance on the **status** instead of the error code → 7 tests
  fail, including every `stay_fatal` case.

### The open item moves to #1983

`embed_search` verification is still outstanding, and #1983 now carries it under
*Knowingly incomplete*: making upload asynchronous means upload-then-search
callers must poll, and the eval fixture — which proceeds at the first indexed
document — now races a corpus that fills over several seconds. The test-side fix
in GitLab `ci-vss-oss` is untouched by any of this.

**This branch is deliberately not deleted** — both because the consolidation
could be reverted and because of that open work.

PR #2037, opened 2026-09-04
fix(rtvi-vlm): Redact URL credentials from logs

nv-amkale:
## Summary

Redact credentials embedded in URLs before RTVI writes request and asset
locations to logs. Preserve the URL host and path so operators retain useful
context without exposing user information, query parameters, fragments, or
URL headers.

## Related Issue

NVBug 6715329

## Changes

- Sanitize structured request data before logging it.
- Sanitize URL-bearing asset, redirect, S3, and server log arguments.
- Apply a logger-level filter as defense in depth for embedded URLs.
- Cover access tokens, signed URLs, user information, and hostile delimiters.

## Validation

- Built `rtvi-vlm-dev:local` from the exact branch worktree.
- Started the RTVI Compose stack and reached healthy service state.
- Sent the credential-bearing `generate_captions` runtime request.
- Confirmed the credential canary was absent from service logs.
- Confirmed the request reached the download path and returned the expected
  HTTP 404 response for the intentionally missing media URL.
- Independent code review approved the change.

### DIFF
--- a/services/rtvi/rt-vlm/src/common/logger.py
+++ b/services/rtvi/rt-vlm/src/common/logger.py
@@ -54,9 +54,7 @@ def sanitize_url_for_logging(url: str) -> str:
         netloc = parsed.netloc.rsplit("@", 1)[-1]
         return urlunsplit((parsed.scheme, netloc, parsed.path, "", ""))
     except (TypeError, ValueError):
-        # Keep malformed values out of logs past the first query/fragment
-        # delimiter even if urllib cannot parse them.
-        return str(url).split("?", 1)[0].split("#", 1)[0]
+        return "[malformed URL redacted]"
 
 
 def sanitize_urls_for_logging(message: str) -> str:
--- a/services/rtvi/rt-vlm/src/common/logger.py
+++ b/services/rtvi/rt-vlm/src/common/logger.py
@@ -17,7 +17,11 @@
 import logging
 import logging.handlers
 import os
+import re
 import time
+from collections.abc import Mapping
+from typing import Any
+from urllib.parse import urlsplit, urlunsplit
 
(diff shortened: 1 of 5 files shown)


nv-amkale:
1. [P1] Malformed network URLs can still expose credentials.
   [logger.py (line 51)](vss-pr2037-redaction-fix/services/rtvi/rt-vlm/src/common/logger.py:51) assumes credentials are in netloc. Inputs with an empty authority do not raise:
   http:////user:CANARY_SECRET@example.com/video.mp4
   → http://user:CANARY_SECRET@example.com/video.mp4
   Require a non-empty authority for http, https, rtsp, and s3; otherwise return the malformed-URL placeholder. Add this exact regression case.
2. [P1] Propagated exception tracebacks bypass the filter.
   [logger.py (line 110)](vss-pr2037-redaction-fix/services/rtvi/rt-vlm/src/common/logger.py:110) sanitizes record.msg, but leaves record.exc_info untouched. Because this logger has propagate=True, a root or observability handler can append the original exception:
   RuntimeError: download failed for
   https://user:password@example.com/video.mp4?token=CANARY_SECRET
   Sanitize the formatted exception text before propagation or prevent unsanitized propagation, and test an exc_info=True record.
3. [P2] Data URL payloads remain visible.
   The URL pattern excludes data: URLs, while [rtvi_vlm_server.py (line 3150)](vss-pr2037-redaction-fix/services/rtvi/rt-vlm/src/server/rtvi_vlm_server.py:3150) logs their first 100 characters before validation. A non-base64 data URL can therefore expose its payload. Log only [data URL redacted] plus media type.

nv-pnikam (approved):
LGTM

PR #2044, opened 2026-09-04
fix(deploy): default RTVI_VLM_BASE_URL; publish the LLM at /llm on the edge

zac-wang-nv:
## Root cause

The 2026-09-03 nightly failed the warehouse report test with:

```
langchain_nvidia_ai_endpoints: The provided url appears incorrect.
  Expected format is: http://host:port  Got: /v1
aiohttp.client_exceptions.InvalidUrlClientError: /v1/chat/completions
```

**`LLM_BASE_URL` was not the empty variable.** Every profile `overrides.env` sets
`LLM_BASE_URL=http://vss-llm-nim:8000`, and the failing job's compose run resolved it
without an unset-variable warning. The gap is **`RTVI_VLM_BASE_URL`**:

| | |
|---|---|
| `services/agent/compose.yml:110` | `RTVI_VLM_BASE_URL: ${RTVI_VLM_BASE_URL}` — no fallback |
| every profile `overrides.env` | no `RTVI_VLM_BASE_URL` line at all |
| warehouse agent `config.yml:197` | `base_url: ${RTVI_VLM_BASE_URL}/v1` → **`/v1`** |
| `services/alert/compose.yml:52` | `${RTVI_VLM_BASE_URL:-http://rtvi-vlm:8000}` — already defaulted |

aiohttp rejects the relative path before opening a socket, so the failure surfaces as
the agent's canned apology rather than a connection error — no signal to anyone
downstream. The agent log confirms the ordering: the langchain URL warning fires
immediately after `Using VLM profile: rtvi_vlm, vlm_mode: none`.

**The fix** gives the agent the same default its sibling already has, rather than
inventing a new one. One line.

## Also in this PR: `/llm` on the ingress

Separate, requested change — not a fix for the above.

The agent reached the LLM by its in-cluster Service name, so the one OpenAI-shaped
backend in the stack had no mount on the canonical ingress route table. Nothing
outside the cluster could reach it and `vss configure` had no route to record. This
mounts it as **one row** in the shared table, so all four developer profiles and the
Docker edge pick it up together:

```
/llm/v1/models
/llm/v1/chat/completions
```

Mounted only where the profile runs the local NIM (`nims.nemotron35.enabled`); a
profile pointed at a remote LLM declares no backend and gets no route rather than a
route to a Service that was never created. The backend carries `timeout server 3600s`
matching `bk_lvs_strip`, since a long completion outlasts the 120s default.

The Helm `LLM_BASE_URL` default now resolves through that edge when
`global.externalHost` is set and `global.llmViaIngress` is left `true`; with no host
it falls back to the in-cluster Service, so it can never render hostless.
`global.llmViaIngress=false` is a kill switch.

## Plan

**1. Repoint the agent's `LLM_BASE_URL` at the edge, or just publish the route?**
- *Recommended:* publish the route only — smallest blast radius, no proxy in the
  streaming path.
- *Answered:* **repoint through HAProxy too.** Done for the four Helm developer
  profiles.

**2. HAProxy defaults to `timeout server 120s`, which would cut long completions.**
- *Recommended:* 3600s, matching `bk_lvs_strip`.
- *Answered:* **3600s.**

### Corrected after review

An earlier revision also repointed `LLM_BASE_URL` in five compose `overrides.env`
files. That was wrong — `LLM_BASE_URL=http://vss-llm-nim:8000` is correct as written,
and changing it was unrelated churn that did not touch the defect. **Reverted**; those
five files are now identical to `develop`.

### Decisions taken without asking

- **Mount path `/llm`, `rewrite: strip`** — the table's convention is that a service is
  mounted at its own name and carries what it itself serves, as `/rtvi-vlm` does.
- **Service name through `vss.subchartFullname`** — a literal renders wrongly under
  `useReleaseNamePrefix=true`. Verified both `nemotron-35-lightning-30b-a3b` and
  `rel-nemotron-35-lightning-30b-a3b`.
- **CLI `INGRESS_SERVICES` entry** — probe `/llm/v1/models`, mirroring `rt_vlm`.

## Verification

- `deploy/helm/scripts/verify-ingress-routes.py` → **OK**: 4 profiles render from one
  table (23 mounts, was 22); CLI probe mounts present; 6 disabled-component cases still
  drop their route; examples current.
- `haproxy -c` on `haproxy:3.4.2-alpine`: **0 errors**, and **5 warnings — identical to
  unmodified `origin/develop`**.
- `LLM_BASE_URL` rendered in all three states:

  | condition | value |
  |---|---|
  | `global.externalHost` set | `http://vss.10.0.0.1.nip.io/llm` |
  | no host set | `http://nemotron-35-lightning-30b-a3b:8000` |
  | `llmViaIngress=false` | `http://nemotron-35-lightning-30b-a3b:8000` |

- `services/agent/compose.yml` parses as YAML; `vss_cli.config` imports cleanly.

## Not covered

The second agent-side failure from the same nightly — NIM serving without
`--enable-auto-tool-choice` / `--tool-call-parser`, returning
`400 "auto" tool choice requires ...` on 15–30 requests per job — is untouched.
Separate change.

## Context

Found while tracing the 7 `ui_e2e` failures in the 2026-09-03 `nightly-develop` run.
The other four of those seven were the storage-ingress 301 already fixed by #2005.

### DIFF
--- a/deploy/docker/services/infra/haproxy/haproxy.cfg.template
+++ b/deploy/docker/services/infra/haproxy/haproxy.cfg.template
@@ -116,6 +116,14 @@ backend bk_rtvi_vlm_strip
     http-request replace-path ^/rtvi-vlm$ /
     server s1 "${RTVI_VLM_SERVICE_HOST}:${RTVI_VLM_SERVICE_PORT}" check resolvers docker init-addr none
 
+backend bk_llm_strip
+    http-request replace-path ^/llm/(.*) /\1
+    http-request replace-path ^/llm$ /
+    # A long completion outlasts the 120s default, and a 504 mid-generation
+    # reads as a backend failure. The caller bounds the wait itself.
+    timeout server 3600s
+    server s1 "${LLM_SERVICE_HOST}:${LLM_SERVICE_PORT}" check resolvers docker init-addr none
+
 backend bk_lvs_strip
     http-request replace-path ^/lvs/(.*) /\1
     http-request replace-path ^/lvs$ /
@@ -234,6 +242,13 @@ frontend fe_http
     acl p_rtvi_vlm path /rtvi-vlm
     acl p_rtvi_vlm path_beg /rtvi-vlm/
     use_backend bk_rtvi_vlm_strip if h_main p_rtvi_vlm
+
+    # The text LLM, mounted at its service name like the RTVI services above.
+    # Profiles pointed at a remote LLM run no local NIM, so the backend stays
+    # DOWN and the route answers 503 rather than proxying to nothing.
+    acl p_llm path /llm
+    acl p_llm path_beg /llm/
+    use_backend bk_llm_strip if h_main p_llm
 
     # Profiles without lvs-server leave the backend DOWN, so this answers 503
     # and `vss configure` records the service absent, not present-but-broken.
--- a/deploy/helm/developer-profiles/dev-profile-alerts/values.yaml
+++ b/deploy/helm/developer-profiles/dev-profile-alerts/values.yaml
@@ -36,6 +36,11 @@ ngc:
 # (verification: RTVI tracks VLM; real-time: RTVI uses remote VLM URL when set, else vss-rtvi-vlm). Override per-agent with agent.vss-agent.llmBaseUrl / vlmBaseUrl.
 # global.llmName / vlmName: shared defaults for vss-agent (override with agent.vss-agent.llmName, etc. when needed).
 global:
+  # Send the agent to the LLM through the HAProxy edge (/llm) instead of
+  # straight to the NIM Service. Requires global.externalHost; with no host
+  # set, the in-cluster Service URL is used, so this never yields a hostless
+  # base_url. Set false to keep agent->LLM traffic off the proxy.
+  llmViaIngress: true
   useReleaseNamePrefix: false
   postgresServiceHost: ""
   imagePullSecrets:
@@ -327,7 +332,7 @@ agent:
       - name: PHOENIX_ENDPOINT
         value: '{{- $g := .Values.global | default dict }}{{- $pfx := default false (coalesce .Values.useReleaseNamePrefix ((index $g "useReleaseNamePrefix") | default false)) }}{{ ternary (printf "http://%s-phoenix:6006" .Release.Name) "http://phoenix:6006" $pfx }}'
       - name: LLM_BASE_URL
-        value: '{{- $g := .Values.global | default dict }}{{- $llm := (default "" (coalesce .Values.llmBaseUrl (index $g "llmBaseUrl"))) | toString | trim }}{{- $pfx := default false (coalesce .Values.useReleaseNamePrefix ((index $g "useReleaseNamePrefix") | default false)) }}{{- $def := ternary (printf "http://%s-nemotron-35-lightning-30b-a3b:8000" .Release.Name) "http://nemotron-35-lightning-30b-a3b:8000" $pfx }}{{- if $llm }}{{ $llm }}{{- else }}{{ $def }}{{- end }}'
+        value: '{{- $g := .Values.global | default dict }}{{- $llm := (default "" (coalesce .Values.llmBaseUrl (index $g "llmBaseUrl"))) | toString | trim }}{{- $pfx := default false (coalesce .Values.useReleaseNamePrefix ((index $g "useReleaseNamePrefix") | default false)) }}{{- $def := ternary (printf "http://%s-nemotron-35-lightning-30b-a3b:8000" .Release.Name) "http://nemotron-35-lightning-30b-a3b:8000" $pfx }}{{- $edgeHost := ((index $g "externalHost") | default "") | toString | trim }}{{- $viaEdge := and (default false (index $g "llmViaIngress")) $edgeHost }}{{- if $llm }}{{ $llm }}{{- else if $viaEdge }}{{ printf "http://%s/llm" $edgeHost }}{{- else }}{{ $def }}{{- end }}'
       - name: VLM_BASE_URL
         value: '{{- if (trim (.Values.rtviVlmBaseUrl | default "")) }}{{ trim .Values.rtviVlmBaseUrl }}{{- else }}{{- $g := .Values.global | default dict }}{{- $pfx := default false (coalesce .Values.useReleaseNamePrefix ((index $g "useReleaseNamePrefix") | default false)) }}{{- $svc := trim (.Values.rtviVlmServiceName | default "vss-rtvi-vlm") }}{{- printf "http://%s:8000" (ternary (printf "%s-%s" .Release.Name $svc) $svc $pfx) -}}{{- end }}'
       - name: VST_EXTERNAL_URL
@@ -546,6 +551,11 @@ vssIngress:
   rtviCvPort: 9010
   rtviEmbedPort: 8000
   rtviVlmPort: 8000
+  # The text LLM is published at /llm, not at the origin root: /llm/v1/models,
+  # /llm/v1/chat/completions. Mounted only when this profile runs the local NIM
+  # (nims.nemotron35.enabled); a profile using a remote LLM gets no route.
+  llmServiceName: "nemotron-35-lightning-30b-a3b"
+  llmPort: 8000
   kibanaHost: ""
   # Default Phoenix exposure is path-based on main host: http://<host>/phoenix
   # Keep empty for path-only mode. Set phoenixHost (for example, phoenix.<host>)
--- a/deploy/helm/developer-profiles/dev-profile-alerts/vss-ingress-example-rewrites.yaml
+++ b/deploy/helm/developer-profiles/dev-profile-alerts/vss-ingress-example-rewrites.yaml
@@ -44,6 +44,8 @@ metadata:
       /elasticsearch /
       /rtvi-cv/(.*) /\1
       /rtvi-cv /
+      /llm/(.*) /\1
+      /llm /
       /phoenix/(.*) /\1
       /phoenix /
 spec:
@@ -94,6 +96,13 @@ spec:
             name: <RELEASE_NAME>-vss-rtvi-cv
             port:
               number: 9010
+      - path: /llm
+        pathType: Prefix
+        backend:
+          service:
+            name: <RELEASE_NAME>-nemotron-35-lightning-30b-a3b
+            port:
+              number: 8000
       - path: /phoenix
         pathType: Prefix
         backend:
--- a/deploy/helm/developer-profiles/dev-profile-base/values.yaml
+++ b/deploy/helm/developer-profiles/dev-profile-base/values.yaml
@@ -35,6 +35,11 @@ ngc:
 # Set global.imagePullSecrets when using a private registry (e.g. nvcr.io). Set global.storageClass for PVCs.
 # When using NGC (or copy from values-ngc-secrets.example.yaml):
 global:
+  # Send the agent to the LLM through the HAProxy edge (/llm) instead of
+  # straight to the NIM Service. Requires global.externalHost; with no host
+  # set, the in-cluster Service URL is used, so this never yields a hostless
+  # base_url. Set false to keep agent->LLM traffic off the proxy.
+  llmViaIngress: true
   # When true, all subcharts default to <release>-<chart> naming again.
   # Per-subchart <subchart>.useReleaseNamePrefix overrides this value.
   useReleaseNamePrefix: false
@@ -162,6 +167,11 @@ vssIngress:
   # these three and still wins where set.
   rtviVlmServiceName: "vss-rtvi-vlm"
   rtviVlmPort: 8000
+  # The text LLM is published at /llm, not at the origin root: /llm/v1/models,
+  # /llm/v1/chat/completions. Mounted only when this profile runs the local NIM
+  # (nims.nemotron35.enabled); a profile using a remote LLM gets no route.
+  llmServiceName: "nemotron-35-lightning-30b-a3b"
+  llmPort: 8000
   elasticsearchPort: 9200
   # Default Phoenix exposure is path-based on main host: http://<host>/phoenix
   # Keep empty for path-only mode. Set phoenixHost (for example, phoenix.<host>)
@@ -257,7 +267,7 @@ agent:
       - name: PHOENIX_ENDPOINT
         value: '{{- $g := .Values.global | default dict }}{{- $pfx := default false (coalesce .Values.useReleaseNamePrefix ((index $g "useReleaseNamePrefix") | default false)) }}{{ ternary (printf "http://%s-phoenix:6006" .Release.Name) "http://phoenix:6006" $pfx }}'
       - name: LLM_BASE_URL
-        value: '{{- $g := .Values.global | default dict }}{{- $llm := ((coalesce (.Values.llmBaseUrl | default "") ((index $g "llmBaseUrl") | default "") "") | default "" | toString | trim) }}{{- $pfx := default false (coalesce .Values.useReleaseNamePrefix ((index $g "useReleaseNamePrefix") | default false)) }}{{- $def := ternary (printf "http://%s-nemotron-35-lightning-30b-a3b:8000" .Release.Name) "http://nemotron-35-lightning-30b-a3b:8000" $pfx }}{{- if $llm }}{{ $llm }}{{- else }}{{ $def }}{{- end }}'
+        value: '{{- $g := .Values.global | default dict }}{{- $llm := ((coalesce (.Values.llmBaseUrl | default "") ((index $g "llmBaseUrl") | default "") "") | default "" | toString | trim) }}{{- $pfx := default false (coalesce .Values.useReleaseNamePrefix ((index $g "useReleaseNamePrefix") | default false)) }}{{- $def := ternary (printf "http://%s-nemotron-35-lightning-30b-a3b:8000" .Release.Name) "http://nemotron-35-lightning-30b-a3b:8000" $pfx }}{{- $edgeHost := ((index $g "externalHost") | default "") | toString | trim }}{{- $viaEdge := and (default false (index $g "llmViaIngress")) $edgeHost }}{{- if $llm }}{{ $llm }}{{- else if $viaEdge }}{{ printf "http://%s/llm" $edgeHost }}{{- else }}{{ $def }}{{- end }}'
       - name: VLM_BASE_URL
         value: '{{- $g := .Values.global | default dict }}{{- $vlm := ((coalesce (.Values.vlmBaseUrl | default "") ((index $g "vlmBaseUrl") | default "") "") | default "" | toString | trim) }}{{- $pfx := default false (coalesce .Values.useReleaseNamePrefix ((index $g "useReleaseNamePrefix") | default false)) }}{{- $def := ternary (printf "http://%s-vss-rtvi-vlm:8000" .Release.Name) "http://vss-rtvi-vlm:8000" $pfx }}{{- if $vlm }}{{ $vlm }}{{- else }}{{ $def }}{{- end }}'
       - name: VST_EXTERNAL_URL
--- a/deploy/helm/developer-profiles/dev-profile-base/vss-ingress-example-rewrites.yaml
+++ b/deploy/helm/developer-profiles/dev-profile-base/vss-ingress-example-rewrites.yaml
@@ -36,6 +36,8 @@ metadata:
       ^/storage /vst/storage
       /rtvi-vlm/(.*) /\1
       /rtvi-vlm /
+      /llm/(.*) /\1
+      /llm /
       /phoenix/(.*) /\1
       /phoenix /
 spec:
@@ -56,6 +58,13 @@ spec:
         backend:
           service:
             name: <RELEASE_NAME>-vss-rtvi-vlm
+            port:
+              number: 8000
+      - path: /llm
+        pathType: Prefix
+        backend:
+          service:
+            name: <RELEASE_NAME>-nemotron-35-lightning-30b-a3b
             port:
               number: 8000
       - path: /phoenix
--- a/deploy/helm/developer-profiles/dev-profile-lvs/values.yaml
+++ b/deploy/helm/developer-profiles/dev-profile-lvs/values.yaml
@@ -28,6 +28,11 @@ ngc:
   apiKey: ""
 
 global:
+  # Send the agent to the LLM through the HAProxy edge (/llm) instead of
+  # straight to the NIM Service. Requires global.externalHost; with no host
+  # set, the in-cluster Service URL is used, so this never yields a hostless
+  # base_url. Set false to keep agent->LLM traffic off the proxy.
+  llmViaIngress: true
   # When false (default), Service/Workload names use chart short names (redis, vss-vios-streamprocessing, vss-summarization, …). Set true for <release>-<chart> naming.
   useReleaseNamePrefix: false
   imagePullSecrets:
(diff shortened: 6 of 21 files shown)


deploy/helm/developer-profiles/dev-profile-base/values.yaml
```diff
@@ -162,6 +167,11 @@ vssIngress:
   # these three and still wins where set.
   rtviVlmServiceName: "vss-rtvi-vlm"
   rtviVlmPort: 8000
+  # The text LLM is published at /llm, not at the origin root: /llm/v1/models,
+  # /llm/v1/chat/completions. Mounted only when this profile runs the local NIM
+  # (nims.nemotron35.enabled); a profile using a remote LLM gets no route.
+  llmServiceName: "nemotron-35-lightning-30b-a3b"
```
kaushikc-nvidia: Can we use the generic llm alias `vss-llm-nim` here so that this does not have to be updated when we switch models. See: https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/develop/deploy/docker/services/nim/nemotron-3.5-lightning-30b-a3b/compose.yml#L29

deploy/helm/developer-profiles/dev-profile-alerts/vss-ingress-example-rewrites.yaml
```diff
@@ -94,6 +96,13 @@ spec:
             name: <RELEASE_NAME>-vss-rtvi-cv
             port:
               number: 9010
+      - path: /llm
+        pathType: Prefix
+        backend:
+          service:
+            name: <RELEASE_NAME>-nemotron-35-lightning-30b-a3b
```
kaushikc-nvidia: Can we use the generic llm alias `vss-llm-nim` here so that this does not have to be updated when we switch models. See: https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/develop/deploy/docker/services/nim/nemotron-3.5-lightning-30b-a3b/compose.yml#L29

deploy/helm/developer-profiles/dev-profile-alerts/values.yaml
```diff
@@ -546,6 +551,11 @@ vssIngress:
   rtviCvPort: 9010
   rtviEmbedPort: 8000
   rtviVlmPort: 8000
+  # The text LLM is published at /llm, not at the origin root: /llm/v1/models,
+  # /llm/v1/chat/completions. Mounted only when this profile runs the local NIM
+  # (nims.nemotron35.enabled); a profile using a remote LLM gets no route.
+  llmServiceName: "nemotron-35-lightning-30b-a3b"
```
kaushikc-nvidia: Can we use the generic llm alias `vss-llm-nim` here so that this does not have to be updated when we switch models. See: https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/develop/deploy/docker/services/nim/nemotron-3.5-lightning-30b-a3b/compose.yml#L29

deploy/helm/developer-profiles/dev-profile-base/vss-ingress-example-rewrites.yaml
```diff
@@ -58,6 +60,13 @@ spec:
             name: <RELEASE_NAME>-vss-rtvi-vlm
             port:
               number: 8000
+      - path: /llm
+        pathType: Prefix
+        backend:
+          service:
+            name: <RELEASE_NAME>-nemotron-35-lightning-30b-a3b
```
kaushikc-nvidia: Can we use the generic llm alias `vss-llm-nim` here so that this does not have to be updated when we switch models. See: https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/develop/deploy/docker/services/nim/nemotron-3.5-lightning-30b-a3b/compose.yml#L29

deploy/helm/developer-profiles/dev-profile-lvs/values.yaml
```diff
@@ -464,6 +469,11 @@ vssIngress:
   vstIngressPort: 30888
   lvsBackendPort: 38111
   rtviVlmPort: 8000
+  # The text LLM is published at /llm, not at the origin root: /llm/v1/models,
+  # /llm/v1/chat/completions. Mounted only when this profile runs the local NIM
+  # (nims.nemotron35.enabled); a profile using a remote LLM gets no route.
+  llmServiceName: "nemotron-35-lightning-30b-a3b"
```
kaushikc-nvidia: Can we use the generic llm alias `vss-llm-nim` here so that this does not have to be updated when we switch models. See: https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/develop/deploy/docker/services/nim/nemotron-3.5-lightning-30b-a3b/compose.yml#L29

deploy/helm/developer-profiles/dev-profile-lvs/vss-ingress-example-rewrites.yaml
```diff
@@ -69,6 +71,13 @@ spec:
             name: <RELEASE_NAME>-vss-rtvi-vlm
             port:
               number: 8000
+      - path: /llm
+        pathType: Prefix
+        backend:
+          service:
+            name: <RELEASE_NAME>-nemotron-35-lightning-30b-a3b
```
kaushikc-nvidia: Can we use the generic llm alias `vss-llm-nim` here so that this does not have to be updated when we switch models. See: https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/develop/deploy/docker/services/nim/nemotron-3.5-lightning-30b-a3b/compose.yml#L29

deploy/helm/developer-profiles/dev-profile-search/values.yaml
```diff
@@ -103,6 +108,11 @@ vssIngress:
   # RT-VLM's cluster Service port. Not 8018 — that is the Compose
   # host-published port and has no meaning in Kubernetes.
   rtviVlmPort: 8000
+  # The text LLM is published at /llm, not at the origin root: /llm/v1/models,
+  # /llm/v1/chat/completions. Mounted only when this profile runs the local NIM
+  # (nims.nemotron35.enabled); a profile using a remote LLM gets no route.
+  llmServiceName: "nemotron-35-lightning-30b-a3b"
```
kaushikc-nvidia: Can we use the generic llm alias `vss-llm-nim` here so that this does not have to be updated when we switch models. See: https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/develop/deploy/docker/services/nim/nemotron-3.5-lightning-30b-a3b/compose.yml#L29

deploy/docker/services/infra/haproxy/haproxy.cfg.template
```diff
@@ -116,6 +116,14 @@ backend bk_rtvi_vlm_strip
     http-request replace-path ^/rtvi-vlm$ /
     server s1 "${RTVI_VLM_SERVICE_HOST}:${RTVI_VLM_SERVICE_PORT}" check resolvers docker init-addr none
 
+backend bk_llm_strip
```
hugoverjus, line 119: **`develop` already has this backend, and the merge of the two is a config HAProxy refuses to load.**

`develop` carries its own `bk_llm_strip` (`haproxy.cfg.template:147` there, added by whichever PR landed the `/llm` route), so this block is a second definition of the same name. GitHub reports #2044 as `CONFLICTING`, but the conflicts it flags are only the four `vss-ingress.yaml` templates — `haproxy.cfg.template` **auto-merges cleanly** into something fatal.

What I ran:

```bash
git merge-tree --write-tree <pr-head> <develop>   # -> tree 75970c8
git show <tree>:deploy/docker/services/infra/haproxy/haproxy.cfg.template > merged/haproxy.cfg
docker run --rm -v merged:/cfg:ro -e HAPROXY_BIND_ADDR='*' -e HAPROXY_PORT=7777 \
  -e VSS_PUBLIC_HOST=... -e EXTERNAL_IP=... -e HOST_IP=... -e LLM_SERVICE_HOST=... (all template vars populated) \
  haproxy:3.4.2-alpine haproxy -c -f /cfg/haproxy.cfg
```

Output:

```
[ALERT] config : Parsing [/cfg/haproxy.cfg:155]: backend 'bk_llm_strip' has the same
        name as backend 'bk_llm_strip' declared at /cfg/haproxy.cfg:129.
        This is no longer supported as of 3.3. Please rename one or the other.
[ALERT] config : Error(s) found in configuration file : /cfg/haproxy.cfg
```

The edge would not start at all on Compose.

To be clear about what this is and isn't: **your verification is accurate against the head.** I ran the same `haproxy -c` on this branch alone with the same populated variables and got 0 errors and the 5 pre-existing warnings you report. This is fatal at merge, not a mistake in the PR as written — `develop` simply moved underneath it.

Scope of the duplication, so the rebase is one decision rather than three:

- this backend block (lines 119-126 here) vs `develop`'s at its line 147
- the frontend `acl p_llm` / `use_backend bk_llm_strip` at lines 249-251 here — the merged frontend carries them twice (merged lines 294-296 and 310-312)
- `haproxy/compose.yml:47-48` — the merged `environment:` mapping has `LLM_SERVICE_HOST` and `LLM_SERVICE_PORT` twice; I confirmed the duplicate keys with a duplicate-key-detecting YAML loader

Suggested fix: on rebase, drop the `haproxy.cfg.template` and `haproxy/compose.yml` hunks entirely. `develop` has all three, plus a `HAPROXY_SERVICE_HOST` entry in `known_host` that this branch doesn't — so taking `develop`'s version is a strict gain.

deploy/docker/services/infra/haproxy/haproxy.cfg.template
```diff
@@ -116,6 +116,14 @@ backend bk_rtvi_vlm_strip
...
+    http-request replace-path ^/llm/(.*) /\1
+    http-request replace-path ^/llm$ /
+    # A long completion outlasts the 120s default, and a 504 mid-generation
+    # reads as a backend failure. The caller bounds the wait itself.
+    timeout server 3600s
```
hugoverjus, line 124: **`develop` chose a different number here, with a stated reason.** Its `bk_llm_strip` carries `timeout server 600s`, commented "first token on a cold NIM outlasts the 120s default, and a 504 there reads to the caller as a model failure". This branch has 3600s, matching `bk_lvs_strip`.

Both rationales are reasonable and I have no view on which is right — but if the HAProxy hunks are dropped on rebase (see my comment on line 119), 3600 vs 600 gets settled by merge order rather than by anyone deciding. If 3600s is the better number, a one-line commit against `develop`'s `bk_llm_strip` makes it a reviewed change.

deploy/helm/developer-profiles/dev-profile-base/values.yaml
```diff
@@ -257,7 +267,7 @@ agent:
...
+        value: '{{- $g := .Values.global | default dict }}{{- $llm := ((coalesce (.Values.llmBaseUrl | default "") ((index $g "llmBaseUrl") | default "") "") | default "" | toString | trim) }}{{- $pfx := default false (coalesce .Values.useReleaseNamePrefix ((index $g "useReleaseNamePrefix") | default false)) }}{{- $def := ternary (printf "http://%s-nemotron-35-lightning-30b-a3b:8000" .Release.Name) "http://nemotron-35-lightning-30b-a3b:8000" $pfx }}{{- $edgeHost := ((index $g "externalHost") | default "") | toString | trim }}{{- $viaEdge := and (default false (index $g "llmViaIngress")) $edgeHost }}{{- if $llm }}{{ $llm }}{{- else if $viaEdge }}{{ printf "http://%s/llm" $edgeHost }}{{- else }}{{ $def }}{{- end }}'
```
hugoverjus, line 270: **This branch ignores `global.externalScheme` and `global.externalPort`.** Same expression in all four profiles (`dev-profile-alerts/values.yaml:335`, `dev-profile-lvs/values.yaml:295`, `dev-profile-search/values.yaml:814`).

Every other externalHost-derived URL in this same file builds `$es://$eh:$ep` from `global.externalScheme` and `global.externalPort` — see `VST_EXTERNAL_URL` at line 274 and `VSS_AGENT_EXTERNAL_URL` at line 288. This branch builds `printf "http://%s/llm" $edgeHost`.

`helm template rel . --set vssIngress.enabled=true` on `dev-profile-base`, reading the rendered `vss-agent` env:

| values | `LLM_BASE_URL` | `VST_EXTERNAL_URL` (sibling, same file) |
|---|---|---|
| `externalHost=10.0.0.1` | `http://10.0.0.1/llm` | `http://10.0.0.1` |
| `+ externalPort=30080` | `http://10.0.0.1/llm` | `http://10.0.0.1:30080` |
| `+ externalScheme=https, externalPort=8443` | `http://10.0.0.1/llm` | `https://10.0.0.1:8443` |

So on any non-default-port or TLS install the agent sends every LLM request to plaintext port 80. `values-base.yaml:35` documents `externalPort` as exactly that: `Non-default port only: e.g. "8080", "8443"`.

Custom host is the same failure by a different route: `--set vssIngress.host=vss.example.com` renders the Ingress with host `vss.example.com` while `LLM_BASE_URL` stays `http://10.0.0.1/llm`, matching no host rule. `README.md` documents that combination, including in the TLS example.

Suggested fix: reuse the `$globalBase` expression already three lines below in this file. Better still, given `develop` just introduced `vss.ingress.llmBackend` in `services/common` to resolve the LLM Service once for all four profiles, factor the external origin out the same way — a `vss.externalBaseUrl` helper — so the four copies can't drift, and fall back to the in-cluster Service when the derived origin matches no rendered host rule.

deploy/helm/developer-profiles/dev-profile-search/values.yaml
```diff
@@ -801,7 +811,7 @@ agent:
...
+        value: '{{- $g := .Values.global | default dict }}{{- $llm := default "" (coalesce (.Values.llmBaseUrl | default "") ((index $g "llmBaseUrl") | default "")) | toString | trim }}{{- $pfx := default false (coalesce .Values.useReleaseNamePrefix ((index $g "useReleaseNamePrefix") | default false)) }}{{- $svc := trim (.Values.llmServiceName | default "nemotron-35-lightning-30b-a3b") }}{{- $def := ternary (printf "http://%s-%s:8000" .Release.Name $svc) (printf "http://%s:8000" $svc) $pfx }}{{- $edgeHost := ((index $g "externalHost") | default "") | toString | trim }}{{- $viaEdge := and (default false (index $g "llmViaIngress")) $edgeHost }}{{- if $llm }}{{ $llm }}{{- else if $viaEdge }}{{ printf "http://%s/llm" $edgeHost }}{{- else }}{{ $def }}{{- end }}'
```
hugoverjus, line 814: **This profile is broken by the repoint in the configuration its own docs prescribe.**

`dev-profile-search` does not use `externalHost` as its main ingress host — `templates/vss-ingress.yaml:50` builds `vss-search.<ip>.nip.io` from it. And `Chart.yaml:36,43,50` plus `README.md:473,494` both document installing with `--set global.externalHost=$NODE_EXTERNAL_IP`, i.e. a bare IP. `values.yaml:22` says so too: "Bare node IP, or full main hostname".

`helm template rel . --set global.externalHost=10.0.0.1 --set vssIngress.enabled=true`:

```
LLM_BASE_URL     = http://10.0.0.1/llm
VST_EXTERNAL_URL = http://vss-search.10.0.0.1.nip.io     <- sibling gets it right
Ingress hosts    = vss-search.10.0.0.1.nip.io, streamer.10.0.0.1.nip.io, kibana.10.0.0.1.nip.io
```

Nothing serves `http://10.0.0.1/llm`, so every agent LLM call 404s at the ingress — the same silent-apology shape as the `/v1` bug this PR fixes. `base`, `alerts` and `lvs` use `externalHost` as the main host verbatim and do render a matching origin, so this one is specific to `search`. The verification table in the description covers `externalHost` already being the full main host, which is why it didn't surface.

Suggested fix: derive the origin from the *rendered* ingress host rather than from `externalHost` — this file already computes `$mainHost` / `$nipBase` for `VST_EXTERNAL_URL` at line 821 — and fall back to the in-cluster Service URL whenever the derived origin matches no host rule, so a misconfiguration degrades to the working direct path instead of a 404.

hugoverjus (commented):
Reviewed at `fcb80b2`, in a worktree off the PR head with `helm dependency build` re-run in all four profiles before anything was rendered. Everything below is either something I ran or something I reasoned about, and I've said which.

## What this gets right

**The `RTVI_VLM_BASE_URL` default is a real fix, and `develop` still needs it** — `services/agent/compose.yml:110` there is still a bare `${RTVI_VLM_BASE_URL}`.

I traced the root cause rather than taking it on trust, and it holds. `deploy/docker/scripts/blueprint-deploy.sh:1334-1336` (and `:1311-1313` for `config --images`) passes exactly three interpolation sources — `--env-file containers.env`, `--env-file .env`, `--env-file generated.env`. `services/rtvi/rtvi.env` *does* define `RTVI_VLM_BASE_URL=http://rtvi-vlm:8000`, but it is attached as an `env_file:` on the RTVI containers only, so it lands inside those containers and never reaches Compose's variable-interpolation environment for `vss-agent`. `${RTVI_VLM_BASE_URL}` therefore resolves empty, with no unset-variable warning because the reference has no default — exactly as you describe. And `http://rtvi-vlm:8000` is the value `services/alert/compose.yml:52`, `services/video-summarization/compose.yml:64` and `rtvi.env` itself already use, so this brings the agent in line rather than inventing anything.

## The main thing

**The `/llm` publication has already landed on `develop`** — the shared route table, the HAProxy backend and frontend rule, and the three `vss-ingress-example-rewrites.yaml` files. `develop`'s version resolves the LLM Service through a new shared `vss.ingress.llmBackend` helper in `services/common` instead of repeating the resolution in each profile template, which is why all four `vss-ingress.yaml` files conflict. My recommendation is to rebase and keep the two still-novel pieces: the `RTVI_VLM_BASE_URL` default, and the CLI `INGRESS_SERVICES` entry (`develop` publishes `/llm` on both edges but never records it, so `vss configure` can't see it).

The blocker is in the inline comment on `haproxy.cfg.template:119`: `haproxy.cfg.template` auto-merges *without conflict* into a config that HAProxy 3.3+ rejects outright for a duplicate backend name. I proved it with `haproxy -c` on the computed merge tree. Worth stressing that this is fatal at merge, not a mistake in the PR as written — I ran the same check on this branch alone and got your result exactly: 0 errors, 5 pre-existing warnings.

Two rendered findings are in inline comments on the `LLM_BASE_URL` lines in `dev-profile-base` and `dev-profile-search`: the edge URL ignores `global.externalScheme` and `global.externalPort`, and `dev-profile-search` breaks outright in the bare-IP install its own `Chart.yaml` and `README.md` prescribe.

## Reasoned about, not executed

**`global.llmViaIngress: true` as the default.** Two things I'd want a second opinion on rather than assert. Operationally, it makes the agent pod dial its own cluster's external address and resolve `nip.io` from inside the pod — that works on some CNI and kube-proxy configurations and not others, and it puts the ingress controller in the critical path of every LLM call. `llmViaIngress=false` is a clean kill switch, but the default is on, so the failure mode is the default one. Architecturally, a per-profile flag plus a hand-assembled `http://<host>/llm` is a second source of truth for the origin that #1983 is centralising — `develop`'s own `bk_llm_strip` comment already writes it as `$VSS_GATEWAY_ORIGIN/llm`, though no such variable exists yet. Worth a word with the #1983 author before this lands, since it's the piece that would have to be unwound later. I had no cluster; one deployment with the flag left on, checking the agent completes a chat turn, would settle the operational half.

**The `vss configure` warmup window.** `/llm/v1/models` answers 503 while the NIM loads (backend DOWN at the edge), and 503 isn't in `_PRESENT_STATUSES`, so `llm` records as *absent* on a deployment that is merely still starting. Explicitly **not** a defect introduced here: `rt_vlm` behaves identically today, so this is consistent with its sibling. Noting it because the `lvs` route already solved the same problem by probing `/lvs/v1/live` instead of readiness, with a comment saying why — that's the precedent if it's ever worth fixing for the OpenAI-shaped routes too.

## One note on the verification section

`verify-ingress-routes.py` passes on this branch — I re-ran it after `helm dependency build`, and got your result: `OK 4 profiles render from one table (23 distinct mounts); CLI probe mounts present; 6 disabled-component cases drop their route; examples current`. I then negative-tested two of its claims:

- Deleting the `- key: llm` row from `_ingress-routes.tpl` → **fails** loudly, from the chart's own `assertBackends`: `"llm" is not a canonical route key`. Real assertion.
- Reverting `dev-profile-base/vss-ingress-example-rewrites.yaml` so it omits the `/llm` mount → still `OK … examples current`. Check 7 does **not** catch an example file missing a newly added mount.

So "examples current" isn't evidence for the example-file edits here. The three files are in fact correct — but for the unrelated reason that `develop` independently added the same rows. That's a gap in the verifier rather than anything for this PR to fix; flagging it so the green isn't over-read.

services/agent/packages/vss_cli/src/vss_cli/config.py
```diff
@@ -762,6 +762,16 @@ def save(deployment: Deployment) -> Path:
             describe="/rtvi-vlm/v1/models",
             describes="models",
         ),
+        # The text LLM behind the agent. OpenAI-shaped like RT-VLM and
+        # RT-Embed, so /v1/models is both the proof it is routed and the
+        # description of what it serves. Absent on profiles pointed at a
+        # remote LLM, which run no local NIM to mount.
+        "llm": ServiceRoute(
```
hugoverjus, line 769: **Reviewability, not correctness — but worth a note.** This package moved to `libs/vss/cli/src/vss_cli/config.py` on `develop`; this path no longer exists there. I checked rather than assumed: git's rename detection carries the hunk across, and the merged `libs/vss/cli/src/vss_cli/config.py` contains this `ServiceRoute` correctly at the right place in `INGRESS_SERVICES`. So nothing breaks — the diff just reads against a deleted file, which costs a reviewer a detour.

Keep this piece. It is the one part of the `/llm` work `develop` does **not** have: `develop` publishes the route on both edges but never records it in `INGRESS_SERVICES`, so `vss configure` can't see it. Re-pointing the hunk at the new path on rebase is all it needs.

hugoverjus:
Taken — #2197, rebased onto current `develop`. **Fine to close this one.**

It carries the two pieces `develop` genuinely still lacks, unchanged in substance from yours:

- the `RTVI_VLM_BASE_URL` default in `deploy/docker/services/agent/compose.yml`, with your comment explaining the relative-`/v1` failure
- the `/llm` entry in `INGRESS_SERVICES`, repointed at the new `libs/vss/cli/src/vss_cli/config.py`

I checked each against `develop` rather than assuming: `compose.yml:110` is still a bare `${RTVI_VLM_BASE_URL}`, the `${RTVI_VLM_BASE_URL}/v1` consumers are still live in the `base`, `search` and `lvs` agent configs (both Docker and Helm copies), and `INGRESS_SERVICES` still has no `llm` key, so `vss configure` could not see the route `develop` already publishes on both edges.

What I did not carry over, and why:

- **HAProxy backend, frontend ACL and `LLM_SERVICE_HOST`/`PORT`** — `develop` has all three, plus a `HAPROXY_SERVICE_HOST` entry in `known_host` this branch doesn't, so taking `develop`'s is a strict gain. This is the duplicate-backend problem from my review comment: the two auto-merge without conflict into a config HAProxy 3.3+ refuses to load.
- **`global.llmViaIngress` and the `LLM_BASE_URL` repoint** — `develop` leaves `LLM_BASE_URL` on the in-cluster Service, so there is nothing to rebase onto. If repointing through the edge is still wanted it's worth its own PR, reusing a shared origin helper rather than four hand-assembled copies, which also avoids the `externalScheme`/`externalPort` and `dev-profile-search` findings from my review.
- **`vss-ingress.yaml` and the example-rewrites files** — already on `develop` via the shared `vss.ingress.llmBackend` helper.

One open question I deliberately left alone: `develop`'s `bk_llm_strip` has `timeout server 600s`, yours had 3600s. Both rationales read fine and I have no view. If 3600s is right it deserves a one-line reviewed commit against `develop` rather than being decided by merge order.

Verification on #2197: `verify-ingress-routes.py` OK (and I negative-tested check 4 — breaking the CLI mount does make it fail, so the green means something here), 443 CLI tests pass, and `vss configure` against a stand-in edge records `llm` with its model list. Details in the description.

Thanks for tracing the root cause — the `env_file:`-versus-interpolation distinction was the non-obvious part and your write-up held up when I retraced it.

PR #2046, opened 2026-09-04
fix(rtvi-vlm): Release callbacks before pipeline replacement

nv-amkale:
## Summary

Release GStreamer callback registrations and the bus signal watch before
replacing an RT-VLM file-decoder pipeline.

## Plan

- Reproduce resource growth across repeated codec transitions.
- Clean up callbacks while the old pipeline and bus are still referenced.
- Validate the fix with the RTVI unit suite and the recorded reproduction.

## Related Issue

NVBug 6697886

## Changes

- Disconnect tracked pad probes and signal handlers during pipeline
  replacement.
- Remove the old bus signal watch before clearing the bus reference.
- Preserve reusable decoders and allow callbacks to be registered again.
- Add focused coverage for callback and bus-watch cleanup ordering.

## Validation

- Independent Compose validation passed eight alternating H.264/H.265
  transitions with stable FD, socket, and handler counts.
- The configured RTVI unit suite introduced no new failures against the exact
  baseline SHA.
- The focused pipeline-replacement cleanup test passed.
- The authoritative fixed-phase Compose reproduction passed.

### DIFF
--- a/services/rtvi/rt-vlm/src/vlm_pipeline/video_file_frame_getter.py
+++ b/services/rtvi/rt-vlm/src/vlm_pipeline/video_file_frame_getter.py
@@ -977,6 +977,25 @@ class VideoFileFrameGetter:
     def _cache_decodebin(self, key, decodebin) -> None:
         self._vdecodebin_cache[key] = decodebin
 
+    def _detach_pipeline_for_replacement(self):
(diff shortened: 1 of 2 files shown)


nv-pnikam (approved):
LGTM

PR #1952, opened 2026-09-01
fix(agent): route RTVI-CV ingest through the internal VST endpoint

ayyappa-dev:
## Summary

 Video-file ingest handed RTVI-CV the raw external VST URL while RTVI-Embed rewrote it to internal. When the external VST ingress was unreachable from in-cluster,
 RTVI-CV got no source → no behavior docs → attribute_search 404'd ("0 results"). Translate the VST URL to internal once; pass the same internal URL to both
 RTVI-CV and RTVI-Embed.

## Plan

 - Translate vst_file_path → internal_video_url once in _run_post_upload_processing; pass to both _register_with_rtvi_cv and _run_rtvi_embedding.
 - Drop the now-redundant rewrite + vst_url param from _run_rtvi_embedding.
 - pytest test_video_ingest.py → 82 passed; new test fails on pre-fix code (stash-verified).

## Related Issue

 VIA-2744

## Changes

 - video_ingest.py: translate to internal_video_url once in _run_post_upload_processing; pass to both RTVI-CV and RTVI-Embed.
 - video_ingest.py: drop vst_url param + internal rewrite_url_host from _run_rtvi_embedding.
 - test_video_ingest.py: add test_rtvi_cv_and_embed_receive_internal_vst_url asserting both consumers get the internal URL.
 - Scope: nvidia-vss-agents only; RTSP path and vss-core/cli unaffected.
 
## Checklist
- [x] I am familiar with the [Contributing Guidelines](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/HEAD/CONTRIBUTING.md).
- [x] I have installed and run [pre-commit hooks](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/HEAD/CONTRIBUTING.md#1-enable-pre-commit-hooks) locally (`uv run pre-commit install` once, then hooks run on every `git commit`).
- [x] Every commit on this PR is [DCO sign-off](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/HEAD/CONTRIBUTING.md#developer-certificate-of-origin-dco)'d (`git commit -s` adds a `Signed-off-by` trailer that certifies you have the right to submit the change under Apache-2.0).
- [x] New or existing tests cover these changes.
- [x] The documentation is up to date with these changes.
- [x] No secrets, API keys, or credentials committed.

PR #2049, opened 2026-09-04
fix(agent): default the LLM endpoint and fail fast when it is empty

hugoverjus:
## The nightly symptom

Nightly integration and UI e2e (2026-09-03, `nightly-develop`) failed broadly because the VSS Agent could not answer any prompt. Every agent-eval prompt — "What sensors are available?", "List the last 3 incidents for sensor id \"Camera_01\"", "Generate a report for the last incident of sensor id \"Camera_01\"" — failed identically, surfacing in the UI as:

```
Error: /v1/chat/completions
```

followed by the canned "Sorry, I wasn't able to complete your request…". Report-generation tests then failed on a missing output link, because no report was ever produced.

## Root cause: an empty `LLM_BASE_URL`, not a misplaced LLM

`Error: /v1/chat/completions` is the signature of an **empty** endpoint variable, not an unreachable one — an unreachable host produces a connection error that names the host. The chain:

1. Every profile config writes `base_url: ${LLM_BASE_URL}/v1` with no `:-` fallback (e.g. `deploy/docker/developer-profiles/dev-profile-base/vss-agent/configs/config.yml:167`).
2. `deploy/docker/services/agent/compose.yml:107` on `develop` is `LLM_BASE_URL: ${LLM_BASE_URL}` — the **only** bare passthrough in that environment block. Its eleven neighbours all carry inline defaults, e.g. `RTVI_CV_ENDPOINT: ${RTVI_CV_ENDPOINT:-http://vss-rtvi-cv:${RTVI_CV_PORT:-9000}}` at `:90`.
3. nvidia-nat's YAML loader substitutes an unset or empty variable with the empty string instead of failing, so `base_url` becomes the literal `/v1`.
4. `NIMModelConfig.base_url` is `str | None` with no validation, so Pydantic accepts it. `langchain_nvidia_ai_endpoints._common._validate_base_url` only **warns** on a URL with no scheme and no netloc. The client's `https://integrate.api.nvidia.com/v1` default applies only when `base_url` is *omitted*; setting it to `/v1` defeats it.
5. The agent starts, **passes its health check**, and fails on the first token with `InvalidUrlClientError("/v1/chat/completions")`, which `services/agent/packages/vss_agents/src/vss_agents/agents/top_agent.py` renders as `f"Error: {ex}"` before the canned apology.

That is why all three prompts failed identically: the failure is at the agent's single reasoning LLM, before any tool selection.

**How it became empty.** Docker Compose ranks the process environment above `--env-file`, and an exported-but-empty variable still wins. Reproduced directly:

```
# env-file value only, variable unset in the shell
LLM_BASE_URL: http://vss-llm-nim:8000
# same env file, LLM_BASE_URL="" exported
LLM_BASE_URL: ""
```

A profile that never sets the variable produces the same result. `industry-profiles/smartcities` sets none of `LLM_MODE`, `LLM_NAME`, `LLM_NAME_SLUG` or `LLM_BASE_URL` in either `.env` or `overrides.env`, yet its `COMPOSE_PROFILES_CV` includes `vss-agent` and `overrides.env:46` derives `EVAL_LLM_JUDGE_BASE_URL=${LLM_BASE_URL}` from it; it relies on an out-of-repo launchable merging it over `dev-profile-alerts/overrides.env`.

**Platform discriminator.** Compose deployments are affected; Helm is structurally immune. The chart always coalesces to an in-cluster NIM default (`deploy/helm/developer-profiles/dev-profile-base/values.yaml:260`), so Helm can render a *wrong* URL but never an empty one, and therefore never this error string. That is what makes the nightly matrix uneven. The H100 gap in that run was a lost host lease and is not evidence about this bug.

## What the three commits do

Ordered smallest-and-most-defensible first, so the hotfix can be taken alone.

**1. `fix(deploy): default the agent LLM endpoint in compose`** — two lines. Gives `LLM_BASE_URL` an inline default in the same style as its eleven neighbours, plus `LLM_MODE` at `:68` for the same reason. Because `${VAR:-default}` treats an empty value as unset, this one line closes both failure modes: an exported-but-empty variable in the deploying shell, and a profile that never sets one. This is the nightly hotfix.

**2. `fix(agent): refuse to serve when the workflow LLM has no endpoint`** — `register_vss_fastapi_front_end` is the one VSS-owned hook that sees the fully env-substituted `Config`, exactly once, at `nat serve` startup. It now refuses to start when the entry named by `workflow.llm_name` has no scheme and host, naming the entry, the value it resolved to, and the endpoint shapes that are valid. Two carve-outs that matter:

- an entry that **omits** `base_url` is left alone — `openai_vlm` does that deliberately so the client uses its own endpoint;
- every other `llms:` entry only **warns**. `llms.rtvi_vlm.base_url` already resolves to `/v1` on base, search, lvs and warehouse as they ship, because `RTVI_VLM_BASE_URL` is defined only in `deploy/docker/services/rtvi/rtvi.env`, an `env_file:` for the RT-VLM containers that is not in the agent's interpolation chain. A blanket rule would refuse to boot every profile shipping today.

**3. `fix(orchestrator): drop empty endpoint vars from the compose env`** — defence in depth for the programmatic deploy path. `docker_compose_util.py` already strips `LLM_MODE`/`VLM_MODE` from the inherited environment unconditionally, which is this same bug class fixed narrowly once before. This adds a separate **empty-only** guard (`_COMPOSE_SHELL_ENV_DROP_IF_EMPTY`) rather than extending that blocklist, because a *non-empty* shell `LLM_BASE_URL` is a legitimate override — it is how a remote endpoint is passed in, and callers outside this repository may rely on it.

## The remote/external LLM case is preserved

`LLM_BASE_URL` stays authoritative everywhere. An explicitly configured endpoint wins over the new default (verified: `LLM_BASE_URL=https://integrate.api.nvidia.com` renders unchanged), the fail-fast accepts any absolute `http(s)` origin, and the orchestrator guard drops only empty and whitespace-only values. Deployments pointing at build.nvidia.com or a hosted NIM are unaffected.

`http://vss-llm-nim:8000` is the default because it is the value all five committed profiles already use, and it is a network alias every LLM NIM compose service registers — dedicated and `-shared-gpu` alike (`deploy/docker/services/nim/nemotron-3.5-lightning-30b-a3b/compose.yml:28-29,72-74`) — so it encodes no GPU placement.

## This changes *how* a misconfigured deployment fails

Worth flagging for whoever owns the nightly: today the agent comes up, passes `/health`, and fails on every prompt. After commit 2 it refuses to start. That is earlier and louder, but it is a different failure shape — a harness that waits on `/health` will now see a container that never becomes healthy rather than one that answers with an error string.

## Tested

- `uv run pytest packages/vss_agents/tests/unit_test/utils/test_llm_endpoint.py packages/vss_agents/tests/unit_test/orchestrator/test_docker_compose_util.py packages/vss_agents/tests/unit_test/api -q` → **303 passed**. New cases cover, for the fail-fast: in-cluster / gateway-style / hosted / omitted `base_url` / `/v1` / whitespace-only / scheme-less, plus warn-only for a non-workflow entry and no warning on a healthy config. For the orchestrator guard: empty and whitespace-only dropped, non-empty honoured, and the existing mode blocklist still unconditional.
- `uv run ruff check packages/vss_agents/src packages/vss_agents/tests` → clean; `ruff format --check` → clean; `uv run mypy packages/vss_agents/src/vss_agents/` → no issues in 132 files.
- `docker compose config` used to reproduce the shadowing and to verify the new defaults across all three cases (profile value honoured, empty shell value defaulted, explicit remote value preserved).

## Not tested

No deployment was stood up: no NIM images pulled, no GPU workload, no end-to-end agent prompt. The behaviour of the compose defaults was verified by rendering configuration, not by running the stack.

## Related but deliberately not included

- **The `/llm` gateway route** is a separate PR stacked on #1983. It addresses topology independence, not emptiness, and would not have prevented this nightly.
- **`unset` guards in `dev-profile.sh` / `blueprint-deploy.sh`** — commit 1 makes them redundant for this variable, and both scripts are already careful (`dev-profile.sh:1713` and `blueprint-deploy.sh:1199` both check `if [[ -n "${llm_base_url}" ]]` before writing, so neither can stamp an empty value). Available as a follow-up if wanted.
- **`smartcities`** setting no LLM variables at all is a latent second exposure that commit 1 makes harmless; restructuring that overlay needs the out-of-repo launchable and is not attempted here.
- **No CI guard covers the LLM endpoint** (`git grep LLM_BASE_URL -- .github` yields one docstring hit). A `check_*.py`-style guard would fit the existing idiom but would bloat this PR.
- **Two Helm mismatches** found while ruling Helm out, left alone: `deploy/helm/developer-profiles/dev-profile-search/override-values.yaml:32` sets a top-level `vss-agent:` key the chart reads at `agent.vss-agent:` and therefore ignores, and `apiKeys.nvidia` in the agent chart values is referenced by no template.
- **The search upload failures in the same nightly are a different bug.** They look like `04d4b25f9` "fix(agent): route RTVI-CV ingest through the internal VST endpoint (#1952)", which landed 2026-09-03T16:06Z — hours before the run — and added a new hard `HTTPException` on `vst_url` in the upload path, and/or the unmerged #2028 rt-embed `409 DuplicateAssetId` race. Nothing in this PR touches either.

### DIFF
--- a/deploy/docker/services/agent/compose.yml
+++ b/deploy/docker/services/agent/compose.yml
@@ -65,7 +65,10 @@ services:
     ports:
     - ${VSS_AGENT_HOST_PORT:-8000}:${VSS_AGENT_PORT:-8000}
     environment:
-      LLM_MODE: ${LLM_MODE} # remote / local / local_shared
+      # Defaulted like the endpoint variables below, so an unset or empty value
+      # from the deploying shell cannot reach the container as a blank. Every
+      # committed profile except warehouse (local) sets local_shared.
+      LLM_MODE: ${LLM_MODE:-local_shared} # remote / local / local_shared
       VLM_MODE: ${VLM_MODE} # remote / local / local_shared
       LLM_MODEL_TYPE: ${LLM_MODEL_TYPE} # nim / openai
       VLM_MODEL_TYPE: ${VLM_MODEL_TYPE} # nim / openai
@@ -104,7 +107,16 @@ services:
 
       # LLM Endpoints
       LLM_NAME: ${LLM_NAME}
-      LLM_BASE_URL: ${LLM_BASE_URL}
+      # The one endpoint in this block that carried no default. Compose ranks the
+      # process env above --env-file and an exported-but-empty variable still
+      # wins, and `${VAR:-default}` treats empty as unset, so this covers both an
+      # empty value from the deploying shell and a profile that never sets one.
+      # It matters here more than elsewhere: the agent config writes
+      # `base_url: ${LLM_BASE_URL}/v1` with no fallback of its own, so a blank
+      # becomes the literal `/v1` and every completion is posted to a bare
+      # `/v1/chat/completions`. An explicit value still wins, which is how
+      # LLM_MODE=remote reaches a hosted endpoint.
+      LLM_BASE_URL: ${LLM_BASE_URL:-http://vss-llm-nim:8000}
       VLM_NAME: ${VLM_NAME}
       VLM_BASE_URL: ${VLM_BASE_URL}
       RTVI_VLM_BASE_URL: ${RTVI_VLM_BASE_URL}
--- a/services/agent/packages/vss_agents/src/vss_agents/api/front_end_config.py
+++ b/services/agent/packages/vss_agents/src/vss_agents/api/front_end_config.py
@@ -27,6 +27,8 @@ from pydantic import BaseModel
 from pydantic import ConfigDict
 from pydantic import Field
 
+from vss_agents.utils.llm_endpoint import validate_chat_llm_endpoints
+
 
 class StreamingIngestConfig(BaseModel):
     """Configuration for the streaming video ingest and RTSP stream endpoints.
@@ -139,4 +141,8 @@ class VSSFastApiFrontEndConfig(FastApiFrontEndConfig, name="vss_fastapi"):  # ty
 async def register_vss_fastapi_front_end(
     _config: VSSFastApiFrontEndConfig, full_config: Config
 ) -> AsyncGenerator[FastApiFrontEndPlugin]:
+    # First point at which the interpolated config is known, and still before the
+    # server binds: an LLM endpoint that no request can reach must stop the agent
+    # here rather than surface later as a stream error naming only the path.
+    validate_chat_llm_endpoints(full_config)
     yield FastApiFrontEndPlugin(full_config=full_config)
--- a/services/agent/packages/vss_agents/src/vss_agents/orchestrator/docker_compose_util.py
+++ b/services/agent/packages/vss_agents/src/vss_agents/orchestrator/docker_compose_util.py
@@ -96,6 +96,14 @@ COMPOSE_PROFILE_REQUIRED_KEYS: Final[tuple[str, ...]] = (
     "VLM_NAME_SLUG",
 )
 _COMPOSE_SHELL_ENV_BLOCKLIST: Final[frozenset[str]] = frozenset({"LLM_MODE", "VLM_MODE"})
+# Dropped only when they arrive empty. Compose ranks the process env above
+# ``--env-file``, and an exported-but-empty variable still wins, so an empty
+# ``LLM_BASE_URL`` in the calling shell blanks the profile's endpoint and
+# everything derived from it -- which is what turns the agent's LLM into a bare
+# ``/v1``. A *non-empty* value is a legitimate override (that is how a remote
+# endpoint is passed in), so it is left alone rather than added to the blocklist
+# above.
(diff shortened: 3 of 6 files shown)


hugoverjus:
Superseded by #2050, which carries the same one-line `LLM_BASE_URL` default without the additional scope. Consolidating on a single minimal PR for the nightly failure.

The other changes here (agent startup fail-fast on a scheme-less LLM endpoint, empty-only orchestrator compose-env guard, and the `LLM_MODE` default) are not lost — they can come back as their own PR once we've confirmed with the CI owners whether the pipeline is meant to supply `LLM_BASE_URL`. That answer decides whether the right behaviour is to default the value or to refuse to start, and those two pull in opposite directions.

PR #2050, opened 2026-09-04
fix: fail fast on an unusable LLM endpoint, order the vios ingress behind its upstreams, and link reports by URL value

hugoverjus:
Three independent defects found while tracing why the nightly's report-generation
check reports "no report link". Each is a separate commit and each stands alone —
the branch is mergeable at every commit.

None of the three is *the* cause of that nightly failure. That cause is
environmental and is written up at the bottom, because the investigation is the
only reason these were found and it should not be lost.

## 1. `da027bd` — refuse to serve when the workflow LLM endpoint is unusable

`LLM_BASE_URL` is interpolated into the workflow config as `${LLM_BASE_URL}/v1`.
When the variable is empty or unset that resolves to the string `/v1` — no
scheme, no host. The agent started anyway, passed its health check, reported
ready, and then failed on the first prompt with a connection error naming a bare
`/v1/chat/completions`. The deployment looks healthy and the failure surfaces one
layer away from its cause.

This validates the workflow's LLM endpoints at startup and refuses to serve if one
cannot be used, with a diagnostic that names the variable, what it resolved to,
why that cannot work, and what a correct value looks like.

**Proven on a live GPU host** against the real deployed profile config, varying
only `LLM_BASE_URL` across four otherwise-identical containers:

| Case | Container | Bound and serving |
|---|---|---|
| `develop` as shipped, `LLM_BASE_URL=''` | running | **yes** — the bug |
| this branch, `LLM_BASE_URL=''` | exit 1 | no |
| this branch, `LLM_BASE_URL` unset | exit 1 | no |
| this branch, `LLM_BASE_URL='http://vss-llm-nim:8000'` | running | yes |

The pre-fix baseline starting and serving with no usable LLM is the defect; cases
2 and 3 are the fix; case 4 shows nothing else regressed.

## 2. `9dad18e` — wait for nginx ingress upstreams before starting

nginx resolves every `proxy_pass` host at config-parse time and exits
`[emerg] host not found in upstream` if one is missing. `vst-ingress` declared
`depends_on` for the sensor services but not for `streamprocessing-ms*` or
`sdr-controller`, which `nginx-${VST_NGINX_MODE}.conf` also names. On a cold start
where those lose the race, the ingress dies instead of waiting.

Adds the missing `depends_on` edges with `condition: service_started` and
`required: false`, matching the pattern already used for the sensor services.
`service_started` is deliberate: nginx needs the DNS name to exist, not the
upstream to be serving, and `service_healthy` would hold the ingress down behind
`streamprocessing`'s deliberately wide (~25 min) healthcheck window.

Chosen over adding an nginx `resolver` with variable `proxy_pass`, which would
have touched 19 `proxy_pass` sites and moved hostname resolution into the request
path for all of them. Ordering is the smaller change and the one with no
data-path consequence.

**Verified live**: without the fix, nginx exits `host not found in upstream`;
with it, the ingress comes up and stays up. `docker compose config` accepts it,
and a negative control with a deliberately misspelled service name is rejected,
confirming the edges are really being resolved.

## 3. `b53f257` — link reports by URL value, not by attribute presence

`TemplateReportGenOutput` declares `http_url` and `pdf_url` as **required**
fields, so `hasattr(report_result, "http_url")` is always true. The download
section was gated on exactly that check, so when the object store returned empty
strings for both URLs the agent emitted

```
**Report Downloads:**
```

— a heading with no anchors under it. To a reader, and to any assertion looking
for a report link, that is indistinguishable from report generation having
silently failed. **It is the same observable as the nightly failure, from an
unrelated cause**, which is worth fixing on its own so the two can never be
confused again.

Truth-tests the values instead. A report with at least one artifact links exactly
those that exist and never emits an empty `()` anchor; a report with nothing to
download says so and logs a warning naming the incident.

Media is deliberately treated differently: an incident need not have a snapshot or
a clip, so an absent media block stays omitted rather than announced. Only a
report with no downloadable copy at all is anomalous.

The block moves to a module-level `_build_report_side_effects` helper so it can be
tested — it previously sat inside a closure nested in the NAT registration
function and was unreachable from a unit test. Tests cover both-present,
markdown-only, pdf-only, both-empty, and media present/absent, plus a guard
asserting the required-field premise against the real output model so a future
optionality change fails loudly here.

## What this PR does *not* fix

**The nightly's report-generation failure itself.** That is environmental, not a
code defect in this repo. The shared LLM endpoint the nightly points at rejects
tool-bound requests, so the report agent is never reached at all.

Measured on a live host, same model, identical tool-bound payload:

- a locally started NIM with `--enable-auto-tool-choice` and a `--tool-call-parser`
  returns **HTTP 200** with a populated `tool_calls`
- the shared endpoint the nightly uses returns **HTTP 400**

The fix belongs to whoever owns that endpoint: it needs to be started with tool
calling enabled, or the nightly needs to point at one that is. Commit `da027bd`
makes a *missing* endpoint fail loudly at startup; it deliberately does not try
to detect a *tool-incapable* one, which would mean issuing a probe tool call on
every boot.

Two further findings from the same investigation — how a sensor can be served by
VIOS but never subscribed by rt-cv, and an ungrounded answer path in the top
agent's planner — are written up in a comment on this PR rather than changed
here, because neither has a fix that is safe to improvise.

### DIFF
--- a/deploy/docker/services/vios/foundational/docker-compose.yaml
+++ b/deploy/docker/services/vios/foundational/docker-compose.yaml
@@ -61,6 +61,13 @@ services:
       timeout: 3s
       retries: 30
       start_period: 2s
+    # Nginx resolves every proxy_pass host at config-parse time and exits with
+    # [emerg] "host not found in upstream" if one is missing, so every host named
+    # by nginx-${VST_NGINX_MODE}.conf must exist first: vss-vios-sensor and
+    # vss-vios-streamprocessing (vst), sdr-controller (vst-sdrc, mms).
+    # service_started suffices -- nginx needs the DNS alias, not a serving
+    # upstream -- and streamprocessing's healthcheck is widened to ~25min to
+    # outlast slow apt mirrors, which service_healthy would hold this edge down for.
     depends_on:
       sensor-ms:
         condition: service_healthy
@@ -70,6 +77,24 @@ services:
         required: false
       sensor-ms-3d:
         condition: service_healthy
+        required: false
+      sensor-ms-mv3dt:
+        condition: service_healthy
+        required: false
+      streamprocessing-ms:
+        condition: service_started
+        required: false
+      streamprocessing-ms-2d:
+        condition: service_started
+        required: false
+      streamprocessing-ms-3d:
+        condition: service_started
+        required: false
+      streamprocessing-ms-mv3dt:
+        condition: service_started
+        required: false
+      sdr-controller:
+        condition: service_started
         required: false
 
 volumes:
--- a/services/agent/packages/vss_agents/src/vss_agents/agents/report_agent.py
+++ b/services/agent/packages/vss_agents/src/vss_agents/agents/report_agent.py
@@ -61,6 +61,59 @@ def _append_artifact_display_note(side_effects: dict[str, Any]) -> None:
         side_effects["artifact_note"] = _ARTIFACT_DISPLAY_NOTE
 
 
+def _build_report_side_effects(report_result: Any, incident_id: str) -> dict[str, Any]:
+    """Build the download and media side effects for a generated incident report.
+
+    ``TemplateReportGenOutput`` declares ``http_url`` and ``pdf_url`` as required
+    fields, so ``hasattr`` is always true for them: only the values say whether the
+    object store produced anything linkable. Gating the section on the attribute
+    instead emitted a bare ``**Report Downloads:**`` heading with no anchors
+    whenever both URLs came back empty, which is indistinguishable -- to a reader
+    and to any link assertion -- from report generation having silently failed.
+
+    Media is treated differently on purpose: an incident need not have a snapshot
+    or a clip, so an absent media block is normal and is omitted rather than
+    announced. A report with nothing at all to download is not normal, so it says
+    so explicitly and logs a warning.
+    """
+
+    side_effects: dict[str, Any] = {}
+
+    downloads = [
+        f"- [{label}]({url})"
+        for label, url in (
+            ("Markdown Report", getattr(report_result, "http_url", "")),
+            ("PDF Report", getattr(report_result, "pdf_url", "")),
+        )
+        if url
+    ]
+    if downloads:
+        side_effects["report_downloads"] = "\n".join(["**Report Downloads:**", *downloads]) + "\n"
+    else:
+        logger.warning(
+            "Report for incident %s produced no downloadable artifact: the object store returned "
+            "neither a markdown nor a PDF URL",
+            incident_id,
+        )
+        side_effects["report_downloads"] = (
+            "**Report Downloads:** none - the report was generated, but the object store returned "
+            "no markdown or PDF URL, so there is nothing to download.\n"
+        )
+
+    media = [
+        f"- {prefix}[{label}]({url})"
+        for prefix, label, url in (
+            ("!", "Incident Snapshot", getattr(report_result, "image_url", "")),
+            ("", "Incident Video", getattr(report_result, "video_url", "")),
+        )
+        if url
+    ]
+    if media:
+        side_effects["media"] = "\n".join(["**Media:**", *media]) + "\n"
+
+    return side_effects
+
+
 # ========== REPORT AGENT MODELS ==========
 
 
@@ -553,21 +606,7 @@ async def report_agent(config: ReportAgentConfig, builder: Builder) -> AsyncGene
         report_result = await template_report_tool.ainvoke(report_tool_args)
         logger.info("Single incident report generated successfully")
 
-        side_effects = {}
-        if hasattr(report_result, "http_url") or hasattr(report_result, "pdf_url"):
-            downloads = ["**Report Downloads:**"]
-            if hasattr(report_result, "http_url") and report_result.http_url:
-                downloads.append(f"- [Markdown Report]({report_result.http_url})")
-            if hasattr(report_result, "pdf_url") and report_result.pdf_url:
-                downloads.append(f"- [PDF Report]({report_result.pdf_url})")
-            side_effects["report_downloads"] = "\n".join(downloads) + "\n"
-        if hasattr(report_result, "image_url") or hasattr(report_result, "video_url"):
-            media = ["**Media:**"]
-            if hasattr(report_result, "image_url") and report_result.image_url:
-                media.append(f"- ![Incident Snapshot]({report_result.image_url})")
-            if hasattr(report_result, "video_url") and report_result.video_url:
-                media.append(f"- [Incident Video]({report_result.video_url})")
-            side_effects["media"] = "\n".join(media) + "\n"
+        side_effects = _build_report_side_effects(report_result, incident_id)
         _append_artifact_display_note(side_effects)
 
         agent_output = AgentOutput(
--- a/services/agent/packages/vss_agents/src/vss_agents/api/front_end_config.py
+++ b/services/agent/packages/vss_agents/src/vss_agents/api/front_end_config.py
@@ -27,6 +27,8 @@ from pydantic import BaseModel
 from pydantic import ConfigDict
 from pydantic import Field
 
+from vss_agents.utils.llm_endpoint import validate_chat_llm_endpoints
+
 
 class StreamingIngestConfig(BaseModel):
     """Configuration for the streaming video ingest and RTSP stream endpoints.
@@ -139,4 +141,8 @@ class VSSFastApiFrontEndConfig(FastApiFrontEndConfig, name="vss_fastapi"):  # ty
 async def register_vss_fastapi_front_end(
     _config: VSSFastApiFrontEndConfig, full_config: Config
 ) -> AsyncGenerator[FastApiFrontEndPlugin]:
+    # First point at which the interpolated config is known, and still before the
+    # server binds: an LLM endpoint that no request can reach must stop the agent
+    # here rather than surface later as a stream error naming only the path.
+    validate_chat_llm_endpoints(full_config)
     yield FastApiFrontEndPlugin(full_config=full_config)
--- a/services/agent/packages/vss_agents/src/vss_agents/utils/llm_endpoint.py
+++ b/services/agent/packages/vss_agents/src/vss_agents/utils/llm_endpoint.py
@@ -1,0 +1,125 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Startup check that the workflow's LLM has an endpoint a request can reach.
+
+Every profile writes its LLM endpoint as ``base_url: ${LLM_BASE_URL}/v1`` with no
+``:-`` fallback. When that variable arrives empty, nothing downstream objects: the
+NAT YAML loader substitutes the empty string rather than failing, so ``base_url``
+becomes the literal ``/v1``; ``NIMModelConfig.base_url`` is ``str | None`` with no
+validation, so Pydantic accepts it; and the LangChain NVIDIA client only *warns*
+about a URL with no scheme and no host. The agent then starts, passes its health
+check, and fails on the first token with ``InvalidUrlClientError`` naming
+``/v1/chat/completions`` -- which the top agent renders as ``Error:
+/v1/chat/completions`` followed by the canned apology, with nothing pointing at
+the variable that was missing.
+
+Note that the client's ``https://integrate.api.nvidia.com/v1`` default only
+applies when ``base_url`` is *omitted*. Setting it to ``/v1`` defeats it, so an
+empty variable is strictly worse than an absent one.
+
+Scope is deliberately narrow:
+
+* The entry named by ``workflow.llm_name`` is fatal. It is the agent's single
+  reasoning LLM, so an unusable endpoint there means no prompt can be answered
+  and there is nothing to be gained by starting.
+* Any other ``llms:`` entry only warns. The profiles declare an entry per
+  LLM/VLM flavour and select one at runtime, so an unused entry can carry an
+  unresolved endpoint without anything being wrong. ``llms.rtvi_vlm`` is the
+  live example: ``RTVI_VLM_BASE_URL`` is defined only in
+  ``deploy/docker/services/rtvi/rtvi.env``, which the agent gets through the
+  ``include:`` chain in ``deploy/docker/services/compose.yml`` -- so it resolves
+  in a Compose deployment but not for ``nat serve --config_file`` run outside
+  the container, where the entry renders as the bare ``/v1``. Refusing there
+  would break a documented dev workflow over an LLM no prompt touches.
+* An entry that omits ``base_url`` is left alone: ``openai_vlm`` does that on
+  purpose so the client uses its own endpoint.
+"""
+
+import logging
+from urllib.parse import urlsplit
+
+from nat.data_models.config import Config
+
+logger = logging.getLogger(__name__)
+
+_VALID_SCHEMES = ("http", "https")
+
+_VALID_SHAPES = (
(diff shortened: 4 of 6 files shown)


hugoverjus:
### CI evidence for `da027bd24b9a69f644ae724f056072a853a3f418`

Re-authorized after the force-push (the mirror was still pinned at the previous
head, so nothing had auto-synced). Mirror `pull-request/2050` now points at
`da027bd2`, and every check reached a terminal state with **no failures**.

The four jobs attributable to this change, all green:

| Check | Result | Evidence |
|---|---|---|
| [Test (pytest)](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33909381507/job/101142705108) | success | `3739 passed, 34 warnings in 344.25s` |
| [Lint (Python)](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33909381507/job/101142704891) | success | `All checks passed!` / `469 files already formatted` |
| [Type Check (mypy)](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33909381507/job/101142705132) | success | `Success: no issues found in 132 source files` |
| [Build vss-agent → GHCR](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33909381079/job/101142136846) | success | `sha256:8cc0725012a344ce8005b8e3e42533e02283cca102f775aa08d545c6f30b734c` |

Colour alone isn't proof, so the new tests were confirmed to have actually
executed rather than been collected and skipped. All 11 cases in
`tests/unit_test/utils/test_llm_endpoint.py` ran and passed, covering both
directions of the contract — the endpoints that must be accepted
(`test_in_cluster_nim_passes`, `test_gateway_route_passes`,
`test_hosted_endpoint_passes`, `test_omitted_base_url_passes`), the ones that
must be refused (`test_empty_llm_base_url_resolves_to_relative_path_and_refuses`,
`test_empty_string_base_url_refuses`, `test_scheme_less_host_refuses`,
`test_scheme_without_host_refuses`, `test_non_http_scheme_refuses`), and the
narrow-scope guarantees (`test_non_workflow_entry_only_warns`,
`test_healthy_config_warns_about_nothing`).

Coverage corroborates that the new module was exercised end to end rather than
merely imported:

```
vss_agents/utils/llm_endpoint.py        31   0   100%
vss_agents/api/front_end_config.py      32   2    94%
```

The built agent image is tied to this commit —
`ghcr.io/nvidia-ai-blueprints/vss/vss-agent:pr-2050-da027bd24b9a` — and the
container-source gate accepted it: `labels match the source tree`.

Remaining runs on the mirror (`OSRB Scan`, `Fern Docs CI`, `Helm Sync`,
`Skills Eval`/`Review`/`Signatures`, `Spatial AI Data Utils`, `Model Routing
Notebook`, `vss-playbook-compliance`) and the public-side `SonarQube Analysis`
and `Tracking Ref Check` are all success.

The one non-terminal check is
[`Trigger Downstream Pipeline`](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/actions/runs/33909381507/job/101146588769),
still pending. That job polls a downstream pipeline on a self-hosted runner
against a four-hour budget and is not a signal about this change, so it was
deliberately left running rather than waited out.

Nothing but review approval is outstanding: `mergeStateStatus: BLOCKED`,
`reviewDecision: REVIEW_REQUIRED`.

hugoverjus:
### Live evidence for `9dad18e2dc9340881af712762756eff13a82b38f`

Validated on a 2×RTX PRO 6000 host against the real `vss-vios-ingress:develop-latest` image (nginx/1.24.0 Ubuntu — open-source, so nginx-plus `resolve` on upstream servers is not available). All proof resources were prefixed `nginxproof-` and removed afterwards; no pre-existing container, volume, network, or image was stopped or modified.

#### 1. `fix(vios)` — the nginx upstream race

The exact nightly failure signature reproduces from `nginx -t` alone, same file and same line as job `425460888`:

```
[emerg] 1#1: host not found in upstream "vss-vios-streamprocessing" in /etc/nginx/nginx.conf:105
nginx: configuration file /etc/nginx/nginx.conf test failed
```

| Case | Setup | Result |
|---|---|---|
| A — upstream absent at start (what CI hit) | ingress started with no upstream present | `Status=restarting ExitCode=1`; `RestartCount` 7→8→9 with the backoff visibly widening (count stalls for the last 18s of a 45s window) |
| B — upstreams present first (what this fix enforces) | stub upstreams started first, then the ingress | `Status=running RestartCount=0`, **zero** `[emerg]`, TCP `:30888` open under the *same* healthcheck the compose file uses, `HTTP 200` proxied to both upstreams |
| C — recovery | case A left running after the upstream appeared | recovered at `RestartCount=10`, then `HTTP 200` |

Case C is the substantive point: this is a **timing** failure, not a broken route. The ingress does come up on its own — just not before the readiness sample. That also confirms the `ci-debug` diagnosis (missing service / upstream should be removed) was wrong; removing the upstream would have deleted a working route.

Upstream paths verified unchanged end to end: `/vst/api/v1/live` arrived at the upstream as `/api/v1/live`, and `/vst/api/v1/sensor/live` as `/api/v1/sensor/live`.

Static validation on this branch: `docker compose config --quiet` exits 0 for the failing `bp_wh_auto_calib` profile set; the resolved graph activates 14 services with all 9 `depends_on` entries present; and a negative control renaming one target to `streamprocessing-ms-TYPO` correctly fails with `depends on undefined service`, proving all nine names are real (`required: false` excuses a *profile-disabled* service, not an *undefined* one).

Fast checks from `services/agent`: `ruff check` clean, `ruff format --check` clean (134 files), `mypy` clean (132 files), `pytest` **2409 passed**.

#### 2. Report-generation failures — confirmed infrastructure, and the fix is now demonstrated

No product change is needed for `[Warehouse flow] Step 2` (jobs `425315116`, `425315399`). The report agent is never invoked because tool calling is unavailable. This host let me prove the remedy rather than just assert it — a **side-by-side probe of the same model on two endpoints, using the identical tool-bound payload** (inference requests only; the shared endpoint was not modified in any way):

| Endpoint | Serving flags | Result |
|---|---|---|
| Local NIM on this host | `--enable-auto-tool-choice --tool-call-parser qwen3_coder` | **HTTP 200**, `finish_reason=tool_calls`, `report_agent(sensor_id="Camera_01")` |
| Shared CI endpoint | neither flag set | **HTTP 400** — `"auto" tool choice requires --enable-auto-tool-choice and --tool-call-parser to be set` |

Both report `id=nvidia/nemotron-3.5-lightning-30b-a3b`, `owned_by=vllm`, and both answer `/v1/models` fine — which is why health checks passed while every tool-bound call failed.

So the model is fully capable of the exact tool call the warehouse report test needs; only the serving configuration differs. A known-good reference for the shared endpoint, taken from the working local deployment:

```
--reasoning-parser nemotron_v3 --enable-auto-tool-choice --tool-call-parser qwen3_coder
```

That is an infra change to a shared endpoint, not a change in this repo.

#### Scope note

Nothing in this PR addresses report generation. This PR is the agent fail-fast commit (`da027bd`) plus the nginx ingress ordering commit (`9dad18e`). The empty-`LLM_BASE_URL` root cause is fixed separately in `ci-vss-oss` MR !511; this PR is not that fix. Known limitations of the ordering fix — restart-time races and nginx's lifetime IP caching — are stated in the PR description and deliberately left for a separate data-path PR.

akshayanv (dismissed):

zac-wang-nv (dismissed):
LGTM

hugoverjus:
## Two findings from this investigation that are **not** changed in this PR

Both were found while tracing the nightly report-generation failure. Neither has a
fix I am willing to improvise, so each is written up here with the code path, what
I could and could not reproduce, and what I believe the correct fix is. Recording
them here rather than in a commit so they are attached to the investigation that
found them.

---

### A. A sensor can be online in VIOS and never subscribed by rt-cv

**Symptom shape**: a camera is registered, VIOS serves it, it shows `state=online`,
and it is simply absent from anything rt-cv produces. Nothing errors. Some cameras
work and one does not.

**Mechanism — observed directly on a live host.** rt-cv is started with
`num-source-bins=0` and `use-nvmultiurisrcbin=1`, so it has no static stream list:
its entire subscription set arrives at runtime as `POST /api/v1/stream/add` on port
9000. Those POSTs come from VIOS webhooks configured in `notification_config.json`.

rt-cv could not start on this host (its TAO model download fails with
`Invalid apikey`), so nothing was listening on that port. I bound a recording
listener onto the `vss-rtvi-cv` network alias on :9000, confirmed
`vss-rtvi-cv:9000` resolved to it, registered an RTSP sensor through `vss vios add`,
and watched what arrived.

**Nothing arrived.** Zero `stream/add` requests, while the sensor went `online` in
VIOS. VIOS says why in its own log:

```
"change" : "camera_add",
webhook_notifier.cpp:799: deliverMessage
    No webhook matched event[camera_status_change/camera_add sensor=11583ccf-...]
```

The deployed webhook table explains it completely:

| enabled | trigger | targets |
|---|---|---|
| **false** | `camera_add` | *(none)* |
| true | `camera_streaming` | `vss-rtvi-cv:9000`, `vss-rtvi-embed:8000`, `rtvi-vlm:8000` |
| true | `camera_remove` | `elasticsearch:9200`, `vss-rtvi-cv:9000`, `vss-rtvi-embed:8000`, `rtvi-vlm:8000` |

**Registering a sensor emits `camera_add`. The webhook that tells rt-cv about it is
gated on `camera_streaming`.** Across the entire VIOS log on this deployment,
exactly one `camera_status_change` was ever emitted — `camera_add` — and it matched
no webhook. So rt-cv is only ever told about a sensor that *transitions into
streaming*, and in this direct-mode deployment (`VST_USE_SDRC=false`) nothing drives
that transition for a newly registered sensor.

That accounts for the "some cameras work, one doesn't" shape exactly: a camera
present and streaming when the stack came up is subscribed; one registered later,
or one that never transitions to streaming, never is — and it stays `online` in
VIOS the whole time, which is why it looks fine everywhere you would think to look.

**Why I did not fix it.** The fix is a change to the webhook trigger table, and
which change is correct is a product decision I should not make unattended.
Enabling the `camera_add` webhook would subscribe rt-cv at registration, but in
SDRC mode the SDR controller also drives subscription, so it risks double-adding.
Making something drive `camera_streaming` on registration is the alternative and
has a different blast radius. It also is not in this repo's product code — it is
deployment configuration. Worth noting the two known-bad states are silent in
opposite directions, which is the real problem.

**What I would ask an owner**: is rt-cv's subscription meant to be driven by
registration or by an actual streaming transition, and what is supposed to drive
that transition in direct (non-SDRC) mode?

---

### B. The top agent's planner can answer the user with no tool and no grounding

This is the one I would most want a second opinion on, because **it fails in a
shape that looks like a good answer** rather than like an error.

**The code path.** `_plan_node` calls the LLM with **no tools bound** —
`self.llm.bind(**llm_kwargs)` carries reasoning kwargs only, never `bind_tools`.
Its output is parsed as a string. If it starts with `[USER]`, the remainder is
assigned straight to `state.final_answer` and the node returns:

```python
if plan_text.strip().startswith(PLAN_CLARIFY_PREFIX):
    clarification = plan_text.strip()[len(PLAN_CLARIFY_PREFIX):].strip()
    state.final_answer = clarification
    return state
```

That return **bypasses the agent node, every tool, and the critic**. And the plan
prompt does not merely permit this for clarifying questions — it explicitly
authorises answering:

```
If user's question can be answered directly without any tools, respond with EXACTLY:
[USER] <your answer>
...
Example direct answer:
what tools are available?
[USER] The available tools are: ... (list of tools)
```

So the planner may answer any question it judges answerable without tools,
straight from model priors, and the result is emitted to the user carrying no
marker that it was never grounded. Downstream it is byte-indistinguishable from a
tool-verified answer.

**Why this is worse in exactly the failure this PR came from.** When the LLM
endpoint rejects tool-bound requests — the environmental cause of the nightly
failure, HTTP 400 on every tool-bound call — the plan node is the one node that
still works, *because it binds no tools*. Its only working output path is the
ungrounded direct answer. A deployment whose tool calling is entirely broken can
therefore keep returning fluent, confident, sourceless answers instead of failing.
That is the worst possible degradation mode and it needs no bug to trigger, just a
misconfigured endpoint.

**Should tools be bound there? No — and I recommend against it.** I was asked to
fix this only if binding tools was genuinely clean. It is not:

1. The plan/clarify protocol is entirely string-parsed. With tools bound the model
   may answer with `tool_calls` and empty content; `plan_text` would be empty, the
   `[USER]` check could never match, and `state.plan` would be empty. It breaks the
   contract of the node it is meant to protect.
2. The plan node runs *before* the agent node, which is the node that owns tool
   execution. Binding tools in both lets a planning step invoke side-effecting
   tools — report generation, sensor deletion — during what is supposed to be a
   read-only planning pass.

The absence of `bind_tools` is not the defect. The defect is that a node with no
tools is permitted to produce a **final answer** rather than only a question.

**What I believe the correct fix is** — in preference order, none of which I would
land unattended:

1. Narrow the `[USER]` path to what its name says: a clarifying *question*. Remove
   the "answer directly" authorisation, and route genuinely tool-free questions
   ("what tools are available?") through a real tool or a deterministic handler
   that answers from the actual registry rather than from the model's memory of it.
2. If a direct-answer path must stay, mark it — carry a flag through to the
   response saying the answer used no tools, so a caller and an eval can tell.
3. Independently: make a tool-incapable endpoint fail loudly rather than
   degrade into this path. Commit `da027bd` catches a *missing* endpoint; it does
   not catch a *tool-incapable* one.

**Reproduction — partial, and I want to be precise about the limit.** I could not
reproduce this specific path on the host. The deployed profile is
`dev-profile-search`, which never instantiates `top_agent` at all: across the whole
agent log, `Plan node produced plan`, `Plan node requesting clarification` and
`Starting Agent Node` each appear **zero** times. Reproducing it needs a profile
that runs `top_agent`, and standing one up needs GPU time I am not willing to spend
unattended.

What I did confirm on the deployed workflow is the *shape*: asking
"what tools are available?" returned a fully-formed, confidently-formatted
description of the available tools with `tool_calls: None` on the response — an
answer about system state produced without consulting system state. Same class of
output, different workflow. The `top_agent` path above is traced statically and the
quoted code and prompt are from this branch.

akshayanv (approved):

hugoverjus:
## CI status on `b53f257` — 51 pass, 16 skipped, 1 blocked

Every GitHub-native check passes. The single failure is `Trigger Downstream
Pipeline`, and it is **blocked by a downstream infrastructure fault, not by
anything on this branch**. Recording the evidence so nobody has to re-derive it.

### What fails

The downstream `ci-vss-oss` pipeline dies in `get_sources`, before a single test
runs:

```
fatal: shallow file has changed since we read it
fatal: Fetched in submodule path 'video-search-and-summarization',
       but it did not contain 52aaf2ca8c1a78167ccb7000e32912547de34921.
       Direct fetching of that commit failed.
ERROR: Job failed: exit code 128
```

### Why it is not this branch

- **`52aaf2ca8c1a78167ccb7000e32912547de34921` does not exist in this repository.**
  `git cat-file -t` on it returns `bad object`. It is not this PR's head
  (`b53f257e97e2c0425cf197165cbe5674a7a5484e`), not an ancestor, and not on
  `develop`. It is a stale gitlink pinned in `ci-vss-oss` at `c3e7f5d8`, fetched
  during `get_sources` before the `VSS_SUBMODULE_HASH` override applies.
- **Exit 128 in `get_sources`** means the failure is in git checkout. No job step
  ran. The whole trace is ~8 KB and contains **zero** references to the agent, LLM
  endpoint validation, report generation, or the vios ingress — I grepped for all
  four. Nothing from this PR is exercised before it dies.
- **Which job reports it moves between runs**, which is the signature of a shared
  git-cache race on the runners rather than a defect in the code under test:

  | Downstream pipeline | Job that lost the race |
  |---|---|
  | #12378 | `test-base` |
  | #12381 | `eval-warehouse-bp-wh-2d` (`test-base` passed) |
  | #12391 | `test-base` again |

  `shallow file has changed since we read it` is git reporting that another process
  mutated `.git/shallow` mid-fetch — concurrent jobs sharing a clone directory.

### What I did about it

Re-ran it three times, producing three fresh downstream pipelines. All three died
the same way in source checkout. Retries reshuffle *which* job loses the race but
never get past it, so I have stopped rather than keep consuming runner capacity —
a fourth attempt has no reason to behave differently.

**This needs an owner of `ci-vss-oss`**, not a change here. Two candidate fixes,
both in that repo: advance or repair the submodule gitlink at `c3e7f5d8` so it
points at a commit that exists on the public remote, and/or stop sharing a shallow
clone between concurrent jobs (`GIT_STRATEGY`/`GIT_DEPTH`, or per-job build dirs).

### Local verification of this branch, for completeness

Run against `b53f257` in `services/agent`:

- `pytest packages/vss_agents/tests/unit_test` — **2416 passed**
- `ruff check` — clean
- `ruff format --check` — clean
- `mypy` — no issues

Plus the live-host evidence for commits 1 and 2 described in the PR body.

PR #2051, opened 2026-09-04
feat(deploy): mount the LLM NIM at the gateway's /llm

hugoverjus:
> **Stacked on #1983** (`feat/via-2713-gateway-consolidated`). Review that first; this branch is one commit on top of it. Base retargets to `develop` when #1983 merges.

## What this is not

**This is not the fix for the 2026-09-03 nightly.** That failure was an *empty* `LLM_BASE_URL` interpolating to a bare `/v1`, and it is fixed in #2049 against `develop` with an inline Compose default. `/llm` addresses topology independence — a caller that has one origin and no second endpoint — and it would not have prevented that nightly. Please do not let the two be conflated.

For the same reason #1983 is not implicated in the nightly either: it is unmerged, is not an ancestor of `develop` (58 commits ahead), `git grep -c VSS_GATEWAY_ORIGIN origin/develop` returns zero matches on the whole branch, and it leaves `LLM_BASE_URL` byte-identical (`develop:107`, PR head `:146`) while moving eleven sibling endpoints onto the gateway. None of its rework was running.

## What it does

Mounts the in-deployment LLM NIM at `/llm`, so `${VSS_GATEWAY_ORIGIN}/llm/v1/chat/completions` reaches the model. Every other service a caller must reach is already on the gateway; the model is the one that still requires a separately communicated endpoint.

**Nothing is pointed at it.** Every profile keeps `LLM_BASE_URL=http://vss-llm-nim:8000`, a colocated agent addresses the model exactly as it does today, and the rule `remote-agent.env.example:64` states — model APIs are configured directly, not mounted — remains the default. This PR makes the other arrangement *possible*; adopting it stays a per-deployment decision made by setting `LLM_BASE_URL`.

That is also the honest test for whether this should merge: if keeping model APIs off the gateway is a deliberate policy rather than an artefact of ordering, close this. Nothing depends on it.

## Conforming to #1983

- Shaped like `bk_rtvi_vlm_strip` and its siblings — the NIM serves its OpenAI-compatible surface at the root, so `/llm/v1/...` must arrive as `/v1/...`. As you note, the Docker edge is a hand-written template with shell-style expansion; a strip backend plus an ACL pair is the idiomatic shape here.
- Uses your `nbsrv() eq 0` optional-mount convention rather than letting HAProxy answer a bare 503, so an `LLM_MODE=remote` deployment reports `x-vss-gateway-unavailable: llm` and stays distinguishable from a model that is deployed and failing.
- Host allowlist, `h_main` gating and `X-Forwarded-Proto` normalisation are inherited unchanged; this adds no Host exceptions.

Two deliberate departures, both stated in the config:

- **`timeout server 600s`**, following the raised timeout on `bk_lvs_strip`. A non-streaming completion sends nothing until generation finishes and a reasoning model can outlast the 120s default; a 504 mid-generation reads as a backend failure. Streaming needs nothing extra — `timeout server` measures inactivity and a token stream is never idle.
- **TCP `check`, not `option httpchk GET /v1/health/ready`**, and this one interacts with your marker. A NIM holds its port open while it loads weights, which takes minutes; a readiness check would leave the backend DOWN for that whole window and the marker would report the model *absent from the deployment* when it is merely warming up. Port-open means deployed, which is precisely the fact `nbsrv` needs. It is also what every other backend in the file uses.

## Verified against a live HAProxy 3.4.2

Stub backend on a Compose network, aliased `vss-llm-nim`, with the real template:

| Request | Result |
|---|---|
| `POST /llm/v1/chat/completions` | backend received `/v1/chat/completions` |
| `GET /llm/v1/models` | backend received `/v1/models` |
| same, backend removed | `503` + `x-vss-gateway-unavailable: llm` and the synthesised body |
| `/rtvi-vlm/v1/models` (control) | `503` + `x-vss-gateway-unavailable: rtvi-vlm`, unchanged |

`haproxy -c` on the expanded template reports no new warnings — only the pre-existing "http-request rule placed after a use_backend rule" notices this file already relies on.

Not verified: no real NIM was loaded, so the 600s timeout and the warm-up behaviour of the TCP check are reasoned from the NIM's documented readiness sequence, not measured.

## Kubernetes deliberately not included

`_ingress-routes.tpl` is declarative and `verify-ingress-routes.py` asserts, among other things, that disabling a component removes its route rather than leaving an Ingress pointed at a Service that was never created. A `/llm` row therefore needs the NIM Service name and its per-profile enable condition threaded through each profile — more than a row, and with no caller yet. Better as a follow-up once something actually uses the Docker mount.

Also not added: `/llm` is not in the CLI's `INGRESS_SERVICES` probe list, since no `vss` command group needs the model endpoint.

### DIFF
--- a/deploy/docker/README.md
+++ b/deploy/docker/README.md
@@ -468,6 +468,7 @@ serve that prefix natively.
 | `/rtvi-cv` | strip | `${VSS_GATEWAY_ORIGIN}/rtvi-cv` | `/rtvi-cv/api/v1/stream/add` | `/api/v1/stream/add` |
 | `/rtvi-embed` | strip | `${VSS_GATEWAY_ORIGIN}/rtvi-embed` | `/rtvi-embed/v1/models` | `/v1/models` |
 | `/lvs` | strip | `${VSS_GATEWAY_ORIGIN}/lvs` | `/lvs/v1/live` | `/v1/live` |
+| `/llm` | strip | opt-in; nothing is pointed here | `/llm/v1/chat/completions` | `/v1/chat/completions` |
 | `/phoenix` | strip | `${VSS_GATEWAY_ORIGIN}/phoenix` | `/phoenix` | `/` (keep `PHOENIX_HOST_ROOT_PATH=/phoenix`) |
 | `/vst` | preserve | `${VSS_GATEWAY_ORIGIN}` (not `/vst`) | `/vst/api/...` | `/vst/api/...` |
 | `/vios` | **rewrite to `/vst`** | `${VSS_GATEWAY_ORIGIN}/vios` | `/vios/api/...` | `/vst/api/...` |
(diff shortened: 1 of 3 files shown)


hugoverjus:
Closing for now to consolidate on a single PR for the nightly failure (#2050).

This route is not the fix for that failure — `/llm` addresses topology independence, not the empty `LLM_BASE_URL` that broke the nightly — and it is stacked on #1983, so it cannot merge before that lands anyway. Worth reopening as a follow-up to #1983 when the gateway work merges.

PR #2052, opened 2026-09-04
docs(rt-vlm): align Compose defaults with GHCR

nv-amkale:
## Summary
- add a tracked RT-VLM Compose `.env.example` and use it from Fern and service docs
- make the standalone Compose fallback use the managed GHCR `develop-latest` image, matching the existing Docker/Helm channel policy
- retain the pinned NVCR image as an explicit release option and update 3.3.0 docs to 26.08.2
- document `MESSAGE_BUS` as the active generated-message bus selector

## Validation
- `python3 .github/scripts/test_helm_release_channel_policy.py`
- environment sample syntax check
- GHCR manifest lookup for `ghcr.io/nvidia-ai-blueprints/vss/vss-rt-vlm:develop-latest`
- `git diff --check`
- pre-commit: TruffleHog, SPDX, LFS, and DCO checks

## Limitation
- Docker and Fern CLIs are not installed on the authoring host, so Compose rendering and Fern link validation are left to Linux CI.


services/rtvi/rt-vlm/docker/compose.yaml
```diff
@@ -33,17 +33,12 @@ services:
       - rtvi-hf-cache:/tmp/huggingface
 
     environment:
-      AZURE_OPENAI_API_KEY: "${AZURE_OPENAI_API_KEY:-}"
-      AZURE_OPENAI_ENDPOINT: "${AZURE_OPENAI_ENDPOINT:-}"
       MODEL_PATH: "${MODEL_PATH:-ngc:nim/nvidia/cosmos3-nano-reasoner:bf16-final}"
```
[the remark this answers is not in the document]
nv-amkale: Thanks, this is valid. Restored the Azure OpenAI and NVIDIA LLM Gateway environment forwarding in 41e0d9c3c. Removing those entries would break existing Compose deployments that provide credentials through .env, even though the current coverage is only manual and credentialed. The documentation now scopes them explicitly to the openai-compat backend. End-to-end validation for those remote backends should be handled as a separate follow-up rather than removing a compatibility contract.

docs/real-time-vlm.mdx
```diff
@@ -401,19 +400,19 @@ export IS_SBSA=true
...
   -f docker/Dockerfile \
-  -t <registry>/<repo>/vss-rt-vlm:3.3.0-26.08.1-custom-sbsa \
+  -t <registry>/<repo>/vss-rt-vlm:3.3.0-26.08.2-custom-sbsa \
   --load .
 ```
 
 For Jetson AGX Thor / IGX Thor (ARM64 but not SBSA), do **not** set
 `IS_SBSA`. The default base image
-(`nvcr.io/nvstaging/vss-core/vss-rt-vlm:3.3.0-26.08.1`) is multi-arch, so a
+(`nvcr.io/nvstaging/vss-core/vss-rt-vlm:3.3.0-26.08.2`) is multi-arch, so a
```
zac-wang-nv: do we still want to point to nvstaging from docs?

nv-amkale: Yes,  I m also not sure, should we use nvcr nightly or ghcr, please recommend

nv-amkale: Updated in 9464ff998: only the overview development reference now uses `ghcr.io/nvidia-ai-blueprints/vss/vss-rt-vlm:develop-latest`. The pinned NGC image references and pull-secret guidance remain unchanged for the final release.

zac-wang-nv: can we change the nvstaging to ghcr?   my worry is for external users checkout the default branch and the doc, they have no access to nvstaging therefore cannot use the container, given that we have gchr hosted dev containers, that might provide an easier way for them to use.  We can document that the dev container(pre-release) is https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization#container-image-availability

zac-wang-nv (approved):
TODO:
### Comment 1 — [INLINE] services/rtvi/rt-vlm/docker/.env.example:5 | P1
> RTVI_IMAGE=ghcr.io/nvidia-ai-blueprints/vss/vss-rt-vlm:develop-latest

**Text:** Please preserve the SBSA image alternative when moving the quick start into this sample. The removed README example told DGX Spark users to select the `-sbsa` image, but the replacement only offers `develop-latest`. In deploy/docker/container-inventory.json, that tag's ARM64 image is built with `ARM_PLATFORM=igpu` (Jetson/Thor dependencies), while Spark/GB300 need the separate `develop-latest-sbsa` build. Both variants report `linux/arm64`, so following the new copy-and-start instructions on those supported hosts selects the wrong runtime. Add the GHCR `develop-latest-sbsa` alternative and identify the hosts that require it; carry that distinction into the mirrored quick-start/Helm guidance too.

PR #2057, opened 2026-09-04
feat(bvai): Enable BVAI to point the NemoClaw harness to the VSS deployment's own local LLM

nvleeryan:
## Summary

`vss-build-vision-ai` can now stand up a NemoClaw harness that runs on the
deployment's **own** LLM NIM instead of a remote provider. Previously the
skill defaulted to remote Claude Opus and had no documented local route, so
"use a local model" could only mean a second, NemoClaw-managed server on its
own GPU. Reusing the build's NIM costs no additional GPU, needs no provider
API key, and requires no egress to a hosted model — the answer for an
air-gapped or key-less host.

What blocked it was mechanical: `NEMOCLAW_INFERENCE_PROXY` was assignable
only as a literal in the notebook's Section 1.3, so a non-interactive run
could not turn the bundled proxy off. That proxy always reaches its upstream
as `https://<host>` on 443, which cannot represent a NIM on plain HTTP on
`30081`.

## Plan

Make the toggle an environment override, then teach BVAI the route it
unlocks — including what it costs, so the user can still choose a separate
local server or the remote default.

The toggle is read from `SHELL_ENV` in the notebook's derived-settings cell
rather than added to `run_setup_notebook.py`'s `NOTEBOOK_PARAMETERS` table:
that table assigns the raw environment string and Section 3.1 accepts only a
real bool. Reading it in the derived cell stays idempotent across cell
re-runs and keeps the existing `True` default.

## Related Issue

N/A

## Changes

- **`skills/vss-build-vision-ai/references/agent-harness.md`** — the substance
  of the PR. A routing table separating "reuse the deployment's LLM" from "use
  a local model" (two different destinations, previously conflated); the four
  variables the NIM route needs, with `LLM_PORT` and `NIM_SERVED_MODEL_NAME`
  read from `resolved.yml` rather than assumed; the shared-KV-cache cost to
  report *before* taking the route, since agent turns then contend with the
  build's own `lvs` and `search` work; and a note that `vss-backend`
  allowlists literal host/port pairs with no entry for `30081` or `30082`, so
  a `403 CONNECT tunnel failed` on the first turn is that gap, not a
  deployment fault.
- **`deploy/docker/scripts/deploy_nemoclaw.ipynb`** — read
  `NEMOCLAW_INFERENCE_PROXY` from the environment, falling back to the 1.3
  literal; say in 1.3 and 3.1 when to disable it and that
  `COMPATIBLE_API_KEY=EMPTY` must then be passed explicitly, since the
  placeholder is applied only on the branch the override turns off.
- **`docs/nemoclaw-configuration.mdx`** — same guidance on that variable's row.

Four lines of executable change; the rest is skill and notebook prose. No
service, image, or Compose change, and the remote-provider default is
untouched.

## Validation

The remote route is unaffected and still works on this branch — the proxy
engages for `inference-api.nvidia.com`, which resolves to private addresses,
the case it exists for.

The local-NIM route is **not yet validated end to end**. It needs a deployed
build whose NIM answers on `30081` before onboarding, and the open question
is whether NemoClaw authorizes the inference endpoint it was onboarded with
given that `vss-backend` carries no entry for that port. The skill documents
that failure as a blocker to report rather than something to work around, so
a `403` there would mean adding a policy entry to `assets/vss_nemoclaw_policy.yaml`
in a follow-up.

## Checklist

- [x] I have installed and run pre-commit hooks locally.
- [ ] New or existing tests cover these changes. — n/a: notebook cells and
      skill references have no test harness.

### DIFF
--- a/deploy/docker/scripts/deploy_nemoclaw.ipynb
+++ b/deploy/docker/scripts/deploy_nemoclaw.ipynb
@@ -200,7 +200,7 @@
         "\n",
         "- `NEMOCLAW_INSTALL_REF` — pinned installer.\n",
         "- `AGENT_RUNTIME` — `openclaw` or `hermes` (env override). Add an `AGENT_HARNESS_PROFILES` entry for a new harness.\n",
-        "- `NEMOCLAW_INFERENCE_PROXY` — SSRF check and host proxy for local endpoints.\n",
+        "- `NEMOCLAW_INFERENCE_PROXY` — SSRF check and host proxy for local endpoints (env override; `0` disables it). The proxy reaches its upstream over `https` on port 443, so turn it off for a self-hosted endpoint served over plain HTTP or on another port — such as this deployment's own LLM NIM.\n",
         "- `AGENT_HOOKS_*` / `AGENT_WEBHOOK_PORT` — inbound webhooks (Hermes default `8644`).\n",
         "- `AGENT_DASHBOARD_PORT` — derived (`18789`; override via `NEMOCLAW_DASHBOARD_PORT`).\n",
         "- `NEMOCLAW_VLLM_PORT` — host port for NemoClaw-managed vLLM. Used by `install-vllm`\n",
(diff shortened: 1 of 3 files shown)


nvleeryan:
/ok to test https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/2057/commits/02fb7ecd81a2dfbc7a55954bb56326f421dfd7d3

zac-wang-nv (approved):

PR #2064, opened 2026-09-07
fix(dash): start playback on a real cushion and publish the segment grid we choose

gous-75:
## What this fixes

DASH playback froze repeatedly in the first seconds of a live or replay session,
and worse the further the viewer was from the server. Two independent causes:

**Client.** Playback started on whatever was buffered when a blind 2.5 s timer
fired — measured at 0.74 s of media, after which the player ran dry four times in
the first four seconds. The autoplay gate and the library's own fill target had
been given the same number, so the buffer settled exactly at the threshold and
whether playback started at all came down to which side rounded first.

**Server.** Sessions republished the camera's access units, which hands the
segment grid to whatever keyframe interval the camera happens to use — typically
one second. That costs a manifest poll and a segment fetch every second, and on a
link with any latency a one second segment takes longer to fetch than the second
of media it returns. The muxer cannot group those keyframes into longer segments
either: asking for longer than the interval returns a mixed timeline (measured as
one 2 s segment followed by three 1 s ones).

## What changed

- **Fill past the gate, and start on evidence.** dash.js is asked to fill to
  twice the cushion the autoplay gate waits for; the gate re-checks while the
  buffer is still growing and starts early only after two consecutive flat
  checks, or after a 12 s cap so a viewer is never stuck on a black player. The
  cushion ceiling is measured in segments rather than a flat number of seconds.
- **Every session encodes**, so the segment grid is ours rather than the
  camera's. Both pipeline builders wire the decoder straight to the packager
  unless a transcode is asked for, which skipped the encoder entirely and left
  the manifest at 202 forever; they are now told when a session encodes.
- **The encoder's keyframe interval comes from the segment length in seconds**,
  not from the WebRTC frame count — a different length at every frame rate. A
  video wall composed at 8 fps read 30 frames as 3.75 s and published 4 s
  segments however short the configured length was.
- **Re-encoding sessions get a timeline they can publish** (live and replay);
  without it the session wrote an init segment and nothing else.
- **The DVR window on disk matches the one the manifest advertises.** Retention
  was a fixed file count while the window is in seconds; at 1 s segments the
  manifest listed 91 segments against 63 present, and every one of the difference
  was a guaranteed 404 for a player seeking back.
- **The player's fallback clock points at this deployment's own server**, not a
  public time service — the wrong dependency for a product installed at sites
  without a route to one.
- **`NvBufSurfTransform` failures log the error code, GPU and geometry.** A host
  was found wedged with 7468 of these and nothing to say why.

## Config

`dash.segment_duration_sec` moves from 1 to 2 in the shipped configs, and
`dash.always_encode` (default on) is stated explicitly. Setting it off trades the
grid back for the decode and encode it costs.

## Testing

Two hosts, one with a hardware encoder and one without, over five-minute soaks of
live, live+overlay, replay, replay+overlay and video wall:

- 0 picture-level gaps (measured with `requestVideoFrameCallback`, which sees
  frozen pictures that `currentTime` does not), playback rate 1.000x.
- Segments a uniform 2.000 s; manifest polled every 2 s rather than every 1 s.
- Time to first frame 10-13 s and 12-15 s respectively, from around 20 s before.
- One 1080p30 session costs about a third of a core where there is no hardware
  encoder, and runs ahead of real time.

Reviewed and confirmed on a third host (aarch64) by the reporter of the original
freezes.

## Known and not addressed here

On a high-latency link the player spends around 3 s between parsing the manifest
and starting its period. It tracks the server's clock differing from the
viewer's; where the two agree the same client and manifest start in 0.04 s, and
no player setting tried against it made any difference.


gous-75:
Superseded by #2065, which carries the same commits from a branch that never touched a lockfile. This one picked up an OSRB reviewer requirement from a package-lock.json line that has since been removed, and that requirement does not reset on force-push.

PR #2072, opened 2026-09-07
docs(alert): document multi-core scaling for the Alerts Microservice

danh-nguyen-nvidia:
Replace the stale Scaling and Performance Tuning section of the Alerts Microservice guide. It described a main thread with a fixed worker pool, presented num_workers as the vertical-scaling control, and recommended the retired chunk_size setting. The new section covers the three pipeline modes and their throughput controls, the alert_agent.processes setting that runs the pipeline across CPU cores, the aggregate concurrency the backends see, the startup validation rules, readiness and shutdown behaviour, rollback, and the multi-process metrics.

Add a subsection to the Alert Verification workflow page with the developer profile file to edit, a conservative four-process example that keeps the instance-wide VLM and VIOS load unchanged, the prerequisites, and a link to the full guide.

Removed the kafka.* rows from the scaling table; they remain in the configuration reference of the same page.

## Checklist

### DIFF
--- a/docs/agent-workflow-alert-verification.mdx
+++ b/docs/agent-workflow-alert-verification.mdx
@@ -670,6 +670,45 @@ If you would like to modify the workflow to work with other videos or usecases, 
 
   * Drop a Python module exposing a parser class with a `parse(self, raw_response: str) -> dict` method, then set `vlm.response_parser` in `config.yml` to its dotted path. Use this to replace the built-in YES/NO verification parsing with classification, analytics, or enrichment outputs. See [Pluggable Response Parser](/vss/system-components/analytics-microservices/alerts-microservice#pluggable-response-parser) for the full contract, examples, and error handling.
 
+#### Scale Alert processing across CPU cores
+
+A single Alerts Microservice process uses about one CPU core. Running the pipeline in several processes adds CPU capacity, and it helps when that process is consistently near one core while Kafka consumer lag keeps growing. A process below one core is limited by offered traffic, by the configured VLM concurrency, or by a downstream service, and gains nothing from additional processes.
+
+Edit the profile's Alerts Microservice configuration and redeploy:
+
+```text
+deploy/docker/developer-profiles/dev-profile-alerts/vlm-as-verifier/configs/config.yml
+```
+
+This setting is not available through `user-overrides.env`. A conservative four-process configuration that keeps the profile's instance-wide VLM and VIOS load unchanged:
+
+```yaml
+alert_agent:
+  pipeline_mode: event_loop
+  processes: 4
+  startup_timeout_seconds: 60
+  async_dispatch_max_in_flight: 4
+  async_io:
+    max_vlm_concurrent: 1
+    max_vst_concurrent: 2
+
+persistence:
+  enabled: true
+```
+
+Every per-process limit is multiplied by the process count, which is why the three limits above are divided by four from the one-process profile values of 16, 4, and 8.
+
+Before raising the process count, confirm the following:
+
+- `pipeline_mode` is `event_loop` and `persistence.enabled` is `true`. Both are the profile defaults.
+- Every Alert input topic has at least as many partitions as `replicas × processes`, counted per topic. The profile creates eight partitions per topic, so one replica can run four processes as shipped.
+- The host has about one CPU core per process, plus capacity for the parent and API processes.
+- The VLM and VIOS backends can serve `processes × max_vlm_concurrent` and `processes × max_vst_concurrent` requests concurrently.
+
+Startup refuses a combination the deployment cannot honor and names the reason. `/health` returns HTTP 503 until every process holds a Kafka assignment. Allow for `startup_timeout_seconds` before the instance reports ready.
+
+For pipeline modes, sizing, monitoring, and rollback, see [Scaling and Performance Tuning](/vss/system-components/analytics-microservices/alerts-microservice#scaling-and-performance-tuning) in the Alerts Microservice guide.
+
 ### Step 2: Add a video stream
 
 Add an RTSP stream by clicking the **+ Add RTSP** button on the **Video Management** tab in the agent UI. If you do not have an RTSP stream, you can use [NVStreamer](/vss/system-components/middleware/video-io-storage-vios/nvstreamer) at `http://<HOST_IP>:31000` to upload a video file and create an RTSP stream.
--- a/docs/agent-workflow-alert-verification.mdx
+++ b/docs/agent-workflow-alert-verification.mdx
@@ -696,13 +696,14 @@ persistence:
   enabled: true
 ```
 
-Every per-process limit is multiplied by the process count, which is why the three limits above are divided by four from the one-process profile values of 16, 4, and 8.
+Every per-process limit is multiplied by the process count, which is why the three limits above are divided by four from the one-process profile values of 16, 4, and 8. `max_vlm_concurrent` caps what each process offers the VLM backend and does not raise the backend's throughput. When the VLM is the limiting service, capacity comes from the VLM side, for example more replicas behind the same endpoint.
 
 Before raising the process count, confirm the following:
 
 - `pipeline_mode` is `event_loop` and `persistence.enabled` is `true`. Both are the profile defaults.
-- Every Alert input topic has at least as many partitions as `replicas × processes`, counted per topic. The profile creates eight partitions per topic, so one replica can run four processes as shipped.
-- The host has about one CPU core per process, plus capacity for the parent and API processes.
+- Every Alert input topic has at least as many partitions as the process count, counted per topic. The profile creates eight partitions per topic, so one instance can run four processes as shipped. When several Alerts Microservice instances share the consumer group, for example as replicas on Kubernetes, the bound applies to `instances × processes`.
+- At least as many active streams as processes. Every message of one stream carries the same `sensorId` and lands on one partition, so work is spread across processes at stream granularity. Fewer streams than processes leaves processes without work, and more streams spread the load more evenly.
+- The host has about one CPU core per process, plus capacity for the parent and API processes. Each process also holds its own Kafka consumers, HTTP connection pools to the VLM and VIOS backends, Elasticsearch clients, and memory, so connection, file-descriptor, and memory use grow with the process count.
 - The VLM and VIOS backends can serve `processes × max_vlm_concurrent` and `processes × max_vst_concurrent` requests concurrently.
 
 Startup refuses a combination the deployment cannot honor and names the reason. `/health` returns HTTP 503 until every process holds a Kafka assignment. Allow for `startup_timeout_seconds` before the instance reports ready.
--- a/docs/alert-verification-service.mdx
+++ b/docs/alert-verification-service.mdx
@@ -604,7 +604,7 @@ Verified alerts include the original fields plus an extended `info` section cont
 
 ## Scaling and Performance Tuning
 
-The Alerts Microservice scales along two independent dimensions: how many CPU cores one instance can use, and how many requests it keeps in flight against the VLM and VIOS backends. This section describes the pipeline modes that govern in-process concurrency, the settings that control each dimension, and how to tell which one is limiting a deployment before changing anything.
+The Alerts Microservice scales along two independent dimensions: how many pipeline processes one instance runs, each using about one CPU core, and how many requests it keeps in flight against the VLM and VIOS backends. This section describes the pipeline modes that govern in-process concurrency, the settings that control each dimension, and how to tell which one is limiting a deployment before changing anything.
 
 ### Pipeline Modes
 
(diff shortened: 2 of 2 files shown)


docs/alert-verification-service.mdx
```diff
@@ -604,70 +604,155 @@ Verified alerts include the original fields plus an extended `info` section cont
...
+| `event_loop` | One coroutine per message on a persistent event loop, with asynchronous VLM, VIOS, and Elasticsearch clients | `alert_agent.async_io.max_vlm_concurrent` | Shipped default. Kafka consumption is decoupled from VLM latency |
 
-   - Retrieves video URL from VIOS for the alert time window
-   - Renders the prompt with alert context
-   - Passes video URL, prompts and other required inputs to VLM backend for verification
-   - Publishes the verified result to configured sinks
+An invalid value fails startup. When the key is unset, the mode is derived from the deprecated `alert_agent.async_io.enabled` flag; set `pipeline_mode` explicitly instead.
 
-Workers operate independently, allowing multiple alerts to be processed in parallel. The main thread blocks when all workers are busy, providing natural backpressure. The end-to-end latency of external services (VLM, VIOS) directly bounds throughput per worker.
+The shipped Alerts developer profile uses `event_loop`. In that mode `num_workers` is not a throughput control, and raising it adds neither VLM concurrency nor CPU capacity.
```
dineshvenkat, line 621: The configuration table later on this page still presents `num_workers` as a general concurrency control and lists the retired `chunk_size` setting as active, which contradicts this mode-specific guidance.

docs/alert-verification-service.mdx
```diff
@@ -604,70 +604,155 @@ Verified alerts include the original fields plus an extended `info` section cont
...
+
+| Metric | Interpretation |
+| --- | --- |
+| `alert_bridge_pipeline_processes_configured` | Requested pipeline-process count. |
+| `alert_bridge_pipeline_processes_alive` | Processes currently running. Should equal the configured count. |
+| `alert_bridge_pipeline_processes_ready` | Processes with a decided Kafka assignment. |
+| `alert_bridge_assigned_partitions` | Partitions currently assigned to the instance. A value below the process count means some processes hold nothing. |
+| `alert_bridge_pipeline_process_exits_total{reason}` | Expected versus unexpected process exits. |
+| `alert_bridge_rebalance_drains_total{outcome}` | Whether revoked-partition work drained before the deadline. |
+| `alert_bridge_records_read_after_revoke_total` | Records observed after partition ownership changed. Investigate sustained growth. |
```
dineshvenkat, line 753: The reason and outcome metric labels are referenced without listing their possible values, making the metrics difficult to interpret.

docs/alert-verification-service.mdx
```diff
@@ -604,70 +604,155 @@ Verified alerts include the original fields plus an extended `info` section cont
...
 ```
 
-<Note>
-For Prometheus metric names, labels, and scrape configuration, see [Metrics](#metrics). In addition, monitor worker utilization (all workers busy indicates a need to scale up), Kafka consumer lag, and host CPU and memory.
-</Note>
+Raising only `processes` multiplies all three. Keep the instance VLM concurrency at or below the backend's measured ceiling, and size the in-flight bound at two to four times the effective VLM concurrency. Startup logs the resolved instance totals. Read them from the log to confirm what the backends are offered.
+
+#### Startup, readiness, and shutdown
+
+- `startup_timeout_seconds` covers topic metadata, prompt seeding, VLM warmup, and every process joining its consumer group. The default is 60 seconds and the minimum is 30. The startup watchdog stops the instance 5 seconds after the budget, so size a startup probe against 65 seconds at the default.
```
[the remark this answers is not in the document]
danh-nguyen-nvidia: Partly right. The 5s extension is applied unconditionally in the multi-process branch (deadline.extend(WATCHDOG_MARGIN_SECONDS) before run_multi_process_pipeline), so 65s holds for processes > 1. At one process there is no extension and the watchdog fires at the budget. Reworded to state both cases.

docs/alert-verification-service.mdx
danh-nguyen-nvidia: Fixed in 19e8cad. num_workers is now scoped to sync, the chunk_size row is removed, and the table gains pipeline_mode, processes, startup_timeout_seconds, and the async concurrency settings, each linking back to the scaling section.

docs/alert-verification-service.mdx
danh-nguyen-nvidia: Added in 19e8cad. reason is shutdown or unexpected; outcome is drained or timed_out. Listed in the monitoring table and also in the Metrics section, which did not cover the multi-process series before.

docs/alert-verification-service.mdx
```diff
@@ -766,6 +851,13 @@ Known `mode` values are `sync`, `async`, `sync_fallback`, and `async_submit`. Kn
 
 | Metric | Labels | Meaning |
 | --- | --- | --- |
+| `alert_bridge_pipeline_processes_configured` | (none) | Pipeline processes requested by `alert_agent.processes`. |
+| `alert_bridge_pipeline_processes_alive` | (none) | Pipeline processes currently running. |
+| `alert_bridge_pipeline_processes_ready` | (none) | Pipeline processes holding a decided Kafka assignment. |
+| `alert_bridge_assigned_partitions` | (none) | Source partitions currently assigned to the instance, summed across its processes. |
+| `alert_bridge_pipeline_process_exits_total` | `reason` | Pipeline process exits, labelled `shutdown` or `unexpected`. |
+| `alert_bridge_rebalance_drains_total` | `outcome` | Drains of revoked partitions, labelled `drained` or `timed_out`. |
```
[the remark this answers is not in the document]
danh-nguyen-nvidia: Correct, nothing_owed is emitted when the revoked partitions had no work in flight. Added to both tables in 629ce79.

dineshvenkat (dismissed):

docs/agent-workflow-alert-verification.mdx
```diff
@@ -670,6 +670,45 @@ If you would like to modify the workflow to work with other videos or usecases,
 
   * Drop a Python module exposing a parser class with a `parse(self, raw_response: str) -> dict` method, then set `vlm.response_parser` in `config.yml` to its dotted path. Use this to replace the built-in YES/NO verification parsing with classification, analytics, or enrichment outputs. See [Pluggable Response Parser](/vss/system-components/analytics-microservices/alerts-microservice#pluggable-response-parser) for the full contract, examples, and error handling.
 
+#### Scale Alert processing across CPU cores
+
+A single Alerts Microservice process uses about one CPU core. Running the pipeline in several processes adds CPU capacity, and it helps when that process is consistently near one core while Kafka consumer lag keeps growing. A process below one core is limited by offered traffic, by the configured VLM concurrency, or by a downstream service, and gains nothing from additional processes.
```
bpisupati-nv, line 675: Besides cores, documentation should capture other per-process limits (file descriptors, open http connections etc)

docs/agent-workflow-alert-verification.mdx
```diff
@@ -670,6 +670,45 @@ If you would like to modify the workflow to work with other videos or usecases,
...
+
+```yaml
+alert_agent:
+  pipeline_mode: event_loop
+  processes: 4
+  startup_timeout_seconds: 60
+  async_dispatch_max_in_flight: 4
+  async_io:
+    max_vlm_concurrent: 1
+    max_vst_concurrent: 2
```
bpisupati-nv, line 693: Do we really want the user to be tweaking max_vlm_concurrent?
If vlm is the bottleneck, we should advice them to add more replicas.

docs/agent-workflow-alert-verification.mdx
```diff
@@ -670,6 +670,45 @@ If you would like to modify the workflow to work with other videos or usecases,
...
+persistence:
+  enabled: true
+```
+
+Every per-process limit is multiplied by the process count, which is why the three limits above are divided by four from the one-process profile values of 16, 4, and 8.
+
+Before raising the process count, confirm the following:
+
+- `pipeline_mode` is `event_loop` and `persistence.enabled` is `true`. Both are the profile defaults.
+- Every Alert input topic has at least as many partitions as `replicas × processes`, counted per topic. The profile creates eight partitions per topic, so one replica can run four processes as shipped.
```
bpisupati-nv: Need to mention about streams being greater than number of partitions as well since message per stream lands on the same paritition.

bpisupati-nv: What is replicas?

docs/alert-verification-service.mdx
```diff
@@ -604,70 +604,155 @@ Verified alerts include the original fields plus an extended `info` section cont
 
 ## Scaling and Performance Tuning
 
-The Alerts Microservice supports concurrent processing to maximize throughput. This section describes the key configuration parameters for scaling the service to meet your deployment requirements.
+The Alerts Microservice scales along two independent dimensions: how many CPU cores one instance can use, and how many requests it keeps in flight against the VLM and VIOS backends. This section describes the pipeline modes that govern in-process concurrency, the settings that control each dimension, and how to tell which one is limiting a deployment before changing anything.
```
bpisupati-nv: It is not number of cores but processes. Need to correct wording accordingly.

docs/alert-verification-service.mdx
```diff
@@ -604,70 +604,155 @@ Verified alerts include the original fields plus an extended `info` section cont
...
+
+The following settings control concurrency and CPU use:
+
+| Setting | Default | Applies To | Meaning |
+| --- | --- | --- | --- |
+| `alert_agent.processes` | `1` | All modes | Independent pipeline processes. The only setting that lets one instance run Python work on more than one CPU core. |
+| `alert_agent.pipeline_mode` | `event_loop` in the shipped profile | — | Dispatch model, as described above. |
+| `alert_agent.async_io.max_vlm_concurrent` | `4` in the shipped profile | `event_loop` | VLM requests in flight per process. Keep at or below the VLM backend's measured ceiling. |
+| `alert_agent.async_io.max_vst_concurrent` | `8` in the shipped profile | `event_loop` | VIOS requests in flight per process. |
+| `alert_agent.async_dispatch_max_in_flight` | `16` in the shipped profile | `thread_bridge`, `event_loop` | Post-deduplication messages accepted but not finished, per process. Bounds memory and applies backpressure to the Kafka consumer. It does not raise the throughput ceiling. |
```
bpisupati-nv: why do we still have so many config parameters? For eg: why is there processes and num_workers settings?

Do we really need async_dispatch_max_in_flight setting?

bpisupati-nv (commented):
pls review comments

docs/agent-workflow-alert-verification.mdx
danh-nguyen-nvidia: Added in dd8f11f. Each process holds its own Kafka consumers, HTTP connection pools to the VLM and VIOS backends, Elasticsearch clients, and memory, so connection, file-descriptor, and memory use grow with the process count. Stated on both pages next to the CPU requirement. The code sets no explicit pool limits, so there are no fixed numbers to quote.

docs/agent-workflow-alert-verification.mdx
danh-nguyen-nvidia: Agreed. Reworded in dd8f11f: max_vlm_concurrent is a cap sized to what the backend serves concurrently, not a throughput lever, and when the VLM is the limiting service capacity comes from the VLM side, for example more replicas behind the same endpoint. Dropped the sentence that suggested raising the per-process caps afterwards. The example keeps the divided values so enabling processes does not multiply the load on the same VLM.

docs/agent-workflow-alert-verification.mdx
danh-nguyen-nvidia: Added in dd8f11f. Every message of one stream carries the same sensorId and lands on one partition, so work is spread across processes at stream granularity: a process receives work only when there are at least as many active streams as processes, and the spread evens out as the stream count grows. Stated as a prerequisite on the workflow page and as a rule on the service page.

Replicas are Alerts Microservice instances sharing the consumer group, not VLM replicas. Both pages now define it before using it: group members are the pipeline processes of every instance, so the bound is replicas × processes per topic. The workflow page states the single-instance bound first, since the profile runs one container.

docs/alert-verification-service.mdx
danh-nguyen-nvidia: Fixed in dd8f11f. The dimension is now described as how many pipeline processes one instance runs, each using about one CPU core.

docs/alert-verification-service.mdx
danh-nguyen-nvidia: num_workers belongs to sync mode only and is kept for the rollback path; processes is a different axis, CPU. async_dispatch_max_in_flight exists because offsets are committed on read: it bounds memory and how many read-but-unfinished messages are lost if a process exits, which max_vlm_concurrent alone does not. In dd8f11f the table is trimmed to the settings in play for event_loop plus that rollback path, and the two descriptions now say why they exist. Reducing the actual config surface would be a code change; happy to open that separately.

docs/alert-verification-service.mdx
```diff
@@ -604,70 +604,155 @@ Verified alerts include the original fields plus an extended `info` section cont
...
+
+### Scaling Controls
+
+The following settings control concurrency and CPU use:
+
+| Setting | Default | Applies To | Meaning |
+| --- | --- | --- | --- |
+| `alert_agent.processes` | `1` | All modes | Independent pipeline processes. The only setting that lets one instance run Python work on more than one CPU core. |
+| `alert_agent.pipeline_mode` | `event_loop` in the shipped profile | — | Dispatch model, as described above. |
+| `alert_agent.async_io.max_vlm_concurrent` | `4` in the shipped profile | `event_loop` | VLM requests in flight per process. Keep at or below the VLM backend's measured ceiling. |
```
bpisupati-nv: Are these configs meant to be exposed to the user or meant to be internal.
If vlm or vst cannot handle rate at which messages are being received then the message count in kafka will keep growing making the system keep falling behind. And so value of exposing this metric is limited.

Instead, I suggest providing guidance for monitoring the input workload versus processing rate in each of the stages and recommend user scale them. Otherwise, alert bridge should support dropping of messages based on user define watermark levels.

For now, consider how to simplify the config parameters.

bpisupati-nv: To that extent, num_processes should be the only knob exposed to the user?

bpisupati-nv (commented):
Pls check comment updates.

docs/alert-verification-service.mdx
danh-nguyen-nvidia: Reworked in 403222a. processes is now the only setting the section asks the user to change. The VLM/VIOS/in-flight caps moved to a "Profile defaults" table at the end: shipped values, what each one is for, and that every per-process cap is multiplied by the process count, so the example divides them to keep backend load unchanged. The sizing formula and the benchmark procedure are gone. Added a "Monitoring throughput" part: input rate vs processed rate, and which wait metric grows when CPU, VLM, or VIOS is the limit, with where capacity comes from for each.

On dropping by lag or watermark: that does not exist today. The current drops are dedup TTL, the end-time-delta filter, and the per-sensor rate limiter, all content-based. That would be a feature request outside this PR.

On num_processes as the only knob: in the doc, yes. In the code, not yet. The image default pipeline_mode still derives to sync, so setting only processes on the image defaults fails startup; the dev profile works because it sets event_loop and the caps. Making it the only knob for real means changing the image defaults, which I can open separately.

docs/agent-workflow-alert-verification.mdx
```diff
@@ -670,6 +670,46 @@ If you would like to modify the workflow to work with other videos or usecases,
...
+
+Before raising the process count, confirm the following:
+
+- `pipeline_mode` is `event_loop` and `persistence.enabled` is `true`. Both are the profile defaults.
+- Every Alert input topic has at least as many partitions as the process count, counted per topic. The profile creates eight partitions per topic, so one instance can run four processes as shipped. When several Alerts Microservice instances share the consumer group, for example as replicas on Kubernetes, the bound applies to `instances × processes`.
+- At least as many active streams as processes. Every message of one stream carries the same `sensorId` and lands on one partition, so work is spread across processes at stream granularity. Fewer streams than processes leaves processes without work, and more streams spread the load more evenly.
```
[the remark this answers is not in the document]
danh-nguyen-nvidia: Correct. Only the Behavior Analytics producer keys mdx-incidents by sensorId; the HTTP submission API keys by event or incident id first. Qualified in 403222a on both pages, including the dedup note under Scaling Across Instances.

bpisupati-nv (approved):

trongp-nvidia (approved):

danh-nguyen-nvidia:
/ok to test [403222a](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/2072/commits/403222aeca3dfce946523923b6c7bcb291b0c966)

PR #2074, opened 2026-09-07
perf(ci): shared GHCR registry buildcache for the VIOS C++ images

gous-75:
## Problem
The three C++ VIOS images (sensor/streamprocessing/nvstreamer) build grpc + the toolchain from source (~11.8 min for grpc). The GHA layer cache (`type=gha`, 10 GB per-repo LRU) evicts that large layer between builds, so grpc keeps cache-missing and every build pays the full rebuild.

## Change (2 files, no Dockerfile changes)
- **`container-inventory.json`**: `registry_buildcache` carries a shared cache **name** (`vss-vios-buildcache`) on the three C++ images. They share the same Dockerfile and differ only by `MODULE`, so their base+grpc layers are identical — one **per-arch** cache stores them once.
- **`build-dev-images.yml`**: a per-image `Select buildcache backend` step resolves the ref to `<registry>/<name>:<arch>` (`type=registry`, a **private** package). Every other image keeps the GHA cache unchanged.
  - cache-from import failure is non-fatal (buildkit builds from scratch)
  - cache-to uses `ignore-error=true` so an export hiccup never fails the build

Ingress keeps the GHA cache (`Dockerfile.ingress`, no grpc, ~1.3 min build).

## Result (validated on the predecessor PR)
grpc imported as **CACHED** for all three from the single shared cache, cutting each build ~2×:

| image | cold | cached |
|---|---|---|
| sensor | 21.5 m | **8.5 m** |
| streamprocessing | 23.5 m | **11.4 m** |
| nvstreamer | 22.8 m | **13.9 m** |

Guards green: golden, release-set closure, detect/helm suites.

- Delete the orphaned `vss-vios-*-buildcache` packages from the earlier per-image iteration.
- A prebuilt toolchain image would keep grpc cached even when the builder apt packages change (the one thing a registry cache can't isolate).


gous-75:
Fixed in 9e3141483's follow-up — the block comment now describes `registry_buildcache` as a cache **package name** (matching the jq selector and the inventory), so it can't be mistaken for a boolean.

kaushikc-nvidia (approved):

PR #2078, opened 2026-09-07
fix(agent): ground sensor answers and surface VST tool errors

hugoverjus:
## Summary
- Ground sensor-list answers and relay VST tool errors instead of inventing cameras
- Preserve terminal subagent outcomes; omit stringified `null` tool arguments
- Default snapshots to the recorded timeline and validate ISO snapshot bounds
- Fail startup on relative LLM endpoints, preserve compose endpoint defaults, and expose the in-deployment `/llm` gateway
- Route explicit long-video-summarization/LVS report requests through `lvs_video_understanding` before presenting the report
- Support noninteractive LVS report execution in the HTTP profile and bind planner tools while preserving the numbered-plan parser
- Also includes the existing `fix(lvs): keep bundled DCGM exporter opt-in` branch commit

## What this PR still is / is not
This is a product-side agent grounding and LVS-report-routing fix. It is not the LVS HITL/GB300 CI-isolation change (that work is already on ci-vss-oss !515), does not hardcode LLM tool-calling flags, and does not broaden the planner DSL beyond accepting the existing text plan from a tool-bound model.

## Validation
- [x] 60 `top_agent` unit tests passed for the LVS route/guard before the follow-up commits
- [x] Focused final-head tests: `test_top_agent.py` + `test_lvs_video_understanding.py`
- [x] Ruff format/lint and mypy on the changed agent modules
- [x] Live 10.63.177.97 smoke for `gwfix6`: `lvs_video_understanding` returned `LVSStatus.SUCCESS` and the streamed response presented the LVS report
- [x] Recreated only `vss-agent`; no `compose down -v`

## Deliberately left out
- ci-vss-oss !515 owns adopt-and-skip LVS HITL isolation and GB300 cleanup
- Infrastructure must configure tool-calling support; this PR does not fake it with product flags
- Unrelated untracked CLI, rt-vlm, skill-eval, and docs work remains outside this PR

### DIFF
--- a/.github/scripts/poll-downstream-pipeline.py
+++ b/.github/scripts/poll-downstream-pipeline.py
@@ -20,9 +20,17 @@ Reporting rules (printed once per job, no duplicates):
 * ``ALLOWED_FAILURE: <job name>`` when a job fails with any other exit
   code while still configured with ``allow_failure: true``.
 * ``FAIL: <job name>`` when any non-``allow_failure`` job reaches status
-  ``failed`` - the script exits 1 immediately.
-* ``CANCELED: <job name>`` when a job is canceled - the script exits 1
-  immediately.
+  ``failed``.
+* ``CANCELED: <job name>`` when a job is canceled.
+
+``FAIL`` / ``CANCELED`` are announcements, not verdicts: polling
+continues until the pipeline itself reaches a terminal state, and the
+exit status is then decided from that final snapshot. A single job
+failing no longer discards the result of every job still running
+alongside it, and because the verdict is re-read at the end, a job that
+failed and was retried into a pass is not held against the pipeline.
+The check still goes red for any job that is genuinely failed or
+canceled once the pipeline is done.
 
 Resolving the exit code is non-trivial: many downstream API versions
 do NOT include ``exit_code`` in either the pipeline-jobs listing or
@@ -34,8 +42,8 @@ cached for the lifetime of the poller.
 Exit codes:
 
 * ``0`` - pipeline finished with no failures.
-* ``1`` - a failing / canceled job was observed, or the poller timed
-  out (see ``MAX_POLL_DURATION_SECONDS``).
+* ``1`` - the finished pipeline had a failing / canceled job, or the
+  poller timed out (see ``MAX_POLL_DURATION_SECONDS``).
 
 Retried jobs are handled by de-duping on ``name`` and keeping only
 the latest attempt (highest ``id``).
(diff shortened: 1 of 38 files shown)


hugoverjus:
Added d8d5c2940 to close the remaining warehouse report-routing gap from the nightly backlog.

Incident-mode `report_agent` supports an unfiltered latest-incident report, so the planner now rejects an unnecessary “Which camera?” clarification for that schema and routes directly to `report_agent`. The guard is schema-specific: uploaded-video/LVS report mode still requires a sensor and keeps the clarification.

Proof: `test_top_agent.py` 57 passed; focused Ruff check/format and mypy passed. UI E2E/Playwright specs are not in this repository, so unique sensor naming/adopt-and-skip and GB300 duplicate cleanup remain in ci-vss-oss !515 commit 11. Shared-LLM HTTP 400 and historical RTVI-VLM 503 cases remain infra.

hugoverjus:
Live PDF 503 diagnosis (10.63.177.97): this was deployment skew, not an HAProxy `/static` routing bug.

- HAProxy correctly routes `/static` to `bk_vss_agent`.
- The live LVS config had `functions.lvs_video_understanding.hitl_enabled: false` from `cbe4e3959`, while the running local image was still labeled `3d159797c` and lacked that schema field.
- NAT rejected the config (`Extra inputs are not permitted`), the agent crash-looped, and HAProxy generated `503 No server is available to handle this request` because `bk_vss_agent` had zero healthy servers.
- Recreated the live agent with the schema-compatible source; agent health is green and `/static` now reaches the app (404 for the two historical keys, not HAProxy 503).
- The two historical PDFs were in the replaced container's ephemeral object storage and are no longer present. There is therefore no valid replacement URL for those exact artifacts; a report must be regenerated.

Regression coverage for the noninteractive LVS schema is included in `52dd0a3da7eaade77fa959b9b93c65dfea089554`. Focused validation: 266 tests passed; Ruff and format checks passed.

hugoverjus:
HITL re-enable proof for `5aa25a4b786a466804426958818ce46ecdb8f16b`:

- LVS Docker and Helm profiles both have `hitl_enabled: true` for `lvs_video_understanding` and `video_report_gen`.
- HTTP behavior is request-scoped: with no registered human callback, the agent logs `HITL is enabled but this request has no human prompt callback; proceeding noninteractively ... for this request only` and uses configured defaults. A live `POST /chat/stream` report for `gwfix6` reached `finish_reason=stop` and `[DONE]` with a non-empty LVS report.
- Websocket behavior remains interactive: live `start captioning gwfix6` produced and waited on all four HITL prompts (scenario, events, objects, confirmation). I submitted the defaults through the websocket proof client. LVS then rejected the caption request with `400 No such resource 07846d71-9253-4fdc-bc71-5b102038d325`, so the modal path is proven but this uploaded-video handle is not an LVS live-stream resource.
- Focused rebased-head validation: 89 tests passed (HITL helper, LVS understanding, report generation, and LVS media configuration); Ruff passed; mypy passed before the isolated replay.
- Live deployment: image `vss-agent:error-grounding-local`, revision label `5aa25a4b786a466804426958818ce46ecdb8f16b`, `healthy`, restart count `0`. Only `vss-agent` was force-recreated; no volume teardown or proof-process cleanup was performed.

kaushikc-nvidia (approved):

PR #2080, opened 2026-09-07
fix(rtvi-vlm): [RTVI-VLM][3.3.0][GHCR]: List live stream API is failing when CV Stream is added to the server with id not being UUID

nv-amkale:
## Summary

Changed LiveStreamInfo.id from UUID-only validation to the bounded string identifier supported by the asset manager, and added a regression test for a CV stream using camera-01. The rebuilt runtime passed the recorded CV-add and legacy-list reproduction.

## Plan

Reproduce NVBug 6735416, apply the minimal scoped fix, and validate it against the recorded reproduction and configured tests.

## Related Issue

NVBug 6735416

## Changes

- Changed LiveStreamInfo.id from UUID-only validation to the bounded string identifier supported by the asset manager, and added a regression test for a CV stream using camera-01. The rebuilt runtime passed the recorded CV-add and legacy-list reproduction.

## Validation

- Baseline reproduction and fixed-worktree validation: PASS

### DIFF
--- a/services/rtvi/rt-vlm/src/api_models/live_stream.py
+++ b/services/rtvi/rt-vlm/src/api_models/live_stream.py
@@ -235,7 +235,11 @@ class DeleteLiveStreamsResponse(CommonBaseModel):
 class LiveStreamInfo(CommonBaseModel):
     """Live Stream Information."""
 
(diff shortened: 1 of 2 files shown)


nv-pnikam (approved):
LGTM

PR #2082, opened 2026-09-08
Feat/search eval skill

abhishek-ak-nv:
## Search eval as a skill

Adds `skills/benchmarking/benchmark-video-search` — measures retrieval quality
and latency of a deployed search profile. It drives the `vss` CLI, the same path
the product uses now that the agent adapter shells out to `vss search run <mode>
--raw`, rather than the REST endpoint being retired.

One command ingests a labelled dataset, decomposes each query live, runs it down
whichever path the decomposition chooses, and reports precision / recall / mAP /
HIT@k with a per-stage latency breakdown.

### Also in here, found while building it

- **`vss_core`: per-stage timings on `SearchOutput`.** There was no way to say
  where a query's time went. Now every stage self-reports, with concurrent
  children accounted for so overlapping legs don't double-count.
- **Decomposition prompt: teach it the empty case.** `attributes` was the only
  field with no empty-case clause and none of the nine few-shot examples showed
  `"attributes": []`, so the model filled it with the query's action phrase.
  That routes every action query to `fusion` — measured at 9/9 on the warehouse
  set, costing mAP −25%, HIT@1 halved and +59% latency.
- **Critic: distinguish "ran and returned nothing" from "never ran".** Both
  showed up as `unverified`.

### Testing

102 unit tests, no deployment needed. End-to-end runs against a live search
profile on vad-r1-v2 (121 queries, all three paths, critic active).

Tracking JIRA: https://jirasw.nvidia.com/browse/VIA-2524

## Checklist

### DIFF
--- a/skills/benchmarking/benchmark-video-search/scripts/flows/metrics.py
+++ b/skills/benchmarking/benchmark-video-search/scripts/flows/metrics.py
@@ -1,0 +1,201 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+# http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Retrieval metrics.
+
+Vendored from ``run_eval.py`` so this flow owns its scoring and that script can
+be deleted without taking the metrics with it. The definitions are deliberately
+byte-for-byte equivalent, not "improved" -- changing them would silently
+invalidate every baseline captured with the old runner.
+
+``tests/test_search_eval_flows.py`` asserts this module and ``run_eval.py``
+produce identical output for the same input, for as long as both exist. When
+``run_eval.py`` goes, that test goes with it and this becomes the only
+definition.
+"""
+
+from __future__ import annotations
+
+import copy
+from datetime import datetime, timedelta
+from typing import Any
+
+#: Ground truth is annotated on a 5-second grid, so retrieved windows are split
+#: onto the same grid before matching.
+SEGMENT_SIZE = 5
+
+#: k values reported as HIT@k.
+HIT_K_VALUES = [1, 3, 5, 10]
+
+
+def parse_ts(ts: str, ceiling: bool = False) -> datetime:
+    """Parse an ISO timestamp, dropping sub-second precision.
+
+    ``ceiling`` rounds a fractional second up, so a window ending at
+    00:00:13.2 covers the segment it partially overlaps rather than losing it.
+    """
+    ts_clean = ts.replace("Z", "").replace("+00:00", "")
+    dt = datetime.fromisoformat(ts_clean)
+    if ceiling and dt.microsecond > 0:
+        dt = dt + timedelta(seconds=1)
+    return dt.replace(microsecond=0)
+
+
+def align_ts_to_segment(start: datetime, end: datetime) -> list[tuple[datetime, datetime]]:
+    """Split ``[start, end)`` onto the 5-second grid the ground truth uses."""
+    if start >= end:
+        return []
+    start = start.replace(microsecond=0)
+    end = end.replace(microsecond=0)
+    sec = start.second
+    floored = (sec // SEGMENT_SIZE) * SEGMENT_SIZE
+    seg_start = start.replace(second=floored, microsecond=0)
+    segments = []
+    while seg_start < end:
+        seg_end = seg_start + timedelta(seconds=SEGMENT_SIZE)
+        segments.append((seg_start, seg_end))
+        seg_start = seg_end
+    return segments
+
+
+def video_name_matches(api_name: str, gt_name: str) -> bool:
+    """Prefix match, because VST renames uploads.
+
+    Ground truth says ``warehouse_sample``; search returns
+    ``warehouse_sample_20250101_000000_e0482.mp4``.
+    """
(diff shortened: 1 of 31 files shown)


ayyappa-dev:
Please add DCO sign-off to the commits

skills/benchmarking/benchmark-video-search/scripts/flows/metrics.py
```diff
@@ -0,0 +1,219 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Retrieval metrics.
+
+Vendored from ``run_eval.py`` so this flow owns its scoring and that script can
+be deleted without taking the metrics with it. The definitions are deliberately
+byte-for-byte equivalent, not "improved" -- changing them would silently
```
ayyappa-dev: Issue: This docstring promises the metric definitions are "byte-for-byte equivalent" to the old run_eval.py and that "changing them would silently invalidate every baseline." But commit 1c9922817 deliberately changed match_segment (equal-start → overlap). So the docstring now claims a contract the code no longer holds. A reader who trusts it will compare a post-merge physicalai-dev number against a pre-merge baseline and believe they're comparable.
 Probable fix: Rewrite the docstring to state the intentional divergence: matching moved to half-open overlap on this commit; baselines on event-bounded datasets (physicalai-dev) captured before it are NOT comparable; chunk-dataset baselines (vad-r1-v2) are unaffected, since overlap and equality accept the same set there.

skills/benchmarking/benchmark-video-search/scripts/run_eval_flows.py
```diff
@@ -0,0 +1,1285 @@
...
+        decompose_s = 0.0
+        if decomposer is not None:
+            decomposition, decompose_s = decomposer.decompose(query, video_sources=sorted(video_names))
+            query_backend.decompositions[query] = decomposition
+
+        raw_results, latency_s = query_backend.search(query)
+        normalized = flows.normalize_results(raw_results)
+        sources_seen.update(flows.verification_sources(normalized))
+        for verdict, count in flows.verdict_counts(normalized).items():
+            verdicts[verdict] = verdicts.get(verdict, 0) + count
```
ayyappa-dev: this updates a shared counter from multiple threads without a lock, so at --concurrency > 1 counts get silently lost — and those counts are exactly what the no_opinion safety check reads, so a healthy critic can be misreported as broken.

Can we move this update into print_lock, which is already defined above. for example,
 ```python
   with print_lock:
       sources_seen.update(flows.verification_sources(normalized))
       for verdict, count in flows.verdict_counts(normalized).items():
           verdicts[verdict] = verdicts.get(verdict, 0) + count
 ```

skills/benchmarking/benchmark-video-search/scripts/run_eval_flows.py
```diff
@@ -0,0 +1,1509 @@
...
+                raise SystemExit(
+                    f"ABORTED: {len(readiness['missing'])} source(s) never registered within "
+                    f"{args.readiness_timeout}s: {readiness['missing']}\n"
+                    "  Querying now would score an indexing delay as a retrieval regression."
+                )
+
+        # Registered in VST is not the same as indexed in Elasticsearch. The
+        # three-step flow already proved the difference with a chunk count; a
+        # flow that reports no proof has to be asked.
+        if not ingest_backend.proves_indexing and not args.skip_index_probe:
```
ayyappa-dev, line 1624:
```
   verify_anchor (flows/ingest.py:381) is defined, unit-tested, and never called in the run — and SKILL.md doesn't tell the user to run it. It's the one check
 that catches a wrong ingest anchor, which its own docstring says "would score every query zero and read as a retrieval collapse." The new default vst-direct
 ingest is fire-and-forget, and the index probe here at line 1456 only confirms a video is indexed — not that it's indexed at the 2025-01-01 anchor the ground
 truth expects. If VIOS doesn't honor the upload's metadata.timestamp, every query scores 0 and reads as broken retrieval.

wire one anchor verification into the post-ingest path, alongside the index probe.
 ```

File: skills/benchmarking/benchmark-video-search/scripts/flows/query.py [hunk 4]
```diff
@@ -0,0 +1,326 @@
...
+        start = time.time()
+        try:
+            proc = subprocess.run(
+                argv,
+                capture_output=True,
+                text=True,
+                timeout=self.timeout,
+                cwd=self.cwd,
+            )
+        except subprocess.TimeoutExpired:
```
ayyappa-dev:
```
   A timed-out query scores 0 and the run continues, while a non-zero exit aborts — the opposite of the PR's own rule that environment faults abort rather than
 score 0 (run_eval_flows.py docstring). A 300s hang is an environment fault; scoring it 0 silently corrupts the aggregate, and the saved JSON has no marker that
 the query was a timeout, so it reads as "no recall."

treat a timeout like a non-zero exit (abort), or exclude the query from aggregation — don't score it 0.
 ```

File: skills/benchmarking/benchmark-video-search/scripts/run_eval_flows.py [hunk 5]
```diff
@@ -0,0 +1,1509 @@
...
+    # metrics come out numerically identical to raw, which reads as "the critic
+    # agreed with retrieval" when it never rendered an opinion at all.
+    #
+    # Observed in practice: a CLI configured with a loopback base_url hands
+    # RT-VLM a `localhost` clip link, its SSRF guard returns 422, and all 121
+    # queries came back unverified at exit 0.
+    real_sources = sources_seen - {flows.VERIFICATION_ABSENT}
+    verdicts = verdicts or {}
+    opinions = verdicts.get("confirmed", 0) + verdicts.get("rejected", 0)
+    if real_sources and not opinions:
```
ayyappa-dev, line 901:
```
   The no_opinion detection at line 787 (real_sources and not opinions -> status "no_opinion", suppressing critic_filtered) and the verdict_counts it reads have
 no tests. This is the PR's headline safety feature — the one that catches a critic that ran but formed no opinion — and the verdict tally that feeds it was just
 fixed for a race, so a regression here would be invisible.

add tests for the three cases — all-unverified with real sources -> status "no_opinion" and critic_filtered suppressed; some confirmed -> status
 "ok"; no verification blocks -> critic key absent.
 ```

PR #2095, opened 2026-09-08
fix(rtvi-vlm): accept string IDs in streams/add

nv-amkale:
## Summary
- accept bounded string identifiers in `POST /v1/streams/add` requests and responses
- cover a `camera-01` add, list, duplicate rejection, and delete lifecycle
- follow up #2080, which fixed listing existing non-UUID stream IDs

## Validation
- Pydantic request/response schema validation on macOS
- Python compilation for both changed files
- `git diff --check`

The focused endpoint test is included but was not executed locally because the macOS environment does not provide `cuda.bindings`; Linux/CUDA CI should execute it.

NVBUG 6735416

### DIFF
--- a/services/rtvi/rt-vlm/src/api_models/live_stream.py
+++ b/services/rtvi/rt-vlm/src/api_models/live_stream.py
@@ -42,6 +42,7 @@ from .common import (
 )
 
 LIVE_STREAM_URL_PATTERN = r"^rtsp://"
+STREAM_ID_PATTERN = r"^[^/\r\n]+$"
 # CV-compatible URL pattern: accepts rtsp://, file://, http://, https://.
 # Empty VIOS camera_add registration URLs are handled by the VIOS-specific pattern.
 CV_STREAM_URL_PATTERN = r"^(rtsp://|file://|https?://)"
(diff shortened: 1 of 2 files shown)


services/rtvi/rt-vlm/src/api_models/live_stream.py
```diff
@@ -43,6 +42,7 @@
 )
 
 LIVE_STREAM_URL_PATTERN = r"^rtsp://"
+STREAM_ID_PATTERN = r"^[^/\r\n]+$"
```
nv-pnikam: P1 — STREAM_ID_PATTERN permits .., enabling deletion outside the stream cache. In live_stream.py (services/rtvi/rt-vlm/src/api_models/live_stream.py:45), ^[^/\r\n]+$ accepts "..". During cleanup, rt-vlm passes the ID directly to shutil.rmtree as /tmp/rtvi/cached_frames/{stream_id} (rtvi_stream_handler.py (services/rtvi/rt-vlm/src/server/rtvi_stream_handler.py:4001)). A processed stream named ".." therefore resolves to /tmp/rtvi, and deleting it can recursively remove all RTVI temporary data, not just that stream’s frames. Reject dot segments (and preferably use a strict identifier allowlist) before allowing arbitrary string IDs.

nv-pnikam (approved):
LGTM

PR #2096, opened 2026-09-08
Change search+alert to make use of webhook [VIA-2733]

prisrivastav-nv:
## Summary

VIOS fans a registered source out to its perception consumers over webhooks, so
nothing has to call RT-CV, RT-Embed or RT-VLM by hand. Two things had not caught up:
the search profile's `notification_config.json` bundled every receiver into three
lifecycle events, so a build deploying a subset could not select its fan-out without
hand-editing JSON, and the headless guidance still read as a direct-REST recipe.

This makes fan-out a build-time selection — one item per capability per event,
resolved by flipping `enabled` — and shrinks provisioning to registering the source.
Delivery is verified once in CI rather than by every caller.

## Related Issue

[VIA-2733: Change search+alert to make use of webhook](https://jirasw.nvidia.com/browse/VIA-2733)

### Notification config

Three bundled items become twelve, one capability per lifecycle event, each carrying
an `id`. Alert Bridge items ship present but disabled, making the file a superset a
build projects from. VIOS already echoes each `id` in its log lines and in the
delivered `webhook_id`, so no consumer changes.

### Build-time resolution

`services/vios.md` adds the receivership test — a service belongs in the config only
if it must act on every newly registered stream unasked — and the rule that the
enabled set equals the build's stream-driven consumer set: inherit the profile config,
project to `patches/notification_config.json`, or take the webhooks-disabled default.
A projection flips `enabled` and nothing else. `composition.md` generalizes `patches/`
to hold payloads an environment variable selects.

### Headless provisioning

`provision-vios-source.md` ends at a registered source. `vss vios add` uploads, pins
the upload anchor and returns only once VIOS has indexed the recording; `vss vios
delete` tears down. Coverage resolution, per-item post-checks and the direct-REST
fallback are gone. It keeps what registration alone does not answer: the agent-tier
guard, the SDRC double-provision rule, the shared-id and upload-date rules, two known
cleanup gaps, the RT-VLM legs no receiver runs, and the logs to read when a consumer
genuinely never got the source.

### Corrected guidance

Critique and visual Q&A are reachable from the CLI against RT-VLM, so they no longer
read as agent-tier features needing a redeploy. Provisioning and query are gated on an
explicit request. Tagging is requested-only, never implied by deploying RT-VLM.

### Tests

`check_notification_supersets.py` guards the superset's item inventory and stock
enable vector. An opt-in runtime check in `test-dev-profile.sh` registers a source and
asserts `camera_streaming` adds it to RT-CV and `camera_remove` withdraws it, failing
closed when RT-CV cannot be read rather than reporting a withdrawal.

## Checklist


skills/operations/vss-manage-video-io-storage/references/provision-vios-source.md
```diff
@@ -165,10 +198,211 @@ the raw NvStreamer/camera URL — that bypasses the VIOS-managed handle and is t
...
+| `rtvi-embed-camera-streaming` | `GET ${RTVI_EMBED_URL}/v1/streams/get-stream-info`; embeddings in `mdx-embed` → `mdx-embed-filtered-*` |
+| `rtvi-vlm-tagging-camera-streaming` | `GET ${RTVI_VLM_URL}/v1/streams/get-stream-info`; captions on `mdx-vlm-captions`, which Logstash writes to `default_<streamId>` (there is no `mdx-vlm-tags` or `mdx-vlm-captions-*` index). Incidents (`mdx-vlm-incidents-*`) appear only when the model emits a trigger token — not a valid liveness check for tagging |
+| An RT-VLM `stream/add` item with no prompt (the LVS profile's config) | `GET ${RTVI_VLM_URL}/v1/stream/get-stream-info` lists the `camera_id` with `inference_active: false`. RT-VLM starts inference only when the add carries a prompt, so registration is the whole contract — do not poll for captions or index writes; nothing produces them until a caller drives one (`vss-build-vision-ai` `references/services/rt-vlm.md`) |
+| `alert-bridge-camera-streaming` | `GET ${ALERT_BRIDGE_URL}/api/v1/realtime/incidents` scoped to the camera; the bridge log shows the always-on POST (see `vss-manage-alerts` `always-on.md`). A repeat delivery returns `STREAM_ADD_ALREADY_ACTIVE` — success, not failure |
```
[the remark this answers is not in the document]
prisrivastav-nv: addressed

ayyappa-dev (dismissed):
LGTM. Can merge after the CIs pass.

skills/operations/vss-manage-video-io-storage/references/provision-vios-source.md
```diff
@@ -36,106 +39,117 @@ capability — search ingestion to `vss-search-archive` (`/api/v1/videos` +
...
-`camera_url` (Step 2); RT-VLM takes it directly — no pre-upload — or registers it via
-`/v1/files`. There is no live proxy on this path.
+The equivalent raw call is
+`PUT /vst/api/v1/storage/file/<filename>?timestamp=…` without the wait
+([`api-reference.md`](api-reference.md) § 8). A bare upload stores bytes only —
+no detections or embeddings.
+
+There is no live proxy on this path. Where a caller drives RT-VLM by hand, the
+media URL is `vss vios clip --sensor <name>` → `media_url`, which the CLI
+re-anchors on the deployment origin so it resolves from inside the consumer's
```
zac-wang-nv: **P1** — Please retain the requirement that the recorded deployment origin be reachable from RT-VLM before passing this handle to the manual RT-VLM calls below. With the documented `vss configure --base-url http://localhost:7777` setup, `vss vios clip` returns `http://localhost:7777/vst/...`: `normalise_media_url()` preserves the configured origin, so the VLM container interprets localhost as itself and cannot fetch the uploaded clip. The previous recipe explicitly required a consumer-reachable upload URL. Require a suitable origin when configuring this path, or use a transfer path that handles loopback; re-anchoring alone does not establish container reachability.

deploy/docker/test-scripts/test-dev-profile.sh
```diff
@@ -2707,6 +2739,116 @@ else
...
+      ((TESTS_FAILED++)) || true
+    else
+      # Returns 0 once RT-CV's stream list matches the wanted presence state.
+      # A failed fetch is not an answer: absence counts only when RT-CV
+      # responded, so an unreachable receiver times out rather than reporting
+      # the stream withdrawn.
+      function await_rtvi_cv_stream() {
+        local _want_present="${1}" _deadline=$(( $(date +%s) + _wh_bound )) _streams
+        while :; do
+          if _streams="$(curl -sf "${VSS_TEST_RTVI_CV_URL%/}/api/v1/stream/get-stream-info" 2>/dev/null)"; then
```
zac-wang-nv: **P2** — Please bound the HTTP requests as well as the polling loop. This curl has neither --connect-timeout nor --max-time, and the deadline is only checked after curl returns. A receiver that accepts the connection but never responds therefore hangs the opt-in test indefinitely instead of reporting failure after _wh_bound; the VIOS delete below is never reached. The sensor add/delete curls have the same unbounded wait. Derive request limits from the remaining deadline (and bound the final sleep) so the test actually terminates within its advertised budget.

skills/operations/vss-manage-video-io-storage/references/provision-vios-source.md
prisrivastav-nv: addressed

deploy/docker/test-scripts/test-dev-profile.sh
prisrivastav-nv: addressed

zac-wang-nv (approved):
lgtm

PR #2101, opened 2026-09-08
refactor(libs): move the CLI library to libs/vss, out of the agent

zac-wang-nv:
## What

Moves the NAT-free CLI library out of the NAT service.

```
libs/vss/
├── pyproject.toml      # virtual uv workspace root — owns the lock, tooling, test config
├── core/               # nvidia-vss-core   (was services/agent/packages/vss_core)
└── cli/                # nvidia-vss-cli    (was services/agent/packages/vss_cli)

services/agent/         # NAT agent only; consumes the library via [tool.uv.sources] path entries
└── packages/vss_agents/
```

The dependency now runs one way — the agent uses the library, never the reverse.

## Why

`vss_core` and `vss_cli` lived inside the NAT service's workspace, so the library that exists
to be NAT-free could only be installed through it. Every caller paid for that: `--extra cli`
to install a CLI that wasn't in the base, and `--no-dev` to keep the agent stack out.

```bash
# before
uv run --project "${VSS_REPO_ROOT}/services/agent" --no-dev --extra cli vss
# after
uv run --project "${VSS_REPO_ROOT}/libs/vss" vss
```

NAT-free by construction rather than by remembering two flags — the agent stack is not in that
workspace at all.

## This is a move, not a rewrite

182 renames; **only 7 carry content**, and all seven are path arithmetic:

| file | change |
|---|---|
| `libs/vss/{core,cli}/pyproject.toml` | hatch vcs root `../../../..` → `../../..` |
| `test_package_boundary.py` | `parents[6]` → `parents[7]/services/agent` |
| `test_search_archive_e2e.py` | `parents[6]` → `parents[7]` |
| `test_search_verification_contract.py` | `parents[7]` → `parents[6]` |
| `test_quiet_stderr.py` | sibling `packages/vss_core` → `core` |
| `cli/README.md` | command paths |

Everything that named the old paths moved with it: `ci.yml`, `sonarqube.yml`, the Sonar project
definition (now scans `libs/vss` and excludes its tests from sources), the OSRB first-party index
test, the Helm ingress-route checker, and the docs.

## The agent image

Two defects here were caught by review, not by the build passing:

- uv refuses to normalise a relative path above the base directory, so a flat `/vss-agent` with
  the library at `/libs` fails: `cannot normalize a relative path beyond the base directory`.
  The builder mirrors the repo layout under `/src` instead.
- Building at `/src` and shipping at `/vss-agent` left every console script with an absolute
  shebang into `/src`, which doesn't exist at runtime — **the `vss` entry point was broken in the
  image** while `python -c "import vss_cli"` still worked. Shebangs are rewritten to `/vss-agent`
  in the builder. `uv venv --relocatable` is *not* the fix: it emits a `#!/bin/sh` wrapper and the
  distroless runtime has no shell.

Build context widens from `services` to the repo root, since the lock names `../../libs/vss/*`.

## Verification

| | result |
|---|---|
| library tests (`libs/vss`) | 1368 passed |
| agent tests (`services/agent`) | 2447 passed |
| repo scripts (`.github/`) | 750 passed |
| ruff + mypy | clean, both workspaces |
| wheels | `nvidia-vss-core`, `nvidia-vss-cli` build |
| CLI | `vss --version` from `libs/vss`, no extras; `nat` absent |
| agent image | builds end to end; all four dists installed; `/vss-agent/.venv/bin/vss` runs; `/src` absent from runtime |

## Follow-up (not in this PR)

`deploy/docker/container-inventory.json` keeps `source_path: "services/agent"`, which feeds
`git rev-parse HEAD:<path>` for the content-addressed `tree-<sha>` tag. **A library-only change
will not retrigger the agent image rebuild.** Fixing it means teaching `release_set.py` to hash
several paths, which changes tag derivation and doesn't belong in a move.

## Plan

- **Where does the library live?** → `libs/vss/{core,cli}` with `libs/vss/pyproject.toml` managing
  the packages. Asked and answered: leave the NAT agent in `services/agent` and consume the library
  from `libs/vss`.
- **Branch base?** → Recommended `develop` over the `vios` feature branch; confirmed. This keeps the
  move out of #1765, which is deep in review.
- **Fix the `source_path` rebuild trigger here?** → Recommended deferring; confirmed — listed above
  as follow-up.
- **Independent review** → Run through `/codex-review` as the workspace requires. It found the two
  image defects above (both fixed) and disputed two further claims that testing showed to be wrong
  (`uv sync` at a virtual root *does* install members; `libs/vss/uv.lock` *is* committed).


zac-wang-nv:
Correction: my previous `/ok to test` carried a SHA I did not have in full and partly invented. Ignore it.

/ok to test 4724fdab1e2d32ef18eb0d9a1f846ed1f2daa0fd

deploy/docker/container-inventory.json
```diff
@@ -39,7 +39,7 @@
       "strategy": "build",
       "ghcr_build": true,
       "source_path": "services/agent",
```
maxervo, line 41: OK if we move the CLI library to libs/vss:
```
services/agent/packages/vss_cli
        ↓
libs/vss/cli
``` 
if container-inventory.json describes how each container image is built, then should this vss-agent entry include "libs/vss" as a watched source in `source_path` instead of the old "services/agent", otherwise CLI changes there will not trigger a new agent image build?

maxervo:
```
Code is pushed
    ↓
.github/workflows/build-dev-images.yml starts
    ↓
It runs .github/scripts/detect_changed_images.py
    ↓
That script reads deploy/docker/container-inventory.json
    ↓
It compares changed files with each image’s source_path
    ↓
Matching images are added to the build list
    ↓
Docker builds and publishes those "vss-agent" images
```

.github/workflows/ci.yml
maxervo: Similarly, is this an old path that needs to be updated to the new libs/vss path?

Before the PR, the CLI tests ran from: `services/agent` => created `services/agent/coverage-lib.xml`
After the PR, the CLI tests now run from: `libs/vss` => create `libs/vss/coverage-lib.xml`
So the coverage artifact may not be uploaded if this is not changed

maxervo:
```
Tests create libs/vss/coverage-lib.xml
                    ↓
CI looks in services/agent/coverage-lib.xml
                    ↓
CI cannot find the report
                    ↓
Coverage artifact is not uploaded
```

libs/vss/pyproject.toml
```diff
@@ -0,0 +1,84 @@
...
+# ---------------------------------------------------------------------------
+# The VSS library: `core` and `cli`, NAT-free by construction.
+#
+# A virtual workspace root -- it ships no code of its own, it only resolves the
+# two distributions together so they share one lock and release in lockstep.
+# The NAT agent lives in services/agent and consumes these as path sources, so
+# the dependency runs one way: the agent uses the library, never the reverse.
+# ---------------------------------------------------------------------------
+
+[tool.uv.workspace]
```
maxervo, line 50: This new file `libs/vss/pyproject.toml` created after the refactor, should it also have the same:
```
constraint-dependencies = [
   ...
]
```
as the other pyproject.toml files. So the pinned versions remain consistent when people install, test, or run the standalone libs/vss workspace, and avoid any issues with slightly different dependencies

maxervo:
```
Before
-------
services/agent/
├── pyproject.toml                       # Workspace root for everything
└── packages/
    ├── vss_agents/pyproject.toml        # Agent package
    ├── vss_cli/pyproject.toml           # CLI package
    └── vss_core/pyproject.toml          # Core package


After
-------
services/agent/
├── pyproject.toml                       # Agent workspace only
└── packages/
    └── vss_agents/pyproject.toml        # Agent package

libs/vss/
├── pyproject.toml                       # NEW CLI/core workspace root <= here missing constraint-dependencies
├── cli/pyproject.toml                   # Moved CLI package
└── core/pyproject.toml                  # Moved core package
```

deploy/docker/container-inventory.json
maxervo: Wondering if it requires two of them so "source_paths" being an array of locations ["services/agent", "libs/vss"] instead of a single string or "." that includes everything

maxervo (dismissed):

libs/vss/pyproject.toml
zac-wang-nv: Good catch, and it was not theoretical — thank you.

Added the block to `libs/vss/pyproject.toml` in 0923cc6ae. Re-locking with it in place moved five packages, so the standalone workspace really was resolving outside the security pins:

| package | was | now | why |
|---|---|---|---|
| `langchain` | 1.4.0 | 1.3.14 | the `<1.4` pin for GHSA-gr75-jv2w-4656 |
| `cffi` | 2.1.1 | 2.0.0 | 2.1.0 reports MIT-0, which the repo licence gate does not accept |
| `langchain-core` | 1.6.2 | 1.4.9 | `<1.5`, required by the fixed LangChain line |
| `tokenizers` | 0.23.2 | 0.22.2 | pinned stable |
| `transformers` | 5.16.1 | 5.15.1 | follows the `tokenizers` pin |

I copied the block **verbatim** rather than filtering it to the entries that currently apply. 19 of the 29 bind in this workspace today; the other 10 are inert, because `constraint-dependencies` only bound versions already in the resolution graph and never add a dependency. Filtering would have meant a dependency added to `core` or `cli` later silently loses its pin, and the two files drifting is the failure mode you are pointing at. There is a comment on the block saying it is kept identical to `services/agent/pyproject.toml` for that reason.

1368 tests pass on the new lock, and `.github/osrb/inventory.csv` is regenerated for the five version changes.

.github/workflows/ci.yml
zac-wang-nv: Correct, and fixed in 0aa735905.

The job has `defaults.working-directory: libs/vss`, so pytest writes `libs/vss/coverage-lib.xml` — but `upload-artifact`'s `path` input is not affected by that default and still resolves from the workspace root, so it was looking for a file that no longer exists. The artifact was going missing silently, exactly as you drew it.

`sonarqube.yml` turned out to be already correct on both counts — its lib step has `working-directory: libs/vss` and line 136 sets `sonar.python.coverage.reportPaths=services/agent/coverage.xml,libs/vss/coverage-lib.xml`. Only `ci.yml` was stale.

deploy/docker/container-inventory.json
zac-wang-nv: You are right that the gap is real, and your flow diagram is exactly it: a library-only change now edits `libs/vss/**`, `source_path` still says `services/agent`, so `detect_changed_images.py` does not add `vss-agent` to the build list. This is listed as a known follow-up in the PR body, and I am deliberately **not** fixing it here. Reasoning, so you can push back if you disagree:

`source_path` is not only change detection. It is also the image tag:

```python
result = run_git(repo, "rev-parse", f"{commit}:{source_path}")
tree_sha = result.stdout.strip()
reference = f"ghcr.io/{owner}/vss/{repository}:tree-{tree_sha}{tag_suffix}"
```

So `"source_path": ["services/agent", "libs/vss"]` cannot work by itself — `git rev-parse <commit>:<path>` yields one tree SHA per path, and the published tag is that single SHA. Supporting a list means deciding how to combine several tree hashes into one stable digest, and that changes the tagging and skip-if-published logic for **all 22 entries** that carry `source_path`, not just `vss-agent`.

That is a shared-CI-tooling change with its own blast radius, and it does not belong inside a move that is otherwise a pure relocation. I would rather land it as its own PR where the tag-derivation change can be reviewed on its own terms.

Your `"."` suggestion would work today without touching the script, but it makes every repo change rebuild the agent image, which is the opposite trade.

Interim: the image still rebuilds whenever anything under `services/agent` changes, and `services/agent/pyproject.toml` pins the library by path, so a library change that matters to the agent usually arrives with a lock or manifest change there. That is not airtight, which is why it is a tracked follow-up rather than a shrug.

maxervo (approved):

PR #2103, opened 2026-09-08
vss-manage-alerts: route bounded event counts to the consolidated incidents view

danh-nguyen-nvidia:
Workflow C gains query variant (c) for "how many events / times in a stated period" asks: GET /api/v1/realtime/incidents with consolidate=true, both window bounds (relative periods resolved in UTC), sensor_id when the ask names a sensor and category= when it names a category, the category read off a raw windowed page rather than guessed. Any non-200 or transport failure is a stop (exit 1), never the VIOS-down raw fallback, since a raw list cannot count events; the UUID fallback re-runs (c) under the alternate identity. Lists, unbounded counts, incident-phrased counts and CV deployments stay on the raw view. The prose now says what each view's count/total/truncated mean, that a consolidated total is exact only while truncated is false, that RT-VLM incidents carry the trigger's verdict "confirmed", and reconciles the older routing table, intent precedence, Workflow B, Cross-Skill Links and exit-code legend with the new view.

The real-time eval seeds six confirmed chunks (four fixture_intrusion on warehouse_sample that fold into two events, one same-sensor fixture_other decoy, one other-sensor decoy) with fixed ids after dropping the fixture index, asserts the timestamps are parseable, rejects a consolidated answer to the unbounded count, and adds a windowed event-count step graded on the number with deterministic probes for the contract, the category decoy and the 400 without a window. The routing catalog gains the matching entry.

## Checklist


skills/operations/vss-manage-alerts/SKILL.md
```diff
@@ -517,6 +556,60 @@ curl -sfG "$AB/api/v1/realtime/incidents" \
...
+# copied verbatim. Read the candidates off a raw page of the SAME window and sensor, never guess
+# them from English — a failed lookup is a stop, not "no categories"; if .count < .total page
+# with offset before deciding:
+#   L=(--data-urlencode "start_time=$START" --data-urlencode "end_time=$END" --data-urlencode "limit=1000")
+#   [ -n "$NAME" ] && L+=(--data-urlencode "sensor_id=$NAME")
+#   RAW=$(curl -sfG "$AB/api/v1/realtime/incidents" "${L[@]}") || { echo "category lookup failed (HTTP error or unreachable) — stop"; exit 1; }
+#   printf '%s' "$RAW" | jq -r '.incidents[].category' | sort -u
+# Exactly one match with the user's wording → CATEGORY. Non-empty page but none or several
+# match → ask the user. EMPTY page (no chunks at all in the window) → the answer is 0 events
+# whatever the category: set CATEGORY= and continue (step 3 still applies). No category in the
```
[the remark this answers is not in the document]
danh-nguyen-nvidia: Confirmed — the no-match branch stopped before the UUID identity check (step 3 only fires on a zero). Fixed in ba8776926: when the name-scoped page is non-empty with no matching category, the lookup is repeated under the sensor's VIOS UUID and, on exactly one match, the consolidated call runs under that identity (the scope is now appended to the request after the identity is decided). NAME= asks (no second identity); once on the UUID identity the branch terminates (no re-resolve, no step 3). Routing catalog entry aligned. Not covered by the live eval — the fixture only has the name identity — listed as deferred coverage.

skills/operations/vss-manage-alerts/SKILL.md
```diff
@@ -517,9 +556,74 @@ curl -sfG "$AB/api/v1/realtime/incidents" \
...
+#   [ -n "$NAME" ] && L+=(--data-urlencode "sensor_id=$NAME")
+#   RAW=$(curl -sfG --max-time 30 "$AB/api/v1/realtime/incidents" "${L[@]}") || { echo "category lookup failed (HTTP error or unreachable) — stop"; exit 1; }
+#   printf '%s' "$RAW" | jq -r '.incidents[].category' | sort -u
+# Exactly one match with the user's wording → CATEGORY. Non-empty page but NO match, and NAME
+# is the sensor's NAME (a sensor was named and this is the first identity) → the category may
+# live under the sensor's other identity (a rule created without sensor_name stores its chunks
+# under the VIOS UUID): resolve $UUID exactly as step 3 does and repeat this lookup with
+# sensor_id=$UUID; exactly one match there → NAME=$UUID, CATEGORY=<match>, and say the events
+# matched the UUID identity. Still none, or several matches under either identity → ask the
+# user. NAME= (no sensor named, or the VIOS-down fallback) → there is no other identity: ask.
```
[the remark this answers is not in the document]
danh-nguyen-nvidia: Agreed — once both the name-scoped and UUID-scoped lookups have been checked, a missing category is a valid zero, not a question. Fixed in 11572f589: no match under either identity now reports 0 events for the asked wording and lists the stored categories the window holds (offer, never guess); "ask" is kept only for several matches, and the same outcome applies to NAME= and to the already-UUID identity. The routing catalog entry says the same. Also merged develop (#2101 libs/vss, #2141 vss-build-vision-ai) to clear the conflicts; the merged branch re-passed the live eval 6/6 (28/28 checks) on the new deploy path.

danh-nguyen-nvidia:
/ok to test [11572f5](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/2103/commits/11572f589e6c51c6b5f1a13b01496c140a56b30b)

skills/operations/vss-manage-alerts/SKILL.md
```diff
@@ -3,7 +3,7 @@ name: vss-manage-alerts
 description: Use this skill when operating VSS alert workflows — real-time monitoring, Alert-Bridge subscriptions, verification verdicts, on-demand verification, always-on operation, Slack notifications, incident queries, or camera onboarding. Not for non-alert analytics.
 license: Apache-2.0
 metadata:
-  version: "3.3.3"
+  version: "3.3.4"
```
zac-wang-nv: can this be changed to 3.3.0?

skills/operations/vss-manage-alerts/evals/alerts_vlm_real_time.json
```diff
@@ -42,19 +42,32 @@
...
...-%m-%dT%H:%M:%SZ; }\nseed() { curl -sf -XPUT \"$ES/_doc/fixture-$1-$3?refresh=true\" -H 'Content-Type: application/json' -d \"{\\\"sensorId\\\":\\\"$1\\\",\\\"category\\\":\\\"${4:-fixture_intrusion}\\\",\\\"timestamp\\\":\\\"$(iso $2)\\\",\\\"end\\\":\\\"$(iso $(($2+10)))\\\",\\\"description\\\":\\\"seeded fixture chunk for $1\\\",\\\"info\\\":{\\\"verdict\\\":\\\"confirmed\\\",\\\"chunkIdx\\\":\\\"$3\\\"}}\" && echo || { echo \"SEED FAILED for $1 chunk $3 — Elasticsearch unreachable or rejected the document; STOP\"; return 1; }; }\nseed warehouse_sample $((NOW-900)) 0 \\\n  && seed warehouse_sample $((NOW-60)) 4 \\\n  && seed warehouse_sample $((NOW-30)) 5 \\\n  && seed warehouse_sample \"$NOW\" 6 \\\n  && seed warehouse_sample \"$NOW\" 9 fixture_other \\\n  && seed cam_decoy \"$NOW\" 0 \\\n  || { echo \"FIXTURE INCOMPLETE — STOP, do not proceed to the next step\"; exit 1; }\n\nAll six must return Elasticsearch \"result\":\"created\" (the seeds are chained, so the first failure stops the rest). If :9200 is unreachable or any seed fails, say so and STOP — do not proceed. The fixture stays in the store for the following steps; its categories (`fixture_*`) are unique to this spec.",
```
zac-wang-nv: this test wont pass for nemoclaw deployment, where vss stack is not reachable from localhost, the current contract doesn't support writing records directly into ES either.

skills/operations/vss-manage-alerts/SKILL.md
```diff
@@ -432,20 +432,49 @@ total. Scope only with `--data-urlencode "sensor_id=..."`.
...
 persist and the `${VAR:?}` / `|| exit` guards fire — each fenced block in its own Bash call
 loses the variables. But this is a **decision tree, not a top-to-bottom script**: run only
-ONE query per the prose (unscoped **vs** name-scoped), run step 3 **only** when the scoped
-count is 0, and take the unfiltered fallback **only** when VIOS is unreachable. The exit code
-says which branch: `exit 1` / a failed `${VAR:?}` = stop and tell the user; the VIOS-down
-`exit 2` = switch to the unfiltered `/incidents` fallback (do **not** report it as an error).
-The explicit guards do the failure detection — do NOT wrap the blocks in `set -e`, which
-(with `pipefail`) would abort the `grep`-no-match branch (an unknown sensor) before it can
-tell the user what exists.
+ONE of the step-2 queries per the prose ((a) unscoped, (b) name-scoped, or (c) consolidated),
```
zac-wang-nv, line 450: this is very verbose,  the intent is,
1) "some queries" ->  chunks, no consolidation
2) some other queries -> consolidate events.

right?  can we simplify it?

skills/operations/vss-manage-alerts/SKILL.md
```diff
@@ -3,7 +3,7 @@ name: vss-manage-alerts
 description: Use this skill when operating VSS alert workflows — real-time monitoring, Alert-Bridge subscriptions, verification verdicts, on-demand verification, always-on operation, Slack notifications, incident queries, or camera onboarding. Not for non-alert analytics.
 license: Apache-2.0
 metadata:
-  version: "3.3.3"
+  version: "3.3.0"
```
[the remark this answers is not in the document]
danh-nguyen-nvidia: This is due to requirement, please don't mark it as the issue

skills/operations/vss-manage-alerts/evals/alerts_vlm_real_time.json
```diff
@@ -42,19 +44,41 @@
...
+        "The trajectory created exactly two new realtime rules by calling Alert Bridge `POST http://localhost:9080/api/v1/realtime` directly, one with `alert_type` `fixture_presence` and one with `alert_type` `fixture_decoy`, both carrying the prompt \"Is an indoor warehouse scene with shelves or floor visible in this video? Answer yes or no.\" (verbatim or near-verbatim), the warehouse_sample RTSP `live_stream_url`, `sensor_name` warehouse_sample and its VIOS UUID as `sensor_id`. It did NOT use the VSS Agent `POST /generate` and did NOT call the `rtvi-vlm` microservice (`:8018`) directly.",
+        "`sh -c 'curl -sf --max-time 15 http://localhost:9080/api/v1/realtime | python3 -c \"import json,sys; rs=json.load(sys.stdin).get(\\\"rules\\\",[]); ts={str(r.get(\\\"alert_type\\\") or \\\"\\\") for r in rs if str(r.get(\\\"sensor_name\\\") or \\\"\\\").lower().startswith(\\\"warehouse_sample\\\")}; sys.exit(0 if {\\\"fixture_presence\\\",\\\"fixture_decoy\\\"} <= ts else 1)\"'` returns exit 0 (both fixture rules are persisted on warehouse_sample; the ladder-PPE rule from step 2 may coexist).",
```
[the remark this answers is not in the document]
danh-nguyen-nvidia: Fair point on the deploy path: the eval's step 1 brings the stack up with docker compose up -d on the resolved build, ES sits on the named elastic-data volume and realtime rules are persisted, so on a box that was not torn down the previous run's fixture rules are still listed and may still be publishing. The agent under test cannot clean them up — the skill's stop/delete gate applies to every DELETE, in CI included — and per-run identities would force the count probes to discover the category. Fixed in 43c8f76aa instead: step 4 tells the agent leftovers may exist, not to delete or reuse them, and to create the two new rules regardless; its probe now requires a rule per fixture alert type whose created_at is within the last 30 minutes, so a leftover pair no longer satisfies it, and step 5 no longer claims the 20-minute bound isolates runs. Impact was already bounded: every count check in steps 6-7 is relational (events ≥ 1 and < chunks, Σ chunkCount == raw total, category-less total > with-category), so a same-alert-type leftover folds into the same events and cannot flip a verdict.

skills/operations/vss-manage-alerts/SKILL.md
danh-nguyen-nvidia: Updated

skills/operations/vss-manage-alerts/SKILL.md
danh-nguyen-nvidia: Yes, exactly that. Collapsed to the two rules in d3671e256:
- Raw → chunks: lists, counts with no period, anything phrased as incidents/alerts, forensics, every CV ask.
- Consolidated (consolidate=true, needs start_time + end_time) → events: only "how many events / times" in a stated period on VLM real-time.
One closing line on where grouping is tuned (rtvi_vlm.consolidation, 60 s gap / 300 s cap) and that nothing is deduplicated in the store or UI. 23 lines → 12.

skills/operations/vss-manage-alerts/evals/alerts_vlm_real_time.json
danh-nguyen-nvidia: Agreed — the ES fixture is gone.

Fix: the fixture is now produced by the stack itself. Step 4 creates two realtime rules on warehouse_sample via POST /api/v1/realtime with a prompt that is true of every warehouse chunk (fixture_presence, fixture_decoy); step 5 waits through the Alert Bridge API only ($VSS_PUBLIC_URL/alert-bridge, else $HOST_IP:9080) until confirmed chunks exist under both. No ES, no localhost.

Since the data is live, the count steps grade relations instead of fixed numbers: consolidated total ≥ 1 and < chunk count for the same filter, events' chunkCount sum to that chunk count, dropping category= adds the decoy's events, consolidate=true without a window is still a 400.

Two deploy settings the spec now asserts, because RT-VLM has to emit the incidents: VIA_VLM_ENDPOINT must be the remote VLM (not the container's own port), and Kafka publishing is gated on MESSAGE_BUS=kafka + KAFKA_BOOTSTRAP_SERVERS — KAFKA_ENABLED in compose/README is not read by the code. The remote VLM also has to be multimodal; the first CI run of steps 4–7 will tell us what the runner has.

danh-nguyen-nvidia:
/ok to test [43c8f76](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/2103/commits/43c8f76aae54beb995a50a82895a911de73c2366)

danh-nguyen-nvidia:
/ok to test [e763845](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/2103/commits/e7638457782375f0f374f69678bd912a4b5def92)

skills/operations/vss-manage-alerts/SKILL.md
```diff
@@ -522,9 +558,82 @@ curl -sfG "$AB/api/v1/realtime/incidents" \
...
+#   [ -n "$NAME" ] && L+=(--data-urlencode "sensor_id=$NAME")
+#   RAW=$(curl -sfG --max-time 30 "$AB/api/v1/realtime/incidents" "${L[@]}") || { echo "category lookup failed (HTTP error or unreachable) — stop"; exit 1; }
+#   printf '%s' "$RAW" | jq -r '.incidents[].category' | sort -u
+# Exactly one match with the user's wording → CATEGORY. Non-empty page but NO match, and NAME
+# is the sensor's NAME (a sensor was named and this is the first identity) → the category may
+# live under the sensor's other identity (a rule created without sensor_name stores its chunks
+# under the VIOS UUID): resolve $UUID exactly as step 3 does and repeat this lookup with
+# sensor_id=$UUID; exactly one match there → NAME=$UUID, CATEGORY=<match>, and say the events
+# matched the UUID identity. Still no match under either identity → both pages were checked, so
+# the answer is 0 events for that wording: report it and list the stored categories the window
```
zac-wang-nv, line 593: **P1** — Please treat a category-discovery scan that reaches the raw 10,000-hit limit as incomplete before taking this zero-result branch. The raw query does not request exact Elasticsearch hit counts, and offset pagination cannot pass the default 10,000-result window, so fetching all reported rows does not prove that a category is absent. With one older `intrusion` chunk followed by 10,000 `routine` chunks in the requested window and no UUID-scoped rows, discovery sees only `routine` and this branch reports 0; a category-filtered consolidated query returns 1 intrusion event with `truncated: false`. Report that discovery is incomplete, or discover categories over smaller intervals covering the full requested window before running the consolidated query with the original bounds. Apply the same guard to the matching expectation in evals/evals.json.

zac-wang-nv (approved):
LGTM

skills/operations/vss-manage-alerts/SKILL.md
danh-nguyen-nvidia: Agreed — the raw page's total saturates at 10,000 (no track_total_hits) and offset cannot page past max_result_window, so a discovery scan that reaches 10,000 rows cannot prove a category is absent. Step 3 already treats 10,000 as a floor; the category lookup should do the same: at 10,000, mark discovery incomplete and either run the consolidated query with the user's category as-is or discover over sub-windows before querying with the original bounds. Tracking as a follow-up (SKILL.md lookup block + the matching evals.json expectation) rather than growing this PR further.

PR #2105, opened 2026-09-08
refactor(vss-generate-video-report): move per-mode steps into references/report-types

danh-nguyen-nvidia:
## What changed

`skills/operations/vss-generate-video-report/SKILL.md` (717 → 388 lines) is now the dispatcher; each report mode's numbered steps live in one file under `references/report-types/`.

vss-generate-video-report/
SKILL.md                            routing, prerequisite gates, shared setup, output contract, error handling
references/
report-types/
video-analysis.md               Mode A — Steps 1-4 (A1 VST clip URL / A2 local file or base64)
incident-range.md               Mode B — Steps 1-3
sop-compliance.md               Mode C — Steps 1-3 + get_sop_report contract
report-templates/                 unchanged
default-vlm-prompt.md             unchanged

Each report-type file opens with a fixed header (trigger phrases → prerequisite gate → inputs → template → output contract → failure modes, every item a pointer to SKILL.md or to the step that owns it), followed by the numbered steps.

### SKILL.md (dispatcher)
- Mode table gains a **Steps** column linking the report-type file; Instructions step 4 reads the file via `$SKILL_DIR/references/report-types/<file>` (open only the routed one).
- Step 0: `SKILL_DIR` fallback = directory containing SKILL.md when no base dir was announced, and the fresh-shell hand-off rule below.
- Ambiguity rule: ask only when the request names a sensor + time range and carries neither incident/alert nor SOP/compliance wording; never run a probe or gate before the mode is settled. Step 1 names Mode A paths A1 / A2.
- Hard gate: Mode A clip ≥ 120 s → stop and prompt for LVS unless LVS is already ready.
- Endpoint resolution runs first for every mode (Kubernetes or Docker Compose), honours the legacy `VSS_ENDPOINT` alias, defaults `HOST_IP` to `localhost` on Docker with a stderr notice, and prints its hand-off.
- Probe block keyed on `REPORT_MODE=A1|A2|B|C`: only that mode's probes run and decide the exit code; on Docker the two standard VLM ports are probed; the Mode C `tools/list` gate is a prose bullet pointing at `sop-compliance.md` Step 1.
- Output contract: Mode B zero results (no invented incidents, no Mode A fallback, no VLM); Mode C empty range keyed to the exact tool result; Mode A rows may hold a literal `N/A (<reason>)` as defined in Step 4.
- `/vss-build-vision-ai` hand-offs (stock Base workflow for A1, stock Alerts workflow for Mode B, SOP profile for Mode C) apply to Docker Compose; on Kubernetes the missing public route is reported to the deployment owner.
- New `## Report types` section: file index, template links, "Adding a report type" touch points.

### Fresh-shell hand-off contract
Every fenced bash block runs in its own shell. Blocks that resolve shared values end by printing shell-quoted `NAME=value` lines (`printf %q`); the agent pastes them at the top of the next block (caller-supplied values after, so they win); consuming blocks guard with `${VAR:?}` and name the source.
- Endpoint resolution → `DEPLOYMENT_KIND`, `VSS_PUBLIC_URL`, `HOST_IP`, `VST_API_BASE`, `VA_MCP_URL`, `VLM_ENDPOINT` (when known).
- Mode A Step 1 (A1) → `VIDEO_URL`, `CLIP_START`, `CLIP_END`, `CLIP_SECONDS`.
- Mode A Step 2 → `VLM_BACKEND`, `VLM_ENDPOINT`, `VLM_MODEL`.
- Clip-URL rewrite (Kubernetes / Docker) → `BROWSER_CLIP_URL`.
- HITL result → `HITL_RESOLVED`, `HITL_PROMPT_FILE` (approved prompt written to a file, never pasted into the shell); `REASONING=true` only when asked.

### Mode A (`video-analysis.md`)
- A1 requires the hand-off origin, bootstraps the CLI from `${VSS_REPO_ROOT}/libs/vss` (guarded), stops when `vss configure` fails, and computes `CLIP_SECONDS` from the resolved `start_time` / `end_time` (GNU or BSD `date`), refusing empty, reversed, or ≥ 120 s windows before printing the hand-off.
- Step 2 is one block: Kubernetes public route or Docker discovery (a caller-supplied `VLM_MODEL` must be served by the candidate), `/models` confirmation, model-in-list check, backend derived from the final endpoint / model.
- Step 3 requires a numeric `CLIP_SECONDS` (A1 hand-off, or ffprobe / stated duration for A2) and re-applies the 120 s gate; it keeps the raw VLM response, fails on transport / non-2xx / non-object body / `finish_reason=length`, removes `<think>…</think>` blocks (and a leading unopened one) and `<answer>` tags, and rejects reasoning-only, null or list content. `VLM_MAX_TOKENS` defaults to 4096 and must be a positive integer. One block serves A1 and A2 (`VIDEO_SRC`); A2 base64 is read from `VIDEO_B64_FILE`.
- Step 4 defines the Clip URL / Clip Range / VLM row values per path (A1, A2, LVS).
- Docker clip rewrite falls back to `http://HOST_IP:7777`, uses resolved `VSS_PUBLIC_*` only when present (Brev), and ignores unresolved `${…}` templates from `generated.env`.

### Mode B (`incident-range.md`)
- `get_incidents` `includes` carries `category` and `place`; `source` / `source_type` documented as a pair; `has_more` is read; zero results = `incidents == []`, error shapes are failures; clip pointer is `info.videoSource`.

### Mode C (`sop-compliance.md`)
- Step 1 and Step 2 consume the pasted `VA_MCP_URL`; both select the JSON-RPC response for their id from SSE or plain JSON, check the HTTP status and separate a VA-MCP failure (`VA-MCP problem`) from a missing tool (`SOP tools absent`).
- Step 2 requires the tool text to be a JSON object; the only `{"error": …}` that passes is the exact empty-range sentinel `{"error": "No VisionLLM messages found for the given filters."}`, which Step 3 renders as the one-line empty-range statement. Step 3 maps `{sensor_id}` / `{start_time}` / `{end_time}` to the Step 1 inputs and `{total_actions}` to `actions_observed.total_action_entries`.

### Other files
- `evals/evals.json`: two Mode A entries now expect the agent to open `references/report-types/video-analysis.md` Step 3.
- `skills/vss-build-vision-ai/references/services/sop.md`, `sop/integrate-ds-sop.md`: pointers to the Mode C step file, worded for any install layout.
- Version stays `3.3.0`; `skill-card.md` unchanged; `skills/README.md` rows are a follow-up.

### Merge with `develop`
The branch merges `develop` (merge commit, no rebase) to pick up the `vss-deploy-profile` retirement (#2141) and the CLI move to `libs/vss`. Every deploy hand-off in SKILL.md and in the text moved to `video-analysis.md` now names `/vss-build-vision-ai`; the A1 CLI bootstrap is `uv run --project "${VSS_REPO_ROOT}/libs/vss" vss`.

Kubernetes LVS readiness probe vs the UI catch-all; auth header for remote VLM endpoints; whole-segment default vs the 120 s gate; stale exported `VLM_ENDPOINT` on Kubernetes; `video_report_gen.hitl_enabled: true` in the shipped profile vs non-interactive runs; Mode B clip-link GET cost at `max_count` 100; sandbox-alias `HOST_IP` in the VLM input URL; literal `<sensor>` / `<sensor-name>` / `<ISO>` placeholders in the A1 and Mode C blocks; fractional `CLIP_SECONDS` truncation and non-UTC offsets in the BSD `date` path; `skills/README.md` rows; eval coverage for Mode C, the hand-off contract and the routing rule.

## Checklist


skills/operations/vss-generate-video-report/references/report-types/sop-compliance.md
```diff
@@ -0,0 +1,80 @@
...
+fi
+MCP="${VA_MCP_URL%/}/mcp"
+CT='Content-Type: application/json'; AC='Accept: application/json, text/event-stream'
+SID=$(curl -si --max-time 10 -X POST "$MCP" -H "$CT" -H "$AC" \
+  -d '{"jsonrpc":"2.0","method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"cli","version":"1.0"}},"id":0}' \
+  | awk 'tolower($1)=="mcp-session-id:"{print $2}' | tr -d '\r')
+[ -n "$SID" ] || { echo "VA-MCP initialize failed (no session id) — is VA-MCP up at ${VA_MCP_URL}?" >&2; exit 1; }
+curl -s --max-time 30 -X POST "$MCP" -H "$CT" -H "$AC" -H "mcp-session-id: $SID" \
+  -d '{"jsonrpc":"2.0","method":"tools/call","params":{"name":"video_analytics__get_sop_report","arguments":{"sensor_id":"<sensor>","start_time":"<ISO>","end_time":"<ISO>"}},"id":2}' \
+  | grep '^data:' | sed 's/^data: //' | jq -r '.result.content[0].text'
```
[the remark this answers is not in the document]
danh-nguyen-nvidia: Fixed in f7b4053b6. Step 2 now keeps the raw tools/call body in a temp file
(removed on exit) and validates the envelope before extracting any text:
- curl transport error or non-200 status -> exit 1 with the raw body on stderr
- selects the JSON-RPC response with id 2 (not just the last SSE event); a
  plain-JSON body is accepted too; a non-SSE / unparseable body -> exit 1 with
  the raw body
- a JSON-RPC `error` envelope, a missing `result`, or `result.isError: true`
  -> exit 1 with the envelope on stderr
Only then is `.result.content[0].text` printed, so the empty-range sentinel
`{"error": "No VisionLLM messages found for the given filters."}` still reaches
Step 3 unchanged. Replayed against fixtures for each of the three failure shapes
you listed plus the two pass-through shapes.

skills/operations/vss-generate-video-report/references/report-types/video-analysis.md
```diff
@@ -0,0 +1,308 @@
...
+}
+
+# FINAL_PROMPT must come from the resolved HITL mode gate (SKILL.md § HITL prompt mode).
+# Resolution order:
+#   1) video_report_gen.hitl_enabled
+#   2) HITL_ENABLED (fallback only when runtime config is unavailable)
+#   3) default false when neither source is set
+# - resolved false: FINAL_PROMPT="$DEFAULT_PROMPT"
+# - resolved true : FINAL_PROMPT comes from the latest EDIT/NEW value after explicit APPROVE.
+FINAL_PROMPT="${FINAL_PROMPT:-$DEFAULT_PROMPT}"
```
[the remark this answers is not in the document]
danh-nguyen-nvidia: Fixed in f7b4053b6. The block now re-establishes the HITL result at its top
instead of relying on inherited shell state:
- `HITL_RESOLVED` is mandatory (exactly true|false; anything else exits 1)
- when true, the approved EDIT/NEW text is handed over as a file
  (`HITL_PROMPT_FILE`, written with the Write tool, read with `cat`) so nothing in
  the prompt is shell-interpreted; a missing, unreadable or blank file exits 1
  instead of silently falling back to `DEFAULT_PROMPT`
- when false, `DEFAULT_PROMPT` is assigned explicitly
The separate A2 inline-bytes curl block, which started at `curl` and skipped
this preamble, is folded into the same Step 3 block (`VIDEO_SRC` takes
`VIDEO_DATA_URL` or `VIDEO_URL`). SKILL.md § HITL prompt mode describes the same
handoff.

skills/operations/vss-generate-video-report/references/report-types/video-analysis.md
```diff
@@ -0,0 +1,306 @@
...
+      ]
+    }
+  ],
+  "max_tokens": 1024,
+  "temperature": 0.0${MM_KWARGS}
+}
+EOF
+```
+
+For Mode A path A2 (inline bytes), run the same Step 3 block with `VIDEO_DATA_URL` (Step 1) set at its top instead of `VIDEO_URL`; the block sends whichever is set, so the HITL guard, prompt resolution, `CFG_JSON` and `MM_KWARGS` apply to A2 unchanged. Because the block is a fresh shell, build the data URL there too — for a local file: `[ -s "$VIDEO_FILE" ] || exit 1; VIDEO_DATA_URL="data:${VIDEO_MIME:-video/mp4};base64,$(base64 < "$VIDEO_FILE" | tr -d '\n')"` (the `tr` strips the line wrapping GNU `base64` adds, which would otherwise corrupt the data URL); for user-supplied base64, `VIDEO_DATA_URL="data:${VIDEO_MIME};base64,${VIDEO_BASE64}"`.
```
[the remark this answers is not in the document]
danh-nguyen-nvidia: Good catch, fixed in 26262365a as suggested: the user-supplied base64 branch now uses
`${VIDEO_MIME:-video/mp4}`, matching Step 1's default and the local-file branch next to it.
The same commit also moves that base64 out of the shell entirely: the payload is written to
a file (`VIDEO_B64_FILE`) and read inside the Step 3 block with `tr -d '[:space:]'`, so it
is never pasted into a shell assignment.

skills/operations/vss-generate-video-report/SKILL.md
```diff
@@ -3,7 +3,7 @@ name: vss-generate-video-report
 description: Use this skill when producing a VSS analysis report — Mode A per-clip VLM, Mode B incident-range via video-analytics, Mode C SOP compliance via the SOP tools. Not for standalone video summarization, real-time alerts or ad-hoc Q&A.
 license: Apache-2.0
 metadata:
-  version: "3.3.0"
+  version: "3.4.0"
```
zac-wang-nv: this version should stick with the current vss release version 3.3.0

skills/operations/vss-generate-video-report/SKILL.md
```diff
@@ -263,437 +271,28 @@ BROWSER_CLIP_URL=$(echo "$RAW_URL" | sed -E "s|^https?://[^/]+|${VSS_PUBLIC_HTTP
...
+1. `references/report-types/<type>.md` (skeleton above) and `references/report-templates/<type>-report.md`.
+2. In this file: frontmatter `description`; the mode table row (Backend + Steps); *Instructions* steps 1–3 (mode pick, deploy hand-off target, HITL applicability); *Examples*; *Runtime prerequisites* (intro sentence, *Mode-by-mode checklist* row, probe / gate line, hand-off sentence, and the mode sentence in *Endpoint resolution* if it shares `${VA_MCP_URL}` or needs a new endpoint); the *HITL prompt mode* scope sentence if the type has a prompt-approval step; *Output contract for evaluators*; the per-mode clip-URL sentence if the report embeds clips; the *Templates* link line in this section; optionally *Error Handling* and *Cross-Reference* lines.
+3. `skill-card.md` (description, use case, references), the skill's `evals/`, and — when the mode list changes — the `skills/README.md` rows that enumerate this skill's modes (not updated in this refactor; follow-up).
```
zac-wang-nv: skill-card.md is for publication, not used by agent/LLM. SKILL.md shouldn't mention skill-card.md or evals/

skills/operations/vss-generate-video-report/SKILL.md
```diff
@@ -133,9 +136,13 @@ curl -sf --max-time 5 "${VLM_ENDPOINT:-http://${HOST_IP}:30082/v1}/models" >/dev
 curl -sf --max-time 5 "${VA_MCP_URL:-http://${HOST_IP}:9901}/health" >/dev/null
 
 # Mode C — reachability is NOT sufficient; also REQUIRE the SOP tools on VA-MCP:
-# tools/list on ${VA_MCP_URL}/mcp (two-step JSON-RPC, see Mode C Step 1) must include
-# video_analytics__get_sop_report. If absent, the deployment lacks the SOP patch —
-# hand off to /vss-build-vision-ai and do NOT proceed with Mode C.
+# tools/list on ${VA_MCP_URL}/mcp must include video_analytics__get_sop_report. The runnable
+# probe is the initialize -> tools/list block in references/report-types/sop-compliance.md
```
zac-wang-nv: does the skill still need these? why commented out?

skills/operations/vss-generate-video-report/SKILL.md
```diff
@@ -13,23 +13,23 @@ metadata:
...
 
 ---
 
 ## Instructions
 
-0. **Set `SKILL_DIR`** to the "Base directory for this skill" path announced when this skill loads. All skill-relative reads (e.g. the default VLM prompt) resolve under `$SKILL_DIR` — never via cwd-relative paths.
-1. **Pick the mode** — Mode A for a single recorded clip/sensor video, Mode B when the request names a time range or incidents/alerts, Mode C when the request asks for an SOP / compliance report (match against *Examples*).
+0. **Set `SKILL_DIR`** to the "Base directory for this skill" path announced when this skill loads. All skill-relative reads (e.g. the default VLM prompt) resolve under `$SKILL_DIR` — never via cwd-relative paths. If no base directory was announced (this file was opened directly), `SKILL_DIR` is the directory containing this `SKILL.md`. Re-set `SKILL_DIR=<that path>` at the top of every shell block you run — each fenced block is its own shell — the same way the *HITL prompt mode* bullet has you set `HITL_RESOLVED` (and `HITL_PROMPT_FILE`) at the top of the Mode A Step 3 block.
```
liamy-nv: **[P1] Endpoint state is lost between fresh-shell blocks**

The new contract says every fenced block runs in a fresh shell, but only SKILL_DIR is explicitly restored. Endpoint resolution and Mode A Step 2 still pass VST_API_BASE, VLM_ENDPOINT, VLM_MODEL, VLM_BACKEND, VA_MCP_URL, and DEPLOYMENT_KIND implicitly between blocks. In the normal Kubernetes/auto-discovery path, the next block therefore probes an empty URL or the Docker fallback and report generation stops.

skills/operations/vss-generate-video-report/SKILL.md
danh-nguyen-nvidia: Reverted to 3.3.0 in 8e40b7d92

skills/operations/vss-generate-video-report/SKILL.md
danh-nguyen-nvidia: Removed in 8e40b7d92. SKILL.md no longer mentions skill-card.md or evals/; the "Adding a report type" list keeps only the README rows note.

skills/operations/vss-generate-video-report/SKILL.md
danh-nguyen-nvidia: They were a pointer to the runnable gate in `references/report-types/sop-compliance.md` Step 1, not disabled code, but inside a bash block they read like dead code. In 8e40b7d92 they are a prose bullet under the probe block instead.

skills/operations/vss-generate-video-report/SKILL.md
danh-nguyen-nvidia: Agreed, fixed in 8e40b7d92. State is now handed over explicitly: a block that resolves shared values ends by printing shell-quoted `NAME=value` lines (`printf %q`), the agent pastes them at the top of the next block (caller-supplied values after, so they win), and consuming blocks guard with `${VAR:?}`.

Endpoint resolution prints `DEPLOYMENT_KIND`, `VSS_PUBLIC_URL`, `HOST_IP`, `VST_API_BASE`, `VA_MCP_URL` and `VLM_ENDPOINT` (when known). The probe block requires them instead of falling back to Docker URLs, so a missing value on Kubernetes fails loudly rather than probing the wrong deployment; it takes `REPORT_MODE` (A1/A2/B/C) so only the chosen mode's probes decide its exit code. Mode A Step 2 is one self-contained block that discovers and confirms the VLM (the model id must be in `/models`) and prints `VLM_BACKEND` / `VLM_ENDPOINT` / `VLM_MODEL` for Step 3, whose guard requires them. The LVS and clip-URL blocks follow the same contract, and the Mode C blocks reuse the pasted `VA_MCP_URL` before falling back to re-deriving it.

Verified end to end with a stub curl on the Kubernetes, Docker (with and without `HOST_IP`), alias-only and caller-supplied paths.

skills/operations/vss-generate-video-report/references/report-types/video-analysis.md
```diff
@@ -0,0 +1,331 @@
...
+# Fresh shell: Step 2's VLM discovery does not arrive on its own — set both at the top of this block.
+: "${VLM_ENDPOINT:?set VLM_ENDPOINT (Step 2) at the top of this block}" "${VLM_MODEL:?set VLM_MODEL (Step 2) at the top of this block}"
+# A1 sends the VST clip URL; A2 (Step 1) sends inline bytes — this one block serves both paths.
+VIDEO_SRC="${VIDEO_DATA_URL:-${VIDEO_URL:?set VIDEO_URL (A1) or VIDEO_DATA_URL (A2) at the top of this block}}"
+case "$VIDEO_SRC" in
+  null|"") echo "ERROR: VIDEO_URL is null/empty — Step 1 clip resolution failed (no media_url)" >&2; exit 1 ;;
+  data:*,) echo "ERROR: VIDEO_DATA_URL carries no base64 payload (empty or unreadable VIDEO_FILE / VIDEO_B64_FILE)" >&2; exit 1 ;;
+esac
+
+curl -s --connect-timeout 5 --max-time 120 -X POST "${VLM_ENDPOINT}/chat/completions" \
```
liamy-nv: **[P1] Validate the VLM response before treating it as report content**
This pipeline masks both transport failures and VLM HTTP/API errors. Because the shell returns jq’s status, an empty body can exit successfully, while an error envelope or missing choices prints null with exit 0. The workflow can consequently render a successful-looking report containing null instead of following the Error Handling contract. Please preserve the raw response and HTTP status, fail on transport/non-2xx responses, and require .choices[0].message.content to be a non-empty string before continuing.

danh-nguyen-nvidia: Agreed, fixed in b29e4f5d2. The VLM call now keeps the raw response in a temp file, fails on a curl transport error or a non-2xx status with the body on stderr, and prints report text only when the body is a JSON object whose `.choices[0].message.content` is a string with something left after stripping `<think>…</think>` blocks and `<answer>` tags; error envelopes, empty bodies, null or list content, reasoning-only answers, an unclosed `<think>` and `finish_reason: "length"` all exit non-zero per Error Handling. Verified with a stub curl across those response shapes on bash 3.2 and zsh. c444e666a and bb35813ad build on it (the fresh-shell hand-off now also carries the resolved clip window and duration into Step 3).

skills/operations/vss-generate-video-report/references/report-types/video-analysis.md
```diff
@@ -0,0 +1,362 @@
...
+     || { echo "vss vios clip returned no media_url for <sensor-name>" >&2; printf '%s\n' "${CLIP}" >&2; exit 1; }
+   # Hand-off (shell-quoted, paste at the top of the Step 3 block and keep for Step 4): the clip URL and the
+   # window the CLI resolved (the whole recorded segment when none was given).
+   CLIP_START=$(printf '%s' "${CLIP}" | jq -r '.start_time // empty'); CLIP_END=$(printf '%s' "${CLIP}" | jq -r '.end_time // empty')
+   printf 'VIDEO_URL=%q\nCLIP_START=%q\nCLIP_END=%q\n' "${VIDEO_URL}" "${CLIP_START}" "${CLIP_END}"
+   ```
+
+The block prints shell-quoted `VIDEO_URL=…`, `CLIP_START=…` and `CLIP_END=…` assignments (signed clip URLs carry `&` and `?`, so paste the lines as printed, never the bare URL). Each fenced block is a fresh shell: at the top of the Step 3 block paste those lines, `CLIP_SECONDS=<CLIP_END minus CLIP_START, in seconds>`, and the `VLM_BACKEND` / `VLM_ENDPOINT` / `VLM_MODEL` hand-off Step 2 printed — the Step 3 guards refuse to run without a video source and a VLM endpoint + model. Set `RAW_URL="$VIDEO_URL"` before applying the report-link rewrite for Step 4.
```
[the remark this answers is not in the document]
danh-nguyen-nvidia: Agreed, fixed in bb35813ad. The A1 block now computes `CLIP_SECONDS` from the resolved `start_time` / `end_time` (GNU or BSD `date`, fractions and `+00:00` normalised), refuses a missing, unparseable, empty or reversed window and a clip of 120 s or longer with the Long-video message, and prints `CLIP_SECONDS` in the hand-off. Step 3 requires a numeric `CLIP_SECONDS` (never a default or an ISO string) and refuses 120 s or longer again before computing `num_frames`; A2 measures the duration with `ffprobe`. Verified with a stub CLI on bash 3.2 and zsh, including 15 s, 119 s, 120 s with fractional bounds, reversed and missing bounds.

skills/operations/vss-generate-video-report/references/report-types/video-analysis.md
```diff
@@ -0,0 +1,374 @@
...
+     echo "vss vios clip failed for <sensor-name>" >&2; exit 1; }
+   VIDEO_URL=$(printf '%s' "${CLIP}" | jq -er '.media_url | select(type=="string" and length>0)') \
+     || { echo "vss vios clip returned no media_url for <sensor-name>" >&2; printf '%s\n' "${CLIP}" >&2; exit 1; }
+   # Hand-off (shell-quoted, paste at the top of the Step 3 block and keep for Step 4): the clip URL and the
+   # window the CLI resolved (the whole recorded segment when none was given).
+   CLIP_START=$(printf '%s' "${CLIP}" | jq -r '.start_time // empty'); CLIP_END=$(printf '%s' "${CLIP}" | jq -r '.end_time // empty')
+   [ -n "${CLIP_START}" ] && [ -n "${CLIP_END}" ] || { echo "vss vios clip returned no start_time / end_time for <sensor-name>" >&2; printf '%s\n' "${CLIP}" >&2; exit 1; }
+   # Duration in whole seconds from the resolved window (GNU date -d or BSD date -j; fractions and +00:00 dropped).
+   # It feeds the Long-video rule and Step 3's frame sampling — never leave it to a default.
+   _iso2s() { _t=$(printf '%s' "$1" | sed -E 's/(\.[0-9]+)?(Z|\+00:00)$/Z/'); date -u -d "$_t" +%s 2>/dev/null || date -j -u -f '%Y-%m-%dT%H:%M:%SZ' "$_t" +%s 2>/dev/null; }
```
zac-wang-nv: **P1** — Preserve fractional seconds until after computing and validating the duration. `_iso2s` strips the fraction from each endpoint before subtraction, so `2026-09-15T00:00:00.900Z` → `2026-09-15T00:02:00.100Z` (119.2 seconds) becomes 120 seconds and incorrectly forces the LVS path. A valid `00:00:00.100Z` → `00:00:00.600Z` clip similarly becomes zero and is rejected as empty. Both failures reproduce with the full A1 block in Bash and zsh using successful CLI responses. Compute the precise delta, check `0 < duration < 120`, and truncate only for frame sampling. This is a distinct false-rejection case introduced by the duration fix; the earlier missing-duration bypass is fixed.

danh-nguyen-nvidia: Fixed in d3fe7b02b. _iso2s re-attaches the ISO fraction to the epoch value (GNU and BSD date), the delta is computed at ms precision and the gate is 0 < CLIP_SECONDS < 120 on that value; truncation only happens in Step 3 for num_frames, after Step 3 re-applies the same precise gate. All awk calls run under LC_ALL=C so a decimal-comma locale cannot alter the hand-off. Verified your two inputs (119.2 s → CLIP_SECONDS=119.2, 0.5 s → 0.5) plus exactly 120.000 s (refused), reversed / equal windows (refused) and integer windows (still 30), in bash and zsh on both date flavours and under C and fr_FR locales.

liamy-nv (approved):
LGTM

zac-wang-nv (approved):

danh-nguyen-nvidia:
/ok to test [d3fe7b0](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/2105/commits/d3fe7b02b3ea2da9cdd9f9b5953ce368c6a992bd)

PR #2160, opened 2026-09-11
skill-eval: add harness-wide evaluated-agent model selection

gbau1:
## Summary

Adds one validated model-routing contract for the agent under evaluation across the Claude Code and NemoClaw harnesses. This is the focused replacement for #1764.

## Plan

- Keep runner defaults when no override is requested.
- Resolve provider, model, endpoint, and runner-owned credentials once before worker allocation.
- Apply the selected route only to the evaluated agent. The coordinator, judge, and NemoClaw setup coding agent keep their existing route.
- Reject incomplete or incompatible endpoint/model/provider combinations with actionable errors.

## Related Issue

Supersedes #1764.

## Changes

- Adds manual workflow inputs for provider, model, and endpoint.
- Adds a shared model-route resolver used by the skill-eval harness.
- Preserves the existing default path when inputs are blank.
- Covers Claude Code, Codex, and NemoClaw route validation without notebook, skill, deployment, compose, or hardware changes.
- Adds focused unit and integration coverage.

## Validation

- `python3 -m pytest -q .github/skill-eval/tests .github/skill-eval/envs/tests` — 336 passed
- Workflow YAML parse — passed
- SPDX header check — passed
- `git diff --check` — passed

## Checklist
- [ ] I have installed and run the complete pre-commit hook suite locally.
- [x] Every commit on this PR has DCO sign-off.


.github/workflows/skills-eval.yml
```diff
@@ -248,8 +275,24 @@ jobs:
...
+          if [ -n "$skills_eval_endpoint" ]; then
+            export SKILLS_EVAL_ENDPOINT_URL="$skills_eval_endpoint"
+          else
+            unset SKILLS_EVAL_ENDPOINT_URL
+          fi
           hash -r
           python3 -c 'import os, sys; expected = tuple(map(int, os.environ["SKILL_EVAL_PYTHON_VERSION"].split("."))); assert sys.version_info[:2] == expected, sys.version'
+          # Fail before run_leg waits for or acquires a Brev worker. This logs
+          # the resolved route but never its credential.
+          python3 .github/skill-eval/model_config.py
```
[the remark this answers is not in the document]
gbau1: Fixed in de553b452: the preflight now runs after the repository-secret fallback is restored, while still running before worker allocation.

.github/skill-eval/model_config.py
```diff
@@ -0,0 +1,179 @@
...
+        )
+    if not api_key:
+        expected = credential_name if runtime == "nemoclaw" else "ANTHROPIC_API_KEY"
+        raise ValueError(f"no API key is configured; set {expected} on the runner")
+
+    return SkillEvalModelConfig(
+        runtime=runtime,
+        provider=provider,
+        model=model,
+        endpoint_url=endpoint_url.rstrip("/"),
```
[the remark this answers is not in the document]
gbau1: Fixed in de553b452: the shared resolver now rejects endpoints without an HTTP(S) scheme and non-empty host, with parameterized tests.

.github/skill-eval/run_leg.py
```diff
@@ -1564,38 +1565,17 @@ def run_invocations(
...
-    if not base_url:
-        print("FATAL: ANTHROPIC_BASE_URL not set", file=sys.stderr)
+    try:
+        model_config = resolve_model_config(os.environ)
+    except ValueError as exc:
+        print(f"FATAL: invalid evaluated-agent model configuration: {exc}", file=sys.stderr)
         return 1
+    agent = model_config.runtime
+    model = model_config.model
+    base_url = model_config.endpoint_url
```
zac-wang-nv, line 1571: **P1** — Wire the selected endpoint into the Claude Code agent's actual environment. With `provider=custom`, `endpoint_url=https://selected.example.test/v1`, and a different runner `ANTHROPIC_BASE_URL`, this only changes the `--ak api_base=...` argument. The pinned Harbor 0.20.0 `ClaudeCode` adapter ignores that kwarg and builds its remote environment from `os.environ["ANTHROPIC_BASE_URL"]`, which `harbor_env()` still copies unchanged. An offline probe through the real adapter therefore sends `selected/model` to the runner endpoint, not the requested endpoint; it either fails there or evaluates the wrong backend. Apply the endpoint through an agent-scoped setting that Harbor actually consumes, preserve the coordinator/judge route, and test the remote command environment rather than only the Harbor argument list.

.github/skill-eval/model_config.py
```diff
@@ -0,0 +1,190 @@
...
+                "SKILLS_EVAL_API_KEY is not supported by claude-code/codex; "
+                "the evaluated agent uses the runner-managed ANTHROPIC_API_KEY"
+            )
+        provider = (
+            "nvidia-inference"
+            if requested_provider == "default"
+            else requested_provider
+        )
+        model = requested_model or _first(
+            env.get("CODEX_MODEL") if runtime == "codex" else "",
```
zac-wang-nv: **P1** — Preserve the Codex-specific missing-model check when no evaluated-model override is supplied. Previously `EVAL_AGENT=codex` with an empty or absent `CODEX_MODEL` failed explicitly; this now falls through to the coordinator's `ANTHROPIC_MODEL`. For example, a normal Claude-configured runner resolves a Codex evaluation to `aws/anthropic/bedrock-claude-opus-4-6`. The unchanged daily workflow exposes Codex and can provide an empty `vars.CODEX_MODEL`, so it now launches that route through NvCodex instead of reporting the missing configuration. Require either `SKILLS_EVAL_MODEL` or `CODEX_MODEL` for Codex, and add coverage for both the missing-default case and an explicit override.

.github/skill-eval/model_config.py
```diff
@@ -0,0 +1,190 @@
...
+            inherited_provider
+            if requested_provider == "default"
+            else requested_provider
+        )
+        model = requested_model or _first(
+            env.get("NEMOCLAW_MODEL"),
+            env.get("ANTHROPIC_MODEL"),
+            env.get("LLM_REMOTE_MODEL"),
+        )
+        if provider == "nvidia-build" or provider == "build":
```
zac-wang-nv, line 136: **P1** — Reject an explicit endpoint for providers that cannot consume it instead of silently dropping it. A manual dispatch with `runner=nemoclaw`, `provider=nvidia-build`, a model, and `endpoint_url=https://selected.example.test/v1` currently passes preflight and resolves to `provider=build` with an empty endpoint. The evaluation consequently uses NVIDIA's hosted build route even though the operator specified another backend. The local-provider branch has the same behavior. Check `requested_endpoint` before clearing it and return an actionable incompatible-provider/endpoint error, with parameterized tests for these routes.

.github/skill-eval/run_leg.py
```diff
@@ -1564,38 +1565,17 @@ def run_invocations(
...
-            "(expected claude-code | codex | nemoclaw)",
-            file=sys.stderr,
-        )
-        return 1
-    model = os.environ.get("ANTHROPIC_MODEL", "")
-    base_url = os.environ.get("ANTHROPIC_BASE_URL", "")
-    if not base_url:
-        print("FATAL: ANTHROPIC_BASE_URL not set", file=sys.stderr)
+    try:
+        model_config = resolve_model_config(os.environ)
```
zac-wang-nv: **P1** — Run this validation in the wrapper before fleet selection and lock acquisition, then pass the resolved configuration into the invocation loop. `main()` enters `hold_pool_lock()` before calling `run_invocations()`, so the shared resolver is still too late for the documented direct run_leg.py command and the unchanged daily workflow, neither of which executes the new skills-eval.yml preflight. A direct call with `SKILLS_EVAL_PROVIDER=custom` and no model/endpoint reaches the worker lock before raising; if the pool is busy it can spend the 21,000-second lock budget and report pool exhaustion without ever reporting the invalid route. A regression test should assert that invalid routing never reaches the fleet or lock helpers.

PR #2165, opened 2026-09-11
fix(agent): treat duplicate LVS RT-VLM stream registration as success

hugoverjus:
## Summary
- VST's LVS webhook can register a sensor with RT-VLM before the agent's transactional RTSP ingest reaches the same service.
- Treat an RT-VLM `add` response whose errors are all `DuplicateStreamId` as idempotent success, and keep failing on any other error.
- Cover both the duplicate-success path and a non-duplicate error path in `test_rtsp_ingest.py`.

## Test plan
- [ ] Agent unit tests for `add_to_rtvi_vlm` (`test_duplicate_stream_is_idempotent_success`, `test_non_duplicate_error_still_fails`)
- [ ] GitHub copy-PR CI via `/ok to test` on the pinned head SHA
- [ ] Confirm other RT-VLM error codes still fail ingest


hugoverjus:

PR #2168, opened 2026-09-11
fix(ci): tell downstream infrastructure failure apart from code failure

hugoverjus:
## Summary

The downstream-pipeline GitHub check renders a 17-second git plumbing error and a genuine product assertion failure identically, with the downstream URL masked as a secret. A PR author cannot tell whether a red gate means "your change is broken" or "the CI system had a bad day", so the only rational response is to ask a maintainer, every time.

This teaches the poller to classify each blocking failure, and to report an infrastructure failure as such instead of as a verdict on the pull request.

## Problem

`blocking_jobs` in `.github/scripts/poll-downstream-pipeline.py` fails the check for any job with `status == "failed" and not allow_failure`, plus any `canceled` job, **without ever consulting `failure_reason`**. A `runner_system_failure`, a `get_sources` git abort, an 80-minute `job_execution_timeout` and a real product failure all come out as the same three lines.

`describe_job` already carried the stage, duration and `failure_reason` that separate these cases. Only the verdict logic ignored them.

## Evidence

Verbatim, from GitHub run 34573793808 (job 103272481086), on commit `47c9ee91` where every native GitHub check passed:

```
FAIL: test-base
FAIL: test-search
##[error]Downstream job failed: test-search (stage blueprint eval docker-compose, 51m53s, script_failure)
Downstream pipeline #67368057 finished 'failed': 2 failed, 0 canceled, 14 succeeded, 0 skipped, 1 allowed failures
```

In that run the failures were real. GitLab pipelines `67253826` and `67393139` emitted the **same message shape** for a 17-second git abort in `get_sources`, where no product code ran at all. The downstream host, project path and pipeline URL are all masked, so the author cannot click through and tell them apart either.

## Change

The terminal snapshot now classifies each blocking failure (`infra_failure_cause`):

**Infrastructure** — reported as `DOWNSTREAM_INFRA`, non-blocking, in words that say plainly the downstream pipeline could not run the job and that this is not a verdict on the PR:

- `failure_reason` in `runner_system_failure` / `stuck_or_timeout_failure`
- a `script_failure` under 90s in a `preflight` stage
- a job whose trace never reached the runner's `step_script` section, which proves it died in `get_sources` / `restore_cache` / artifact download before running anything

**Product** — everything else. Still a hard red.

Classification has to *prove* infrastructure. Anything it cannot prove stays a product failure, so an unavailable trace, an empty trace, a missing `failure_reason`, a slow `preflight` job, and `job_execution_timeout` (a job that ran to its wall clock may have been hung by the product) all keep the check red. A false green hides a regression; a false red costs a retry.

Infrastructure failures are still printed when the check goes red for another reason — as warnings rather than as the verdict — so they are never silently dropped.

The report carries a triage handle built only from non-secret values: the downstream pipeline IID and the GitHub run id / attempt that prefix the `VSS_TRIGGER_CORRELATION_ID` sent with the trigger. Nothing new is masked, because a masked handle would be redacted in the log and therefore useless.

`fetch_job_trace_exit_code` was split so the new classifier reuses its trace fetch rather than duplicating it. The trace is fetched at most once per job and only on the terminal snapshot, so the added API cost is bounded by the number of blocking failures.

## Checks run

- `python3 .github/scripts/test_poll_downstream_pipeline.py` — 47 tests, OK. This is the repo's own check for this path (`.github/workflows/ci.yml`). 20 of them are new, one per classification branch.
- `python3 .github/scripts/test_trigger_downstream_pipeline.py`, `python3 .github/scripts/test_cancel_downstream_pipeline.py` — OK, for the shared trace-fetch refactor.
- `python3 .github/scripts/check_copyright_headers.py` — OK.
- `uv run ruff check` on both touched files — identical findings before and after this branch, so no new lint debt. (`.github/scripts` is outside the configured ruff scope.)

## Out of scope

- `trigger_downstream_pipeline.py` re-run behaviour — a separate concern.
- Exporting the full `VSS_TRIGGER_CORRELATION_ID` as a trigger step output. It is generated inside the trigger and not written to `$GITHUB_OUTPUT`, so the handle names its `gh-<run id>-<attempt>-` prefix rather than the exact id. Wiring the output through would make the handle exact, and would touch `trigger_downstream_pipeline.py`.
- The per-tick `FAIL: <job>` announcements are unchanged. Classifying mid-flight would mean fetching traces that are still being written; the final report explicitly retracts those lines when everything turned out to be infrastructure.
- Unmasking the downstream pipeline URL.

## Checklist

- [x] The documentation is up to date with these changes (the module docstring documents the classification and the new exit-code meaning).

### DIFF
--- a/.github/scripts/poll-downstream-pipeline.py
+++ b/.github/scripts/poll-downstream-pipeline.py
@@ -32,6 +32,23 @@ failed and was retried into a pass is not held against the pipeline.
 The check still goes red for any job that is genuinely failed or
 canceled once the pipeline is done.
 
+Not every blocking failure is the pull request's fault, though, and the
+report used to render them identically: a 15-second git abort in
+``get_sources`` on a dirty runner and a genuine product assertion both
+came out as ``FAIL: <job>``, with the downstream URL masked so the
+author could not click through and see which. So the terminal snapshot
+classifies each blocking failure (see ``infra_failure_cause``):
+
+* **Infrastructure** - the downstream CI system could not run the job.
+  Reported as ``DOWNSTREAM_INFRA``, in words that say plainly this is
+  not a verdict on the pull request, and *not* fatal on its own. A
+  non-secret triage handle is printed so a maintainer can find the run.
+* **Product** - everything else. Still a hard red.
+
+Classification has to prove infrastructure; anything unproven stays a
+product failure. A false green hides a regression, a false red costs a
+retry.
+
 Resolving the exit code is non-trivial: many downstream API versions
 do NOT include ``exit_code`` in either the pipeline-jobs listing or
 the per-job detail endpoint, so we fall back to fetching the job's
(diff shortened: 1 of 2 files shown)


hugoverjus:

PR #2170, opened 2026-09-11
fix(agent): cap identical tool calls at two retries

liamy-nv:
The top agent could re-issue the same tool call with the same arguments until it hit the graph recursion limit and returned the canned apology, because a successful result that the plan updater failed to mark complete left the step looking pending. Fingerprint each call by tool name and canonical arguments and allow the original invocation plus two retries; a further exact duplicate returns a skip message pointing the agent at the previous result instead of executing again.

Reservations are taken sequentially before the calls are launched so parallel duplicates in one turn cannot race past the cap.

## Checklist


liamy-nv:
/ok to test [f8c6897](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/2170/commits/f8c6897d76c276545751756ac5de5025cf82f5e0)

services/agent/packages/vss_agents/src/vss_agents/agents/top_agent.py
```diff
@@ -1123,6 +1169,31 @@ async def run_tool(tool: BaseTool | None, tool_call: dict[str, Any]) -> ToolMess
...
+                                    f"Tool: {tool_name}\n"
+                                    f"Args: {_llm_tool_args(tool_call.get('args'))}\n"
+                                    f"Result: {skip_content}"
+                                ),
+                            )
+                        )
+                        return ToolMessage(
+                            name=tool_name,
+                            tool_call_id=tool_call["id"],
+                            content=skip_content,
```
zac-wang-nv, line 1245: **P1** — Preserve the previous failure when an identical call is skipped. This `ToolMessage` defaults to `status="success"`, so after three identical calls raise an exception, the fourth call is skipped but still makes `any_tool_succeeded` true and clears `state.tool_failure`. The next `agent_node` can then return an ungrounded answer instead of relaying the backend error. In planning mode, `_plan_update_node` also treats this skip as a completed tool; with a reasoning-only plan update I reproduced it appending “already completed successfully” even though all three executions failed. Please carry forward the last execution outcome for this fingerprint, or handle skipped calls separately from successful executions, and cover the failed-attempts-then-skip path. The current tests only reach the cap with a tool that succeeds.

Suggested change:

- Treat a skipped call separately from a successful execution in both `any_tool_succeeded` and the plan updater's `completed_tools` bookkeeping.
- Make `DUPLICATE_TOOL_CALL_SKIP_MESSAGE` reflect the recorded outcome for the same tool/argument fingerprint. If all three executions failed, return an error-status `ToolMessage`, preserve the unrecovered failure, and tell the agent: “This tool failed three times: {last_error}. Retry limit reached; no successful result is available.” Only direct it to reuse a successful result when one exists.
- Add a regression test that runs three failing attempts followed by a skipped fourth call, asserting that the failure survives and the plan does not claim successful completion.

liamy-nv: A skipped identical call is no longer treated as success. After three failures the fourth skip stays an error, keeps tool_failure, and the plan updater will not mark the step completed.

services/agent/packages/vss_agents/src/vss_agents/agents/top_agent.py
```diff
@@ -1339,11 +1462,15 @@ async def run_tool(tool: BaseTool | None, tool_call: dict[str, Any]) -> ToolMess
                         logger.warning(f"Tool {tool_call['name']} returned empty content, using placeholder")
                         tool_content = "Tool returned empty content"
 
+                    status = "error" if _tool_response_failed(tool_response) else "success"
+                    _store_identical_tool_call_outcome(
+                        state, tool_call["name"], tool_call.get("args"), status, tool_content
+                    )
```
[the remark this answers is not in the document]
liamy-nv: Fixed

liamy-nv:
/ok to test [e931dec](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/2170/commits/e931dec8f6aa37d60b764fcbee6022abeeeb57ea)

zac-wang-nv:
**P1** — Please only allow the agent to retry failed calls. Once a tool call succeeds within the current request, prevent any further execution with the same tool name and canonical arguments and reuse the successful result. Allow the original attempt plus at most two retries only while calls keep failing. The current attempt-count-only guard still allows a successful call to execute twice more.

Please cover success → skip (one execution), failure → success → skip (two executions), and three failures → stop.

zac-wang-nv (approved):
LGTM

PR #2173, opened 2026-09-12
fix(search): wait for the RTVI-CV text embedder, and retry its warm-up window

hugoverjus:
## The fault

On a fully healthy `dev-profile-search` deployment — every container up,
Elasticsearch populated, `embed_search` returning 42 results — search still
failed:

```
{"code": "workflow_error",
 "message": "Invalid RTVI CV response format: RTVI CV response missing or empty 'data' field",
 "details": "ExecutionFailed"}
```

Only the paths that embed a *query* through RTVI-CV failed. `embed_search`,
which goes through Cosmos Embed, passed with 42 results in the same run;
`attribute_search` retried for its full 180s and then reported the message
above.

`vss-rtvi-cv` had answered `POST /api/v1/generate_text_embeddings` with HTTP
200 and an empty `data` list. It serves HTTP long before its DeepStream
`text-embedder` TensorRT engine is loaded, and nothing waited for that state or
retried it.

## Two halves of one fault

**1. `vss-rtvi-cv` had no healthcheck**, so the agent could start and serve
searches against an embedder that was not ready.

The probe issues a real embedding, because that is the only thing which
distinguishes "loaded" from "listening". `/api/v1/live` and `/api/v1/ready`
both come up earlier and report `YES` in exactly the state that fails — the
eval harness's existing `ds-liveness` check passed in the failing run. The
agent now gates on it with `required: false`, matching how it already waits for
`rtvi-vlm`.

It lives in the `dev-profile-search` override rather than the shared
`services/rtvi/rtvi-cv/compose.yaml` because `DS_VISION_ENCODER` is `true` only
there. The shared definition defaults it to `false`, so a text-embedding probe
in the shared file would mark `perception-2d`, `perception-3d` and
`perception-alerts` **permanently unhealthy** — and, with the agent gated on
`service_healthy`, would deadlock those profiles. That is worth stating plainly
because it is the same trap as the next paragraph, reached by a different route.

`python3`/`urllib.request`, not `curl`: the rt-cv image never installs curl —
it pulls in `netcat-openbsd`, and `services/rtvi/rt-cv/docker/Dockerfile:168`
says so in as many words ("netcat (healthchecks)"). `python3` is present and
already used at build time. A curl probe would have failed permanently and
marked a healthy service unhealthy forever, which is worse than no healthcheck.
This also matches the agent service's own `urllib.request` healthcheck, so the
shape is not new to this repo.

**2. `_fetch_text_embedding` had no retry**, so a single 200-with-empty-data
failed a user-facing search. It now retries that case alone — three attempts,
linear backoff, about 6s — and leaves transport errors and malformed payloads
failing on the first occurrence, since neither is transient. The final message
records that the emptiness persisted across attempts, so a log reader can tell
a slow start from a broken embedder.

Both `RTVICVEmbedClient` copies are patched. `vss_agents/embed` is the one
`attribute_search` uses today; `vss_core/search_core/clients` is the port that
supersedes it. Behaviour should not depend on which one is wired. The
duplication itself is pre-existing and left for separate cleanup.

## Why the healthcheck and the retry are complementary

The healthcheck alone cannot cover a service that degrades *after* it reports
healthy, and the retry alone cannot cover a cold start that takes minutes. They
address the two ends of the same window.

The same point applies to the CI-side gate in the internal harness
(`ci-vss-oss!576`), which fails a search run fast when its pre-flight probe
finds RTVI-CV unable to embed. Measured across 339 `test-search` failures since
1 September, 78 carried this `ExecutionFailed` envelope. That gate catches the
**23** where pre-flight already knew. In the other **55**, pre-flight logged
`✓ ready` and the backend degraded afterward — which is precisely what this PR
addresses. Complementary, not redundant.

One caveat on that envelope, for anyone matching on it: `details:
"ExecutionFailed"` is a generic NeMo Agent Toolkit wrapper and covers more than
one fault. A second cluster of failures with the same string is corpus
truncation on the `embed_search` path (1 indexed document instead of 12–42),
which is unrelated to RTVI-CV and not touched here.

## The probe

```yaml
healthcheck:
  test:
  - CMD
  - python3
  - -c
  - >-
    import json, sys, urllib.request;
    body = json.dumps({'text_input': 'healthcheck', 'model': ''}).encode();
    req = urllib.request.Request(
    'http://localhost:${RTVI_CV_PORT:-9000}/api/v1/generate_text_embeddings',
    data=body, headers={'Content-Type': 'application/json'});
    data = json.load(urllib.request.urlopen(req, timeout=10)).get('data');
    sys.exit(0 if isinstance(data, list) and len(data) > 0 else 1)
  interval: 30s
  timeout: 15s
  retries: 5
  start_period: 1200s
```

`start_period: 1200s` covers phase-0 model download plus the ONNX → TensorRT
build, the same budget `rtvi-embed` and `rtvi-vlm` already allow. The probe
accepts both documented response shapes, `{"data": [{"embedding": [...]}]}` and
`{"data": [[...]]}`.

## Kubernetes / Helm

The `Helm Sync` check flagged that the Compose change had no Helm counterpart,
and its bot prepared one on `helm-sync-bot/pr-2173`. I reviewed that proposal
and adopted it here rather than leaving the drift for a follow-up, so the two
deployment paths do not disagree about what "ready" means.

In Kubernetes the readiness probe does the job that `depends_on:
service_healthy` does in Compose, so no separate agent-side change is needed.
It is gated the same way, on `profileMode: search` and `DS_VISION_ENCODER`:

| Rendering | startupProbe | readinessProbe | livenessProbe |
|---|---|---|---|
| `search`, encoder `true` | embedding probe, 30s x 40 = 1200s | embedding probe, 150s | TCP |
| `search`, encoder `false` | TCP (unchanged) | TCP (unchanged) | TCP |
| `alerts` | untouched | untouched | untouched |

`helm lint` is clean, `helm template` renders all three cases as above, and the
`alerts` profile contains no `python3` probe at all.

## Verification

`helm lint` and `helm template` are covered above. `docker compose config` accepts the healthcheck, interpolates
`${RTVI_CV_PORT:-9000}`, and renders `start_period` as `20m0s`; the folded
block scalar compiles as Python on one line. The full profile cannot be
resolved standalone (`perception-2d-fusion` depends on `sensor-ms` and
`broker-health-check` from sibling files, which the deploy script assembles) —
that is pre-existing and reproduces identically on unmodified `develop`.

The probe was executed against a stub server in every state that matters:

| State | Exit |
|---|---|
| engine loaded, `{"data": [{"embedding": [...]}]}` | 0 |
| engine loaded, `{"data": [[...]]}` | 0 |
| warming up, `{"data": []}` | 1 |
| warming up, no `data` key | 1 |
| service down / connection refused | 1 |

Repository gates:

```
python3 .github/scripts/check_copyright_headers.py   # OK, all 8490 tracked files
ruff check .            (services/agent)             # All checks passed
ruff format --check .   (services/agent)             # 475 files already formatted
pytest packages/vss_core/tests packages/vss_agents/tests
                                                     # 3449 passed, 1 pre-existing failure
```

The one failure is
`test_cosmos_embed.py::TestGetVideoEmbeddingsFromUrls::test_get_video_embeddings_url_formatting`,
which fails identically before this change and is in `cosmos_embed.py` — a file
this PR does not touch.

Eight new tests cover the retry: that an empty `data` list is retried and
recovers, that persistent emptiness fails after exactly three attempts with the
reason preserved, and that neither a malformed payload nor a transport error is
retried.

### DIFF
--- a/deploy/docker/developer-profiles/dev-profile-search/video-analytics-2d-app/compose.yml
+++ b/deploy/docker/developer-profiles/dev-profile-search/video-analytics-2d-app/compose.yml
@@ -58,6 +58,40 @@ services:
       VISION_ENCODER_VERSION: ${VISION_ENCODER_VERSION:-v1.1}
       NGC_CLI_API_KEY: ${NGC_CLI_API_KEY}
       DS_MODEL_DOWNLOAD: auto
+    # Only a real embedding proves readiness. The REST server answers long
+    # before the text-embedder TensorRT engine is built, and until it is this
+    # endpoint returns 200 with an empty `data` list -- which the agent's
+    # RTVICVEmbedClient reports as a failed search, not as "not ready yet".
+    # `/api/v1/live` and `/api/v1/ready` both come up earlier and cannot
+    # distinguish that state.
+    #
+    # This lives here rather than in services/rtvi/rtvi-cv/compose.yaml because
+    # DS_VISION_ENCODER is true only in this profile; the shared definition
+    # defaults it to false, so a text-embedding probe there would mark
+    # perception-2d, perception-3d and perception-alerts permanently unhealthy.
+    #
+    # python3, not curl: the rt-cv image never installs curl (it pulls in
+    # netcat-openbsd for probes) while python3 is present and used at build
+    # time. Matches the agent service's urllib.request healthcheck.
+    healthcheck:
+      test:
+      - CMD
+      - python3
+      - -c
+      - >-
+        import json, sys, urllib.request;
+        body = json.dumps({'text_input': 'healthcheck', 'model': ''}).encode();
+        req = urllib.request.Request(
+        'http://localhost:${RTVI_CV_PORT:-9000}/api/v1/generate_text_embeddings',
+        data=body, headers={'Content-Type': 'application/json'});
+        data = json.load(urllib.request.urlopen(req, timeout=10)).get('data');
+        sys.exit(0 if isinstance(data, list) and len(data) > 0 else 1)
+      interval: 30s
+      timeout: 15s
+      retries: 5
+      # Covers the phase-0 model download plus the ONNX -> TensorRT engine
+      # build, the same budget rtvi-embed and rtvi-vlm allow.
+      start_period: 1200s
     depends_on:
       sensor-ms:
         condition: service_started
--- a/deploy/docker/services/agent/compose.yml
+++ b/deploy/docker/services/agent/compose.yml
@@ -198,6 +198,13 @@ services:
       rtvi-vlm:
         condition: service_healthy
         required: false
+      # Search embeds every query through RTVI-CV, so starting the agent before
+      # the text-embedder engine is built means the first searches fail rather
+      # than wait. Only the search profile declares a healthcheck for it;
+      # required: false keeps every other profile unaffected.
+      perception-2d-fusion:
+        condition: service_healthy
+        required: false
       rtvi-embed:
         condition: service_healthy
         required: false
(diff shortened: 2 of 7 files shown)


hugoverjus:

PR #2176, opened 2026-09-13
fix(rtvi-embed): stop decode-retry telemetry from killing the chunk watcher

hugoverjus:
## Problem

The `rt-embed` service permanently stops delivering chunk results, which wedges
embedding mid-job. The service keeps reporting healthy.

The failure chain:

1. A text-embeddings request (`generate_text_embeddings`) enqueues its chunk
   straight onto a VLM process, skipping the decoder, so its item dict has no
   `decode_retry_count` key. Decoded video-file chunks always carry it.
2. A VLM batch is assembled by appending **per key** across the queued item
   dicts (`vlm_pipeline/process_base.py`). Co-batching a text chunk with a
   decoded chunk therefore leaves `decode_retry_count`'s list **shorter than the
   chunk list**.
3. Unbatching indexes each key by chunk position and falls back to passing the
   value through when it cannot index (`IndexError`), so the overflow chunk
   receives the **whole list**. `PipelineChunkResult.decode_retry_count` is
   declared `int = 0` but ends up holding a list.
4. A list is truthy even when it holds only zeros, so the `if
   chunk_result.decode_retry_count:` gate fires for chunks that never retried,
   and `Counter.add(<list>, ...)` raises `TypeError`.
5. That emission sits in `_on_vlm_chunk_response`, which runs on the pipeline's
   processed-chunk **watcher thread**. The raise kills the thread:

```
Exception in thread Thread-3 (_watch_processed_chunk_queue):
  File ".../vlm_pipeline/vlm_pipeline.py", line 1861, in _watch_processed_chunk_queue
  File ".../server/rtvi_stream_handler.py", line 2644, in _on_vlm_chunk_response
  File ".../opentelemetry/sdk/metrics/_internal/instrument.py", line 166, in add
TypeError: '<' not supported between instances of 'list' and 'int'
```

After that, chunk results are never delivered again: the in-flight video-file
embedding job halts mid-flight, the Elasticsearch index stops filling, and
subsequent `generate_text_embeddings` requests hang forever.

### CI impact

This is the root cause of an internal `test-search` job sitting at roughly a
23% pass rate (about 102 passed / 339 failed / 90 canceled across 545
attempts). The crash happens on essentially every run; only its *timing*
decides pass or fail. If the watcher survives long enough for the first
42-chunk embedding job to populate the index, the job passes and the crash
lands in a later job; if it lands during that first job, the job fails.

## Changes

Narrowest first, in `services/rtvi/rt-embed`:

- **`vlm_pipeline/vlm_pipeline.py:2033,2066`** — the two enqueue paths that skip
  the decoder (`enqueue_text_only_chunk`, `enqueue_text_chunk`) now pass
  `decode_retry_count=0`, so batches stay key-homogeneous and unbatching keeps
  producing ints. This is the root-cause fix.
- **`vlm_pipeline/vlm_pipeline.py:87` and `:1782`** — `_coerce_decode_retry_count`
  reduces the value to an `int` where the `PipelineChunkResult` is built. A
  leaked batch list is **summed** rather than dropped: its entries are real
  retry counts, so summing keeps the signal instead of silently losing it, and
  the common `[0]` case reduces to `0`. This is defence in depth so a future
  producer cannot re-introduce a non-int.
- **`server/rtvi_stream_handler.py:2643`** — the counter emission is guarded and
  logs at `error` with a traceback on failure. Telemetry must never be able to
  kill the watcher thread.
- **`vlm_pipeline/vlm_pipeline.py:1692`** — every chunk-result callback in
  `_watch_processed_chunk_queue` now goes through
  `_invoke_chunk_result_callback`, which logs a faulting callback instead of
  letting it end the thread, so one bad chunk cannot stop delivery for the rest
  of the process lifetime.

Genuine pipeline faults are untouched — only the observability side effect is
made non-fatal, and every swallowed exception is logged with its traceback.

## What was verified vs. what is bounded

**Verified** by driving the real `ProcessBase` batching and unbatching code in
process (this is what the new `TestRaggedBatchUnbatching` pins):

- Ragged per-key batching plus position-indexed unbatching *does* yield a list
  for the overflow chunk — reproduced deterministically, not inferred.
- `bool([0]) is True`, so the counter fires even with zero real retries.
- A real OpenTelemetry counter raises `TypeError` on a list and accepts the
  coerced value.
- Only `rt-embed` is reachable: both its models (`cosmos-embed1`, `ce1-nim`)
  return `can_batch() == True` unconditionally, so batches of >1 form. No
  `rt-vlm` model implements `can_batch`, so `_can_batch` refuses to batch there
  and the ragged path cannot occur today. `rt-vlm` carries the same counter
  emission and watcher code but is deliberately left alone to keep this change
  scoped — worth a follow-up if a batching `rt-vlm` model ever lands.

**Bounded uncertainty** — stated plainly rather than tidied up:

- I did **not** capture a live instance of the misaligned batch from a failing
  CI run's logs. The mechanism above is proven *sufficient and reachable* from
  the code, and it matches the observed timing behaviour, but the causal link to
  any specific CI failure is inference.
- Whether a text chunk and a decoded chunk actually co-batch in a given run
  depends on queue timing and `vlm_batch_size`; I have not measured how often
  that window is hit.
- The exact `TypeError` message differs by OpenTelemetry version (the reported
  traceback shows `'<' not supported ...` from `if amount < 0:`; current SDK
  raises `must be real number, not list`). The defect and its fatal outcome are
  version-independent.
- There may be other producers of a non-int `decode_retry_count` that I have not
  found. The coercion and the callback guard make the fix robust to both shapes
  regardless of origin, which is why they are included rather than relying on
  the enqueue fix alone.

## Tests

New `tests/rtvi_embed/test_decode_retry_telemetry.py` (17 tests) pins:

- Coercion: int/zero pass through; list and tuple reduce by sum; `[0]` → `0`;
  empty → `0`; unusable values → `0`.
- The ragged-batch mechanism itself, and that uniform keys unbatch to ints.
- That the text-chunk enqueue path supplies the key, and that a text item then
  batches evenly with a decoded item.
- That a raising callback does **not** kill the watcher and the next chunk is
  still delivered; that `_invoke_chunk_result_callback` logs rather than raises.
- End to end over the watcher: a leaked list reaches the callback as an `int`.
- That a real OTEL counter rejects the pre-fix value, accepts the coerced one,
  and records `3` for `[2, 1]`.

Each of the three production changes was reverted individually to confirm the
corresponding tests fail without it, so none of them is vacuous.

The tests are hermetic — no network, no Docker, no GPU — so they are added to
the existing CPU-only `rtvi-vllm-compatible-test` lane, along with the
pure-Python imports `vlm_pipeline` pulls in (`aiofiles`, `boto3`, `grpcio`,
`nvtx`, `opentelemetry-*`, `pyyaml`, `requests`). The lane's exact command was
simulated locally against that dependency set: **21 passed**.

## Verification

- `tests/model` on pristine `develop`: 4 passed. Same plus the new file on this
  branch: **21 passed**.
- The exact CI lane command: **21 passed**.
- Copyright headers: `check_copyright_headers.py` → all 8489 tracked files OK.
- `yamllint` with the repo's CI options on `ci.yml`: clean.
- `ruff check` / `ruff format --check` report findings on the two touched `src`
  files, but **identically on pristine `develop`** — the repo's ruff gate is
  scoped to `services/agent`, and the `rtvi` tree has never been ruff-managed.
  The new test file is clean under `ruff format`. Baselined, not introduced here.
- `tests/rtvi_embed` (beyond the new file), `tests/kafka` and `tests/redis`
  cannot be collected outside the container (a redis module `sys.exit(1)`s at
  import; kafka needs `protobuf`). Confirmed identical on pristine `develop`.


hugoverjus:

PR #2178, opened 2026-09-14
feat(ui): wire HITL responses through the OpenClaw agent adapter

ashan-nv:
## Summary
- VSS UI's chat sidebar could see that an OpenClaw agent run paused to ask the user a question (`interaction.required`), but had no way to answer it — `interaction_responses` was hardcoded `false`, the OpenClaw connector silently dropped any paused-turn signal, and `POST /runs/:runId/respond` was a stub always returning `409`.
- Wires a full `respond()` path through the adapter: the OpenClaw connector now detects a paused chat state and emits `interaction.required` with the question text; `Connector.respond()` sends the reply back over the same WebSocket session; the run/thread stay "active" (not terminal) while paused so a reply doesn't race a new message; the `/respond` route dispatches into the service.
- On the frontend, the `agent-api` transport now emits the same `{kind: 'interaction'}` chat event the legacy transport already has working inline-prompt UI for (`ChatPanel.tsx`), routed to the new `POST /runs/:runId/respond` REST endpoint instead of the legacy query-param scheme.

## Known limitation — please read before relying on this
No live OpenClaw gateway was reachable to observe real frames during an `AskUserQuestion`-style pause, and the public OpenClaw protocol docs don't document this case (they cover `chat.send`/`chat.abort`/`sessions.send` and a separate tool-exec `approval.*` family, but nothing for a paused chat turn awaiting structured input).

The connector's guess (flagged inline in `openClaw.ts`):
- A paused turn arrives as a `chat` event with `state: "awaiting_input"` or `state: "interaction"`, using the same `message` shape as a final response for the question text.
- A reply continues the paused turn via another `chat.send` RPC on the same `sessionKey` (there's no `chat.respond` in the handshake's advertised methods).

**This needs to be verified against a live OpenClaw gateway** (capture real frames during a genuine `AskUserQuestion` pause) before depending on it in production. If the real state name / payload shape differs, only the two flagged blocks in `openClaw.ts` need updating — the service/store/API/frontend plumbing around it is protocol-agnostic.

## Test plan
- [x] `packages/nv-metropolis-bp-vss-ui/chat`: `npx jest --runInBand agentApi useChatStream sse` — 23/9 passed (added 2 new tests covering the interaction event shape and the missing-prompt fallback)
- [x] `packages/nv-metropolis-bp-vss-ui/chat`: `npx tsc --noEmit` — no new errors (one pre-existing unrelated failure in `ChatUpload.tsx`, confirmed present on `develop` too)
- [ ] Backend `apps/nv-metropolis-bp-vss-ui` jest suite (`agentAdapter*.test.ts`) could not be run — pre-existing `ts-jest`/TypeScript version mismatch (`moduleResolution: "bundler"` unsupported by the pinned TS 4.9.5) unrelated to this change, confirmed present on `develop`. Backend changes were manually reviewed instead.
- [ ] End-to-end verification against a live OpenClaw gateway (see limitation above)


ashan-nv:
Superseded by #2182, which implements OpenClaw's documented `question.requested`/`question.resolved`/`question.resolve` protocol instead of the `awaiting_input` + second `chat.send` guess this PR was built on. That guess was speculative (no live gateway was reachable to verify it) and turned out not to match OpenClaw's actual protocol. Closing this in favor of the protocol-correct implementation.

PR #2183, opened 2026-09-14
fix(nemoclaw): expose gateway to containerized UI

ashan-nv:
## Summary

- bind the OpenClaw dashboard/gateway forward to the Docker bridge gateway when the VSS agent adapter is enabled outside Brev
- keep the default non-adapter deployment loopback-only and preserve the existing Brev `0.0.0.0` behavior
- validate the discovered Docker bridge address and probe the actual selected interface
- print the matching SSH tunnel destination for non-Brev access
- cover loopback, Brev, adapter, discovery-failure, and rebind paths

## Why

The VSS UI runs in a container and reaches the host through `host.docker.internal`. A loopback-only OpenClaw forward therefore appears unreachable even though the host-local health probe succeeds.

This operational networking fix is intentionally separate from the structured HITL implementation in #2182.

## Testing

- `uv run --with pytest --with nbformat pytest -q deploy/docker/scripts/tests/test_run_setup_notebook.py` — 44 passed
- Ruff lint passed for the updated test file
- notebook JSON and diff validation passed


deploy/docker/scripts/deploy_nemoclaw.ipynb
```diff
@@ -1606,9 +1615,34 @@
...
+        "\n",
+        "_docker_bridge = (\n",
+        "    _docker_bridge_gateway() if VSS_AGENT_ADAPTER_ENABLED and not _chat_fqdn else None\n",
+        ")\n",
+        "_desired_bind = \"0.0.0.0\" if _chat_fqdn else (_docker_bridge or \"127.0.0.1\")\n",
         "# Onboard starts the dashboard forward, but it dies with the kernel or terminal that\n",
         "# launched it and a rebuild/recover drops it; the UI then 503s. Re-establish it here.\n",
-        "_health = f\"http://127.0.0.1:{AGENT_DASHBOARD_PORT}/health\"\n",
+        "_health_host = \"127.0.0.1\" if _desired_bind == \"0.0.0.0\" else _desired_bind\n",
+        "_health = f\"http://{_health_host}:{AGENT_DASHBOARD_PORT}/health\"\n",
```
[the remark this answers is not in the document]
ashan-nv: Addressed in 0c908daf6. Added behavioral tests that execute the checked-in dashboard-forwarding cell with mocked host I/O for loopback, Brev wildcard, adapter bridge selection, bridge lookup failure, and mismatched-bind rebind/readiness. Bridge lookup failures now also report a concise runtime error. All 44 notebook contract tests pass.

ayyappa-dev:
Are notebook updates still in scope ?

deploy/docker/scripts/deploy_nemoclaw.ipynb
```diff
@@ -1527,9 +1534,28 @@
         "\n",
         "_chat_fqdn = brev_secure_link_fqdn(AGENT_DASHBOARD_PORT)\n",
         "\n",
+        "\n",
```
ayyappa-dev: vss-ui connects through host.docker.internal, but this binds to the default bridge gateway; those addresses can differ when Docker’s host-gateway-ip is configured. In that configuration, the UI cannot reach the gateway even though the notebook health check passes.
  Please derive the bind IP from the same host-gateway configuration used by vss-ui (or make one explicit shared setting drive both), and add coverage for a non-default host-gateway IP.

deploy/docker/scripts/deploy_nemoclaw.ipynb
```diff
@@ -1527,9 +1534,28 @@
...
+        "        raise RuntimeError(f\"Could not resolve Docker's bridge gateway: {detail}\")\n",
+        "    candidate = inspected.stdout.strip()\n",
+        "    try:\n",
+        "        return str(ipaddress.IPv4Address(candidate))\n",
+        "    except ipaddress.AddressValueError as exc:\n",
+        "        raise RuntimeError(f\"Docker returned an invalid bridge gateway: {candidate!r}\") from exc\n",
+        "\n",
+        "\n",
+        "_docker_bridge = _docker_bridge_gateway() if VSS_AGENT_ADAPTER_ENABLED and not _chat_fqdn else None\n",
+        "_desired_bind = \"0.0.0.0\" if _chat_fqdn else (_docker_bridge or \"127.0.0.1\")\n",
```
zac-wang-nv: **P1** — When the adapter is enabled off Brev, the forward now listens only on the Docker bridge, but the optional verification cell still posts to `http://127.0.0.1:{AGENT_DASHBOARD_PORT}/hooks/agent` at deploy/docker/scripts/deploy_nemoclaw.ipynb:1799. With OpenClaw hooks enabled (the default), running section 3.6 after a successful setup therefore reports a transport failure and tells the user to restart a healthy forward. I reproduced this by executing the checked-in forwarding cell and webhook verification block with mocked host I/O: the health check used `172.18.0.1:18789`, while verification used `127.0.0.1:18789` and printed `hooks test: FAIL`. Update the host-side webhook probe to use the selected reachable interface (including proxy bypass), or run it inside the sandbox, and cover verification after bridge rebinding.

ashan-nv:
Yes—notebook updates remain in scope for this PR. The dashboard-forward selection and webhook verification live in `deploy_nemoclaw.ipynb`, with behavioral coverage in `test_run_setup_notebook.py`. Follow-up commit 02649370d makes the bind follow the actual `host.docker.internal` mapping used by `vss-agent-ui` and verifies hooks through that interface.

deploy/docker/scripts/deploy_nemoclaw.ipynb
ashan-nv: Addressed in 02649370d. The notebook now resolves `host.docker.internal` from inside the running `vss-agent-ui` container and binds the forward to that exact IPv4 address, so Docker daemon `host-gateway-ip` overrides no longer diverge from the notebook bind. The behavioral test uses a non-default `192.0.2.44` mapping and verifies selection, health probing, and rebinding.

deploy/docker/scripts/deploy_nemoclaw.ipynb
ashan-nv: Addressed in 02649370d. OpenClaw hook verification now posts through the selected dashboard health host instead of hard-coded loopback and passes `curl --noproxy "*"`. The new behavioral test first rebinds from loopback to a non-default `host.docker.internal` address, then executes the checked-in hooks verification block and asserts that address and proxy bypass are used. All 45 notebook contract tests pass.

deploy/docker/scripts/deploy_nemoclaw.ipynb
```diff
@@ -1480,9 +1481,33 @@
...
     "\n",
+    "\n",
+    "def _vss_ui_host_gateway():\n",
+    "    resolver = (\n",
+    "        \"require('dns').lookup('host.docker.internal',{family:4},\"\n",
+    "        \"(error,address)=>{if(error){console.error(error.message);process.exit(1)}\"\n",
+    "        \"process.stdout.write(address)})\"\n",
+    "    )\n",
+    "    resolved = subprocess.run(\n",
+    "        [\"docker\", \"exec\", \"vss-agent-ui\", \"node\", \"-e\", resolver],\n",
```
nvleeryan: wouldn't it be better with `docker network inspect bridge --format '{{(index .IPAM.Config 0).Gateway}}'` instead of docker exec into the vss-agent-ui so the notebook doesn't depend on the running vss-agent-ui?

nvleeryan (dismissed):
I would prefer something that doesn't require the nemoclaw notebook to depend on a running vss-agent-ui

deploy/docker/scripts/deploy_nemoclaw.ipynb
ashan-nv: Addressed in a767ff800. The notebook no longer execs into the running UI. It gets the already-local UI image ID with `docker inspect` (which works when the UI is stopped), then runs a disposable `docker run --rm --add-host host.docker.internal:host-gateway` probe to read the mapping Docker actually materializes. I did not use the default bridge gateway directly because it can differ from `host.docker.internal` when dockerd has a custom `--host-gateway-ip`, which is the case Ayyappa identified. The focused suite now covers a non-default mapping; all 47 notebook tests pass.

ashan-nv: You are right: my previous fix removed `docker exec`, but it still inspected the `vss-agent-ui` container for its image ID. Corrected in b50d01ed2. The notebook now inspects the already-required OpenShell `gateway_container` image and runs one short-lived `docker run --rm` probe with `host.docker.internal:host-gateway`; a POSIX shell builtin reads the mapping Docker materializes in `/etc/hosts`. There is no `vss-agent-ui` reference in the lookup path now. I retained a Docker-materialized probe rather than using only the default bridge gateway because a daemon-level `--host-gateway-ip` override can make those addresses differ. The test asserts the OpenShell image is used and that no Docker command mentions `vss-agent-ui`; all 47 notebook tests pass.

ayyappa-dev:
# Follow-up TODOs — `fix(nemoclaw): expose gateway to containerized UI` (#2183)

All four are non-blocking for the primary happy-path flow; none break the
core networking fix. Log as follow-up items unless noted.

- [ ] **TODO-1: Docker bridge discovery runs eagerly on every cell execution**
  - Location: `deploy/docker/scripts/deploy_nemoclaw.ipynb:1520` (`_docker_gateway = _docker_host_gateway() if VSS_AGENT_ADAPTER_ENABLED and not _remote_bind else None`; function at L1486, probe `docker run` ~L1504–1512)
  - Detail: discovery is computed unconditionally before the health/ownership check, so every re-run of the UI cell spawns a throwaway `docker run --rm` container. A transient Docker hiccup, an unpullable/unrunnable gateway image, or `gateway_container` not being set yet raises `RuntimeError` and aborts the whole UI setup — even when a correct forward is already up and reusable.
  - Root cause: discovery is eager and unconditional; needed to compute `_desired_bind`, so it can't simply be deferred, but it isn't cached or guarded.
  - Probable fix: cache the resolved gateway (env var / small state file) across cell runs, and/or resolve only when the existing bind isn't already a reachable non-loopback IPv4; on discovery failure, degrade to loopback with a warning instead of hard-aborting.

- [ ] **TODO-2: `--entrypoint sh` assumes the OpenShell gateway image ships a POSIX shell**
  - Location: `deploy/docker/scripts/deploy_nemoclaw.ipynb:1507` (the `--entrypoint sh` probe inside `_docker_host_gateway`)
  - Detail: the probe runs `docker run --rm --entrypoint sh <gateway-image> -c <resolver>`. If the gateway image is minimal/distroless with no `sh` in PATH, discovery fails with "Could not resolve Docker's host-gateway mapping" — misleading, since the actual `host.docker.internal` mapping is fine.
  - Root cause: reusing the gateway image to avoid a new dependency assumes it contains a shell; the failure surfaces as a host-gateway resolution error.
  - Probable fix: verify the gateway image has `sh` before merge (`docker run --rm --entrypoint sh <gateway-image> -c 'echo ok'`); if not, use a shell-bearing image (busybox/alpine) for the probe or add a `docker network inspect bridge --format '{{(index .IPAM.Config 0).Gateway}}'` fallback. **Cheap to verify — check this one before deferring.**

- [ ] **TODO-3: Binding to the default-bridge gateway exposes the dashboard to all default-bridge containers**
  - Location: `deploy/docker/scripts/deploy_nemoclaw.ipynb:1521` (`_desired_bind = "0.0.0.0" if _remote_bind else (_docker_gateway or "127.0.0.1")`)
  - Detail: the listen address is the IP Docker maps `host.docker.internal` to — i.e. the default-bridge gateway. Any container on the default `bridge` network (not just the VSS UI container) can then reach the dashboard port on that host IP.
  - Root cause: the shared default bridge's gateway is a broad address; there's no network isolation scoping reachability to just the UI container.
  - Probable fix: place the UI container and the forward on a dedicated user-defined Docker network and bind to that network's gateway, so only the UI container (and the host) can reach the forward.

- [ ] **TODO-4: Hooks-verification cell depends on `_health_host` from the earlier forwarding cell**
  - Location: `deploy/docker/scripts/deploy_nemoclaw.ipynb:1775` (`hooks_url = f"http://{_health_host}:..."`; `_health_host` defined at L1524 in the `s37-ui-code` cell)
  - Detail: the verify cell now builds the webhook URL from the runtime variable `_health_host` instead of the old hard-coded `127.0.0.1`. Running the verify cell standalone (or after a kernel restart without re-running the forwarding cell) raises `NameError` — a regression from the prior self-contained loopback probe.
  - Root cause: the two cells are now coupled through a module-level derived name rather than each computing their own host.
  - Probable fix: make the verify cell self-contained — recompute the bind host there, or fall back via `namespace.get("_health_host", "127.0.0.1")`. Small, isolated; can be done inline in the follow-up.

hugoverjus (approved):

PR #2187, opened 2026-09-15
feat(agent): make structured HITL opt-in

ashan-nv:
## Summary

- default structured HITL to off across vss-agent tools, Docker profiles, Helm profiles, and both VSS UI chat surfaces
- expose one explicit HITL_ENABLED opt-in for legacy vss-agent deployments while suppressing the legacy modal on adapter-backed surfaces
- configure NemoClaw/OpenClaw to ask required follow-up questions as ordinary completed chat turns and continue on the same session
- wire the launch-safe setting through both deployment notebooks and document the behavior for skills and operators

This supersedes the structured OpenClaw HITL direction in #2182 for launch. The two PRs should not be merged together.

## Validation

- uv run ruff check packages/vss_agents/src
- uv run ruff format --check packages/vss_agents/src
- uv run mypy packages/vss_agents/src/vss_agents
- 53 focused vss-agent HITL tests passed
- UI typecheck passed across all 10 packages
- 94 chat package tests passed
- 8 focused UI integration/connector tests passed
- 37 notebook contract tests passed
- live legacy vss-agent smoke: interaction prompt rendered, response returned HTTP 204, and the run resumed
- live OpenClaw smoke: follow-up rendered as a normal assistant message and the run completed; adapter reports interaction_responses=false


skills/operations/vss-generate-video-report/SKILL.md
```diff
@@ -179,6 +179,12 @@ Resolve HITL mode for **Mode A only** in this order:
 2. Harness override `HITL_ENABLED=true|false` (fallback only when runtime config is unavailable)
 3. If neither source is set, default to `false`
 
+The agent cannot introspect `video_report_gen`'s in-process Pydantic config.
```
zac-wang-nv: why do we need this?

skills/vss-build-vision-ai/SKILL.md
```diff
@@ -64,6 +64,20 @@ metadata:
 
 ## Entry Mode (Step 0)
 
+### Question transport
+
+Obey the harness environment's `HITL_ENABLED` setting. Unless it is explicitly
+`true`, every instruction below to use `AskUserQuestion` means: render the same
+question and choices as ordinary assistant chat text, end the turn, and
```
zac-wang-nv: we shouldnt need this, no?  documenting from ui reference shall be enough

deploy/docker/scripts/deploy_vss_orchestrator.ipynb
```diff
@@ -153,6 +162,25 @@
...
+        "VSS_AGENT_ADAPTER_ENABLED = (\n",
+        "    os.environ.get(\"VSS_AGENT_ADAPTER_ENABLED\", str(VSS_AGENT_ADAPTER_ENABLED)).strip().lower()\n",
+        "    in (\"1\", \"true\", \"yes\", \"on\")\n",
+        ")\n",
+        "VSS_AGENT_BACKEND_PROTOCOL = (\n",
+        "    VSS_AGENT_BACKEND_PROTOCOL or os.environ.get(\"VSS_AGENT_BACKEND_PROTOCOL\", \"openclaw-ws\")\n",
+        ").strip().lower()\n",
+        "VSS_AGENT_BACKEND_URL = (\n",
+        "    VSS_AGENT_BACKEND_URL or os.environ.get(\"VSS_AGENT_BACKEND_URL\", \"\")\n",
+        ").strip() or f\"ws://host.docker.internal:{DASHBOARD_PORT}\"\n",
```
zac-wang-nv: **P1** — On a non-Brev Linux host, the companion deploy_nemoclaw.ipynb keeps an existing loopback dashboard forward (line 1711) or starts the default loopback forward when there is no secure-link FQDN (line 1735). Enabling VSS_AGENT_ADAPTER_ENABLED with the new default URL therefore makes both VSS chat surfaces connect to the host gateway address, where nothing is listening: host.docker.internal is mapped to host-gateway in the UI Compose service, not host loopback. The notebook can finish and the OpenClaw UI can work through SSH while every VSS chat request fails to connect. Establish and verify a Docker-reachable private forward when enabling the adapter, or require a reachable backend URL before starting it.

deploy/docker/scripts/deploy_vss_orchestrator.ipynb
```diff
@@ -776,6 +810,14 @@
         "    env[\"NVIDIA_API_KEY\"] = NVIDIA_API_KEY\n",
         "if HARDWARE_PROFILE:\n",
         "    env[\"HARDWARE_PROFILE\"] = HARDWARE_PROFILE\n",
+        "env[\"HITL_ENABLED\"] = \"true\" if HITL_ENABLED else \"false\"\n",
+        "if VSS_AGENT_ADAPTER_ENABLED:\n",
+        "    env[\"VSS_AGENT_ADAPTER_ENABLED\"] = \"true\"\n",
+        "    env[\"VSS_AGENT_BACKEND_PROTOCOL\"] = VSS_AGENT_BACKEND_PROTOCOL\n",
```
zac-wang-nv, line 826: **P1** — This replaces backend settings that the previous notebook inherited through os.environ.copy(). For example, exporting VSS_AGENT_ADAPTER_ENABLED=true, VSS_AGENT_BACKEND_PROTOCOL=responses, VSS_AGENT_BACKEND_URL=http://agent.local:8642 and VSS_AGENT_BACKEND_PATH=/v1/responses now produces protocol openclaw-ws and path /. The settings cell always keeps its nonempty openclaw-ws literal before checking the protocol environment override (lines 169–171), so the adapter rejects the unchanged HTTP URL with "AGENT_BACKEND_URL must use ws: or wss:". Even editing the notebook protocol to responses still overwrites its endpoint path with /. Honor the explicit protocol/path overrides and select OpenClaw defaults only when those settings are absent.

skills/vss-build-vision-ai/SKILL.md
ashan-nv: Agreed. Removed the skill-local question-transport section in f2a7b9d95. The harness-wide behavior remains owned by `.openclaw/workspace/_nemoclaw/AGENTS.md`, while the adapter/UI documentation explains the surface behavior, so `vss-build-vision-ai` stays harness-independent.

deploy/docker/scripts/deploy_vss_orchestrator.ipynb
ashan-nv: Fixed in f2a7b9d95. Explicit environment values now win for `VSS_AGENT_BACKEND_PROTOCOL`, `VSS_AGENT_BACKEND_URL`, `VSS_AGENT_BACKEND_PATH`, and the token. The server environment propagates the resolved path instead of hard-coding `/`. Added a regression test using the exact `responses` / `http://agent.local:8642` / `/v1/responses` combination from this review.

deploy/docker/scripts/deploy_vss_orchestrator.ipynb
ashan-nv: Fixed in f2a7b9d95. The orchestrator notebook no longer invents `ws://host.docker.internal:<dashboard-port>`. Enabling the adapter now requires an explicit `VSS_AGENT_BACKEND_URL` documented as Docker-reachable, and fails before startup when it is absent. This preserves the non-Brev loopback-only dashboard forward instead of exposing it just to satisfy the adapter default. Added a contract test for the missing-URL failure.

skills/operations/vss-generate-video-report/SKILL.md
ashan-nv: Agreed. Removed this paragraph in 36867a3c8. OpenClaw does not need to inspect the in-process `video_report_gen` configuration; it follows its own harness `HITL_ENABLED` setting, while legacy `vss-agent` reads the runtime field internally.

ashan-nv: yeah, I had this because the agent might try to use the pydantic config file to get the HITL value. That config isn’t accessible in OpenClaw, so it needs to use the HITL_ENABLED environment variable instead

ashan-nv: It's already handled at the harness level so we can remove

skills/vss-build-vision-ai/SKILL.md
ashan-nv: okay sure, will remove

deploy/docker/scripts/deploy_vss_orchestrator.ipynb
ashan-nv: done

deploy/docker/scripts/deploy_vss_orchestrator.ipynb
ashan-nv: done

zac-wang-nv (dismissed):

nv-zhliu (approved):

deploy/helm/services/ui/values.yaml
```diff
@@ -201,6 +201,12 @@ env:
     value: "true"
   - name: NEXT_PUBLIC_CHAT_HISTORY_DEFAULT_ON
     value: "true"
+  # Structured interactions are opt-in. Set both flags to "true" only for a
+  # vss-agent deployment configured to emit legacy interaction frames.
+  - name: NEXT_PUBLIC_ENABLE_HITL
+    value: "false"
+  - name: NEXT_PUBLIC_SIDEBAR_CHAT_ENABLE_HITL
+    value: "false"
```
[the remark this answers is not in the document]
ashan-nv: Fixed

PR #2211, opened 2026-09-15
fix(alerts): Update alert service realtime to prevent stream removal race condition

hfarooq-nv:
## Summary
Bug: In services/alert/src/realtime/services/realtime_service.py, when an always-on camera is removed (camera_remove from VIOS), Alert Bridge tore down the RT-VLM side by firing two calls concurrently via asyncio.gather:
- stop_captions() → RT-VLM DELETE /v1/generate_captions/{stream_id}
- stop_stream() → RT-VLM DELETE /v1/streams/delete/{stream_id}

Both calls route through the same underlying RT-VLM method (remove_rtsp_stream), which is guarded by a per-stream mutex that rejects a second concurrent teardown request with HTTP 409. Whichever call lost the race got the 409. When it was /streams/delete that lost, RT-VLM's asset cleanup (cleanup_asset) never ran, so the stream record was permanently orphaned — still visible in GET /v1/streams/get-stream-info forever. Alert Bridge logged the failure but removed its own rule tracking anyway and reported success, so a retried camera_remove found nothing to retry and never reconciled the orphan.

Fix: stop running the two calls concurrently — sequence them so stop_captions fully completes before stop_stream starts (in both _safe_teardown_rtvi and _safe_teardown_rtvi_with_outcome). Since both calls ultimately do the same underlying pipeline teardown, there was never a real concurrency benefit; sequencing eliminates the self-inflicted race entirely — by the time stop_stream runs, RT-VLM's mutex is already free, so cleanup_asset always executes and the stream record is never orphaned.

## Checklist


services/alert/src/realtime/services/realtime_service.py
```diff
@@ -2317,9 +2323,14 @@ async def _stop_stream() -> bool:
...
-            captions_ok, stream_ok = await asyncio.gather(
-                _stop_captions(), _stop_stream(),
-            )
+            # Sequenced, not gathered: both calls tear down the same
+            # RTVI-side pipeline for rtvi_stream_id, and RTVI 409s a
+            # second concurrent teardown for the same stream — which
+            # can abort /streams/delete before it removes the stream
+            # record, orphaning it. Waiting for stop_captions to
+            # finish first means stop_stream never contends with it.
+            captions_ok = await _stop_captions()
```
dineshvenkat, line 2516: Each RT-VLM call uses the 10s timeout from config.yaml:324, so running them one after the other makes a last-reader teardown take up to 20s instead of 10s. The caller that drives this path gives up after 10s (WDM_ADD_REMOVE_REQUEST_TIMEOUT in the warehouse sdrc config), so it can time out and retry while the first teardown is still in progress.

dineshvenkat:
Two notes on code this change does not touch, both about whether the stale-stream case is fully closed.

**`realtime_service.py` around 1456-1461 (rule pop, then ref-count, then teardown)**

Two alert rules can share one RT-VLM stream. If both are stopped at the same time, each one removes its own rule first, then checks whether anyone else is still using the stream — and both see nobody left. So both go on to delete the same stream, and the two deletes reach RT-VLM together. That is the same collision this change fixes for a single request, still reachable when two requests overlap.

**`realtime_service.py:1456` (rule deleted before teardown)**

The rule is deleted before the teardown runs, and a failed stream delete is only logged — the call still returns success. The always-on layer then drops its camera entry on that success. So if the delete genuinely fails, nothing is left that knows the stream is still there, and when VIOS retries the removal it finds no rule and reports it as already done.

hfarooq-nv:
> Two notes on code this change does not touch, both about whether the stale-stream case is fully closed.
> 
> **`realtime_service.py` around 1456-1461 (rule pop, then ref-count, then teardown)**
> 
> Two alert rules can share one RT-VLM stream. If both are stopped at the same time, each one removes its own rule first, then checks whether anyone else is still using the stream — and both see nobody left. So both go on to delete the same stream, and the two deletes reach RT-VLM together. That is the same collision this change fixes for a single request, still reachable when two requests overlap.
> 
> **`realtime_service.py:1456` (rule deleted before teardown)**
> 
> The rule is deleted before the teardown runs, and a failed stream delete is only logged — the call still returns success. The always-on layer then drops its camera entry on that success. So if the delete genuinely fails, nothing is left that knows the stream is still there, and when VIOS retries the removal it finds no rule and reports it as already done.

addressed in latest commit

services/alert/src/realtime/services/realtime_service.py
hfarooq-nv: WDM_ADD_REMOVE_REQUEST_TIMEOUT is an SDR variable set for the warehouse config. 

We can either increase that to 20sec, give each of these calls 1/2 of rtvi_vlm.timeout to timeout, or change the rtvi_vlm.timeout config value to 5 though this will affect other calls as well. is there a preferred method?

dineshvenkat:
Two follow-ups on the new commit. Not repeating the two Greptile P1 comments — both look right to me.

**`always_on_service.py` around 541 (tracked removal path)**

The new recovery path only runs when a later camera_remove arrives and finds nothing tracked. The removal that actually failed still returns success and drops its camera entry, so there is no second attempt for that failure on its own — recovery depends on another removal event arriving later. If the probe or delete inside that later attempt also fails, it returns success too.

**`test/unit/api/test_realtime_service.py` around 979 (persistent teardown tests)**

The new tests cover ordering and concurrent deletes on the persistent path, but none makes stop_captions fail there and check that stop_stream still runs. That error handling is duplicated in the persistent variant, and the only test covering it is on the in-memory helper, so the persistent copy can change without any test failing.

services/alert/src/realtime/services/always_on_service.py
```diff
@@ -509,6 +509,21 @@ async def stop_camera(self, camera_id: str) -> AlwaysOnResult:
...
 
+        if not tracked:
+            # Nothing tracked for this camera — either already removed,
+            # or its rule lost the last-reader race in
+            # RealtimeAlertService and got dropped despite RTVI teardown
+            # failing. Always-on rules register RTVI streams under
+            # sensor_id=camera_id, so reconcile directly against RTVI:
+            # otherwise a retried camera_remove reports success forever
+            # on a stream nothing is tracking anymore.
+            await self._realtime.reconcile_orphaned_stream(camera_id)
```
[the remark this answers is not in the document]
hfarooq-nv: reuse _count_other_rules_for_stream inside reconcile_orphaned_stream to check for any active referencing rule before tearing the stream down, skipping the delete if one still exists

services/alert/src/realtime/services/realtime_service.py
```diff
@@ -1278,69 +1287,81 @@ async def _stop_alert_persistent(
...
         rtvi_outcome = "n/a"
         if rtvi_stream_id:
-            other_count = await self._count_other_rules_for_stream(
-                rtvi_stream_id, alert_rule_id,
-            )
-            ctx["other_active_rules"] = other_count
             rtvi_outcome = await self._safe_teardown_rtvi_with_outcome(
-                rtvi_stream_id, ctx, stop_stream=(other_count == 0),
+                rtvi_stream_id, ctx, stop_stream=stop_stream,
             )
```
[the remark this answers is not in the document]
hfarooq-nv: extend the per-stream lock to hold through the actual RTVI HTTP calls (not just the delete + last-reader decision), so no two callers can have RTVI calls for the same stream in flight at once

hfarooq-nv:
> Two follow-ups on the new commit. Not repeating the two Greptile P1 comments — both look right to me.
> 
> **`always_on_service.py` around 541 (tracked removal path)**
> 
> The new recovery path only runs when a later camera_remove arrives and finds nothing tracked. The removal that actually failed still returns success and drops its camera entry, so there is no second attempt for that failure on its own — recovery depends on another removal event arriving later. If the probe or delete inside that later attempt also fails, it returns success too.


FIX: call reconcile_orphaned_stream unconditionally after every camera_remove (not just when nothing was tracked), so a stop_alert that reported success while its own RTVI teardown silently failed gets caught on the same request

> **`test/unit/api/test_realtime_service.py` around 979 (persistent teardown tests)**
> 
> The new tests cover ordering and concurrent deletes on the persistent path, but none makes stop_captions fail there and check that stop_stream still runs. That error handling is duplicated in the persistent variant, and the only test covering it is on the in-memory helper, so the persistent copy can change without any test failing.

FIX: add a test_stop_captions_failure_continues test to the persistent (ES-backed) test class mirroring the existing in-memory one, asserting stop_stream still runs

services/alert/src/realtime/services/realtime_service.py
```diff
@@ -2445,6 +2597,68 @@ async def _safe_stop_stream_with_ctx(
...
+                )
+                return False
+
+            if not any(s.get("id") == rtvi_stream_id for s in streams):
+                return False
+
+            # "" never matches a real alert_rule_id (always a uuid4 hex
+            # string), so this counts every currently-active rule
+            # referencing the stream, not "every rule but one".
+            other_count = await self._count_other_rules_for_stream(
```
dineshvenkat, line 2692: When the rule lookup fails, `_count_other_rules_for_stream` logs "assuming 0" and continues, so `other_count` comes back 0 and this path treats the stream as orphaned and deletes it. A failed store query, a rule still being created whose pending reference is private to the other instance, and persistence being disabled all produce 0 the same way — so the stream can be deleted while a rule is still using it.

dineshvenkat:
`realtime_service.py` around 2707, in `_cleanup_failed_rule`: this path counts and then calls stop_stream without taking the per-stream teardown lock, unlike the stop paths which now hold it across the RTVI calls. A cleanup running here can overlap a concurrent stop on the same stream, which is the overlap the lock was added to prevent.

hfarooq-nv:
> `realtime_service.py` around 2707, in `_cleanup_failed_rule`: this path counts and then calls stop_stream without taking the per-stream teardown lock, unlike the stop paths which now hold it across the RTVI calls. A cleanup running here can overlap a concurrent stop on the same stream, which is the overlap the lock was added to prevent.

 wrapped its ref-count-and-stop_stream call in the same _get_stream_teardown_lock every other RTVI teardown path already uses, so it can no longer race a concurrent stop_alert/reconcile_orphaned_stream on the same stream

services/alert/src/realtime/services/realtime_service.py
hfarooq-nv: added a fail_open=False mode to the ref-count query that reconcile_orphaned_stream now uses, so a failed ES lookup causes it to skip the delete instead of silently treating "couldn't check" the same as "verified zero readers."

PR #2212, opened 2026-09-15
fix: support Base skills with Nemotron 3.5 Lightning with OpenClaw

jiayin-nvidia:
## Summary

- default Nemotron 3.5 Lightning to direct OpenClaw tool disclosure
- route Base sensor, snapshot, video question, and report prompts to the appropriate VSS skills
- honor the NemoClaw sandbox proxy for VIOS requests and inline sandbox-only clips for VLM access
- document the Lightning tool-disclosure setting and cover sandbox-host clip detection

## Validation

- replayed the four target prompts through the VSS UI agent API
- `157 passed` across focused VIOS and VLM CLI tests
- validated `deploy_nemoclaw.ipynb` as JSON
- `git diff --check`


.openclaw/workspace/_nemoclaw/AGENTS.md
```diff
@@ -2,6 +2,35 @@
 
 This folder is home. Treat it that way.
 
+## VSS Base prompt routing
```
zac-wang-nv, line 5: do we need this ? skills should still have the descriptions fused into agent's system prompt?

libs/vss/core/src/vss_core/vios/client.py
```diff
@@ -198,7 +198,7 @@ async def get_timelines_map(
     timelines_url = f"{base}/vst/api/v1/storage/timelines"
     timeout = aiohttp.ClientTimeout(total=timeout_seconds)
```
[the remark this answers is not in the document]
jiayin-nvidia: Thanks for flagging this. We are keeping `trust_env=True` intentionally: NemoClaw/OpenShell sandboxes reach host-side VIOS through the managed proxy, so disabling ambient proxy handling breaks this deployment path. aiohttp resolves proxy settings per request and honors `NO_PROXY`; sharing a session does not force the external download and internal VIOS request onto the same route. Deployments that require direct access to a private VIOS host should include that host in `NO_PROXY`, while `host.openshell.internal` is intentionally proxied in the NemoClaw sandbox. Therefore we do not plan a code change for this finding.

.openclaw/workspace/_nemoclaw/AGENTS.md
```diff
@@ -2,6 +2,36 @@
...
+Do not override its long-video rule; if it requires LVS, report that dependency.
+The command shape is always:
+
+```bash
+uv run --project /sandbox/video-search-and-summarization/libs/vss vss vlm run \
+  --no-persist --prompt "<question>" --sensor "<listed-name>" \
+  --start-time "<timeline-start>" --end-time "<timeline-end>" --fps 2
+```
+
+Keep the `vss` executable and `--no-persist` in the command. Obtain the listed
```
zac-wang-nv: this will disable memory function entirely.

if writing to memory causes issues, can you please work with @liamy-nv  to get it resolved?

.openclaw/workspace/_nemoclaw/AGENTS.md
```diff
@@ -2,6 +2,36 @@
...
+skills after a failure.
+
+The Base profile intentionally has no Elasticsearch or LVS. For Base smoke
+tests, add `--no-persist` to every `vss vlm run`; unified memory is not a VLM
+prerequisite. Generate a requested named-video report with the direct Mode A
+VLM path and `--no-persist` only when the report skill permits the direct path.
+Do not override its long-video rule; if it requires LVS, report that dependency.
+The command shape is always:
+
+```bash
```
zac-wang-nv: the openclaw Dockerfile is supposed to get vss executable by the agent.

.openclaw/workspace/_nemoclaw/AGENTS.md
jiayin-nvidia: I tested Nemotron 3.5 Lightning in direct mode with the four prompts sequentially in one conversation, removing only the explicit routing block. Sensor listing initially routed to VA-MCP/analytics and made 86 tool calls, the snapshot returned the first frame instead of 0:05, the PPE answer reused memory instead of running fresh VLM analysis, and the report reused prior sensor data rather than following the report skill. Thus, I think we need the routing blocks.

.openclaw/workspace/_nemoclaw/AGENTS.md
jiayin-nvidia: I checked the current `vss-openclaw` sandbox. After applying the configured PATH, command `-v vss` returns nothing, while `uv run --project /sandbox/video-search-and-summarization/libs/vss vss --version` works. The existing `TOOLS.md` also documents the project-local invocation. Is the standalone vss executable expected in a newer Docker image? cc @nvleeryan

.openclaw/workspace/_nemoclaw/AGENTS.md
jiayin-nvidia: I removed the hardcoded `--no-persist`. The CLI now works without Elasticsearch and reports that memory was not persisted

.openclaw/workspace/_nemoclaw/AGENTS.md
```diff
@@ -2,6 +2,33 @@
...
+
+Never route a named-video PPE question to analytics or VA-MCP. Resolve names
+from the sensor listing; if one unambiguous result corrects a typo, state the
+correction and use the listed identifier. Obtain the sensor's exact recorded
+timeline before time-based requests; never substitute the current date.
+A report for a named video shorter than 120 seconds uses
+`vss-generate-video-report` Mode A, never `vss summarize`: resolve the exact
+timeline, call `vss_cli` with `vlm run --prompt "Generate a structured video
+analysis report." --sensor <listed-name> --start-time <timeline-start>
+--end-time <timeline-end> --fps 2`, then render the structured report.
```
zac-wang-nv: **P1** — A clean notebook/onboard build still installs the CLI from `VSS_REF=5212368036879feb704a258be0db012ce518aeaa` in `.openclaw/Dockerfile:49`; it only takes workspace Markdown from this checkout. That pinned CLI has no `--fps` option, so this newly required report command exits 2 (`No such option '--fps'`) before inference. It also lacks this PR's sandbox-host inlining and missing-memory fallback, so the other Base fixes are absent from the shipped executable. Please update the source pin to a commit containing the required CLI changes, or stage those sources during the normal notebook build, and verify the command against the resulting image.

.openclaw/workspace/_nemoclaw/AGENTS.md
```diff
@@ -2,6 +2,33 @@
...
+invoke the skill's `vss` arguments through the `vss_cli` tool. Do not look for a
+repository checkout or replace the CLI with raw HTTP.
+
+Never route a named-video PPE question to analytics or VA-MCP. Resolve names
+from the sensor listing; if one unambiguous result corrects a typo, state the
+correction and use the listed identifier. Obtain the sensor's exact recorded
+timeline before time-based requests; never substitute the current date.
+A report for a named video shorter than 120 seconds uses
+`vss-generate-video-report` Mode A, never `vss summarize`: resolve the exact
+timeline, call `vss_cli` with `vlm run --prompt "Generate a structured video
```
zac-wang-nv: **P1** — Please pass the prompt selected by the report skill through `vss_cli` instead of fixing it to this sentence. Mode A Step 3 loads `references/default-vlm-prompt.md` (which requires chronological events with timestamps) or the user's explicitly approved prompt when HITL is enabled. Following this command replaces either selection with a generic request, so edits/approval have no effect on inference and the VLM is no longer asked for the timestamped evidence needed by the report. Keep the CLI transport, but preserve the skill's prompt selection and approval step before calling it.

jiayin-nvidia:
/ok to test [a4aface](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/2212/commits/a4aface185ebefd9ca6f0a9046e8a3527fc8896a)

zac-wang-nv (approved):
lgtm.

one todo for hermes

.hermes/Dockerfile
```diff
@@ -32,7 +32,7 @@
 ARG BASE_IMAGE=ghcr.io/nvidia/nemoclaw/hermes-sandbox@sha256:32d7ed6a6f1b682991818a2acaf5636d9fb066ae33265b011837b0ba601c0970
 ARG BUILDER_IMAGE=node:22-trixie-slim@sha256:db8a96a63e5264607ada2d206758876ebbed6a12be2ada7517793cbfb0c2a29c
 ARG VSS_REPO=https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization
-ARG VSS_REF=5212368036879feb704a258be0db012ce518aeaa
+ARG VSS_REF=9dc06553d67fcc21882fe0620b33f7a094b38f72
```
zac-wang-nv, line 35: **P2** — TODO (non-blocking follow-up): make the Base routing instructions work with Hermes' standalone `vss` CLI, or provide a Hermes-specific workspace overlay. This pin copies the new `_nemoclaw/AGENTS.md` into `/sandbox`, where it requires OpenClaw's `vss_cli` tool for sensor, video-question, and report requests. Hermes does not register that tool. Track this compatibility limitation as a Hermes follow-up to the OpenClaw changes.

PR #2225, opened 2026-09-16
skill-eval: split coding and operational model routes

gbau1:
## Scope

Supersedes #2160 with two independent skill-eval routes:

- **Coding route** controls the harness/provider/model used for Build Vision AI and operational setup (`expects[0]`).
- **Operational route** independently controls the harness/provider/model used for the remaining operational queries.
- NVIDIA-hosted model IDs come from the public [NVIDIA inference catalog](https://inference.nvidia.com/); remote routes use the fixed `https://inference.nvidia.com/v1` API base.
- Coordinator, judge, worker allocation, cleanup, skills, eval specs, adapters, and notebooks are unchanged.

## Execution contract

- Specs under `skills/operations/`: coding route for `expects[0]`, operational route for `expects[1:]`.
- `vss-build-vision-ai` and other non-operational specs: coding route throughout.
- NemoClaw keeps the existing sandbox setup, but the setup task can use either Claude Code or Codex independently of the NemoClaw model.
- Both routes are validated before Brev worker selection; credentials are not logged.

## Endpoint and credential boundaries

- Manual dispatch no longer accepts endpoint URLs or the unrestricted `custom` provider.
- Endpoint environment overrides are ignored; NVIDIA inference routes are fixed to `https://inference.nvidia.com/v1`.
- The provider-managed NemoClaw `nvidia-build` route exports its resolved credential as `NVIDIA_API_KEY`; OpenAI-compatible NemoClaw routes use `COMPATIBLE_API_KEY`.

## Workflow inputs

- `coding_harness`, `coding_provider`, `coding_model`
- `operational_harness`, `operational_provider`, `operational_model`

Blank model fields preserve the configured default when the selected provider permits it.

## Validation

- `python3 -m pytest -q .github/skill-eval/tests .github/skill-eval/envs/tests`
- **342 passed**
- Workflow YAML parsed locally and passes the repository yamllint policy.
- Rebased onto current `develop` (`df380b6e5`).


.github/skill-eval/run_leg.py
```diff
@@ -1710,17 +1686,31 @@ def finish(rc: int) -> int:
...
         )
+        invocation_env = env.copy()
+        if invocation_agent == "claude-code":
+            # Harbor 0.20's ClaudeCode adapter reads the endpoint from the
+            # child environment, not its api_base kwarg. Scope this per
+            # invocation so NemoClaw's setup coding agent keeps the runner
+            # route while a directly evaluated Claude agent gets its selected
+            # endpoint.
+            invocation_env["ANTHROPIC_BASE_URL"] = invocation_base_url
+            invocation_env["ANTHROPIC_API_KEY"] = invocation_config.api_key
```
zac-wang-nv: **P1** — Keep the judge's endpoint and credential separate from the evaluated Claude route. These assignments replace the environment of the entire Harbor process. The unchanged adapters emit `[verifier.env]` entries referencing `${ANTHROPIC_BASE_URL}` and `${ANTHROPIC_API_KEY}`, and Harbor 0.20 resolves those from this same process environment. Selecting a different coding or operational endpoint therefore sends the verifier to that endpoint with the route's key, while `ANTHROPIC_MODEL` still names the runner's judge model. A valid agent route can consequently fail all LLM checks because that endpoint does not serve the judge model, or silently change the judge's provider. I reproduced the changed verifier endpoint/key using Harbor's resolver and a mocked invocation. Preserve the runner settings for verification and scope route overrides to the evaluated agent's environment.

.github/skill-eval/run_leg.py
```diff
@@ -1606,20 +1576,22 @@ def run_invocations(
...
-    # Build Vision AI's own specs are coding-agent evaluations. For an
-    # operational spec, its first normal Harbor task is the setup contract:
-    # the coding agent follows Build Vision AI, deploys from expects[0], and
-    # attaches NemoClaw. The remaining normal tasks run through that sandbox.
+    # Skill location is the phase contract: coding/deployment/tool specs use
+    # the coding route throughout. Operational specs use expects[0] as their
+    # coding-agent deployment contract, then switch to the operational route.
+    spec_path = os.environ.get("EVAL_SPEC_PATH", "")
+    operational_eval = spec_path.startswith("skills/operations/")
+    coding_setup = invocations[0] if operational_eval and invocations else None
```
zac-wang-nv: **P1** — Select a coding setup task for each chain, rather than only for the whole invocation list. The operational adapters support multiple modes per platform, and `discover_invocations()` concatenates those separate `step-1..N` chains. With two chains, this code selects models `[coding, operational, operational, operational]`: the second mode's `step-1` is sent to the operational route. It also inherits `SKILL_EVAL_PRESERVE_DEPLOYMENT=1` from the first chain, so its first task skips the fresh deployment reset that previously applied to each mode. This evaluates the wrong setup model against leftover state. Track setup and preservation per `chain_key`, so each chain runs its own `expects[0]` through the coding route before switching to operations. Current specs use one mode, but the existing multiple-mode adapter/discovery contract exposes this regression.

.github/skill-eval/run_leg.py
gbau1: Addressed in fd5272197. Route credentials and endpoints now use Harbor agent-only `--ae` variables with environment-template indirection, while the parent `ANTHROPIC_*` environment remains unchanged for `[verifier.env]`. Added regression coverage that uses distinct route and judge credentials, plus verified the separation with Harbor 0.20’s resolver.

.github/skill-eval/run_leg.py
gbau1: Addressed in fd5272197. Setup selection is now keyed by `chain_key`; every chain starts with the coding route, clears deployment preservation for a fresh reset, and enables preservation only for its later steps. NemoClaw setup preparation and deferred cleanup are also applied per chain. A failed setup skips only its own chain. Multi-chain route, preservation, and failure-isolation regressions are covered locally.

zac-wang-nv:
Follow-up TODOs from review of `4af8a2d4`:

- [ ] **P1** — [.github/skill-eval/model_config.py:16](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/4af8a2d410d51f95216f33f36b83a49cd15d5bd7/.github/skill-eval/model_config.py#L16): Use the inference API host for the fixed route instead of deriving it from the model catalog URL. On this head, every runtime is forced to `https://inference.nvidia.com/v1`, but unauthenticated POSTs to its `/messages`, `/chat/completions`, and `/responses` paths return HTTP 200 `text/html` containing the NVIDIA Model Catalog, not an inference response. The catalog's own frontend configuration sets `LITELLM_INFERENCE_URL` to `https://inference-api.nvidia.com`; `/v1/models` on that host returns the expected JSON authentication error. As written, coding/setup and operational model calls are sent to the website and cannot execute their evaluations. Keep the endpoint fixed, but use the actual API base and cover that distinction in validation.

- [ ] **P1** — [.github/skill-eval/model_config.py:55](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/4af8a2d410d51f95216f33f36b83a49cd15d5bd7/.github/skill-eval/model_config.py#L55): Migrate the Daily workflow to the new coding-route input when changing this default. The unchanged `.github/workflows/skills-eval-daily.yml:270` still implements its `agent` dropdown by setting only `EVAL_AGENT`. Selecting `codex` there now leaves the coding route at `claude-code`, and `run_invocations()` sends every Build Vision AI/non-operational task through that route. I reproduced a Daily-style environment with `EVAL_AGENT=codex` and a configured `CODEX_MODEL`: the generated Harbor command uses `-a claude-code` and `ANTHROPIC_MODEL`. This silently evaluates the wrong agent for those specs. Pass the Daily selection into the explicit coding and operational harness settings, or preserve that legacy selection for coding when no explicit override is supplied.

zac-wang-nv (approved):
Approved with the two follow-up TODOs tracked in the [general review comment](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/2225#issuecomment-5733451785).

gbau1:
Addressed both follow-up TODOs on rebased head `b80a030fc`:

- Kept `https://inference.nvidia.com/` as the model-catalog source and fixed evaluated routes to `https://inference-api.nvidia.com/v1`, with an exact-host regression test.
- Preserved the Daily workflow single-agent UI while explicitly applying its selection to both `SKILLS_EVAL_CODING_HARNESS` and `SKILLS_EVAL_OPERATIONAL_HARNESS`; `EVAL_AGENT` remains set for compatibility. Added a Daily-style Codex regression proving a non-operational task generates a Codex invocation.

Focused validation: 102 tests passed; yamllint and copyright validation passed.

.github/skill-eval/run_leg.py
```diff
@@ -334,18 +337,23 @@ def build_harbor_command(
...
-        # claude-code reads ANTHROPIC_API_KEY), so it never lands on the CLI.
+        # `--ak api_base`; route endpoint/key values are resolved from the
+        # Harbor process environment into agent-only `--ae` variables below,
+        # so the key never lands on the CLI and the verifier route stays intact.
         agent_flags = [
             "-a", "agents.nv_codex:NvCodex",
             "--model", model,
             "--ak", f"api_base={_api_base_v1(anthropic_base_url)}",
+            "--ae", f"OPENAI_API_KEY=${{{AGENT_ROUTE_API_KEY_ENV}}}",
+            "--ae", f"OPENAI_BASE_URL=${{{AGENT_ROUTE_BASE_URL_ENV}}}",
```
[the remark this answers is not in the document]
gbau1: Harbor 0.20 resolves these templates itself; shell expansion is neither expected nor required. The CLI parses each agent-env argument into config.agent.env in harbor/cli/trials.py, then AgentFactory.create_agent_from_config() calls resolve_env_vars(config.env). That resolver replaces a full environment-variable template from Harbor's process environment. I reproduced this against the pinned harbor==0.20.0 with sentinel route values: both the agent key and base URL resolved correctly. run_leg.py supplies those source variables in the Harbor child environment while leaving the judge's ANTHROPIC variables unchanged. The earlier live Claude setup run also authenticated successfully through this path.

PR #2234, opened 2026-09-16
Fix(vss-cli):vss cli analytics

liamy-nv:
## Summary

- Add a read-only `vss analytics` CLI command group backed by the Video Analytics API.
- Allow NemoClaw and host-side coding agents to query analytics without deploying the in-stack `vss-agent` or legacy `vss-va-mcp`.
- Preserve `vss-agent` and `vss-va-mcp` in stock Alerts deployments where the existing agent configuration still depends on VA-MCP.

## What Changed

- Added the following CLI commands:
  - `vss analytics incidents`
  - `vss analytics incident`
  - `vss analytics sensors`
  - `vss analytics places`
  - `vss analytics fov-histogram`
  - `vss analytics average-speed`
  - `vss analytics analyze`
- Added a Video Analytics API client under `vss_core.analytics`.
- Added `/video-analytics-api` discovery to `vss configure`.
- Registered `analytics` through the existing lazy `vss.commands` plugin interface.
- Added typed handling for invalid input, missing records, timeouts, unavailable backends, and missing deployment configuration.
- Kept analytics operations read-only:
  - No job IDs
  - No memory records
  - No direct Elasticsearch fallback
  - JSON results on stdout
- Added complete analytics place paths using `name=value/...` formatting.
- Added incident pagination protection with `has_more`.
- Added deterministic incident overlap, average-speed, occupancy, people-count, and vehicle-count analysis.
- Updated the analytics and video-report skills to use `vss analytics` instead of VA-MCP for supported read-only incident workflows.
- Updated deployment composition guidance so:
  - Stock Alerts keeps `vss-agent` and `vss-va-mcp`.
  - NemoClaw and host-CLI builds omit both unless explicitly requested.
  - Removing an agent harness does not prune unrelated Foundation services.
- Added analytics readiness and service-graph validation.
- Added and updated CLI, core-client, configuration, composition, and skill-eval tests.

  ```bash
  vss configure --base-url <deployment-origin>
## Checklist


liamy-nv:
/ok to test [ecd0261](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/2234/commits/ecd0261bf7ef1b6c459280b2acfa8a6286c92285)

liamy-nv:
/ok to test [4a88633](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/2234/commits/4a8863372a9bb6f8c66dcb2614a4eed2d36f6bb0)

libs/vss/core/src/vss_core/analytics/client.py
```diff
@@ -0,0 +1,373 @@
...
+
+def _sensor_rows(calibration: object) -> list[dict[str, Any]]:
+    body = _object(calibration, "calibration read")
+    sensors = body.get("sensors", [])
+    if not isinstance(sensors, list):
+        raise AnalyticsError("Video Analytics API calibration read returned a non-array 'sensors'")
+    return [row for row in sensors if isinstance(row, dict)]
+
+
+def _place_values(sensor: dict[str, Any]) -> list[str]:
```
prisrivastav-nv: This discards each entry's `name` key, so the place token that ends up in `places()` output is a bare value — `{"Warehouse": ["Room-1"]}`. But the API's `place` parameter is a hierarchy path prefix-matched against `place.name.keyword`: `"building=Warehouse/room=Room-1"`. See [`Services/Incidents.js:72`](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/4a8863372a9bb6f8c66dcb2614a4eed2d36f6bb0/services/analytics/video-analytics-api/src/web-api-core/Services/Incidents.js#L72), [`Metrics/Behavior.js:847`](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/4a8863372a9bb6f8c66dcb2614a4eed2d36f6bb0/services/analytics/video-analytics-api/src/web-api-core/Metrics/Behavior.js#L847), and the service's own request fixture [`incidents_place_only/input.json`](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/4a8863372a9bb6f8c66dcb2614a4eed2d36f6bb0/services/analytics/video-analytics-api/test/integration-test/fixtures/warehouse_2d_app/incidents_place_only/input.json#L2).

So `incidents` and `average_speed` send `place=Warehouse`, which prefix-matches nothing and returns `{"count":0,"incidents":[]}` at exit 0. The skill then instructs the agent to report that as "no matching analytics data is indexed" — a wrong answer delivered confidently. `fov_histogram` is unaffected because it resolves sensor IDs from calibration instead, which is why the breakage is asymmetric and easy to miss.

The tests currently bless the broken shape, which is why this got through: [`test_client.py:151`](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/2234/files#diff-073965e26ddf711a336385ba7ba7542da413e2c7b778e8917015ff519a0c7ba2R151) and [`test_analytics_smoke.py:38`](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/2234/files#diff-634a52936035c625fac11563e8f955a8d34886c0473df9391c18e79dadfa6b8eR38) both model a calibration place as `[{"value": "San Jose"}, {"value": "First"}]`, omitting the `name` key that real calibration always carries — compare [`nv-warehouse-4cams/calibration.json`](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/4a8863372a9bb6f8c66dcb2614a4eed2d36f6bb0/deploy/docker/industry-profiles/warehouse-operations/warehouse-2d-app/calibration/sample-data/nv-warehouse-4cams/calibration.json#L1), which has `[{"name":"building","value":"Warehouse"},{"name":"room","value":"Room-1"}]`. Both fixtures need updating alongside the fix, or the corrected code will fail tests that are themselves wrong.

Suggested fix: build the `name=value/...` path here and emit that as the place token throughout. That also resolves two follow-on problems — the `len(values) >= 2` guard in `_places_from_sensors`, which today drops single-level places entirely and ignores any level past index 1, and the `count` reported by `places`, which counts top-level keys rather than places. It also removes the current disagreement between commands: `sensors(place=...)` matches at any level, while `places()` only ever exposes the first two.

libs/vss/core/src/vss_core/analytics/client.py
```diff
@@ -0,0 +1,373 @@
...
+    count = 0
+    maximum = 0
+    minimum: int | None = None
+    maximum_at: datetime.datetime | None = None
+    minimum_at: datetime.datetime | None = None
+    for instant, delta in events:
+        count += delta
+        if delta > 0 and count > maximum:
+            maximum, maximum_at = count, instant
+        if minimum is None or count < minimum:
```
prisrivastav-nv: `minimum_overlap` is structurally always 0. `count` starts at 0 and the last event in a sorted sweep is always an end, so `count` returns to 0 and this check always lands there. For any non-empty incident set the answer is `minimum_overlap: 0`, with `minimum_overlap_at` pointing at the first gap — and it is interpolated into the user-facing `summary`.

Either compute the minimum over intervals where incidents actually exist, or drop the field. Reporting a constant as an analysis result is worse than not reporting it.

Separately, `minimum or 0` below should be an explicit `is None` check — the `or` idiom hides the None case behind a value that is also falsy.

libs/vss/core/src/vss_core/analytics/client.py
```diff
@@ -0,0 +1,373 @@
...
+    """Issue one bounded GET and return parsed JSON."""
+    url = f"{base_url.rstrip('/')}/{path.lstrip('/')}"
+    timeout = aiohttp.ClientTimeout(total=timeout_seconds)
+    try:
+        async with (
+            aiohttp.ClientSession(timeout=timeout, trust_env=True) as session,
+            session.get(url, params=params) as response,
+        ):
+            text = await response.text()
+            if response.status < 200 or response.status >= 300:
```
prisrivastav-nv, line 77: Every non-2xx becomes `AnalyticsError`, which extends `BackendUnreachableError` and therefore exits 3. So a 400 from a malformed `--source` and a dead service are indistinguishable to the caller: nothing maps 4xx to `Exit.INVALID_INPUT` or an API 404 to `Exit.NOT_FOUND`, and `AnalyticsNotFoundError` is only ever raised by the `incident` lookup.

The skill's exit-3 text ("rejected the request or is unreachable") documents the conflation rather than fixing it, which leaves the agent no way to tell "fix your options" from "retry later" — the exact distinction the exit-code contract exists to carry. Since the PR claims it maps API errors to CLI exit codes, branch on status: 4xx to invalid input, 404 to not found, 5xx and transport failures to unreachable.

libs/vss/core/src/vss_core/analytics/client.py
```diff
@@ -0,0 +1,373 @@
...
+    params: dict[str, QueryValue] | None,
+    operation: str,
+    timeout_seconds: float,
+) -> object:
+    """Issue one bounded GET and return parsed JSON."""
+    url = f"{base_url.rstrip('/')}/{path.lstrip('/')}"
+    timeout = aiohttp.ClientTimeout(total=timeout_seconds)
+    try:
+        async with (
+            aiohttp.ClientSession(timeout=timeout, trust_env=True) as session,
```
prisrivastav-nv: A fresh `ClientSession` (and connector, and DNS resolution) per GET, with no reuse. `fov_histogram` compounds it: for a place it loops sensors serially, so N sensors means N sessions and N sequential round trips, each granted the full `timeout_seconds` budget — worst case N x 30s with no bound on the command as a whole.

Hold one session on `AnalyticsClient` and `asyncio.gather` the per-sensor requests. As written, the async plumbing buys nothing over blocking calls: every entry point is driven by a one-shot `asyncio.run`, so the only concurrency opportunity in the client is the one place that doesn't take it.

libs/vss/core/src/vss_core/analytics/client.py
```diff
@@ -0,0 +1,373 @@
...
+                "fromTimestamp": start_time,
+                "toTimestamp": end_time,
+                "bucketCount": bucket_count,
+            }
+            if object_type:
+                params["objectType"] = object_type
+            results.append(
+                _object(await self._get("metrics/occupancy/fov/histogram", "FOV histogram", params), "FOV histogram")
+            )
+        return results[0] if len(results) == 1 else _merge_histograms(results)
```
prisrivastav-nv: This returns two different schemas from one command: the raw API object when there is exactly one sensor, the normalized `{bucketSizeInSec, histogram}` when there are more. Two consequences — `--source-type sensor` never exercises the merge path at all, and a place with one camera returns a different shape than the same place after a second camera is added.

An unstable output contract is especially costly here because the consumer is an LLM reading JSON. Normalize unconditionally.

skills/vss-build-vision-ai/scripts/resolve_service_graph.py
```diff
@@ -0,0 +1,65 @@
...
+    foundation_profiles: Iterable[str],
+    requested_profiles: Iterable[str] = (),
+    *,
+    host_cli: bool,
+    legacy_va_mcp: bool = False,
+) -> tuple[str, ...]:
+    """Return an ordered profile set after applying explicit ownership rules."""
+    profiles = dict.fromkeys((*foundation_profiles, *requested_profiles))
+
+    if host_cli:
```
prisrivastav-nv, line 54: `requested_profiles` is merged and then `vss-agent` and `vss-va-mcp` are popped regardless of whether they were explicitly requested, so the `legacy_va_mcp` boolean is the only way to select VA-MCP.

That contradicts the docs added in this PR: `references/services/video-analytics-api.md:31` says "Another explicitly selected capability may independently re-add a key it owns," and `references/profiles/alerts.md:27` says VA-MCP is added "only in a delta that explicitly requests" it — and `requested_profiles` is the delta channel.

Two mechanisms for one decision, with the boolean silently overriding the explicit request. Pick one: either honour `requested_profiles` for these two keys, or drop the boolean and say so in the docs.

skills/vss-build-vision-ai/scripts/tests/test_analytics_composition_contract.py
```diff
@@ -0,0 +1,184 @@
...
+    value = _env_value(overrides, f"COMPOSE_PROFILES_{mode}")
+    value = value.replace("${LLM_MODE}", _env_value(overrides, "LLM_MODE"))
+    value = value.replace(
+        "${LLM_NAME_SLUG}",
+        _env_value(overrides, "LLM_NAME_SLUG"),
+    )
+    return tuple(value.split(","))
+
+
+def _compose_config(
```
prisrivastav-nv, line 80: This shells out to `docker compose config` and asserts `returncode == 0` below, with no availability guard, while living under `scripts/tests/` alongside pure unit tests. On any machine or CI lane without Docker — or without the profile env files it reads — this hard-fails rather than skipping, and the failure looks like a composition regression rather than a missing tool.

Add a skip marker gated on Docker being present, or move it to an integration lane.

prisrivastav-nv (changes requested):
The shape of this is right: read-only, no job grammar, capability-gated through `REQUIRES`, endpoints only from the recorded deployment, and a test that proves `vss --help` doesn't import the group. Error mapping reuses the existing name-based table rather than importing `vss_core` eagerly, and the exit codes documented in the skill (3/4/5/7) match `exits.py`. The SKILL.md rewrite from raw MCP JSON-RPC to exit-code branching is a clear improvement.

Blocking concern is #1 below: place-scoped `incidents` and `average-speed` send a place token the API cannot match, so those queries return `{"count":0,...}` at exit 0 — a wrong answer the skill then instructs the agent to report as "no matching analytics data is indexed." #2 and #3 are also correctness issues rather than polish.

deploy/docker/developer-profiles/dev-profile-alerts/overrides.env
```diff
@@ -219,11 +219,12 @@ RTVI_VLM_OPENAI_MODEL_DEPLOYMENT_NAME=${VLM_NAME}
...
 # COMPOSE PROFILE SELECTION
 # =============================================================================
-# Services for CV mode (2d_cv): perception pipeline
-COMPOSE_PROFILES_CV=vss-behavior-analytics-alerts,nvstreamer-alerts,perception-alerts,kibana-init-container-alerts,vss-video-analytics-api,vss-va-mcp,vss-agent,alert-bridge,phoenix,elasticsearch,elasticsearch-init-container,kafka,kafka-topic-init-container,redis,kibana,logstash,broker-health-check,vss-haproxy-ingress,rtvi-vlm,vss-ui,centralizedb,vst-ingress,sensor-ms,streamprocessing-ms,llm_${LLM_MODE}_${LLM_NAME_SLUG}
+# Services for CV mode (2d_cv): perception pipeline. Legacy vss-va-mcp is
+# selected only by an explicit MCP/SOP-report delta.
+COMPOSE_PROFILES_CV=vss-behavior-analytics-alerts,nvstreamer-alerts,perception-alerts,kibana-init-container-alerts,vss-video-analytics-api,vss-agent,alert-bridge,phoenix,elasticsearch,elasticsearch-init-container,kafka,kafka-topic-init-container,redis,kibana,logstash,broker-health-check,vss-haproxy-ingress,rtvi-vlm,vss-ui,centralizedb,vst-ingress,sensor-ms,streamprocessing-ms,llm_${LLM_MODE}_${LLM_NAME_SLUG}
```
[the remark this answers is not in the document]
liamy-nv: Stock Alerts still ships the in-stack agent, and that agent’s config still calls video_analytics_mcp to list incidents and build reports. VA-MCP is therefore back in the stock profile; NemoClaw/host-CLI builds still drop both the agent and VA-MCP.

libs/vss/core/src/vss_core/analytics/client.py
liamy-nv: Places now use complete name=value/... paths at every depth, including single-level places. Fixtures, sensor filtering, counts, CLI output, and documentation were updated.

libs/vss/core/src/vss_core/analytics/client.py
liamy-nv: Events are grouped by timestamp and minima are calculated only across intervals containing incidents. The None fallback is now explicit.

libs/vss/core/src/vss_core/analytics/client.py
liamy-nv: HTTP 404 now maps to not-found exit 5, other 4xx responses to invalid-input exit 2, and 5xx/transport failures remain exit 3. Skill guidance reflects this distinction.

libs/vss/core/src/vss_core/analytics/client.py
liamy-nv: Each operation now reuses one client session. Per-sensor histogram requests use asyncio.gather, with a timeout bounding the entire operation.

libs/vss/core/src/vss_core/analytics/client.py
liamy-nv: Every histogram response now passes through the same normalization/merge path, regardless of sensor count.

skills/vss-build-vision-ai/scripts/resolve_service_graph.py
liamy-nv: requested_profiles is now the sole explicit selection channel. Host-CLI defaults remove both services, but explicitly requested profiles are retained.

skills/vss-build-vision-ai/scripts/tests/test_analytics_composition_contract.py
liamy-nv: Only Compose-backed tests are skipped when docker compose is unavailable; pure resolver tests continue running.

libs/vss/core/src/vss_core/analytics/client.py
```diff
@@ -0,0 +1,428 @@
...
+    async def _calibration_sensors(self) -> list[dict[str, Any]]:
+        return _sensor_rows(await self._get("config/calibration", "calibration read"))
+
+    async def sensors(self, place: str | None = None) -> list[str]:
+        rows = await self._calibration_sensors()
+        sensors = {
+            str(row["id"])
+            for row in rows
+            if row.get("id") is not None
+            and (place is None or ((sensor_place := _place_path(row)) is not None and sensor_place.startswith(place)))
```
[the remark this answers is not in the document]
liamy-nv: A place now matches only the exact token or place + "/", so descendants still resolve and Room-1 no longer includes Room-10.

libs/vss/cli/src/vss_cli/analytics/group.py
```diff
@@ -0,0 +1,309 @@
...
+    _validate_pair(values, "source", "source_type")
+    _validate_pair(values, "start_time", "end_time")
+    requested = values["limit"]
+    incidents = _run(
+        _client(ctx).incidents(
+            source=values.get("source"),
+            source_type=values.get("source_type"),
+            start_time=values.get("start_time"),
+            end_time=values.get("end_time"),
+            limit=requested + 1,
```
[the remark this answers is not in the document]
liamy-nv: The user-visible --limit is now 1–9999, so the sentinel request stays at 10000. --limit 9999 is accepted locally and --limit 10000 is rejected before any HTTP call.

liamy-nv:
/ok to test [897b182](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/2234/commits/897b1820bf856560051e5ed5a34a6f2ebc8e6f59)

liamy-nv:
/ok to test [5efa544](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/2234/commits/5efa544db4fef7d4e775f53e04ab685434f63c41)

skills/vss-build-vision-ai/SKILL.md
```diff
@@ -302,9 +322,31 @@ After the selection, ask in one typed-values message only for that provider's st
...
+
+   **Harness-only delta invariant (apply here, before capability pruning).** When the selected capabilities exactly equal the Foundation, Q3 is the only customization, and the user did not explicitly add or remove a capability, set `ADDED_PROFILES=∅` and set `REMOVED_PROFILES` only to the harness-owned removal: `vss-agent`, plus `vss-va-mcp` only when it is present and the existing VA-MCP rule says it is unrequested. Compute `FINAL_PROFILES = (FOUNDATION_PROFILES ∪ ADDED_PROFILES) − REMOVED_PROFILES` and bypass generic forward-closure/unused-service pruning. A Q3 **no** is host-CLI driven, not headless; it does not authorize removing `vss-ui`, `phoenix`, `vss-haproxy-ingress`, the Foundation `llm_*` peer, `redis`, VIOS, models, or any other Foundation capability service. A Q3 **yes** has the same Compose preservation rule; NemoClaw is added outside Compose.
+
+   Before continuing, run this exact check against the Foundation and final profile lists:
+
+   ```bash
+   uv run "$REPO/skills/vss-build-vision-ai/scripts/resolve_service_graph.py" \
+     --foundation "$FOUNDATION_PROFILES" \
+     --final "$FINAL_PROFILES"
```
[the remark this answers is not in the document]
liamy-nv: Both skill invocations and composition.md now pass --requested "$REQUESTED_PROFILES" (empty is fine)

skills/vss-build-vision-ai/scripts/resolve_service_graph.py
```diff
@@ -0,0 +1,102 @@
...
+        profiles.pop(profile, None)
+
+    return tuple(profiles)
+
+
+def validate_harness_only_delta(
+    foundation_profiles: Iterable[str],
+    final_profiles: Iterable[str],
+    requested_profiles: Iterable[str] = (),
+) -> None:
```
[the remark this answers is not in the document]
liamy-nv: Resolved

liamy-nv:
/ok to test [7bcf9da](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/2234/commits/7bcf9dad16c271cc4eef394838d8c9a5d8b81b96)

skills/vss-build-vision-ai/eval/vdr_3_nemoclaw_read_only_analytics.json
zac-wang-nv, line 1: we don't need to add this.

libs/vss/cli/src/vss_cli/analytics/group.py
zac-wang-nv, line 1: can you put it under `analytics` subfolder?

skills/operations/vss-generate-video-report/references/report-types/incident-range.md
```diff
@@ -20,35 +20,37 @@ Numbered steps for Mode B, loaded on demand from [`SKILL.md`](../../SKILL.md) (`
...
-  "id": 1
-}
+Hand off to `/vss-query-analytics`, which configures and invokes the project-local
+VSS CLI. It uses `vss analytics incidents`; use this command shape after its
+bootstrap:
+
+```bash
+VSS_REPO_ROOT="${VSS_REPO_ROOT:-$HOME/video-search-and-summarization}"
+vss() { uv run --project "${VSS_REPO_ROOT}/libs/vss" vss "$@"; }
+vss analytics incidents \
```
zac-wang-nv, line 30: **P1** — Please preserve the VLM-incident selection when replacing the Alerts MCP call. The previous Alerts VA-MCP configuration sets `vlm_verified: true` (deploy/docker/developer-profiles/dev-profile-alerts/vss-agent/configs/va_mcp_server_config.yml:26), but this command omits `--vlm-verdict`, so the new client leaves `vlmVerified` unset and the REST API queries `mdx-incidents`, not `mdx-vlm-incidents`. A real-time Alerts deployment with only VLM incidents therefore reaches the mandated “No incidents found” response even when verified incidents exist; on CV deployments the verdict/reasoning fields can also disappear. Select the intended incident source explicitly here (for the previous Alerts behavior, `--vlm-verdict all`) and cover a report with an empty raw index and a populated VLM index.

libs/vss/core/src/vss_core/analytics/client.py
```diff
@@ -0,0 +1,435 @@
...
+            incidents = await self.incidents(
+                source=source,
+                source_type=source_type,
+                start_time=start_time,
+                end_time=end_time,
+                limit=1001,
+                includes=("timestamp", "end"),
+            )
+            has_more = len(incidents) > 1000
+            result = _overlap_result(incidents[:1000])
```
zac-wang-nv: **P1** — Clip incident intervals to `start_time`/`end_time` before this sweep. The API returns whole incidents that intersect the window (`timestamp <= toTimestamp` and `end >= fromTimestamp` in Services/Incidents.js:68–69), so their endpoints can lie outside it. With incidents 00:00–01:00 and 00:10–01:00, querying 00:30–00:40 currently returns `minimum_overlap: 1` at 00:00 and `maximum_overlap_at: 00:10`, although both incidents are active throughout the requested window and its minimum is 2. Pass the bounds into the overlap calculation and test incidents spanning both boundaries.

skills/operations/vss-query-analytics/evals/query_analytics.json
```diff
@@ -12,54 +12,42 @@
...
-      "checks": [
-        "`curl -sf --max-time 15 http://localhost:8000/docs` returns exit 0 (Agent REST API responsive)",
-        "`docker ps --format '{{.Names}}' | grep -qx vss-agent` returns exit 0",
-        "`docker ps --format '{{.Names}}' | grep -qx redis` returns exit 0",
-        "`docker ps --format '{{.Names}}' | grep -qx vss-rtvi-vlm` returns exit 0 (real-time VLM processor)"
-      ]
-    },
-    {
-      "query": "Verify the VA-MCP analytics endpoint is reachable and ready to answer queries.",
+      "query": "Show me the 10 most recent incidents across all sensors.",
```
zac-wang-nv: **P1** — Keep an explicit setup step, or provision an API fixture, before these read-only queries. In the normal CI path, `BrevEnvironment.start()` resets Docker for step 1 (.github/skill-eval/envs/brev_env.py:470–477) and explicitly does not pre-deploy a stack (lines 546–552). This removes the spec’s deployment query while the adapter now tells every query not to trigger deployments, yet the checks still require a successful incidents response. A compliant fresh run can only report an unavailable/unconfigured backend, which is not the successful empty result required here. The generated preamble still authorizes prerequisite deployment, so it also contradicts the new restriction. Separate setup from the read-only cases and keep both instructions consistent.

zac-wang-nv:
@liamy-nv  can you also post a screenshot of some example outputs? 

1)  incidents from a time range
2) historygram/count?

liamy-nv:
/ok to test [155a6d4](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/2234/commits/155a6d4f662a0a73f1d5cdc514758722bd2c19ae)

skills/operations/vss-generate-video-report/references/report-types/incident-range.md
```diff
@@ -20,35 +20,40 @@ Numbered steps for Mode B, loaded on demand from [`SKILL.md`](../../SKILL.md) (`
...
+  --source "<sensor-id>" --source-type sensor \
+  --start-time "<ISO>" --end-time "<ISO>" \
+  --vlm-verdict all \
```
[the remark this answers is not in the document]
liamy-nv: The REST API already documents vlmVerdict=all and treats it as a no-op on the VLM index (Incidents.js: if (vlmVerdict === "all") { // do nothing }). Sending vlmVerified=true&vlmVerdict=all is equivalent to the old MCP mapping of vlmVerified=true with no verdict. Confirmed on this deployment: --vlm-verdict all returns both confirmed and rejected incidents.

liamy-nv:
Incidents from a time range:
`$ vss analytics incidents \
    --source sample-warehouse-ladder --source-type sensor \
    --start-time 2026-09-17T22:50:00Z --end-time 2026-09-17T22:55:00Z \
    --vlm-verdict all --limit 3 --include info --include category --pretty
{
  "count": 3,
  "incidents": [
    {
      "sensorId": "sample-warehouse-ladder",
      "timestamp": "2026-09-17T22:54:51.922Z",
      "end": "2026-09-17T22:54:53.922Z",
      "category": "Ladder PPE Violation",
      "info": {
        "objectTimeline": "{\"1325\": [{\"start\": \"2026-09-17T22:54:51.922Z\", \"end\": \"2026-09-17T22:54:53.922Z\"}]}",
        "videoSource": "http://vst-ingress:30888/vst/storage/temp_files/sample-warehouse-ladder_20260917_225451_c3f5d.mp4",
        "verificationResponseCode": "200",
        "verificationResponseStatus": "OK",
        "reasoning": "",
        "verdict": "confirmed",
        "description": ""
      },
      "Id": "3782beb737655074a275ae79a4688e7c6e65e624"
    },
    {
      "sensorId": "sample-warehouse-ladder",
      "timestamp": "2026-09-17T22:54:28.522Z",
      "end": "2026-09-17T22:54:36.622Z",
      "category": "Ladder PPE Violation",
      "info": {
        "objectTimeline": "{\"1325\": [{\"start\": \"2026-09-17T22:54:28.522Z\", \"end\": \"2026-09-17T22:54:36.622Z\"}]}",
        "videoSource": "http://vst-ingress:30888/vst/storage/temp_files/sample-warehouse-ladder_20260917_225428_7f2a3.mp4",
        "verificationResponseCode": "200",
        "verificationResponseStatus": "OK",
        "reasoning": "",
        "verdict": "confirmed",
        "description": ""
      },
      "Id": "264c93e23f01dad5b33538497a232657cd0f34cb"
    },
    {
      "sensorId": "sample-warehouse-ladder",
      "timestamp": "2026-09-17T22:53:06.922Z",
      "end": "2026-09-17T22:53:33.022Z",
      "category": "Ladder PPE Violation",
      "info": {
        "objectTimeline": "{\"1324\": [{\"start\": \"2026-09-17T22:53:06.922Z\", \"end\": \"2026-09-17T22:53:33.022Z\"}]}",
        "isComplete": "true",
        "videoSource": "http://vst-ingress:30888/vst/storage/temp_files/sample-warehouse-ladder_20260917_225323_5f34c.mp4",
        "verificationResponseCode": "200",
        "verificationResponseStatus": "OK",
        "reasoning": "",
        "verdict": "rejected",
        "description": ""
      },
      "Id": "1314f8a8c1c69e7d538798c064fb51f3246871e2"
    }
  ],
  "has_more": true
}`

liamy-nv:
fov counts:
`$ vss analytics fov-histogram \
    --source warehouse_sample --source-type sensor \
    --start-time 2026-01-01T00:00:00Z --end-time 2026-01-01T00:01:00Z \
    --object-type Person --bucket-count 3 --pretty
{
  "bucketSizeInSec": 20,
  "histogram": [
    {
      "start": "2026-01-01T00:00:00Z",
      "end": "2026-01-01T00:00:20Z",
      "objects": [{"type": "Person", "averageCount": 2.4}]
    },
    {
      "start": "2026-01-01T00:00:20Z",
      "end": "2026-01-01T00:00:40Z",
      "objects": [{"type": "Person", "averageCount": 3.1}]
    },
    {
      "start": "2026-01-01T00:00:40Z",
      "end": "2026-01-01T00:01:00Z",
      "objects": [{"type": "Person", "averageCount": 1.7}]
    }
  ]
}`

skills/vss-build-vision-ai/eval/vdr_3_nemoclaw_read_only_analytics.json
liamy-nv: deleted

libs/vss/cli/src/vss_cli/analytics/group.py
liamy-nv: moved to analytics/

skills/operations/vss-generate-video-report/references/report-types/incident-range.md
liamy-nv: incident-range.md now passes --vlm-verdict all, restoring the old VA-MCP vlm_verified: true behavior so the API reads mdx-vlm-incidents. Without it, a VLM-only Alerts deployment hit the mandated "No incidents found" reply and lost verdict/reasoning fields.

libs/vss/core/src/vss_core/analytics/client.py
liamy-nv: _overlap_result now takes the query bounds and clamps each incident to them, since the API returns whole incidents that merely intersect the window. Your 00:00–01:00 / 00:10–01:00 case queried at 00:30–00:40 now reports minimum overlap 2 instead of 1 at 00:00.

zac-wang-nv:
https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/2234#issuecomment-5722586887 

is this due to deployment time flakiness? should we loosen the checks?

liamy-nv:
/ok to test [c3356dd](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/2234/commits/c3356dde104ba83a9447405c20db44ddc7949dc0)

liamy-nv:
/ok to test [4658769](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/2234/commits/4658769de27876d4ba9335098bd9be3995e5983b)

zac-wang-nv (approved):

libs/vss/core/src/vss_core/analytics/client.py
```diff
@@ -0,0 +1,447 @@
...
+                float(item["averageCount"])
+                for bucket in histogram.get("histogram", [])
+                if isinstance(bucket, dict)
+                for item in bucket.get("objects", [])
+                if isinstance(item, dict)
+                and item.get("type") == object_type
+                and isinstance(item.get("averageCount"), int | float)
+                and not isinstance(item.get("averageCount"), bool)
+            ]
+            average = sum(counts) / len(counts) if counts else None
```
zac-wang-nv, line 431: **P2** — **Non-blocking TODO:** In a follow-up, weight occupancy averages by each returned bucket's duration (or obtain the window average directly). The API rounds histogram intervals to five-second multiples and clips boundary buckets to the requested window, so bucket durations can differ. For a 00:00:04–00:00:10 query with 100 people in the first one-second bucket and zero in the remaining five-second bucket, this calculation reports 50 instead of the time-weighted 16.67. This affects both `avg-num-people` and `avg-num-vehicles`. Preserve the bucket bounds, weight `averageCount` by `end - start`, and add a regression covering unequal boundary-bucket durations. This TODO does not block this PR.

liamy-nv: ACK will open PR for this follow up

PR #2235, opened 2026-09-17
docs(vss-build-vision-ai): drop manual VIOS fan-out recipe, document webhook requirement

zac-wang-nv:
## Summary

`skills/operations/vss-manage-video-io-storage/references/provision-vios-source.md` was a manual, direct-REST recipe for fanning a registered VIOS source out to RT-CV/RT-Embed/RT-VLM on a headless (no-agent) build. Its premise — "no agent tier, so the operator does it by hand" — turned out to be stale for the `search` and `lvs` Foundations: each ships its own `notification_config.json` with `webhooks.enabled: true`, wired unconditionally via `VST_NOTIFICATION_CONFIG_PATH` in that Foundation's `overrides.env`. On `camera_streaming` / `camera_remove`, VIOS itself POSTs directly to RT-CV, RT-Embed, and RT-VLM (and cleans up Elasticsearch on removal) — exactly the fan-out the manual recipe duplicated, no agent involved.

## Plan

| # | Question | Recommendation | Decision |
|---|---|---|---|
| 1 | Does `vss vios add` alone replace the manual recipe? | No — `vss vios add` only performs registration (VIOS Step 1). Fan-out is a separate mechanism (the webhook), which is what actually replaces the recipe. | Confirmed via code: `notification_config.json` webhooks call RT-CV/RT-Embed/RT-VLM `/stream/add` directly. |
| 2 | Should the manual recipe be kept as a fallback for builds without a wired webhook (e.g. a custom delta off `base`)? | Treat a missing/incomplete webhook config as a **build-time blocker** in `vss-build-vision-ai` composition validation, rather than keeping a manual escape hatch — a silent no-op fan-out (source registers, nothing downstream fires) is worse than a hard stop. | Approved — "manual recipe should be dropped... BVA should document that it needs this config." |
| 3 | What about `alerts` and `base`? | `alerts` ships a narrower webhook (RT-CV / Alert-Bridge only, no RT-Embed) — unaffected. `base` ships `webhooks.enabled: false` — any delta resolving RT-CV/RT-Embed/RT-VLM off `base` now must write/edit a `notification_config.json`, enforced by the new Validate-gate bullet in `composition.md`. | Applied as designed. |

## Changes

- **Delete** `provision-vios-source.md` (311 lines).
- **`vss-build-vision-ai/SKILL.md`** — routing table, the Q3 no-agent bullet, and step 9: provisioning is `vss vios add`, fan-out is VIOS's own webhook, not a manual step.
- **`references/services/vios.md`** — document `VST_NOTIFICATION_CONFIG_PATH` / `webhooks.enabled` (previously only the unrelated `VST_ENABLE_NOTIFICATION`, which publishes an event but makes no HTTP call, was documented). This is the gap that caused the confusion in the first place.
- **`references/composition.md`** — new Validate-gate bullet: a delta changing the resolved RT-CV/RT-Embed/RT-VLM set (or adding any of them to `base`/`alerts`) must carry a matching `notification_config.json`, or the build silently drops perception.
- **`references/deployment_resolution.md`, `deployment.md`, `agent-harness.md`** — the write path no longer resolves per-consumer loopback ports; `vss vios` reads the same `vss configure`-recorded origin as the read path (verified in `vios_group.py`).
- **`references/services/rt-vlm.md`, `references/profiles/search.md`** — describe the webhook-driven single-call `/stream/add` (auto-starts inference when `metadata` is present) instead of the old two-call manual recipe.
- **`vss-manage-video-io-storage/SKILL.md`, `vss-search-archive/SKILL.md`, `cli_usage.md`, `source_lifecycle.md`** — routing/boundary language updated to `vss vios add`/`delete` + webhook.
- **Left untouched:** `docs/designs/vlm-tagging-search.md` — a dated design record, not a living spec.

## Test plan

- [x] `grep -rln "provision-vios-source" --include="*.md" .` — only the historical design doc remains.
- [x] Confirmed via `deploy/docker/developer-profiles/dev-profile-search/vios/configs/notification_config.json` that `camera_streaming` fires for both `file` and `rtsp` camera types (covers upload and live), and `camera_remove` cleans up Elasticsearch + all three consumers.
- [x] Confirmed via `libs/vss/cli/src/vss_cli/vios_group.py` that `vss vios` commands resolve `ctx.deployment.base_url`, the same `vss configure`-recorded origin used for reads.
- [x] Confirmed via `skills/deployment/vss-deploy-video-embedding/references/rest-api.md` and `vss-deploy-dense-captioning/references/integrate-rt-vlm.md` that `/stream/add` (singular, CV-compatible envelope) auto-starts embedding/inference when the request carries `model`/prompt metadata — matching what the webhook config sends.
- [ ] Not run: an actual `search`/`lvs` deploy + `vss vios add` + confirmation that hits land in `vss search tag`/`fusion` with no manual follow-up call. Doc-only change; flagging this as the one claim not empirically re-verified end-to-end in this PR.


ayyappa-dev:
- The search notification_config.json camera_remove cleans mdx-embed-filtered-2025-01-01, mdx-behavior-2025-01-01, mdx-raw-2025-01-01
   (_delete_by_query) and calls /v1/stream/remove on CV/embed/VLM — but has no _delete_by_query against the default_<streamId> tag index.
 - vss vios delete → vios.delete_media (libs/vss/core/.../vios/client.py:1536) only does sensor + storage delete; it does no
   Elasticsearch/tag cleanup (it's a VIOS/storage client).

 So deleting a source — or deleting + re-adding the same sensorId — leaves stale/duplicate tag docs in default_<streamId> → vss search tag and fusion return stale hits. The author does flag this in services/vios.md ("…does not delete the VLM tag index… needs a follow-up fix"), so it's documented, not hidden — but this PR also makes vss vios delete the sole documented teardown, so stale tag hits are now the expected outcome of the documented path.

PR #2273, opened 2026-09-18
Make the NemoClaw setup log usable and correct prerequisite facts

nvleeryan:
## Summary

The NemoClaw harness bring-up had two opposite problems in one place, and fixing
either alone makes the other worse.

The log was empty of anything useful. `run_setup_notebook.py` held every cell
output in memory, printed one summary line, and discarded the rest, so a `tee`d
setup log carried none of section 3.5's `Sandbox:` and `Agent UI:` lines, nor the
`WARNING:` that cell prints — without failing — when the dashboard forward does
not come up. `--require-output` matched the in-memory copy either way, so a run
exited 0 with a log that proved nothing.

The log was also where a live credential was expected to live. `agent-harness.md`
told the agent to strip the `#token=` fragment from its summary and then point the
user at `_builds/<name>/nemoclaw-setup.log` for the complete authenticated URL —
a gateway token at rest in a file, as the documented happy path, which is also why
that block set `umask 077`.

So the runner now echoes the notebook's output *with the `#token=` fragment
redacted*. That makes the log worth keeping and worth reading, and it removes the
reason to send anyone there for a secret: the doc hands over
`nemoclaw <name> gateway-token --quiet` and the `/#token=` form instead of a value.

The branch also corrects hardware prerequisite facts confirmed against the VSS 3.4
requirements — driver floors, the Fabric Manager package name and scope, the Docker
bound on DGX Station GB300, and which platforms are Compose-only.

## Plan

1. Add echoing behind an explicit opt-in rather than on by default — cell output can
   carry credentials in shapes the `#token=` pattern does not match, so the caller
   chooses.
2. Redact the one shape that is known to appear, so the log can be shared.
3. Delete the "read the log for the URL" instruction that the redaction obsoletes,
   and replace it with the token command plus the link form.
4. Fix the notebook comments that misdescribe where the remote bind is established.
5. Correct the prerequisite facts.

Follow-up, not in this PR: the new flag and its redaction have no test. The suite at
`deploy/docker/scripts/tests/test_run_setup_notebook.py` (681 lines, run in CI)
covers the parameter contract and the HITL launch contract, but not output echoing.

## Related Issue

None.

## Changes

**`deploy/docker/scripts/run_setup_notebook.py`**

- New `--echo-output` flag, threaded through `run_notebooks` into `execute_notebook`
  as a keyword defaulting to `False`. Default behaviour is byte-for-byte unchanged.
- `_TOKEN_FRAGMENT` rewrites `#token=<value>` to `#token=<redacted>` before printing.
  Only the printed copy is scrubbed; `--require-output` still matches the unredacted
  in-memory text, so a marker containing a token would still work.

**`skills/vss-build-vision-ai/references/agent-harness.md`** (the bulk of the diff)

- Bring-up block passes `--echo-output`, and the `umask 077` comment now says what
  the log actually holds — the notebook's settings dump, not an authenticated URL.
- Replaces the "keep the whole run" paragraph: names the flag as the thing that puts
  output in the log at all, and warns that `--require-output` passes without it, so
  omitting it yields a green run and a worthless log.
- Removes the instruction to point the user at the setup log for the complete
  authenticated URL. Requires carrying **all three** of origin, token command, and
  the `/#token=` form into the summary, because a token-free origin on its own lands
  the user on an unauthenticated page.
- Adds the log's one unique value: which origin the notebook *chose*, which can
  differ from what the Brev context file publishes if the notebook's own read of
  `/etc/brev` was denied.
- Treats a `127.0.0.1` origin as a claim to prove, not a fallback to take: resolve
  `brev_origin <dashboard-port>` and read the result first, since an unread lookup is
  not an empty one.
- Names a denied `/etc/brev` read as a third outcome distinct from both. Under a user
  namespace the directory appears owned by `nobody`, so a confined agent's denial
  looks like an ordinary missing file; it must be escalated (unconfined re-run, or the
  approved `docker run -v /etc/brev:/brev:ro` read in `brev.md`) rather than resolved
  as "no link". Only a lookup that ran with access and came back empty justifies the
  loopback origin and its SSH tunnel.

**`skills/vss-build-vision-ai/references/brev.md`**

- Adds the harness dashboard port (`NEMOCLAW_DASHBOARD_PORT`, default 18789) to the
  secure-link table for NemoClaw builds: it is the Agent UI's own origin and is not
  behind HAProxy, so it needs its own link, and the answer decides whether the summary
  reports a secure-link URL or a loopback one behind a tunnel.

**`deploy/docker/scripts/deploy_nemoclaw.ipynb`** (comments and markdown only)

- Corrects the `CHAT_UI_URL` comment: onboard derives the UI allowed origins from it
  but *not* the forward bind. `NEMOCLAW_DASHBOARD_BIND=0.0.0.0` is refused for a
  custom `--from` image, which is every image here, so onboard always binds loopback.
- Section 3.5 now says a loopback bind is the normal case rather than a fault to
  investigate, and that this cell is where the `0.0.0.0` bind a remote origin needs is
  actually established — re-reading the Brev context itself instead of trusting what
  3.1 baked into `CHAT_UI_URL`.
- Spells out the OpenClaw link form, where the token comes from, that it must be
  re-read after a gateway restart, and why it rides in a fragment: a fragment is never
  sent to the server, so the token stays out of the secure-link edge's request logs.

**`skills/vss-build-vision-ai/references/prerequisites.md`**

- New platform row for x86 dGPU H100 (SXM / PCIe) and H200 at `595.58.03`; the table
  previously had no H100/H200 entry outside the Kubernetes row.
- DGX Station GB300 driver floor `595.58.03` → `610.43.03`, the DGX OS 7.6.0 bundled
  driver (host-confirmed, VSS-TEGRA-REQ-002 §4.9).
- Fabric Manager: package corrected to `cuda-drivers-fabricmanager-<branch>` from
  `nvidia-fabricmanager-<branch>`, and scope corrected from "H100 SXM HBM3" to H100
  SXM *and* H200 SXM, with the exclusion restated as PCIe form factors rather than
  "multi-GPU PCIe-only".
- Notes that DGX Station GB300 ships Docker `29.6.1`, above the tested upper bound, so
  the pin script's downgrade branch is what runs there; the override is the fallback
  only on a *version not found*, and it settles the snapshotter alone, leaving
  `cgroupfs` to be applied separately.
- New Kubernetes row in the dependency table: GPU Operator `v26.3.3`, K8s
  `v1.29`–`v1.36`, x86_64 and ARM SBSA only, with every Tegra platform, DGX Spark,
  GH200/GB200 and DGX Station GB300 called out as Docker Compose only.
- Drops a line asserting the expected GPUs of one specific machine.

## Checklist

- [x] I have installed and run [pre-commit hoo_


deploy/docker/scripts/run_setup_notebook.py
```diff
@@ -56,6 +57,10 @@
 
 _MINIMUM_TIMEOUT_SEC = 60
 
+# The OpenClaw control UI authenticates from this fragment, so an echoed
+# Agent UI line would otherwise leave a live gateway token in the caller's log.
+_TOKEN_FRAGMENT = re.compile(r"(#token=)[^\s\"'<>]+")
```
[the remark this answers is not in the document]
nvleeryan: did'tn https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/2273#:~:text=%E2%80%A6-,71cfc3a,-bp%2Dsre%2Dsonarqube resolve this?

nvleeryan:
/ok to test https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/2273/commits/2fc6bd2e9adcd0e47cc4b1f961a8a8181662f76d

deploy/docker/scripts/deploy_nemoclaw.ipynb
```diff
@@ -1435,12 +1437,14 @@
...
     "\n",
-    "<span style=\"color:red\"><strong>Not on Brev?</strong> If you are accessing the UI from a different machine, run the exact SSH tunnel command printed by the next cell before opening the Agent web UI.<br/></span>\n"
+    "A loopback bind here is the normal case, not a fault to investigate: NemoClaw refuses `NEMOCLAW_DASHBOARD_BIND=0.0.0.0` for a sandbox built from a custom `--from` Dockerfile, which is every sandbox this notebook creates, so onboard can only bind `127.0.0.1`. This cell is therefore where the `0.0.0.0` bind a remote origin needs is actually established — it re-reads the Brev context itself (falling back to `BREV_ENV_ID`) rather than trusting what 3.1 baked into `CHAT_UI_URL`.\n",
+    "\n",
+    "<span style=\"color:red\"><strong>Not on Brev?</strong> If you are accessing the UI from a different machine, open an SSH tunnel before opening the Agent web UI from below:<br/><code>ssh -L &lt;AGENT_DASHBOARD_PORT&gt;:127.0.0.1:&lt;AGENT_DASHBOARD_PORT&gt; &lt;user&gt;@&lt;agent-host&gt;</code><br/></span>\n"
```
zac-wang-nv: **P1** — Keep the instruction to use the tunnel printed by the next cell instead of hard-coding `127.0.0.1` here. Off Brev, `VSS_AGENT_ADAPTER_ENABLED=true` makes section 3.5 bind the dashboard only to the address Docker maps to `host.docker.internal` (and the build skill exports that flag). The new command therefore connects to an interface with no dashboard listener, so a remote user cannot open the Agent UI even when setup succeeds. The cell already prints a tunnel using `_health_host`, which handles both adapter and loopback modes.

skills/vss-build-vision-ai/references/agent-harness.md
```diff
@@ -549,15 +558,32 @@ Confirm the two things that exit code cannot cover:
...
-   for the complete authenticated URL. Never copy the token or the
-   secret-bearing URL into the response. Hand over the recipe rather than the
-   value: on the deployment host `nemoclaw <name> gateway-token --quiet` prints
-   a fresh token, and the UI URL is the reported origin plus `/#token=<token>`.
-   On Brev the host is the secure-link FQDN. A `127.0.0.1` origin only resolves
-   on the deployment host, so pair it with the SSH tunnel the same section
-   prints.
+   fragment and report only the token-free origin as a markdown link. Never
+   copy the token or the secret-bearing URL into the response, and do not send
+   the user to the setup log for it — `--echo-output` redacts the fragment, so
```
zac-wang-nv, line 563: **P1** — Update the matching final-summary instruction in skills/vss-build-vision-ai/SKILL.md:354 in this PR too. Step 10 still explicitly tells the agent to point the user to the local setup log for the authenticated URL. Once this change enables `--echo-output` and redacts that URL, following the main skill sends users to a log that cannot authenticate them and contradicts this reference. Replace that instruction with the same origin + `gateway-token` command + `/#token=<token>` recipe so the entry-point workflow produces a usable handoff.

deploy/docker/scripts/deploy_nemoclaw.ipynb
nvleeryan: address https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/2273/commits/14365190326311c7e604de86c66ba37696d0ba1e

skills/vss-build-vision-ai/references/agent-harness.md
nvleeryan: https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/2273/commits/0705eb09b14e4daaea28b4564eb54dc434eb8af3

nvleeryan:
/ok to test https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/2273/commits/9fff56bf1d74c17a1b9c5cedef7fe564e1bd4e17

zac-wang-nv (approved):

PR #2276, opened 2026-09-18
fix(agent-ui): relicense UI sources to Apache-2.0 with SPDX headers

akshayanv:
## Summary
- Relicense NVIDIA-authored files under `services/ui/` from leftover MIT/`MIT AND Apache-2.0` SPDX headers to Apache-2.0 with NVIDIA copyright notices.
- Replace the UI MIT `LICENSE` file with Apache-2.0 `LICENSE.md`, update the Docker copy path, and drop duplicate UI-local contributing/security/code-of-conduct stubs that pointed at repo-wide docs.

## Test plan
- [ ] Spot-check SPDX headers on a few app, package, and test files (`Apache-2.0` + NVIDIA copyright).
- [ ] Confirm `services/ui/LICENSE.md` is Apache-2.0 and the image still copies `LICENSE*` as documented in the Dockerfile.
- [ ] Confirm no remaining `SPDX-License-Identifier: MIT` under `services/ui/`.


services/ui/Dockerfile
```diff
@@ -98,7 +98,7 @@ COPY --from=builder --chown=${USER_ID}:${GROUP_ID} --chmod=555 /repo/services/ui
 # Copy custom server (read-only)
 COPY --from=builder --chown=${USER_ID}:${GROUP_ID} --chmod=555 /repo/services/ui/custom-server.js ./custom-server.js
 
-# Copy license files (read-only): LICENSE (MIT) and LICENSE-3rd-party.txt.
+# Copy license files (read-only): LICENSE.md (Apache-2.0) and LICENSE-3rd-party.txt.
```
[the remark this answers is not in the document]
akshayanv: Fixed in 7a7305bbe: `Dockerfile.dockerignore` now has `!**/LICENSE.md` so the Apache-2.0 file stays in the build context and the existing `COPY LICENSE*` into `/repo` can pick it up.

services/ui/LICENSE.md
zac-wang-nv, line 1: this is the same license as the top level license. can we remove this copy?

akshayanv: Done

services/ui/apps/nv-metropolis-bp-vss-ui/__tests__/utils/server/agentAdapterConnectors.test.ts
```diff
@@ -1,3 +1,5 @@
+// SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// SPDX-License-Identifier: Apache-2.0
```
zac-wang-nv: **P1** — Keep the environment docblock first, or put the SPDX fields inside it. Jest only extracts a leading block comment, so these new `//` lines hide `@jest-environment node` and this file falls back to the app's `testEnvironment: 'jsdom'`. With the lockfile's Jest packages, the current connector tests fail four of seven cases with `ReferenceError: Response is not defined`; removing only these two added lines makes all seven pass. The existing SPDX comments immediately below the pragma already cover the file, so the new prefix can simply be removed.

LICENSE
```diff
@@ -1,9 +1,6 @@
 Copyright (c) 2025-2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 
-This project is dual-licensed:
-  - Apache-2.0 applies to all code in the repository except the services/ui/ directory.
-  - MIT applies to the original code under the services/ui/ directory, which is
-    derived from upstream projects (see MIT appendix below).
+This project is licensed under Apache-2.0.
```
zac-wang-nv, line 3: **P1** — Preserve the upstream MIT notices for the code that is still derived from those projects. For example, services/ui/apps/nv-metropolis-bp-vss-ui/next-i18next.config.js is unchanged apart from its SPDX header and matches [NeMo Agent Toolkit UI's configuration](https://github.com/NVIDIA/NeMo-Agent-Toolkit-UI/blob/41cb8cf843091a24bd06d0a5f844a3cb824049d8/next-i18next.config.js) exactly; it also matches [chatbot-ollama's configuration](https://github.com/ivanfioravanti/chatbot-ollama/blob/main/next-i18next.config.js) apart from whitespace and quote style, and the app still bundles it. This PR deletes both copies of Ivan Fioravanti's and Mckay Wrigley's notices, while services/ui/LICENSE-3rd-party.txt contains neither attribution. The [upstream MIT license](https://github.com/NVIDIA/NeMo-Agent-Toolkit-UI/blob/41cb8cf843091a24bd06d0a5f844a3cb824049d8/LICENSE) retains those authors and requires preservation of the notices for copied material. Keep the applicable attribution/license text in a file shipped with source and images, and retain accurate provenance for derived files while applying Apache-2.0 to NVIDIA-authored changes.

zac-wang-nv:
**P2** — Update README.md:157 with the license-file removal. The repository's License section still links to `services/ui/LICENSE` and says the original UI code is MIT with an appendix in the root LICENSE. This PR removes both targets, leaving the main landing page with a broken link and stale licensing guidance. Point it to the surviving license/attribution files and describe their current scope.

akshayanv:
Done in 1308ac690. The License section no longer links to the removed `services/ui/LICENSE`; it now points at the surviving root `LICENSE`, `LICENSE-3rd-party.txt` (including `services/ui/LICENSE-3rd-party.txt`), and `LICENSE.DATA`.

LICENSE
akshayanv: Only a next js config file. Even though it is same as NAT UI repo, this kind of a config file can be same in multiple projects. I dont think this suggests same codebase

zac-wang-nv: Agreed: the matching configuration file alone is insufficient evidence to support the licensing P1 I posted. I am withdrawing that finding and resolving this thread. I also verified the fixes at 1308ac690: the isolated connector suite now passes all seven tests, and the README license links point to existing files.

zac-wang-nv (approved):
Verified the fixes at 1308ac690: all seven connector tests pass in the isolated reproduction, and README license links are corrected. The licensing P1 is withdrawn because the configuration-file comparison was insufficient evidence. No remaining code-review findings; CI completion is still pending.

PR #2290, opened 2026-09-18
feat: report one version — the installed `nvidia-vss-core` — plus version-endpoint follow-ups

hugoverjus:
## Summary

Follow-ups to **#2241** (`feat/vss-version-endpoint`), stacked on that branch. It does **not** add the version endpoint or the preflight gate — that is #2241 / [VIA-2760](https://jirasw.nvidia.com/browse/VIA-2760). Merge **after** #2241.

The headline change is the last commit: `GET /api/v1/version` now reports **one** version, the installed `nvidia-vss-core`, instead of a number written down in several places. The rest are the smaller follow-ups the branch started with.

### One version: the library the deployment is running

Review feedback on #2241 asked whether the VSS library or CLI could supply the version, since `vss-agent` is not in every deployment and the library already tracks the repo's SemVer. The right reading of that is *unification*, not another field in the payload — the response shape is unchanged.

As merged in #2241, the endpoint resolved from `VSS_DEPLOYMENT_VERSION`, then `VSS_AGENT_VERSION`, then the checkout's git tags. The first of those was defaulted by hand in `containers.env`, in the agent `compose.yml`, and again via `.Chart.Version` in the Helm values — three copies of one number, with a CI drift test policing them. Meanwhile the packages inside the image carried a placeholder version, so the library could not answer for itself.

Now:

- `vss_core.library_version()` reads the installed `nvidia-vss-core` distribution and normalises PEP 440 to strict SemVer.
- The image build stamps that distribution with `<release line>+tree.<source tree sha>`, where the release line is read from `deploy/helm/services/agent/Chart.yaml` — the repo's one declared release line, the same field Helm already consumes.
- So the version a deployment reports is the version of the code it is running, established at build time.

Resolution order becomes: `VSS_DEPLOYMENT_VERSION` (an operator override for a wrong stamp, **unset everywhere by default**) → the installed `nvidia-vss-core` → `git describe` for a source run with no install metadata → HTTP 503.

`VSS_AGENT_VERSION` **leaves the resolution chain**. It doubles as a fallback image tag, so it routinely carries `develop-latest` rather than a version; keeping it as a second source defeated the point. Its unrelated jobs — Phoenix telemetry project naming and image-tag fallback — are untouched.

Two properties worth keeping in view when reviewing:

- **The stamp stays tree-derived, never commit-derived.** `build-dev-images.yml` re-tags an already-published manifest onto a later commit whose source tree is identical. A commit-derived stamp would go stale the moment that reuse fires, and a re-tagged image would report the commit it was first built from.
- **SemVer precedence ignores build metadata**, so `3.3.0+tree.<sha>` satisfies a range like `>=3.3.0,<4.0.0`. That is what makes the stamp usable by the benchmark compatibility check rather than merely informative.

The Compose/chart drift test is deleted along with the duplication it guarded. The check that the chart version is strict SemVer stays, and matters more than before: that string is now baked into every image.

**Deliberately not solved here:** the endpoint still lives on the agent, so a deployment without it still answers 404. Unifying the version *source* does not change which service serves it.

### The image stamp

`services/agent/docker/Dockerfile` baked `SETUPTOOLS_SCM_PRETEND_VERSION=0.1.0`, plus a `..._FOR_VSS_AGENT=0.1.1` that named a distribution this repo does not build (`vss-agent`; the packages are `nvidia-vss`, `nvidia-vss-agents`, `nvidia-vss-core`, `nvidia-vss-cli`), so it set nothing.

A value is mandatory — the `COPY` lines name individual paths and never `.git`, so `hatch-vcs` has no history to derive from and the install fails without one. The problem is that `0.1.0` looks like a release, and `importlib.metadata` inside the container reported it as fact.

It is now `ARG VSS_PACKAGE_VERSION`, composed by the workflow as `<release line>+tree.<sha>`. The local default is `0.0.0+local`: it cannot be mistaken for a shipped build, cannot satisfy any skill's version range, and does not name a tree it did not verify.

The workflow reads the chart version once in the `changes` job and both build jobs consume it, so there is no per-matrix duplication and no new CI dependency (`grep`/`cut`, no `yq`). An empty or malformed chart version fails the step loudly rather than publishing an immutable image that claims no release.

### `vss configure check` reports the version

`check` re-probes a recorded deployment; it now also reports the version that deployment says it is, or why it cannot:

```
  agent          ok           http://example/api  HTTP 200
  version        3.3.0+tree.4d9b2c1
```

```
  version        not reported  HTTP 404 — no agent behind /api here, or a deployment predating the endpoint
```

- **Probed per check, never recorded by `configure`.** A redeploy changes the version without changing the config, so a recorded value would be the one field in the file guaranteed to go stale.
- **It does not fail the check.** A lean stack without the agent is a legitimate deployment, not drift. The point is that the operator learns it from the prober they already run, rather than from a benchmark that stops on the same fact much later.
- 404 and 503 are distinguished, because they call for different actions: no agent behind this origin (or a deployment predating the endpoint) versus a deployment reporting no usable version.

### `benchmark-vlm-qa` documentation corrections

Both claims were wrong in the same way — treating a client-side capability as a deployment one.

- The prerequisites named `uv run --project libs/vss vss`, but `benchmark_vlm_qa.py` runs `uv run --project services/agent --no-dev --extra cli vss`. Documenting an invocation the script does not use sends anyone debugging an environment problem to the wrong workspace.
- The `requires-vss` floor was justified by a 3.2 deployment lacking the `vss vlm` command group. That group comes from this checkout, so it is present whatever the deployment reports; the real floor is the Cosmos Reason 3 stack the published baselines were measured on.

## Validation

- `pytest core/tests/unit_test/lib/test_version.py -q` in `libs/vss` — **26 passed**, covering `library_version()` (normal metadata, `PackageNotFoundError`, unnormalisable value) and the new resolution order. `importlib.metadata.version` is patched, so nothing depends on what is installed in the venv.
- `pytest packages/vss_agents/tests/unit_test/api -q` in `services/agent` — **294 passed**, including the endpoint reporting the library version with nothing configured, the override winning, and an invalid override giving 503.
- `pytest cli/tests/unit_test -q` in `libs/vss` — **449 passed**, including 5 in `test_configure_version.py`.
- Against the real working tree, the endpoint resolves `3.2.1-dev.1506+g884936ae8` from the editable `nvidia-vss-core` install — the library answering for itself, end to end.
- `ruff check` + `ruff format --check` clean on every touched source and test; `mypy` clean on `vss_core/version.py`.
- The `ARG` → `ENV` expansion was built for real: `--build-arg VSS_PACKAGE_VERSION=3.3.0+tree.abc123` bakes that exact string, and the default bakes `0.0.0+local`. Both parse as PEP 440 **and** match `SEMVER_PATTERN`.
- The chart reader was exercised against the real file (reads `3.3.0`) and against `''`, `3.3` and `3.3.0+meta` — all three rejected. The `+meta` case matters: a chart version already carrying build metadata would collide with the appended `+tree.<sha>`.
- `docker compose config` with nothing set renders `VSS_DEPLOYMENT_VERSION: null` — passed through only when the operator actually sets it, no empty-string injection. `helm template` on default values no longer emits the chart version.
- `.github/scripts/test_agent_config_defaults.py` — 2 tests pass after the drift test's removal.
- `build-dev-images.yml` parses, and `release-line` is wired from `changes` into both build jobs.

## Test plan

- [ ] Base is `feat/vss-version-endpoint` (#2241); merge only after it
- [ ] An agent image built from this branch reports `<chart version>+tree.<sha>` from `GET /api/v1/version`, and the sha matches the image's `com.nvidia.vss.source_tree_sha` label
- [ ] Two commits with an identical `services/agent` tree still hit the re-tag path and carry the same stamp
- [ ] A stock Compose deployment answers 200 with nothing set, from the image stamp alone
- [ ] Setting `VSS_DEPLOYMENT_VERSION` overrides it; setting it to a non-SemVer value gives 503 rather than falling back
- [ ] `vss configure check` against a deployment with the agent prints its version; against one without, prints the 404 line and still exits 0


hugoverjus:
Tracking: [VIA-2760](https://jirasw.nvidia.com/browse/VIA-2760)

Follow-up polish only — not the ticket's required deliverable. Parent PR: https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/2241

/ok to test c85c4a4e8038bba3c3473eeb4941807ea15627fa

hugoverjus:
Parent / VIA-2760 deliverable: https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/2241

zac-wang-nv (dismissed):

zac-wang-nv (approved):
lgtm

PR #2293, opened 2026-09-18
fix: add locked VLM request policy

maxervo:
## Summary
- add a configuration-only `rt-vlm` / `vllm` backend choice that translates the same locked policy into the appropriate request schema
- pass `seed`, `enable_reasoning`, and `chunk_duration` through `vss vlm run` for URL and base64 requests
- persist `timeout`, `temperature`, `max_tokens`, `seed`, `enable_reasoning`, `chunk_duration`, and `fps` with `vss configure vlm`
- support locked policies that reject conflicting per-call overrides and survive base-URL reconfiguration

The lock is to deterministically prevent agents from overriding the defaults (edge case happening at scale during eval, when agent was following `vss-ask-video` and doing exploratory targeted queries)  

This PR is needed to run evals for Vantage-Bench and now Qwen 3.8 27b on RT-VLM, for proper server side video processing and sampling.

## Validation
- 491 CLI unit tests passed locally
- Ruff checks passed for all changed files
- Mypy passed for all changed source files


libs/vss/cli/src/vss_cli/config.py
```diff
@@ -596,21 +614,193 @@ def from_json(cls, raw: object) -> MemoryConfig:
...
+
+
+def _parse_vlm_environment_value(field_name: str, environment_name: str, raw: str) -> object:
+    """Parse one explicitly defined VLM environment override."""
+    value = raw.strip()
+    if not value:
+        raise ConfigError(f"{environment_name} is set but empty")
+    if field_name == "backend":
+        if value not in {"rt_vlm", "vllm"}:
+            raise ConfigError(f"{environment_name} must be 'rt_vlm' or 'vllm'")
```
zac-wang-nv: we need to support cosomos reason NIM as a backend as well

libs/vss/cli/src/vss_cli/config.py
```diff
@@ -596,21 +614,193 @@ def from_json(cls, raw: object) -> MemoryConfig:
...
+
+def effective_vlm_config(configured: VlmConfig | None) -> VlmConfig | None:
+    """Overlay per-field environment values on a persisted VLM policy.
+
+    Environment variables have precedence over values written by
+    ``vss configure vlm``. An absent environment variable leaves its persisted
+    value unchanged. When neither source defines a policy, return ``None`` so
+    existing built-in request defaults retain their current behavior.
+    """
+    overrides = {
```
zac-wang-nv: this should be the other way around.  VLM_ENV provides default.  vss commands can override

zac-wang-nv:
**P1** — Please define and enforce a clear sampling contract between the CLI, vLLM's video loader, and Qwen's processor. Both sampling modes are valid:

- `--num-frames 30` requests 30 frames total spread across the clip, subject to available frames. The current fixed-count branch sends this to the loader as `media_io_kwargs.video.num_frames`.
- `--fps 4` requests approximately four frames per second: about 80 frames for a 20-second clip. If sampling happens in the loader, send `media_io_kwargs.video.fps` with a compatible `num_frames` budget. In vLLM 0.17's OpenCV loader, a positive `num_frames` caps the FPS-derived count; `{"num_frames": -1, "fps": 4}` removes that count cap while retaining FPS sampling. Any intended limit should be explicit and documented.
- After the loader selects frames, preserve its `do_sample_frames=False` metadata so the processor consumes those frames with their original timestamps. Alternatively, processor-side sampling with `do_sample_frames=True` requires arranging for all original frames to reach it.

The current FPS branch only sets `mm_processor_kwargs.fps` and forces `do_sample_frames=True`, leaving the loader's frame budget uncontrolled. With vLLM 0.17's default OpenCV configuration, a 20-second, 30-FPS source is reduced from 600 frames to 32 before Qwen sees it, but its metadata still describes the original 600 frames. Qwen then selects indices up to 599 from that 32-frame array and raises `IndexError`. Here, "already sampled" means this internal loader step, not a special kind of input video. A deployment configured to deliver all original frames avoids this specific failure.

Please test the loader-output/metadata-to-processor boundary for both frame-count and FPS requests, covering URL and base64 inputs; payload-field assertions alone miss this mismatch.

References: [vLLM loader defaults](https://github.com/vllm-project/vllm/blob/v0.17.0/vllm/multimodal/media/video.py), [loader metadata](https://github.com/vllm-project/vllm/blob/v0.17.0/vllm/multimodal/video.py), [Qwen processor sampling](https://github.com/huggingface/transformers/blob/main/src/transformers/models/qwen3_vl/video_processing_qwen3_vl.py). The repository's vendored Qwen3 adapter also explicitly requires `do_sample_frames=False` for presampled videos at services/rtvi/rt-vlm/docker/rtvi_vlm/patches/evs_vllm_public_files/model_executor/models/qwen3_vl.py:997.

Location: [libs/vss/cli/src/vss_cli/vlm/group.py:440](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/72e8e78ead284c55124be60ddab3a3ff9db4ee47/libs/vss/cli/src/vss_cli/vlm/group.py#L440).

zac-wang-nv:
**P2** — Please map VLM configuration failures to `Exit.CONFIGURATION` (4), as `configure memory` does. `vss configure vlm --fps 4` with no saved deployment, and `vss configure vlm --lock` with no configured request values, currently return Click's generic exit 1. Callers following the CLI's documented exit-code contract cannot classify these as configuration problems. The load-error handler here and the policy-validation handler at libs/vss/cli/src/vss_cli/configure.py:759 need the same mapping; assert the exact exit in the tests rather than only nonzero.

Location: [libs/vss/cli/src/vss_cli/configure.py:709](https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/blob/72e8e78ead284c55124be60ddab3a3ff9db4ee47/libs/vss/cli/src/vss_cli/configure.py#L709).

zac-wang-nv:
**P2** — Please update the `--fps` row at libs/vss/cli/README.md:333. It still promises a 60-frame cap and conversion to a fixed frame count, but this PR deliberately removes that behavior from direct `vlm run`. A two-minute clip at 2 FPS now sends FPS through unchanged instead of requesting 60 frames. Describe the backend-enforced limit and distinguish the separate introspection path, which still uses the bounded analyzer.

zac-wang-nv (approved):

PR #2305, opened 2026-09-19
Recommend the build's own LLM for the harness on edge devices

nvleeryan:
## Summary

On DGX Spark and AGX/IGX Thor with a `local` or `local_shared` VSS LLM, Q3a
still defaulted the NemoClaw sandbox to a remote Claude Opus 5 endpoint, so an
edge box already serving a model took on a needless remote dependency and a
second inference path.

Q3a now offers the build's own LLM as the first, recommended choice on that
hardware. It stays a recommendation the user confirms, and it is not offered on
non-edge hardware or against a remote VSS LLM, where the three existing choices
and the Claude Opus 5 default are unchanged.

## Plan

Documentation-only change to the `vss-build-vision-ai` skill. The route itself
already exists — `agent-harness.md` option *(a) against the build's own LLM
NIM*, with `NEMOCLAW_INFERENCE_PROXY=0` and `COMPATIBLE_API_KEY=EMPTY` — so this
selects it at Q3a on the hardware where it applies rather than adding any new
mechanism. Port and served model are read from `resolved.yml` (`LLM_PORT`,
`NIM_SERVED_MODEL_NAME`) rather than assumed.

## Related Issue

None.

## Changes

- `skills/vss-build-vision-ai/SKILL.md` — adds the edge rule to **Harness model
  — Q3a**: on DGX Spark or AGX/IGX Thor with a `local`/`local_shared` VSS LLM,
  the build's own LLM becomes the first and recommended option, taken only when
  the user selects it. Also drops a stale "Otherwise" from the Q3a-default lead
  sentence.
- `skills/vss-build-vision-ai/references/agent-harness.md` — **Default
  provider** gains the edge case ahead of the existing remote default; the
  harness-requirements bullet now states both recommendations and that each is
  user-confirmed; the GPU-budget bullet notes the edge route reuses the VSS LLM
  with no second model process, and scopes its warning to the
  NemoClaw-managed local provider; the environment table's `NEMOCLAW_PROVIDER`
  row reflects both accepted-recommendation outcomes.
- `skills/vss-build-vision-ai/references/edge.md` — restates the recommendation
  and the no-second-process rationale at the point where the edge LLM is
  configured.

## Checklist
- [ ] I am familiar with the [Contributing Guidelines](https://github.com/N


nvleeryan:
/ok to test https://github.com/NVIDIA-AI-Blueprints/video-search-and-summarization/pull/2305/commits/dc08d557017a082101809fc7d022ba60f6a43af3

